PR #66, opened 2023-06-01
helmChart: add runtimeClass for device-plugin daemonset

gsakun:
k8s node may have multiple runtimes coexisting and nvidia-runtime is not the default runtime.But device-plugin needs to use nvidia-runtime，So provide a configuration item for users to configure.


archlitchi:
approved, thanks for this pr:)

PR #117, opened 2024-01-10
fix device-plugin panic

chaunceyjiang:
Fixes: https://github.com/4paradigm/k8s-vgpu-scheduler/issues/116


archlitchi:
thanks for contributing:)

PR #119, opened 2024-01-11
Fix go.sum file

chaunceyjiang:
After using go mod vendor and go mod tidy, I found that the go.sum file has changed.

```
# go version             
go version go1.21.0 darwin/arm64

```


archlitchi:
indeed, thanks for contributing:)

PR #122, opened 2024-01-12
install nvidia-vgpu-schduler as default scheduler

calvin0327:
If we install the nvidia-vgpu-scheduler as default scheduler, we need to remove the k8s defualt
scheduler pod from the cluster first, we must specified node name to skip the shcedule workflow.

And use the following helm command to install nvidia-vgpu-scheduler:

```shell
helm install nvidia-vgpu -n nvidia-vgpu  --create-namespaces \
         --set schedulerName=default-scheduler \
         --set scheduler.nodeName=<node_name> \
         --set scheduler.kubeScheduler.image=<kube_scheduler> \
         --set scheduler.kubeScheduler.tag=<kube_scheduler_tag> \
```

### DIFF
--- a/charts/vgpu/values.yaml
+++ b/charts/vgpu/values.yaml
@@ -31,6 +31,10 @@ global:
   annotations: {}
 
 scheduler:
+  # @param nodeName defines the node name and the nvidia-vgpu-scheduler-scheduler will schedule to the node.
+  # if we install the nvidia-vgpu-scheduler-scheduler as default scheduler, we need to remove the k8s defualt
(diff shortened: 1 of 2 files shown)


calvin0327:
@archlitchi /cc

charts/vgpu/values.yaml
archlitchi, line 1: please fix 'shcedule' -> 'schedule'

calvin0327:
@archlitchi Thanks for review, done.

archlitchi:
> @archlitchi Thanks for review, done.

i never thought there would be a bootstrapping issue, thanks for contributing:)

PR #123, opened 2024-01-13
fix device-plugin panic

chaunceyjiang:
When `util.GetNode(val.Name)` returns an error, a panic will occur.


archlitchi:
thanks for fixing:)

PR #129, opened 2024-01-17
fix: loss of metrics after vdeivce restart

CoderTH:
Problem description.

After the scheduler component restarts unexpectedly, some metrics are lost, which can only be obtained when the scheduling is triggered.
image
image

Solution
When scheduler starts, all node information is obtained and injected into scheduler object

The test results are as follows

image


CoderTH:
@archlitchi cc

archlitchi:
thanks for contributing:)

PR #131, opened 2024-01-22
add metrics bind address flag for scheduler

calvin0327:
Why do we need the flage?
The metrics port is confilct with vGPU monitor if we deploy the scheduler with `HostNetwork` mode.
In the scenario, we want to deploy scheduler with k8s native scheduler in one static pod with `HostNetwork` mode. but the  vGPU monitor is same port with scheduler.


calvin0327:
@archlitchi I want to test the feature, but Idon't know how to build the scheduler image.

calvin0327:
@archlitchi Thanks,  Here are the test results：

```bash
helm template ./charts/vgpu
```

image

archlitchi:
thanks ~

PR #135, opened 2024-01-23
add vendor-check to github actions

chaunceyjiang:
I introduced the `vendor `directory because, based on my experience, incorporating this directory into `git` allows for a quick review of `changes` in a certain `dependency library `without needing to browse through upstream code.

PR #147, opened 2024-02-01
cleanup: remove io/ioutil package

chaunceyjiang:
Part of #https://github.com/Project-HAMi/HAMi/pull/135

### DIFF
--- a/cmd/device-plugin/nvidia/vgpucfg.go
+++ b/cmd/device-plugin/nvidia/vgpucfg.go
@@ -3,7 +3,6 @@ package main
 import (
 	"encoding/json"
 	"fmt"
-	"io/ioutil"
 	"os"
 	"strings"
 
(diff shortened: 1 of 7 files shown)


haitwang-cloud:
Add more details here about the diff between  ioutil.ReadFile & os.ReadFile from GPT-4.

_Both ioutil.ReadFile and os.ReadFile are functions in Go language used to read data from files. However, as of Go version 1.16, the ioutil package has been deprecated. Most functions previously in ioutil have been moved to new packages such as os and io. Thus, ioutil.ReadFile is deprecated, and it's recommended to use os.ReadFile.
While these two functions serve similar purposes, reading data from a specified file path, they do differ slightly in usage:
**ioutil.ReadFile:**
This function reads the entire content of a file and returns a byte slice and an error. If the operation is successful, the error is nil. The function closes the file after use. This function was deprecated in Go 1.16.
`data, err := ioutil.ReadFile("path/to/file")`
**os.ReadFile:**
This function, introduced in Go 1.16, operates almost identically to ioutil.ReadFile. It reads the entire content of a file and returns a byte slice and an error. If the operation is successful, the error is nil. The function also closes the file after use.
`data, err := os.ReadFile("path/to/file")`
In comparison, os.ReadFile is preferred because ioutil.ReadFile has been deprecated and their functionalities are nearly identical with little difference in usage_.

archlitchi:
agree, thanks for fixing that

PR #150, opened 2024-02-04
add a ci job to build dev image

lengrongfu:
Fixes: https://github.com/Project-HAMi/HAMi/issues/149


archlitchi:
Thanks, a dev image is necessary indeed

Issue #89, opened 2023-10-12
`kubescheduler.config.k8s.io/v1beta2` should be `kubescheduler.config.k8s.io/v1` on k8s v1.28.2


morlay:
may consider to make it  configable in helm chart.

archlitchi:
Thanks, i'll fix that in next version

morlay:
be v1 since v1.25
https://kubernetes.io/docs/reference/scheduling/config/

PR #151, opened 2024-02-04
upgrade kubescheduler version

lengrongfu:
Fixes: https://github.com/Project-HAMi/HAMi/issues/89


archlitchi:
We need to  compat v1.16, so we need to check the 'scheduler.kubeScheduler.imageTag', if > 1.25, then we use v1, else use v1beta2

lengrongfu:
Thanks you suggest, PLAT.

PR #162, opened 2024-02-21
add a option to disable kube scheduler in chart

calvin0327:
in our scenario, we have instanlled component `scheduler-plugins` replace the kube-scheduer, and want only install the vgpu-scheduler as a extender. we need to suport this way to deploy vgpu scheduler only. 

the test results:
```shell
helm install hami -n hami-system ./charts/hami --create-namespace --set scheduler.kubeScheduler.enabled=false --dry-run
```
image


calvin0327:
@archlitchi PTAL, thanks

archlitchi:
thanks for your contributing, you could provide a document regarding how to connect hami-extender to custom scheduler if you had time:)

Issue #171, opened 2024-03-05
add nvidia.com/gpu field default value


lengrongfu:
when the user does not set this `vidia.com/gpu` resource value, the scheduler can't schedule success, so we can add a default value.

lengrongfu:
done

PR #172, opened 2024-03-05
add default gpu num value

lengrongfu:
It mainly simplifies user settings of resource limits. By default, the core and memory in one GPU are used.

issue: https://github.com/Project-HAMi/HAMi/issues/171


lengrongfu:
cc @archlitchi

wawa0210:
thx, It would be better if there were some test results~

lengrongfu:
> thx, It would be better if there were some test results~

done.

archlitchi:
nice PR, thanks:)

PR #176, opened 2024-03-11
Remove redundant assignment code for DefaultResourceNum

lengrongfu:
`default-gpu` is defined in two files, so there will be values that cannot be configured by the user. remove a define in `device.go` file.
issue: https://github.com/Project-HAMi/HAMi/issues/171

### DIFF
--- a/pkg/device/nvidia/device.go
+++ b/pkg/device/nvidia/device.go
@@ -32,7 +32,6 @@ var (
 	ResourceMemPercentage string
 	ResourcePriority      string
 	DebugMode             bool
-	DefaultResourceNum    int
 )
 
 type NvidiaGPUDevices struct {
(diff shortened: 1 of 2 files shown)


lengrongfu:
@archlitchi PTAL

pkg/device/nvidia/device.go
```diff
@@ -51,7 +50,6 @@ func (dev *NvidiaGPUDevices) ParseConfig(fs *flag.FlagSet) {
 	fs.StringVar(&ResourceMemPercentage, "resource-mem-percentage", "nvidia.com/gpumem-percentage", "gpu memory fraction to allocate")
 	fs.StringVar(&ResourceCores, "resource-cores", "nvidia.com/gpucores", "cores percentage to use")
 	fs.StringVar(&ResourcePriority, "resource-priority", "vgputaskpriority", "vgpu task priority 0 for high and 1 for low")
-	fs.IntVar(&DefaultResourceNum, "default-gpu", 1, "default gpu to allocate")
```
wawa0210, line 54: Do we need to delete the "DefaultResourceNum" value directly? Is it better if both the default value and config.DefaultResourceNum are assigned to "DefaultResourceNum".

In this way, we can also check whether the value of config meets expectations, etc.

lengrongfu: I suggest deleting the field, because have two points:
- current `DefaultMem` and `DefaultCores` field only in `config.go` file.
- in `main.go` file having a about `default-gpu` flags, so result in the `device.go` flag invalid.
https://github.com/Project-HAMi/HAMi/blob/efb342ffc6166de4b6ccfebab714f1bb9fc3012f/cmd/scheduler/main.go#L58

wawa0210: I see. Because the value has been assigned before 

https://github.com/Project-HAMi/HAMi/blob/efb342ffc6166de4b6ccfebab714f1bb9fc3012f/cmd/scheduler/main.go#L58

This is redundant code and meaningless

wawa0210:
lgtm

PR #178, opened 2024-03-11
fix EncodePodSingleDevice method

lengrongfu:
fix `EncodePodSingleDevice` method, when one pod having more then one container use device, this method encode device string will error.

- add many case to run unit tests.
- fix this bug.
- use const `OneContainerMultiDeviceSplitSymbol` and `OnePodMultiContainerSplitSymbol` to replace hardcode.


lengrongfu:
cc @archlitchi Please help to check this issue

archlitchi:
thanks for fixing this issue:)

PR #179, opened 2024-03-11
replace hami.sh to hami.io

lengrongfu:
fix: https://github.com/Project-HAMi/HAMi/issues/177

use `hami.io` to replace `hami.sh`.


archlitchi:
thank for fixing

PR #196, opened 2024-03-13
scheduler server add leader elect

lengrongfu:
Cureent kube-scheduler and scheduler-extender can't run more than one pod because they do not use leader election, This PR adds `leaderElect` to controller `kube-scheduler` and `hami-scheduler` servers, whether use leader election.  default: use helm `.Release.Namespace` value to set `leader-elect-resource-namespace`.

Fixes: https://github.com/Project-HAMi/HAMi/issues/195

Test:
-  Start two scheduler pods; only one pod can execute the scheduler flow.
image

- Start three use gpu pods that can running success.
image

### DIFF
--- a/charts/hami/templates/scheduler/deployment.yaml
+++ b/charts/hami/templates/scheduler/deployment.yaml
@@ -47,6 +47,9 @@ spec:
             - {{ . }}
             {{- end }}
             {{- end }}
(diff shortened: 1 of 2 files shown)


lengrongfu:
@archlitchi @wawa0210 PTAL

Issue #206, opened 2024-03-15
fix devices registry name error for Cambricon and Hygon this two type


lengrongfu:
In this code, the device map writes the key which is `Cambricon` data, but the key used in the place where it is used is indeed `MLU`, so I think this code has a bug.
https://github.com/Project-HAMi/HAMi/blob/58275e5b53bf8e541bf399f6a08865ab9216d722/pkg/device/devices.go#L47-L52

`devhandsk` var value is `Cambricon`,
https://github.com/Project-HAMi/HAMi/blob/58275e5b53bf8e541bf399f6a08865ab9216d722/pkg/scheduler/scheduler.go#L164

Init HandshakeAnnos map use key is `MLU`.
https://github.com/Project-HAMi/HAMi/blob/58275e5b53bf8e541bf399f6a08865ab9216d722/pkg/device/cambricon/device.go#L35

lengrongfu:
@archlitchi pleses confirm this is a bug?

PR #207, opened 2024-03-15
fix devices registry name error for Cambricon and Hygon this two type

lengrongfu:
Fixes: https://github.com/Project-HAMi/HAMi/issues/206


archlitchi:
yes, indeed, thanks

Issue #209, opened 2024-03-16
webhook api error


lengrongfu:
I found this error info in the scheduler pod.
```
I0316 06:02:50.508490       1 route.go:131] Into webhookfunc
[controller-runtime] log.SetLogger(...) was never called; logs will not be displayed.
Detected at:
	>  goroutine 281 [running]:
	>  runtime/debug.Stack()
	>  	/usr/local/go/src/runtime/debug/stack.go:24 +0x5e
	>  sigs.k8s.io/controller-runtime/pkg/log.eventuallyFulfillRoot()
	>  	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.16.3/pkg/log/log.go:60 +0xcd
	>  sigs.k8s.io/controller-runtime/pkg/log.(*delegatingLogSink).WithName(0xc0000c3280, {0x1a0a379, 0x9})
	>  	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.16.3/pkg/log/deleg.go:147 +0x45
	>  github.com/go-logr/logr.Logger.WithName(...)
	>  	/go/pkg/mod/github.com/go-logr/logr@v1.4.1/logr.go:345
	>  sigs.k8s.io/controller-runtime/pkg/webhook/admission.(*Webhook).getLogger.func1()
	>  	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.16.3/pkg/webhook/admission/webhook.go:182 +0x66
	>  sync.(*Once).doSlow(0x0?, 0xc000430190?)
	>  	/usr/local/go/src/sync/once.go:74 +0xbf
	>  sync.(*Once).Do(...)
	>  	/usr/local/go/src/sync/once.go:65
	>  sigs.k8s.io/controller-runtime/pkg/webhook/admission.(*Webhook).getLogger(0xc000561860, 0xc00079f000?)
	>  	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.16.3/pkg/webhook/admission/webhook.go:180 +0x4d
	>  sigs.k8s.io/controller-runtime/pkg/webhook/admission.(*Webhook).ServeHTTP(0xc000561860, {0x1c7d368?, 0xc0004d2d20}, 0xc000696000)
	>  	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.16.3/pkg/webhook/admission/http.go:96 +0xbab
[...]

PR #210, opened 2024-03-16
fix: init controller-runtime log.SetLogger()

lengrongfu:
Fixes: https://github.com/Project-HAMi/HAMi/issues/209


archlitchi:
yeah, i noticed that log, thanks for fixing:)

Issue #197, opened 2024-03-13
use can specify device UUID in pod resource


lengrongfu:
Current use can use set annotation`nvidia.com/use-gputype` or `nvidia.com/nouse-gputype`  to filter by card type, we hope use can specify device UUID to use specify device or not use specify device. 
We can add `nvidia.com/use-gpuuuid` or `nvidia.com/nouse-gpuuuid` to config.

PR #211, opened 2024-03-17
add use can specify use or not use device id feature

lengrongfu:
Fixes: https://github.com/Project-HAMi/HAMi/issues/197


archlitchi:
thanks a lot, please
1. add an example into /examples/nvidia folder
2. modify "device type specification" in README.md  introduction part to "device specification", describe this feature besides type specification

lengrongfu:
@archlitchi PTAL.

Issue #213, opened 2024-03-19
fix klog error


lengrongfu:
In this https://github.com/Project-HAMi/HAMi/pull/170 pr, we upgrade `klog` version from v1 to v2, but v2 version have error info in project build process,  error info as follows, so need fix it.

```shell
k8s.io/klog/v2.Infof call has arguments but no formatting directives
```

PR #214, opened 2024-03-19
fix klog upgrade after verify error

lengrongfu:
Fixes: https://github.com/Project-HAMi/HAMi/issues/213


chaunceyjiang (approved):
Thanks~

/lgtm

Issue #212, opened 2024-03-19
创建多个 container 的 pod 报错


chilianyi:
创建 pod 报错：

![企业微信截图_17107486711442](https://github.com/Project-HAMi/HAMi/assets/5917832/844a1c35-4b5e-45c5-b849-9ed11c850a99)

hami-scheduler 中 vgpu-scheduler-extender 容器报错：
```
2024-03-18T18:53:37.543856878+08:00 2024/03/18 10:53:37 http2: panic serving 127.0.0.1:49194: runtime error: index out of range [1] with length 1
2024-03-18T18:53:37.543869390+08:00 goroutine 5398 [running]:
2024-03-18T18:53:37.543873665+08:00 net/http.(*http2serverConn).runHandler.func1()
2024-03-18T18:53:37.543876731+08:00  /usr/local/go/src/net/http/h2_bundle.go:6186 +0x13b
2024-03-18T18:53:37.543879026+08:00 panic({0x1902880?, 0xc000ec17b8?})
2024-03-18T18:53:37.543881375+08:00  /usr/local/go/src/runtime/panic.go:914 +0x21f
2024-03-18T18:53:37.543884332+08:00 github.com/Project-HAMi/HAMi/pkg/scheduler.calcScore(0xc00009a2a0, 0xc0016fdc20?, {0xc000e4dc10, 0x2, 0xc0012a081c?}, 0x0?, 0x0?)
2024-03-18T18:53:37.543886755+08:00  /k8s-vgpu/pkg/scheduler/score.go:208 +0xaae
2024-03-18T18:53:37.543889394+08:00 github.com/Project-HAMi/HAMi/pkg/scheduler.(*Scheduler).Filter(0xc000458000, {0xc000d82900?, 0x0?, 0xc0016fdc20?})
2024-03-18T18:53:37.543898751+08:00  /k8s-vgpu/pkg/scheduler/scheduler.go:366 +0x525
2024-03-18T18:53:37.543901385+08:00 github.com/Project-HAMi/HAMi/pkg/scheduler/routes.PredicateRoute.func1({0x1c7d308, 0xc0001282b0}, 0xc000229800, {0x0?, 0x0?, 0x0?})
2024-03-18T18:53:37.543903715+08:00  /k8s-vgpu/pkg/scheduler/routes/route.go:59 +0x33b
[...]

Issue #198, opened 2024-03-13
在 kubeflow 1.18 中使用 HAMi 调度 GPU 


TerenceLiu98:
Version:
1. cluster version:  v1.28.2
2. kubeflow version: 1.18.0
3. Driver Version: 545.29.06    
4. CUDA Version: 12.3

在 HAMi 正常启动后安装 kubeflow， 但是无法正常启动 notebook with GPU，节点 description 和 logs 如下：

```
Name:               nexus
Roles:              <none>
Labels:             beta.kubernetes.io/arch=amd64
                    beta.kubernetes.io/os=linux
                    gpu=on
                    kubernetes.io/arch=amd64
                    kubernetes.io/hostname=nexus
                    kubernetes.io/os=linux
Annotations:        NVIDIA: Deleted_2024.03.13 14:41:58
                    csi.volume.kubernetes.io/nodeid: {"kadalu":"nexus"}
                    hami.io/node-handshake: Requesting_2024.03.13 14:41:43
                    hami.io/node-nvidia-register: GPU-f08602a4-7e07-b51d-6f9f-cd7289527ca2,10,12288,100,NVIDIA-NVIDIA GeForce GTX TITAN X,0,true:
                    kubeadm.alpha.kubernetes.io/cri-socket: unix:///var/run/containerd/containerd.sock
                    node.alpha.kubernetes.io/ttl: 0
                    volumes.kubernetes.io/controller-managed-attach-detach: true
CreationTimestamp:  Wed, 13 Mar 2024 13:51:50 +0000
Taints:             <none>
Unschedulable:      false
Lease:
  HolderIdentity:  nexus
  AcquireTime:     <unset>
  RenewTime:       Wed, 13 Mar 2024 14:41:55 +0000
Conditions:
  Type                 Status  LastHeartbeatTime                 LastTransitionTime                Reason                       Message
  ----                 ------  -----------------                 ------------------                ------                       -------
  NetworkUnavailable   False   Wed, 13 Mar 2024 13:52:32 +0000   Wed, 13 Mar 2024 13:52:32 +0000   CiliumIsUp                   Cilium is running on this node
  MemoryPressure       False   Wed, 13 Mar 2024 14:40:59 +0000   Wed, 13 Mar 2024 13:51:50 +0000   KubeletHasSufficientMemory   kubelet has sufficient memory available
[...]

PR #215, opened 2024-03-19
fix: when container nou user device resource index out of range

lengrongfu:
Fixes: https://github.com/Project-HAMi/HAMi/issues/212
Fixes: https://github.com/Project-HAMi/HAMi/issues/198

When multi-container doesn't have to use device resources,  fix the index out of range.


archlitchi:
thanks for fixing:)

PR #216, opened 2024-03-20
Optimize scheduling cache. Accelerate the discovery and deletion of nodes

chaunceyjiang:
1. Add a node informer. To notify the scheduler of cache updates. Faster discovery of new nodes and removal of old nodes.  
2. When there is no device on a node in the scheduling cache, remove it from the scheduling cache to avoid scheduling pods to these nodes.


archlitchi:
thanks for contributing:)

PR #217, opened 2024-03-20
fix: when multi container use device allocate fail

lengrongfu:
Fixes: https://github.com/Project-HAMi/HAMi/issues/212#issuecomment-2007150458

Kubelet is a loop to allocate GPU devices for multiple container requests using GPU resources, currently in `PodAllocationTrySuccess` method will check whether need release node lock, but `hami.io/devices-to-allocate` this annotation key is not use, so check failed, because current key is use `hami.io/vgpu-devices-to-allocate`.

### DIFF
--- a/pkg/device/devices.go
+++ b/pkg/device/devices.go
@@ -56,9 +56,13 @@ func init() {
 	DevicesToHandle = append(DevicesToHandle, iluvatar.IluvatarGPUCommonWord)
 }
 
-func PodAllocationTrySuccess(nodeName string, pod *v1.Pod) {
-	refreshed, _ := client.GetClient().CoreV1().Pods(pod.Namespace).Get(context.Background(), pod.Name, metav1.GetOptions{})
-	annos := refreshed.Annotations[util.AssignedIDsToAllocateAnnotations]
+func PodAllocationTrySuccess(nodeName string, devName string, pod *v1.Pod) {
+	refreshed, err := client.GetClient().CoreV1().Pods(pod.Namespace).Get(context.Background(), pod.Name, metav1.GetOptions{})
+	if err != nil {
+		klog.Errorf("get pods %s/%s error: %+v", pod.Namespace, pod.Name, err)
+		return
+	}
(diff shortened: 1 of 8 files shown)


pkg/util/types.go
archlitchi, line 1: please remove unused parameter here, "AssignedIDsToAllocateAnnotations = "hami.io/devices-to-allocate""

pkg/device/devices.go
archlitchi, line 1: since there may be multiple devices invoking this method, we need to change the parameter into 
func PodAllocationTrySuccess(nodeName string, devName string,pod *v1.Pod){
...
annos:=refreshed.Annotations[util.InRequestDevices[devName]]
}

instead of refreshed.Annotations["hami.io/vgpu-devices-to-allocate"]

and modify corresponding devs to invoke this method accordingly

pkg/util/types.go
lengrongfu: done

pkg/device/devices.go
lengrongfu: update code.

lengrongfu:
Test result: 
- test.yaml
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: gpu-pod
spec:
  containers:
    - name: ubuntu-container
      image: chrstnhntschl/gpu_burn
      args:
      - "6000"
      resources:
        limits:
          nvidia.com/gpu: 1
          nvidia.com/gpumem: 500
          nvidia.com/gpucores: 3
    - name: ubuntu-container1
      image: ubuntu:18.04
      command: ["bash", "-c", "sleep 86400"]
      resources:
        limits:
          nvidia.com/gpu: 1
          nvidia.com/gpumem: 1000
          nvidia.com/gpucores: 10
    - name: ubuntu-container2
      image: ubuntu:18.04
      command: ["bash", "-c", "sleep 86400"]
      resources:
        limits:
          nvidia.com/gpu: 1
          nvidia.com/gpumem: 2000
          nvidia.com/gpucores: 30
```

- run `kubectl apply -f test.yaml` result.
image

archlitchi:
Thanks :)

PR #225, opened 2024-03-22
add lint script & fix code that fails static checking

CoderTH:
Add a script for static checking and fix static check errors in the code

### DIFF
--- a/hack/util.sh
+++ b/hack/util.sh
@@ -1,0 +1,25 @@
+#!/usr/bin/env bash
+set -o errexit
+set -o nounset
+set -o pipefail
+
+MIN_Go_VERSION=1.21.0
+
+function util::cmd_exist {
+  local CMD=$(command -v ${1})
+  if [[ ! -x ${CMD} ]]; then
+    return 1
+  fi
+  return 0
+}
+
+function util::verify_go_version {
+    local go_version
+    IFS=" " read -ra go_version <<< "$(GOFLAGS='' go version)"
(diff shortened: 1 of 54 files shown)


CoderTH:
@lengrongfu @chaunceyjiang @calvin0327 @wawa0210 cc

I have added ci action for static checking and have been able to work properly, but I have only opened part of the linter. Because of the code history, there are many static checking errors. These bug fixes may affect the existing logic. I suggest that projects such as E2e test have relatively perfect unit tests and open them again after e2e check.

Please help review.

hack/verify-staticcheck.sh
```diff
@@ -0,0 +1,29 @@
...
+cd "${REPO_ROOT}"
+source "hack/util.sh"
+
+if util::cmd_exist golangci-lint ; then
+  echo "Using golangci-lint version:"
+  golangci-lint version
+else
+  echo "Installing golangci-lint ${GOLANGCI_LINT_VER}"
+  # https://golangci-lint.run/usage/install/#other-ci
+  curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin ${GOLANGCI_LINT_VER}
```
chaunceyjiang: I don't highly recommend using the code from the `master` branch, you should use a fixed version, such as `v1.57.1` that you defined above.

CoderTH: done

hack/util.sh
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env bash
+set -o errexit
+set -o nounset
+set -o pipefail
+
+MIN_Go_VERSION=go1.21.0
```
wawa0210: why "go1.12.0" not directly "1.21.0"

CoderTH: done

PR #295, opened 2024-05-06
fix gpu type matching mechanism,  Enable the combination of GPUInUse and GPUNoUse

coldzerofear:
At present, the selection of GPU type can be matched with device type by commenting `nvidia.com/use-gputype` and `nvidia.com/nouse-gputype` , similar to black and white lists,  But both cannot take effect simultaneously. 

The code uses substring matching as a filtering criterion，this may lead to users wanting to use A10 type GPUs, but ultimately being assigned to A100

To solve this problem, we can use `nvidia.com/use-gputype` and `nvidia.com/nouse-gputype` together

```yaml
metadata.:
  annotations:
    nvidia.com/use-gputype: A10
    nvidia.com/nouse-gputype: A100
```
This specifies the required device types and excludes unnecessary device types

### DIFF
--- a/pkg/device/nvidia/device.go
+++ b/pkg/device/nvidia/device.go
@@ -160,35 +160,33 @@ func (dev *NvidiaGPUDevices) MutateAdmission(ctr *corev1.Container) bool {
 }
 
 func checkGPUtype(annos map[string]string, cardtype string) bool {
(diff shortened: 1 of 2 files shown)


archlitchi:
That's great, thanks:)

PR #364, opened 2024-06-24
feat: Update webhook.yaml to whitelist kube-system and istio-system 

haitwang-cloud:
…amespaces

The code changes in `webhook.yaml` modify the namespace whitelist for the webhook. The `kube-system` and `istio-system` namespaces are added to the whitelist. This change aims to allow the webhook to function properly in these namespaces.

**What type of PR is this?**

The purpose of this change is to whitelist `kube-system` and `istio-system` namespaces, following [Kubernetes best practices for webhooks](https://kubernetes.io/zh-cn/docs/reference/access-authn-authz/extensible-admission-controllers/#avoiding-operating-on-the-kube-system-namespace). This prevents the Kubernetes system components and Istio components from being unnecessarily intercepted and possibly affected by our application's webhook. 
Fixes #

- https://kubernetes.io/zh-cn/docs/reference/access-authn-authz/extensible-admission-controllers/#avoiding-operating-on-the-kube-system-namespace
No

### DIFF
--- a/charts/hami/templates/scheduler/webhook.yaml
+++ b/charts/hami/templates/scheduler/webhook.yaml
@@ -24,6 +24,12 @@ webhooks:
         operator: NotIn
         values:
         - ignore
+      {{- if .Values.scheduler.customWebhook.whitelistNamespaces }}
+      - key: kubernetes.io/metadata.name
+        operator: NotIn
+        values:
+        {{- toYaml .Values.scheduler.customWebhook.whitelistNamespaces | nindent 10 }}
+      {{- end }}
     objectSelector:
       matchExpressions:
       - key: hami.io/webhook
--- a/charts/hami/values.yaml
+++ b/charts/hami/values.yaml
@@ -85,6 +85,11 @@ scheduler:
     host: 127.0.0.1 # hostname or ip, can be your node'IP if you want to use https://<nodeIP>:<schedulerPort>/<path>
     port: 31998
     path: /webhook
+    whitelistNamespaces:
+    # Specify the namespaces that the webhook will not be applied to.
+      # - default
+      # - kube-system
+      # - istio-system
   patch:
     image: docker.io/jettech/kube-webhook-certgen:v1.5.2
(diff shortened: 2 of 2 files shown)


charts/hami/templates/scheduler/webhook.yaml
```diff
@@ -20,10 +20,10 @@ webhooks:
     name: vgpu.hami.io
     namespaceSelector:
       matchExpressions:
-      - key: hami.io/webhook
+      - key: kubernetes.io/metadata.name
         operator: NotIn
         values:
-        - ignore
```
wawa0210, line 26: I suggest adding a check if whitelistNamespaces is empty and keeping the existing logic

At the same time, it is recommended to keep the `hami.io/webhook: ignore` logic, and you can add one or more matchExpressions

charts/hami/values.yaml
```diff
@@ -85,6 +85,9 @@ scheduler:
     host: 127.0.0.1 # hostname or ip, can be your node'IP if you want to use https://<nodeIP>:<schedulerPort>/<path>
     port: 31998
     path: /webhook
+    whitelistNamespaces:
```
wawa0210, line 88: I recommend leaving the default value empty. This is just an entry point for end users to customize.

charts/hami/templates/scheduler/webhook.yaml
haitwang-cloud: Done, add the empty check logic

charts/hami/values.yaml
haitwang-cloud: Done, updated it with a comment using the template.

haitwang-cloud:
Already addressed the latest comment, PTAL again @wawa0210

wawa0210:
Nice work~~ Can you rebase your 2 commits to 1 ?

haitwang-cloud:
@wawa0210 All done, PTAL

wawa0210:
> @wawa0210 All done, PTAL

thx

Issue #375, opened 2024-07-08
 Add feature to use the node selector in the scheduler to filter BM Nodes with GPUs


haitwang-cloud:
#### Description
In our cluster, we have both VM nodes and BM (Bare Metal) nodes. However, only the BM nodes have GPUs. Therefore, I would like to request a new feature that reads the current scheduler's node selector and uses this node selector to filter BM nodes.

#### Rationale
This feature will help optimize our scheduling by ensuring that tasks requiring GPUs are only scheduled on BM nodes that actually have GPUs, thereby improving resource allocation and performance.

#### Proposed Implementation
- Modify the scheduler to read the node selector configuration.
- Implement a filter mechanism that uses the node selector to identify BM nodes.
- Ensure that only BM nodes with GPUs are considered for tasks requiring GPU resources.
#### Benefits
- Improved scheduling efficiency.
- Better resource utilization.
- Enhanced performance for GPU-dependent tasks.

PR #354, opened 2024-06-14
feat: add liveness probe on scheduler extender

JasonHe-WQ:
**What type of PR is this?**

/kind feature

**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:
Fixes #353 

**Special notes for your reviewer**:
Test needed

**Does this PR introduce a user-facing change?**:
Nothing

PR #376, opened 2024-07-08
feat: Add GPU node selector to scheduler deployment

haitwang-cloud:
The code changes in `deployment.yaml` add the GPU node selector to the scheduler deployment. This change allows the scheduler to select nodes with the `gpu` label set to `"on"`. The environment variables are updated accordingly to include the `NODE_SELECTOR_GPU` variable.

**What type of PR is this?**

/kind feature

This PR introduces a new feature to enhance our scheduler by allowing it to filter BM (Bare Metal) nodes based on a node selector. This is particularly useful in our cluster environment where only BM nodes have GPUs, ensuring that tasks requiring GPUs are scheduled appropriately. 

Fixes # https://github.com/Project-HAMi/HAMi/issues/375

### DIFF
--- a/charts/hami/values.yaml
+++ b/charts/hami/values.yaml
@@ -40,11 +40,15 @@ global:
   gpuHookPath: /usr/local
   labels: {}
   annotations: {}
+  vmBmMixMode: false
+  vmBmMixModeSelector:
+    gpu: "on"
+
 
 scheduler:
   # @param nodeName defines the node name and the nvidia-vgpu-scheduler-scheduler will schedule to the node.
-  # if we install the nvidia-vgpu-scheduler-scheduler as default scheduler, we need to remove the k8s defualt
-  # scheduler pod from the cluster first, we must specified node name to skip the schedule workflow.
+  # if we install the nvidia-vgpu-scheduler-scheduler as default scheduler, we need to remove the k8s default
+  # scheduler pod from the cluster first, we must specify node name to skip the schedule workflow.
   nodeName: ""
   defaultMem: 0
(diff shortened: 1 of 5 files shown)


haitwang-cloud:
@wawa0210 @archlitchi

charts/hami/values.yaml
archlitchi, line 1: we'd better not add gpu:on selector here, because there are some clusters which don't have NVDIA GPU, user can add this field when they need.

pkg/scheduler/utils.go
archlitchi, line 1: have you tested what will happen if you don't specify .value.nodeSelector, or there is an empty .value.nodeSelector?

pkg/scheduler/utils_test.go
archlitchi, line 1: better add a test case with an empty nodeSelector here in Test_filterNodesBySelector

charts/hami/values.yaml
haitwang-cloud: Rename it to `vmBmMixMode`

haitwang-cloud:
There will be an extra env set in the `vgpu-scheduler-extender` is we enable the `vmBmMixMode`
```
        - name: vgpu-scheduler-extender
          image: projecthami/hami:v2.3.12
          imagePullPolicy: "IfNotPresent"
          env:
            - name: NODE_SELECTOR_GPU
              value: "on"
```

haitwang-cloud:
Fix the bug in this pull request: https://github.com/Project-HAMi/HAMi/pull/354, as it is causing the scheduler to restart.

PR #384, opened 2024-07-15
fix: CUDA_DISABLE_CONTROL not working properly

### LINKED ISSUE #383: env CUDA_DISABLE_CONTROL is not working properly
### 1. Issue or feature description

even setting CUDA_DISABLE_CONTROL=false,  HAMi will not take care this container.

which is not correct.

### 2. Reason

HAMi only check whether this env existed, but no care it's value.

### 2. What do we want:

if setting CUDA_DISABLE_CONTROL to false,  HAMi should Hijack CUDA for this container.

joy717:
**What type of PR is this?**

/kind bug

Fixes #383


lengrongfu:
Thanks~

PR #391, opened 2024-07-18
feat:Update k8s-device-plugin to v0.14.5 to Resolve nanoGPT Runtime Issue

haitwang-cloud:
**What type of PR is this?**

During an offline debugging session with @archlitchi , we identified that the current NVIDIA device plugin (v1.4.0) is causing compatibility issues with nanoGPT, preventing it from running properly. This issue persists even after setting CUDA_DISABLE_CONTROL to true and removing ld.so.preload from the GPU node.
We've confirmed that this problem also occurs in version 0.14.0 of the k8s-device-plugin. To resolve this, we need to update the k8s-device-plugin to at least version 0.14.5

/kind bug

This update is crucial for ensuring that our GPU resources are utilized effectively and that applications like nanoGPT can run without the encountered hindrances. 

Fixes # https://github.com/Project-HAMi/HAMi/issues/347

I can not build the HAMi in my local env, so I am not be able to run the E2E testing , could u plz help to build a test img with this PR?

No

### DIFF
--- a/pkg/device-plugin/nvidiadevice/nvinternal/cdi/cdi.go
+++ b/pkg/device-plugin/nvidiadevice/nvinternal/cdi/cdi.go
@@ -1,18 +1,36 @@
 /*
-Copyright 2024 The HAMi Authors.
+ * Copyright 2024 The HAMi Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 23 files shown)


haitwang-cloud:
Replace it with new PR https://github.com/Project-HAMi/HAMi/pull/855

PR #413, opened 2024-08-02
fix china goproxy.cn on github

wawa0210:
**What type of PR is this?**
/kind bug

GitHub does not need a China goproxy.  Using an proxy instead leads to abnormal operation. Fix this problem.

image

Fixes #


archlitchi:
fixed, related commit: https://github.com/Project-HAMi/HAMi/commit/38313a439860ce141677bc7eccbcab30261c7414

PR #422, opened 2024-08-08
fix: fix duplicate resource keys in configmap

devenami:
**What type of PR is this?**

/kind bug

fix duplicate resource keys in scheduler's configmap

Fixes #

No


archlitchi:
thanks:), that's a good fix:)

PR #489, opened 2024-09-11
Remake README.md

archlitchi:
The current HAMi Readme.md is not very good organized, i remake that to make it more clear


README.md
```diff
@@ -219,76 +126,8 @@ kubectl get pods -n kube-system
...
-</details>
-
-### Submit Task
-
-<details> <summary> Task example </summary>
-
-Containers can now request NVIDIA vGPUs using the `nvidia.com/gpu`` resource type.
-
-```
-apiVersion: v1
```
wawa0210, line 255: Would it be better if the quick start demo is retained? After the current changes, there is only the installation and introduction of hami, but no first example.

This yaml retention may give users a good first impression of hami, which is so easy to use.

archlitchi: I add a link to examples here for user to verify their installation.

PR #492, opened 2024-09-13
refactoring ascend device code

zoyopei:
**What type of PR is this?**

/kind feature

refactoring ascend device code

Fixes #


charts/hami/templates/scheduler/configmap.yaml
archlitchi, line 1: better rename to 'device-spec-configmap', we'll need to support other devices if necessary

charts/hami/ascend-config.yaml
```diff
@@ -0,0 +1,59 @@
+vnpus:
```
archlitchi: please change name to 'ascend'

Issue #491, opened 2024-09-13
Add filter specify Nvidia GPU register to node when device-plugin start


lengrongfu:
**What would you like to be added**:

**What type of PR is this?**

/kind feature

**What this PR does / why we need it**:

When there are multiple cards in a cluster, the user only wants to manage certain cards.

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

lengrongfu:
use charts/hami/templates/device-plugin/configmap.yaml config

lengrongfu:
done

PR #495, opened 2024-09-14
Feat: add filter device when register node by uuid or index

lengrongfu:
**What type of PR is this?**
/kind feature

When there are multiple cards in a cluster, the user only wants to manage certain cards.

Fixes # https://github.com/Project-HAMi/HAMi/issues/491

Users can customer decide whether device-plugin should filter certain devices, whether it is due to equipment failure or other reasons.


lengrongfu:
@wawa0210 @archlitchi PTAL

PR #496, opened 2024-09-14
fix: error happen when allocate iluvatar device 

lijm87:
**What type of PR is this?**
/kind bug

fix: error 'index out of range [1] with length 1' happen when allocate iluvatar device

Fixes #


wawa0210:
@lijm87 

This PR seems to have some problems on ut and needs to be fixed. It will be released today. Do you have time to follow up?

lijm87:
> @lijm87
> 
> This PR seems to have some problems on ut and needs to be fixed. It will be released today. Do you have time to follow up?

done

PR #593, opened 2024-11-04
fix: wrong quota for fitInDevices

### LINKED ISSUE #592: Wrong Calculate for 1 pod 2+ container with gpus
**What happened**:
when 1 pod with 2+ container request gpus, wrong score.device `used resources` calculated.

assume we have 2 gpu cards, each one has 10Gi memory, there is no task now.
1 pod with 2+ containers, and each container request a gpu (let's say it memory)
container1 request 10Gi memory
container2 request 1Gi memory

we want this result:
container1 got card1
container2 got card2

wrong result now:
container1 got card1
container2 got card1

root reason:
there is a `sort` function in `fitInDevices` function, which will change the order of node.Devices
but the result return by `fitInCertainDevice` records `the device index` before sorted, 
so after `fitInCertainDevice`,  we calculate `used`, `usedcores`, `usedmem`, then add them to a wrong device.

should got correct device when 1 pod with 2+ containers

we metioned above

**Environment**:
- HAMi version:
- Others:

joy717:
**What type of PR is this?**

/kind bug

fix wrong quota for fitInDevices

Fixes #592

### DIFF
--- a/pkg/scheduler/score.go
+++ b/pkg/scheduler/score.go
@@ -164,16 +164,21 @@ func fitInDevices(node *NodeUsage, requests util.ContainerDeviceRequests, annos 
 		fit, tmpDevs := fitInCertainDevice(node, k, annos, pod, devinput)
 		if fit {
 			for _, val := range tmpDevs[k.Type] {
-				total += node.Devices.DeviceLists[val.Idx].Device.Count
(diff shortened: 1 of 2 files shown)


archlitchi:
yes, thats' a nice fix:)

PR #623, opened 2024-11-16
Fix Kubernetes version string handling by stripping metadata

### LINKED ISSUE #621: InvalidImageName when deploying HAMi after latest update
**What happened**: I noticed that since the last update to 2.4.1, the helm chart has changed significantly, for instance no longer it's there the following fields: `scheduler.kubeScheduler.imageTag`, `devicePlugin.deviceMemoryScaling` or `devicePlugin.deviceSplitCount`. The worse outcome is that now when I deploy, I get an InvalidImageName error in the hami-vgpu-scheduler pod.

Here's the pod description:

`Failed to apply default image tag "registry.cn-hangzhou.aliyuncs.com/google_containers/kube-scheduler:v1.31.1+k3s1": couldn't parse image name "registry.cn-hangzhou.aliyuncs.com/google_containers/kube-scheduler:v1.31.1+k3s1": invalid reference format`

**What you expected to happen**: I expect the HAMi scheduler to be deployed as it was before the change. Also it would be good to have a guide on how to transition to the newer version (how do I specify device split count and memory scaling?), plus how to be able to stick to the old version (I've tried with helm version 2.4.0 but I see the same error).

**How to reproduce it (as minimally and precisely as possible)**: Run in a 1.31.1 kubernetes cluster the installer: 

`helm install hami-vgpu hami-charts/hami -n kalavai`

N/A

**Environment**:
- Kubernetes: k3s 1.31.1
- HAMi version: `2.4.1 and 2.4.0`
- Docker version from `docker version`: `27.3.1`
- Kernel version from `uname -a`: `Linux pop-os 6.9.3-76060903-generic #202405300957~1726766035~22.04~4092a0e SMP PREEMPT_DYNAMIC Thu S x86_64 x86_64 x86_64 GNU/Linux`

Nimbus318:
**What type of PR is this?**
/kind bug

This PR fixes an issue where Kubernetes version strings with additional metadata (e.g., `+k3s1`) caused problems in dynamically generated image tags. The new `strippedKubeVersion` function removes the part after the `+` in the Kubernetes version string, ensuring consistent and predictable image tagging.

Fixes #621

@archlitchi
@wawa0210

Yes, this PR restores compatibility for specifying certain parameters during `helm install/upgrade`. 

After version 2.4.1, users were unable to specify parameters like `scheduler.kubeScheduler.tag` (scheduler image version), `devicePlugin.deviceSplitCount`, `devicePlugin.deviceMemoryScaling`, and `devicePlugin.deviceCoreScaling` directly during Helm operations. 

With this PR merged:
- Users can now specify the scheduler image version using `scheduler.kubeScheduler.tag` during `helm install` or `helm upgrade`. If the parameter is not provided, the image version will continue to be auto-generated based on the Kubernetes version.
- Parameters like `deviceSplitCount`, `deviceMemoryScaling`, and `deviceCoreScaling` are also restored for direct specification during Helm operations.

### DIFF
--- a/charts/hami/templates/_helpers.tpl
+++ b/charts/hami/templates/_helpers.tpl
@@ -85,3 +85,24 @@ Image registry secret name
 imagePullSecrets: {{ toYaml .Values.imagePullSecrets | nindent 2 }}
 {{- end }}
 
+{{/*
+    Resolve the tag for kubeScheduler.
+*/}}
+{{- define "resolvedKubeSchedulerTag" -}}
+{{- if .Values.scheduler.kubeScheduler.imageTag }}
+{{- .Values.scheduler.kubeScheduler.imageTag | trim -}}
+{{- else }}
+{{- include "strippedKubeVersion" . | trim -}}
+{{- end }}
+{{- end }}
+
+
+{{/*
+    Remove the part after the `+` in the Kubernetes version string.
+    v1.31.1+k3s1 -> v1.31.1
+    v1.31.1 -> v1.31.1
+*/}}
+{{- define "strippedKubeVersion" -}}
+{{- $parts := split "+" .Capabilities.KubeVersion.Version -}}
+{{- print $parts._0 -}}
+{{- end -}}
--- a/charts/hami/templates/scheduler/deployment.yaml
+++ b/charts/hami/templates/scheduler/deployment.yaml
@@ -33,7 +33,7 @@ spec:
       containers:
       {{- if .Values.scheduler.kubeScheduler.enabled }}
         - name: kube-scheduler
-          image: {{ .Values.scheduler.kubeScheduler.image }}:{{ .Capabilities.KubeVersion.Version }}
+          image: "{{ .Values.scheduler.kubeScheduler.image }}:{{ include "resolvedKubeSchedulerTag" . }}"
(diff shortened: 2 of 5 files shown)


wawa0210:
Very good bug fix, have we done some testing? For example, can it work properly in the K3s environment?

wawa0210 (approved):

Nimbus318:
I have rebased the commits into one as requested.

Regarding testing, I have not tested this in a real K3s environment. However, I conducted mock testing for this function locally by injecting a mock Kubernetes version value in _helpers.tpl for validation. Below are the steps I followed:

Mock the Kubernetes version by temporarily injecting a value for testing:

```yaml
{{/*
    Mock Kubernetes version for testing
*/}}
{{- define "mockedCapabilitiesKubeVersion" -}}
v1.31.1+k3s1
{{- end }}
```
Modify the existing logic to replace Capabilities.KubeVersion.Version with the mock value:

```yaml
{{- define "strippedKubeVersion" -}}
{{- $parts := split "+" (include "mockedCapabilitiesKubeVersion" .) -}}
{{- print $parts._0 -}}
{{- end }}
```
Validated that the output correctly strips the metadata after +.

PR #624, opened 2024-11-18
Update vGPUmonitor to add dynamic adjustment on core and memory limit

archlitchi:
/kind feature

add dynamic adjustment API used for future works


kosumic:
Should we make the  Go assignments to shared memory atomic?

Even if we don't need to worry about data tearing, invalidating caches is good.

Issue #658, opened 2024-11-29
HAMI2.4.1 带有gpu的pod运行一段时间后莫名的就会出现无法使用gpu的情况，重启pod之后就会恢复正常


linsistqb:
一、环境
    hami版本  2.4.1 和 2.3.12 
    k8s 版本  1.22.12 docker方式
二、日志及问题描述
1.hami-device-plugin device-plugin日志如下
      "failed to get numa information" err="exit status 255" idx=0
        nvml registered device id=1, memory=15360, type=Tesla T4, numa=0
2.正常时候pod里面nvidia-smi是正常的，一段时间之后pod里面再执行nvidia-smi 报错 Failed to initialize NVML: Unknown Error
3.有两个集群都是这种情况，几乎是必现

linsistqb:
hami-device-plugin  device-plugin日志： "failed to get numa information" err="exit status 255" idx=0

linsistqb:
hami-device-plugin  device-plugin日志如下
 "failed to get numa information" err="exit status 255" idx=0
 nvml registered device id=1, memory=15360, type=Tesla T4, numa=0

leeyuejun:
一、环境
hami版本: hami-2.4.1    projecthami/hami:v2.4.1
k8s with docker, nvidia as runtime版本: v1.26.15
docker:  Server Version: 20.10.24
nvidia-Driver Version: 550.67
OS: Ubuntu 22.04 LTS
linux kernel: 5.15.0-112-generic

二、日志及问题描述
遇见相同的问题 "register.go:148] "failed to get numa information" err="exit status 255" idx=1"
在报错后, pod中的任务就失败了,nvidia-smi也不能起来了; "Failed to initialize NVML: Unknown Error"
但是重新创建的pod可以运行:

> 
hami-device-plugin log: 
1129 08:29:58.044268  639047 register.go:197] Successfully registered annotation. Next check in 30s seconds...
2024-11-29T08:30:28.046041214Z I1129 08:30:28.045863  639047 register.go:132] MemoryScaling= 1 registeredmem= 40960
I1129 08:30:28.378554  639047 register.go:160] nvml registered device id=1, memory=40960, type=NVIDIA A100-SXM4-40GB, numa=0
I1129 08:30:28.378693  639047 register.go:132] MemoryScaling= 1 registeredmem= 40960
[...]

PR #671, opened 2024-12-03
set ENV PASS_DEVICE_SPECS to device-plugin

jingzhe6414:
…PECS=true as an environment variable.

**What type of PR is this?**
bug

Mitigate the issue of GPU unavailability during the pod runtime.
Fixes #
[https://github.com/Project-HAMi/HAMi/issues/658](url)Containerized GPU workloads may suddenly lose access to their GPUs. This situation occurs when systemd is used to manage the cgroups of the container and it is triggered to reload any Unit files that have references to NVIDIA GPUs (e.g. with something as simple as a systemctl daemon-reload).
[https://github.com/NVIDIA/gpu-operator/issues/485](url)According to the solution provided in the k8s-nvidia-plugin, adding the PASS_DEVICE_SPECS ENV to the plugin was tested and found that not setting privileged can also mitigate the issue.
**Does this PR introduce a user-facing change?**:  yes

### DIFF
--- a/cmd/vGPUmonitor/metrics.go
+++ b/cmd/vGPUmonitor/metrics.go
@@ -18,7 +18,6 @@ package main
 
 import (
 	"fmt"
-	"github.com/Project-HAMi/HAMi/pkg/util"
 	"log"
 	"net/http"
 	"os"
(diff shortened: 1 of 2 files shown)


charts/hami/templates/device-plugin/daemonsetnvidia.yaml
```diff
@@ -63,6 +63,10 @@ spec:
               value: all
             - name: HOOK_PATH
               value: {{ .Values.global.gpuHookPath }}
+            {{- if typeIs "bool" .Values.devicePlugin.compatWithCPUManager }}
+            - name: PASS_DEVICE_SPECS
```
lengrongfu: Because hami no implement read this env PASS_DEVICE_SPECS logic, so even if add it after code is invalid.

jingzhe6414: Yes, I also had this issue at the beginning, including the fact that the nvidia-k8s-device plugin only recognized PASS_DEVICE_SPECS but did not use it. However, adding PASS_DEVICE_SPECS env is effective, perhaps due to compatibility with nvidia-k8s-device plugin causing it to take effect?

PR #681, opened 2024-12-05
fix: scheduler ignore KUBECONFIG env even if this environment variable is set

Shouren:
**What type of PR is this?**
/kind bug

The implementation of `NewClient` function in `pkg/k8sutil/client.go` gets `*restclient.Config` from `InClusterConfig` api first even if the `KUBECONFIG` env has been set which is out of expect. This PR fix it.


pkg/scheduler/scheduler.go
```diff
@@ -125,7 +126,7 @@ func (s *Scheduler) onDelPod(obj interface{}) {
 }
 
 func (s *Scheduler) Start() {
-	kubeClient, err := k8sutil.NewClient()
+	kubeClient, err := client.NewClient()
```
lengrongfu: you can use `client.GetClient()` to get kubeClient, because this kubeClient having init function Initialization completed.

lengrongfu:
suggest title: fix: scheduler ignore KUBECONFIG env even if this environment variable is set

pkg/scheduler/scheduler.go
Shouren: Good idea. By doing so, some the unused code can also be removed

lengrongfu:
@Shouren can you rebase main branch? because having resolve code lint question.

Shouren:
> @Shouren can you rebase main branch? because having resolve code lint question.

All right, rebase is done.

Issue #569, opened 2024-10-24
[Call to action] UT coverage improvement task


wawa0210:
**Task description**:

Currently, the overall test coverage of the HAMi project is 26.9%, which is far from enough. Interested partners are urgently needed to contribute to improve the ut coverage.

| go file      | ut coverage |
| ----------- | ----------- |
| pkg/scheduler/scheduler.go | 28.30% |
| pkg/util/util.go | 29.45% |
| pkg/device/cambricon/device.go | 13.43% |
| pkg/device/ascend/device.go | 2.92% |
| pkg/device/nvidia/device.go | 31.58% |
| pkg/monitor/nvidia/cudevshr.go | 0.00% |
| pkg/device/mthreads/device.go | 14.47% |
| pkg/device/hygon/device.go | 15.17% |
| pkg/util/nodelock/nodelock.go | 0.00% |
| pkg/device/iluvatar/device.go | 29.85% |
| pkg/scheduler/routes/route.go | 0.00% |
| pkg/monitor/nvidia/v1/spec.go | 0.00% |
| pkg/monitor/nvidia/v0/spec.go | 0.00% |
| pkg/device/devices.go | 27.14% |
| pkg/scheduler/score.go | 67.05% |
| pkg/oci/spec.go | 0.00% |
| pkg/oci/spec_mock.go | 0.00% |
| pkg/scheduler/webhook.go | 46.81% |
| pkg/scheduler/event.go | 44.83% |
| pkg/k8sutil/client.go | 0.00% |
| pkg/scheduler/nodes.go | 67.86% |
| pkg/device/ascend/vnpu.go | 0.00% |
| pkg/util/client/client.go | 54.17% |
| pkg/scheduler/pods.go | 88.89% |
| pkg/version/version.go | 0.00% |
| pkg/scheduler/policy/gpu_policy.go | 83.87% |
| pkg/k8sutil/pod.go | 76.47% |
| pkg/scheduler/policy/constant.go | 100.00% |
| pkg/scheduler/policy/node_policy.go | 100.00% |
| pkg/oci/runtime_exec.go | 100.00% |
| pkg/oci/runtime_mock.go | 100.00% |

![image](https://github.com/user-attachments/assets/39c1d9ff-2b3c-4826-beae-599345cefe2d)

**Solution**:

**Who can join or take the task**:

The good first issue is intended for `first-time contributors` to get started on his/her contributor journey.

After a contributor has successfully completed 1-2 good first issue's,
[...]

PR #697, opened 2024-12-10
random time may be zero

shijinye:
**What type of PR is this?**
/kind bug

1、fix When i is 0, it will cause the following error

`panic: invalid argument to Intn [recovered]
        panic: invalid argument to Intn
`

2、add ut for the device/cambricon（setNodeLock、LockNode、ReleaseNodeLock）

Fixes # https://github.com/Project-HAMi/HAMi/issues/569


wawa0210:
/lgtm 

thanks your contribution

PR #763, opened 2024-12-30
fix get device ids when a specific number of devices are to be replicated

KubeKyrie:
**What type of PR is this?**
/kind bug

When a specific number of devices for this resource type are to be replicated, every time we get device ids different. Because the traversal map is unordered.

Fixes #

### DIFF
--- a/pkg/device-plugin/nvidiadevice/nvinternal/rm/device_map.go
+++ b/pkg/device-plugin/nvidiadevice/nvinternal/rm/device_map.go
@@ -34,6 +34,7 @@ package rm
 
 import (
 	"fmt"
+	"sort"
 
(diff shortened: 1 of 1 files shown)


wawa0210:
@lengrongfu Can you help review this pr ?

lengrongfu:
There is no logical problem with the code, but I still do not recommend modifying it.
1. The code in the `pkg/device-plugin` directory comes from the repository `https://github.com/NVIDIA/k8s-device-plugin`. The upstream logic is also handled in this way. Do not modify the code in this directory unless necessary to facilitate subsequent tracking of the upstream code.
2. In the current logic, the order of maps does not affect the results.

Mainly because of the first point I do not recommend making any changes.

PR #767, opened 2025-01-02
feat: Restart/Upgrade device-plugin will not affect services.

chaunceyjiang:
**What type of PR is this?**

The most important reason that restarts affect business is that libvgpu.so is rewritten, leading to the application loading a partially written .so file. Currently, an MD5 calculation is used to determine if the file is the same, and it is not rewritten if it is different.

Fixes #

Test:

OCI config:

image

Pod: 

image

### DIFF
--- a/pkg/device-plugin/nvidiadevice/nvinternal/info/version.go
+++ b/pkg/device-plugin/nvidiadevice/nvinternal/info/version.go
@@ -57,3 +57,8 @@ func GetVersionString(more ...string) string {
 	v := append(GetVersionParts(), more...)
 	return strings.Join(v, "\n")
 }
+
+// GetVersion returns the version of the binary.
+func GetVersion() string {
+	return version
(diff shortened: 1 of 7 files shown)


pkg/device-plugin/nvidiadevice/nvinternal/info/version.go
```diff
@@ -57,3 +57,8 @@ func GetVersionString(more ...string) string {
 	v := append(GetVersionParts(), more...)
 	return strings.Join(v, "\n")
 }
+
+// GetVersion returns the version of the binary.
+func GetVersion() string {
+	return version
```
chaunceyjiang, line 63: This version has already been injected at compile time.

https://github.com/Project-HAMi/HAMi/blob/master/Makefile#L38

Mrxuexi (approved):

Issue #793, opened 2025-01-09
optimize MetricsBindAddress to MetricsBindPort


phoenixwu0229:
**What happened**:
The targetPort specified in the service.yaml is hardcoded, while the metricsBindAddress is configurable, leading to inconsistent behavior. Additionally, the metrics interface should ideally be exposed at 0.0.0.0, so flexible configuration may not be necessary. Therefore, it is recommended to optimize this to metricsBindPort.

**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
hami 2.4.1

PR #796, opened 2025-01-09
optimize MetricsBindAddress to MetricsBindPort

phoenixwu0229:
What type of PR is this?

/kind bug

The targetPort specified in the service.yaml is hardcoded, while the metricsBindAddress is configurable, leading to inconsistent behavior. Additionally, the metrics interface should ideally be exposed at 0.0.0.0, so flexible configuration may not be necessary. Therefore, it is recommended to optimize this to metricsBindPort.

I have successfully tested it locally
image

Which issue(s) this PR fixes:
Fixes https://github.com/Project-HAMi/HAMi/issues/793

### DIFF
--- a/charts/hami/templates/scheduler/service.yaml
+++ b/charts/hami/templates/scheduler/service.yaml
@@ -21,7 +21,7 @@ spec:
       protocol: TCP
     - name: monitor
       port: {{ .Values.scheduler.service.monitorPort }}
-      targetPort: 9395
+      targetPort: {{ (split ":" (printf "%s" .Values.scheduler.metricsBindAddress))._1 }}
       nodePort: {{ .Values.scheduler.service.monitorPort }}
       protocol: TCP
   selector:


phoenixwu0229:
I have reopen a pr, the pr before have some problems

cmd/scheduler/main.go
```diff
@@ -61,7 +61,7 @@ func init() {
 	rootCmd.Flags().Int32Var(&config.DefaultResourceNum, "default-gpu", 1, "default gpu to allocate")
 	rootCmd.Flags().StringVar(&config.NodeSchedulerPolicy, "node-scheduler-policy", util.NodeSchedulerPolicyBinpack.String(), "node scheduler policy")
 	rootCmd.Flags().StringVar(&config.GPUSchedulerPolicy, "gpu-scheduler-policy", util.GPUSchedulerPolicySpread.String(), "GPU scheduler policy")
-	rootCmd.Flags().StringVar(&config.MetricsBindAddress, "metrics-bind-address", ":9395", "The TCP address that the scheduler should bind to for serving prometheus metrics(e.g. 127.0.0.1:9395, :9395)")
```
wawa0210: This change will cause problems. Previously, MetricsBindAddress allowed users to pass the entire address (address+port) as a parameter, not just the port.

For example, users do not want MetricsBindAddress to be exposed as "0.0.0.0", which will bring other risks, security, etc.

cmd/scheduler/metrics.go
```diff
@@ -257,5 +258,7 @@ func initMetrics(bindAddress string) {
 	NewClusterManager("vGPU", reg)
 
 	http.Handle("/metrics", promhttp.HandlerFor(reg, promhttp.HandlerOpts{}))
-	log.Fatal(http.ListenAndServe(bindAddress, nil))
+	addr := net.JoinHostPort("0.0.0.0", bindPort)
```
wawa0210: Same question as above, it cannot be hardcoded to "0.0.0.0"

charts/hami/templates/scheduler/deployment.yaml
```diff
@@ -84,7 +84,7 @@ spec:
             - --cert_file=/tls/tls.crt
             - --key_file=/tls/tls.key
             - --scheduler-name={{ .Values.schedulerName }}
-            - --metrics-bind-address={{ .Values.scheduler.metricsBindAddress }}
```
wawa0210: I don't understand what the specific problem is with the current setup. If you want to flexibly adjust MetricsBindAddress, just change values.yaml metricsBindAddress

charts/hami/templates/scheduler/service.yaml
```diff
@@ -21,7 +21,7 @@ spec:
       protocol: TCP
     - name: monitor
       port: {{ .Values.scheduler.service.monitorPort }}
-      targetPort: 9395
```
wawa0210, line 24: There is a bug here. We can split .Values.scheduler.metricsBindAddress and get the port information after ":"

charts/hami/templates/scheduler/deployment.yaml
phoenixwu0229: image
The targetPort specified in the service.yaml is hardcoded, while the metricsBindAddress is configurable

wawa0210: I understand. This is a bug. The configuration information is inconsistent. At this time, the targetport can be obtained by splitting the `.Values.scheduler.metricsBindAddress` information.

charts/hami/templates/scheduler/service.yaml
phoenixwu0229: i can fix 0.0.0.0 hardcoded problem, please wait

wawa0210: The metricsBindAddress declaration method is a standard CNCF ecosystem practice and there is no particular reason to change it.

phoenixwu0229: Then I'll narrow the scope of the modifications to only change the Helm chart, and I won't modify the startup parameters.

phoenixwu0229: please review again

phoenixwu0229: @wawa0210 please review again

wawa0210: image

wawa0210: We can reuse  `.Values.scheduler.metricsBindAddress` directly instead of defining it again

charts/hami/templates/scheduler/deployment.yaml
phoenixwu0229: > I understand. This is a bug. The configuration information is inconsistent. At this time, the targetport can be obtained by splitting the `.Values.scheduler.metricsBindAddress` information.

@wawa0210 

I have modified it to use only metricsBindAddress, while also considering the scenario where metricsBindAddress does not include an IP address. Therefore, I use _1 to index the last element. Please review.

If you need any further assistance or adjustments, feel free to let me know! 🌟😊

wawa0210: Very timely follow-up, thumbs up for you, I have no other suggestions

ps, it would be better if you could post the test results of different scenarios

phoenixwu0229: > Very timely follow-up, thumbs up for you, I have no other suggestions
> 
> ps, it would be better if you could post the test results of different scenarios

image

wawa0210:
/lgtm

thanks your contribution

wawa0210:
ignore e2e fail test

PR #799, opened 2025-01-10
refactor: update service configurations for device plugin and scheduler

haitwang-cloud:
**What type of PR is this?**

/kind design

### Dynamic Configuration:

- Ports and Service types are now dynamically configured using values.yaml, making the charts more flexible and easier to customize.
- Default values are provided for httpPort, monitorPort, and nodePort to ensure backward compatibility.

### Removed Hardcoded Values:

- Hardcoded values for ports and nodePorts have been replaced with dynamic references to values.yaml.

### Improved Readability:

- Added comments to explain the purpose of each section, making the templates easier to understand and maintain.

These changes are necessary to make the Helm charts more adaptable to different deployment scenarios and easier to maintain in the long term
Fixes #

This PR focuses on optimizing the Service configurations in the Helm charts without changing the core functionality.

Yes, this PR introduces the following user-facing changes:

Users can now customize Service types (`NodePort` or `ClusterIP`) and ports via values.yaml.

The nodePort field is only set when the Service type is `NodePort`, avoiding unnecessary configurations for ClusterIP Services.

### DIFF
--- a/charts/hami/values.yaml
+++ b/charts/hami/values.yaml
@@ -119,9 +119,10 @@ scheduler:
     tolerations: []
     runAsUser: 2000
   service:
-    httpPort: 443
-    schedulerPort: 31998
-    monitorPort: 31993
+    type: NodePort  # Default type is NodePort, can be changed to ClusterIP
+    httpPort: 443   # HTTP port
+    schedulerPort: 31998  # NodePort for HTTP
+    monitorPort: 31993    # Monitoring port
     labels: {}
     annotations: {}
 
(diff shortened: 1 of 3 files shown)


charts/hami/values.yaml
```diff
@@ -119,9 +119,10 @@ scheduler:
     tolerations: []
     runAsUser: 2000
   service:
-    httpPort: 443
-    schedulerPort: 31998
-    monitorPort: 31993
+    type: ClusterIP  # Default type is NodePort, can be changed to ClusterIP
```
haitwang-cloud: Change default 2 ClusterIP

charts/hami/values.yaml
```diff
@@ -141,7 +142,10 @@ devicePlugin:
     - -v=4
   
   service:
+    type: ClusterIP  # Default type is NodePort, can be changed to ClusterIP
```
haitwang-cloud: Change default 2 ClusterIP

Issue #791, opened 2025-01-08
我无法在我的EKS集群上调度我的GPU


cbwcbwcbw:
EKS version: 1.24.17
hami version: 2.4.1
我的scheduler和device都是正常运行的
```bash
NAMESPACE     NAME                              READY   STATUS    RESTARTS   AGE
default       cbwtesteip                        1/1     Running   0          148m
kube-system   aws-node-hg8tv                    2/2     Running   0          3h21m
kube-system   coredns-f556495c8-99r79           1/1     Running   0          6d2h
kube-system   coredns-f556495c8-tsrhx           1/1     Running   0          6d2h
kube-system   eks-pod-identity-agent-fd6fn      1/1     Running   0          3h21m
kube-system   gpu-test-cbw2                     0/1     Pending   0          4m42s
kube-system   hami-device-plugin-jqgj8          2/2     Running   0          35m
kube-system   hami-scheduler-864d9fdd5b-4wt6j   2/2     Running   0          9m1s
kube-system   kube-proxy-lk8bk                  1/1     Running   0          3h21m
```
而且没有错误日志，测试调用gpu的pod也没有event，节点也没有调度
apiserver log 也没有报错
现在就是没有任何有用信息能继续调查

scheduler日志
```
Defaulted container "kube-scheduler" out of: kube-scheduler, vgpu-scheduler-extender
I0108 09:48:08.816767       1 flags.go:64] FLAG: --add-dir-header="false"
I0108 09:48:08.816836       1 flags.go:64] FLAG: --allow-metric-labels="[]"
I0108 09:48:08.816844       1 flags.go:64] FLAG: --alsologtostderr="false"
I0108 09:48:08.816848       1 flags.go:64] FLAG: --authentication-kubeconfig=""
I0108 09:48:08.816852       1 flags.go:64] FLAG: --authentication-skip-lookup="false"
I0108 09:48:08.816857       1 flags.go:64] FLAG: --authentication-token-webhook-cache-ttl="10s"
I0108 09:48:08.816864       1 flags.go:64] FLAG: --authentication-tolerate-lookup-failure="true"
I0108 09:48:08.816867       1 flags.go:64] FLAG: --authorization-always-allow-paths="[/healthz,/readyz,/livez]"
I0108 09:48:08.816876       1 flags.go:64] FLAG: --authorization-kubeconfig=""
[...]

Issue #744, opened 2024-12-26
can not install HAMi on AWS EKS


cataglyphis:
**Please provide an in-depth description of the question you have**:

**What do you think about this question?**:

**Environment**:
- HAMi version: 2.4.1
- Kubernetes version:  v1.29.11-eks-56e63d8
- Others:

```
helm install hami hami-charts/hami --set scheduler.kubeScheduler.imageTag=v1.29.11 -n hami
Error: INSTALLATION FAILED: chart requires kubeVersion: >= 1.16.0 which is incompatible with Kubernetes v1.29.11-eks-56e63d8
```

elrondwong:
The version `v1.29.11-eks-56e63d8` is not a standard semantic version, so it cannot be recognized by Helm. To resolve this issue, there are two approaches:

- Directly remove or comment out the `kubeVersion` field in `charts/hami/Chart.yaml`

```yaml
# kubeVersion: ">= 1.16.0"
```

- Use `--set kubeVersionOverride=true` when installing the Helm Chart to bypass the Kubernetes version check.

```yaml
helm install hami hami-charts/hami --set kubeVersionOverride=true
```

cataglyphis:
> The version `v1.29.11-eks-56e63d8` is not a standard semantic version, so it cannot be recognized by Helm. To resolve this issue, there are two approaches:
> 
> * Directly remove or comment out the `kubeVersion` field in `charts/hami/Chart.yaml`
> 
> ```yaml
> # kubeVersion: ">= 1.16.0"
> ```
> 
> * Use `--set kubeVersionOverride=true` when installing the Helm Chart to bypass the Kubernetes version check.
> 
> ```yaml
> helm install hami hami-charts/hami --set kubeVersionOverride=true
> ```

Since there is no user-managed kube-scheduler component in AWS EKS clusters, should I install the HAMi kube-scheduler or only the vgpu-scheduler-extender?

elrondwong:
I have not installed HAMI on AWS EKS before. Normally, both components need to be installed:

- kube-scheduler will register a scheduler extender with the cluster.
- vgpu-scheduler-extender handles the actual scheduling requests.

cataglyphis:
> The version `v1.29.11-eks-56e63d8` is not a standard semantic version, so it cannot be recognized by Helm. To resolve this issue, there are two approaches:
> 
> * Directly remove or comment out the `kubeVersion` field in `charts/hami/Chart.yaml`
> 
> ```yaml
[...]

PR #845, opened 2025-01-28
Fix: Update handling of version strings in Helm template and helpers.tpl

HJJ256:
**What type of PR is this?**
/kind bug

The current master branch was not able to handle version strings like **v1.30.8-eks-2d5f260** which is prevalent for AWS EKS. This causes HAMi charts installation failure as the minor version checking condition defaults to false and old parameters are used.
Also currently, scheduler imageTag needs to be set every time we have versions like **v1.30.8-eks-2d5f260**. However changing the definition of **strippedKubeVersion** in _helpers.tpl to a more general regexReplaceAll command solves this while being backwards compatible.

Fixes #
https://github.com/Project-HAMi/HAMi/issues/744
https://github.com/Project-HAMi/HAMi/issues/791
**Special notes for your reviewer**: This PR will help Project HAMi to get deployed on various cloud-based Kubernetes clusters.

### DIFF
--- a/charts/hami/templates/_helpers.tpl
+++ b/charts/hami/templates/_helpers.tpl
@@ -98,11 +98,11 @@ imagePullSecrets: {{ toYaml .Values.imagePullSecrets | nindent 2 }}
 
 
 {{/*
-    Remove the part after the `+` in the Kubernetes version string.
+    Return the stripped Kubernetes version string by removing extra parts after semantic version number.
     v1.31.1+k3s1 -> v1.31.1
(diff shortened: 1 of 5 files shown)


HJJ256:
@archlitchi @wawa0210 any timeline on review and merging of this PR?

archlitchi:
thanks, a nice fix, i assume you have tested on your AWS cluster?

HJJ256:
> thanks, a nice fix, i assume you have tested on your AWS cluster?

Thanks. Yes, that's correct

PR #873, opened 2025-02-16
enhancement: use placeholder for node name in device plugin ConfigMap template

### LINKED ISSUE #811: deviceSplitCount not honored
Context:

We are using NVIDIA GPUs and are trying to limit 2 pods max per GPU. Reading through the documentation, we've learned this should be possible with the deviceSplitCount parameter. After following the guide, changing configmaps, restarting containers, the parameter is not honoured (though being correctly modified). 

Issue:

We are able to schedule more than 2 pods. On the other hand if we try to schedule more than 10 pods on a GPU, it also sticks to the default value of 10. 

https://github.com/Project-HAMi/HAMi/blob/master/charts/hami/templates/scheduler/device-configmap.yaml#L23C7-L23C23

Nimbus318:
**What type of PR is this?**
/kind enhancement

Several users have reported issues where the device plugin configuration doesn't take effect because they didn't modify the default node name in the ConfigMap. This PR replaces the specific node name with a placeholder to make it more obvious that users need to replace it with their actual node name.

Changes:
- Changed the default node name from `m5-cloudinfra-online02` to `<your-node-name>` in the device plugin ConfigMap template to better indicate that this value needs to be replaced

Fixes #811

Yes. Users will see a more intuitive placeholder in the ConfigMap template that clearly indicates they need to replace it with their actual node name.

### DIFF
--- a/charts/hami/templates/device-plugin/configmap.yaml
+++ b/charts/hami/templates/device-plugin/configmap.yaml
@@ -10,7 +10,7 @@ data:
     {
         "nodeconfig": [
             {
-                "name": "m5-cloudinfra-online02",
+                "name": "<your-node-name>",
                 "operatingmode": "hami-core",
                 "devicememoryscaling": 1.8,
                 "devicesplitcount": 10,


lengrongfu:
https://github.com/Project-HAMi/HAMi/pull/493 
I dealt with this issue a long time ago. my idea is to delete this useless configuration directly.

FouoF:
/lgtm
cc @archlitchi

PR #889, opened 2025-02-20
fix parseNvidiaNumaInfo index out of range

flpanbin:
**What type of PR is this?**

/kind bug

This PR fixes a potential array index out of bounds issue in the `parseNvidiaNumaInfo` function. The issue occurs when accessing NUMA Affinity data for the target GPU where the row has fewer columns than expected.

Current code assumes the NUMA Affinity column exists in every row, but when processing the target GPU's data, if this column is missing, it will cause an index out of bounds panic.

The fix adds proper bounds checking before accessing the NUMA Affinity value, making the NVIDIA device plugin more robust when handling various topology output formats.

Fixes #


archlitchi:
a nice fix, thanks:)

PR #938, opened 2025-03-13
Separate options from client to make the responsibility more clear.

yangshiqi:
Remove the magic number in the main function and define it as a constant.

**What type of PR is this?**

/kind feature

add kubeclient timeout parameter.
In our scenarios, biz logic needs to be coordinated with multi-node scheduling. When the nodes' number is large, the kube API's timeout must be controlled to prevent it from being stuck.

Fixes # nope

Depends on PR #681
Base on this pr, I merged my version and splited the params according to the option feature of k8s, to separate the options-related extensions(maybe more like request/transports) from the client. I also isolated the default params in client, to prepare for the future needs of hami, and avoided the magic numbers.

nope


ouyangluwei163:
image

Burst/QPS 未生效

PR #963, opened 2025-03-28
feat: Add support for profiling via net/http/pprof package

Shouren:
**What type of PR is this?**
feature

This feature helps debugging performance issue and potential memory leaks in HAMi

Fixes #

### DIFF
--- a/cmd/scheduler/main.go
+++ b/cmd/scheduler/main.go
@@ -18,7 +18,7 @@ package main
 
 import (
 	"net/http"
-	_ "net/http/pprof"
+	"net/http/pprof"
 
 	"github.com/julienschmidt/httprouter"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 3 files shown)


cmd/scheduler/main.go
```diff
@@ -68,9 +71,23 @@ func init() {
 	// qps and burst settings for the client-go client
 	rootCmd.Flags().Float32Var(&config.QPS, "kube-qps", 5.0, "QPS to use while talking with kube-apiserver.")
 	rootCmd.Flags().IntVar(&config.Burst, "kube-burst", 10, "Burst to use while talking with kube-apiserver.")
+	// Add profiling related flags
+	rootCmd.Flags().BoolVar(&enableProfiling, "enable-profiling", false, "Enable pprof profiling via HTTP server")
+	rootCmd.Flags().StringVar(&profilingBindAddr, "profiling-bind-address", "127.0.0.1:6060", "The TCP address for the pprof profiling server to listen on (e.g., 127.0.0.1:6060)")
```
lengrongfu: This two var name, can we reference kubernetes define? like use `--profiling` to set enable pprof.
https://kubernetes.io/docs/reference/command-line-tools-reference/kube-apiserver/

Shouren: Updated.

PR #964, opened 2025-03-28
scheduler: fix blocked the nodeNotify channel when node changes

Iceber:
**What type of PR is this?**

/kind bug
There's no need to block the informer process when a new node event comes in, we just need to make sure there's a item in the nodeNotify

Fixes #

### DIFF
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -73,16 +73,11 @@ func NewScheduler() *Scheduler {
 	return s
 }
 
-func (s *Scheduler) onUpdateNode(_, newObj interface{}) {
-	s.nodeNotify <- struct{}{}
-}
-
-func (s *Scheduler) onDelNode(obj interface{}) {
-	s.nodeNotify <- struct{}{}
-}
-
-func (s *Scheduler) onAddNode(obj interface{}) {
-	s.nodeNotify <- struct{}{}
+func (s *Scheduler) doNodeNotify() {
+	select {
+	case s.nodeNotify <- struct{}{}:
+	default:
+	}
 }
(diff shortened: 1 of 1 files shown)


pkg/scheduler/scheduler.go
```diff
@@ -73,16 +73,11 @@ func NewScheduler() *Scheduler {
...
-
-func (s *Scheduler) onDelNode(obj interface{}) {
-	s.nodeNotify <- struct{}{}
-}
-
-func (s *Scheduler) onAddNode(obj interface{}) {
-	s.nodeNotify <- struct{}{}
+func (s *Scheduler) doNodeNotify() {
+	select {
+	case s.nodeNotify <- struct{}{}:
```
lengrongfu, line 78: Current nodeNotify channel length is 1, if we use select to write, maybe will lost event.

Iceber: Since RegisterFromNodeAnnotations actually handles all nodes, we just need to make sure that there is a signal in nodeNotify when the node changes, so that the node will not be ignored during a RegisterFromNodeAnnotations execution.

lengrongfu: After reading the RegisterFromNodeAnnotations code, I understand it, 
but I think we can create an issue to optimize the implementation of RegisterFromNodeAnnotations here; handle each node separately. no need to process all nodes every time.

Iceber: yeah, that would be a large modification, maybe we can fix this first, otherwise the number of nodes needed during the initialization phase of the program leads to a lot of duplicate RegisterFromNodeAnnotations executions

PR #965, opened 2025-03-28
scheduler: fix duplicate handling of the node label selector

Iceber:
**What type of PR is this?**

Fixes #

### DIFF
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -151,6 +151,10 @@ func (s *Scheduler) Stop() {
 func (s *Scheduler) RegisterFromNodeAnnotations() {
 	klog.InfoS("Entering RegisterFromNodeAnnotations")
 	defer klog.InfoS("Exiting RegisterFromNodeAnnotations")
+
+	labelSelector := labels.Set(config.NodeLabelSelector).AsSelector()
+	klog.InfoS("Using label selector for list nodes", "selector", labelSelector.String())
+
 	ticker := time.NewTicker(time.Second * 15)
 	defer ticker.Stop()
 	printedLog := map[string]bool{}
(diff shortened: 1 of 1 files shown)


pkg/scheduler/scheduler.go
```diff
@@ -151,6 +151,12 @@ func (s *Scheduler) Stop() {
 func (s *Scheduler) RegisterFromNodeAnnotations() {
 	klog.InfoS("Entering RegisterFromNodeAnnotations")
 	defer klog.InfoS("Exiting RegisterFromNodeAnnotations")
+
```
Shouren, line 154:
``` Go
// SelectorFromValidatedSet returns a Selector which will match exactly the given Set.
// A nil and empty Sets are considered equivalent to Everything().
// It assumes that Set is already validated and doesn't do any validation.
// Note: this method copies the Set; if the Set is immutable, consider wrapping it with ValidatedSetSelector
// instead, which does not copy.
func SelectorFromValidatedSet(ls Set) Selector {
    ...
}
func SelectorFromSet(ls Set) Selector {
	return SelectorFromValidatedSet(ls)
}

func (ls Set) AsSelector() Selector {
	return SelectorFromSet(ls)
} 
```

According to the docs in `"k8s.io/apimachinery/pkg/labels"` as above, I think we can simplify the following code to 
```Go
labelSelector = labels.Set(config.NodeLabelSelector).AsSelector()
klog.InfoS("Using label selector for list nodes", "selector", labelSelector.String())
```

pkg/scheduler/scheduler.go
```diff
@@ -151,6 +151,12 @@ func (s *Scheduler) Stop() {
...
+	labelSelector := labels.Everything()
+	if len(config.NodeLabelSelector) > 0 {
+		labelSelector = labels.Set(config.NodeLabelSelector).AsSelector()
+		klog.InfoS("Using label selector for list nodes", "selector", labelSelector.String())
+	}
```
Shouren:
``` Go
// SelectorFromValidatedSet returns a Selector which will match exactly the given Set.
// A nil and empty Sets are considered equivalent to Everything().
// It assumes that Set is already validated and doesn't do any validation.
// Note: this method copies the Set; if the Set is immutable, consider wrapping it with ValidatedSetSelector
// instead, which does not copy.
func SelectorFromValidatedSet(ls Set) Selector {
    ...
}
func SelectorFromSet(ls Set) Selector {
	return SelectorFromValidatedSet(ls)
}

func (ls Set) AsSelector() Selector {
	return SelectorFromSet(ls)
} 
```

According to the docs in `"k8s.io/apimachinery/pkg/labels"` as above, I think we can simplify code to 
```Go
labelSelector = labels.Set(config.NodeLabelSelector).AsSelector()
klog.InfoS("Using label selector for list nodes", "selector", labelSelector.String())
```

Iceber: great suggestion, updated

Iceber:
@archlitchi PTAL, Thanks

PR #966, opened 2025-03-28
scheduler: fix removing node devices from node manager

Iceber:
**What type of PR is this?**

/kind bug
This PR mainly fixes the following issues in rmNodeDevice:

1. Incorrectly deleting some devices from the node.
https://github.com/Project-HAMi/HAMi/blob/38bed6fb23cfd913b6ce009ee8f8a79cb014db23/pkg/scheduler/nodes.go#L85-L93
This will cause anything inconsistent with the vendor to be deleted, which doesn't seem to align with the function's purpose. Please let me know if I've misunderstood.

2. Concurrent read/write issues for `nodes`. Although this problem is frequent in the code, this PR fixes the issue in this specific function :smile:
3. The log exists before the `rmNodeDevice` function, so the printed device information is incorrect.
https://github.com/Project-HAMi/HAMi/blob/38bed6fb23cfd913b6ce009ee8f8a79cb014db23/pkg/scheduler/scheduler.go#L198-L200

the following issues have been optimized:
1. Modified the function name `rmNodeDevice` to `rmNodeDevices`. Since multiple devices are involved, it is changed to rmNodeDevices.
2. Modified the parameters. Repeatedly passing nodeInfo makes the entire function difficult to understand. Even if you want to restrict the device ID in the future, there are other better design methods.

Fixes #

### DIFF
--- a/pkg/scheduler/nodes.go
+++ b/pkg/scheduler/nodes.go
@@ -75,39 +75,28 @@ func (m *nodeManager) addNode(nodeID string, nodeInfo *util.NodeInfo) {
 	}
 }
 
-func (m *nodeManager) rmNodeDevice(nodeID string, nodeInfo *util.NodeInfo, deviceVendor string) {
+func (m *nodeManager) rmNodeDevices(nodeID string, deviceVendor string) {
 	m.mutex.Lock()
 	defer m.mutex.Unlock()
-	_, ok := m.nodes[nodeID]
-	if ok {
-		if len(m.nodes[nodeID].Devices) == 0 {
-			delete(m.nodes, nodeID)
-			return
-		}
-		klog.V(5).Infoln("before rm:", m.nodes[nodeID].Devices, "needs remove", nodeInfo.Devices)
-		tmp := make([]util.DeviceInfo, 0, len(m.nodes[nodeID].Devices)-len(nodeInfo.Devices))
-		for _, val := range m.nodes[nodeID].Devices {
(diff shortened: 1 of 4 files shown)


pkg/scheduler/nodes.go
```diff
@@ -75,39 +75,28 @@ func (m *nodeManager) addNode(nodeID string, nodeInfo *util.NodeInfo) {
...
-			return
+
+	nodeInfo := m.nodes[nodeID]
+	if nodeInfo == nil {
+		return
+	}
+
+	devices := make([]util.DeviceInfo, 0)
+	for _, val := range nodeInfo.Devices {
+		if val.DeviceVendor != deviceVendor {
```
lengrongfu, line 89: Before logic is when Val.DeviceVendor != deviceVendor will filter device; but new logic is on the contrary.

Iceber: Yes, that's exactly what this pr mainly fixes, according to the caller's logic **we can only remove the same device as the vendor**. please correct me if I'm wrong.

lengrongfu: only remove the same device as the vendor, that's right.

lengrongfu:
look like this OK. @chaunceyjiang please review this pr.

Iceber:
@archlitchi PTAL, Thanks

chaunceyjiang (commented):
/lgtm

Thanks~

PR #978, opened 2025-04-02
refactor: simplify code by using modern constructs

Shouren:
**What type of PR is this?**
cleanup

By using modernizer, we can simplify code in this project

Fixes #


wawa0210:
This PR causes UT coverage to decrease, so you need to pay attention to this.

Shouren:
> This PR causes UT coverage to decrease, so you need to pay attention to this.

Yeah, it is a pain when simplifying codes that are covered by UT unless that file is fully covered. Maybe i could work on some unit tests later in another pull request.

PR #1041, opened 2025-05-07
fix: Skip admission webhook when Pod's scheduler is already assigned.

ghostloda:
**What type of PR is this?**

/kind bug

If pod has a scheduler already assigned, we should not override user's explicit scheduler choice.

### DIFF
--- a/pkg/scheduler/webhook.go
+++ b/pkg/scheduler/webhook.go
@@ -60,6 +60,10 @@ func (h *webhook) Handle(_ context.Context, req admission.Request) admission.Res
 		klog.Warningf(template+" - Denying admission as pod has no containers", req.Namespace, req.Name, req.UID)
 		return admission.Denied("pod has no containers")
 	}
+	if pod.Spec.SchedulerName != "" {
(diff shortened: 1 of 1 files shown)


pkg/scheduler/webhook.go
```diff
@@ -60,6 +60,10 @@ func (h *webhook) Handle(_ context.Context, req admission.Request) admission.Res
 		klog.Warningf(template+" - Denying admission as pod has no containers", req.Namespace, req.Name, req.UID)
 		return admission.Denied("pod has no containers")
 	}
+	if pod.Spec.SchedulerName != "" {
```
Shouren: There is a chance that `pod.Spec.SchedulerName` equals to `config.SchedulerName` when `len(config.SchedulerName) > 0` is true.

ghostloda: Fixed

Shouren (approved):
/lgtm

fyp711:
@ghostloda If there are multiple schedulers within the cluster, this commit will cause HAMI's pods to not be scheduled. This is a destructive commit. 
cc @archlitchi

PR #1072, opened 2025-05-22
fix(chart): support GKE pre-release versions via kubeVersion '-0'

### LINKED ISSUE #1071: Helm install fails on GKE due to kubeVersion mismatch with pre-release suffix
When installing HAMi on a GKE cluster running Kubernetes `v1.32.3-gke.1785003`, Helm fails with the error:

```
INSTALLATION FAILED: chart requires kubeVersion: >= 1.18.0 which is incompatible with Kubernetes v1.32.3-gke.1785003
```

This is because Helm uses [Masterminds/semver](https://github.com/Masterminds/semver), which follows strict semantic versioning (SemVer) rules. According to the specification:

> “SemVer's comparisons using constraints without a pre-release comparator will skip pre-release versions. For example, >=1.2.3 will skip pre-releases when looking at a list of releases, while >=1.2.3-0 will evaluate and find pre-releases.”
 — [Masterminds/semver#Working-With-Prerelease-Versions](https://github.com/Masterminds/semver#working-with-prerelease-versions)
> 

GKE appends a suffix (e.g. `-gke.1785003`) to Kubernetes versions, and this is treated as a SemVer **pre-release**, making it ineligible for the constraint `>= 1.18.0`.

Nimbus318:
**What type of PR is this?**
/kind bug

This PR fixes Helm installation failure on GKE clusters where Kubernetes versions include a pre-release suffix (e.g., `v1.32.3-gke.1785003`). Helm uses [Masterminds/semver](https://github.com/Masterminds/semver), which by spec skips pre-releases when comparing against a constraint like `>= 1.18.0`.

From the semver spec used by Helm:

> “>=1.2.3 will skip pre-releases when looking at a list of releases, while >=1.2.3-0 will evaluate and find pre-releases.”
> 

Therefore, changing the constraint to `>= 1.18.0-0` ensures compatibility with Kubernetes versions that use pre-release identifiers, including those from GKE and other distributions.

Additionally, this PR updates the minimum supported Kubernetes version in the `README` from 1.16 to 1.18, reflecting the current chart constraint.

References:

- https://helm.sh/docs/topics/charts/#the-chart-yaml-file
- https://github.com/Masterminds/semver#working-with-prerelease-versions

Fixes #1071

no

### DIFF
--- a/charts/hami/Chart.yaml
+++ b/charts/hami/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: hami
 version: 2.5.0
-kubeVersion: ">= 1.16.0"
+kubeVersion: ">= 1.18.0-0"
 description: Heterogeneous AI Computing Virtualization Middleware
 keywords:
   - vgpu
--- a/README.md
+++ b/README.md
@@ -94,7 +94,7 @@ The list of prerequisites for running the NVIDIA device plugin is described belo
 - NVIDIA drivers >= 440
 - nvidia-docker version > 2.0
 - default runtime configured as nvidia for containerd/docker/cri-o container runtime
-- Kubernetes version >= 1.16
+- Kubernetes version >= 1.18
 - glibc >= 2.17 & glibc < 2.30
 - kernel version >= 3.10
 - helm > 3.0
--- a/README_cn.md
+++ b/README_cn.md
@@ -90,7 +90,7 @@ HAMi 由多个组件组成，包括统一的 mutatingwebhook、统一的调度�
 - NVIDIA 驱动 >= 440
 - nvidia-docker 版本 > 2.0
 - containerd/docker/cri-o 容器运行时的默认运行时配置为 nvidia
-- Kubernetes 版本 >= 1.16
+- Kubernetes 版本 >= 1.18
 - glibc >= 2.17 & glibc < 2.30
 - 内核版本 >= 3.10
 - helm > 3.0


charts/hami/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: hami
 version: 2.5.0
-kubeVersion: ">= 1.16.0"
+kubeVersion: ">= 1.18.0-0"
```
Shouren, line 4: Is this means HAMi is no longer support kubernetes v.1.16.x ?
If so, we should update the Prerequisites section in README.md

Shouren (changes requested):

charts/hami/Chart.yaml
archlitchi: yes, there is some issue in v1.16, so i change the requirement to v1.18, you can submit a PR to fix that in document

Nimbus318: I was using Helm 2.5.1 and noticed the chart already requires kubeVersion >= 1.18.0.
https://github.com/Project-HAMi/HAMi/blob/83b9b23639b50399ca21517ee5776bc991b63ad3/charts/hami/Chart.yaml#L4
I asked @archlitchi  about it—apparently there were some issues with 1.16.x in earlier testing. So yes, HAMi doesn't officially support 1.16.x anymore, and you're totally right that the README.md should reflect that. I’ll update it soon. Appreciate the reminder!

PR #1092, opened 2025-05-28
fix: wrong Pod's UID and emtpy Pod's name in log of webhook.go

Shouren:
**What type of PR is this?**
/kind bug

This PR updates the `Handle()` func in `pkg/scheduler/webhook.go`. The previous implementation using `req.UID` and `req.Name` when logging. However, according to the doc of [AdmissionRequest](https://pkg.go.dev/k8s.io/api/admission/v1#AdmissionRequest) as below:
> // UID is an identifier for the individual request/response. It allows us to distinguish instances of requests which are
   // otherwise identical (parallel requests, requests when earlier requests did not modify etc)
   // The UID is meant to track the round trip (request/response) between the KAS and the WebHook, not the user request.
   // It is suitable for correlating log entries between the webhook and apiserver, for either auditing or debugging.
   UID [types](https://pkg.go.dev/k8s.io/apimachinery/pkg/types).[UID](https://pkg.go.dev/k8s.io/apimachinery/pkg/types#UID) `json:"uid" protobuf:"bytes,1,opt,name=uid"`

`req.UID` is not equal to `pod.UID`, as shown in the following screenshot.

CleanShot 2025-05-27 at 17 36 39

And the `req.Name` might be empty according the above [doc](https://pkg.go.dev/k8s.io/api/admission/v1#AdmissionRequest):
> // Name is the name of the object as presented in the request.  On a CREATE operation, the client may **omit name** and
   // rely on the server to generate the name.  If that is the case, this field will contain an empty string.
   // +optional
   Name [string](https://pkg.go.dev/builtin#string) `json:"name,omitempty" protobuf:"bytes,5,opt,name=name"`

Fixes #

@archlitchi 

### DIFF
--- a/pkg/scheduler/webhook.go
+++ b/pkg/scheduler/webhook.go
@@ -57,20 +57,20 @@ func (h *webhook) Handle(_ context.Context, req admission.Request) admission.Res
 		return admission.Errored(http.StatusBadRequest, err)
 	}
 	if len(pod.Spec.Containers) == 0 {
-		klog.Warningf(template+" - Denying admission as pod has no containers", req.Namespace, req.Name, req.UID)
+		klog.Warningf(template+" - Denying admission as pod has no containers", pod.Namespace, pod.Name, pod.UID)
 		return admission.Denied("pod has no containers")
 	}
 	if pod.Spec.SchedulerName != "" && (len(config.SchedulerName) == 0 || pod.Spec.SchedulerName != config.SchedulerName) {
 		klog.Infof(template+" - Pod already has different scheduler assigned", req.Namespace, req.Name, req.UID)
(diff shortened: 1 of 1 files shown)


Shouren:
@archlitchi @wawa0210 PTAL

archlitchi:
yes, we've discussed on weekly meeting

/lgtm

archlitchi:
CC @ouyangluwei163

Shouren:
> CC @ouyangluwei163

@ouyangluwei163 PTAL

Shouren:
Conflict resolved.
@ouyangluwei163 Please take a look.

Shouren:
@wawa0210 PTAL

PR #1097, opened 2025-05-30
Optimize Fit-in-device logic to make it device-dependent

archlitchi:
/kind feature

Right now, a unified fitInCertainDevice is responsible for all devices, which is a very bad design, it causes many unnecessary logic among different types of device.

We need to optimize that by make fit logic device-specific, implemented by device interface.So We did several adjustment to  device API, such as 

Add interface Fit
Remove interface CheckType, CheckUUID, CustomFilterRule. Because these logics can be implemented in Fit directly


Shouren:
@archlitchi After a quick review of this pr, i find there are lots of duplicated code in the current implementation, especially in the `Fit` func of each device. I have not compared every implementation of device in detail, but i think it is better to create some helper functions in `pkg/util` or `pkg/device/util` to deal with common processing logic for each reasons.

pkg/device/nvidia/device.go
```diff
@@ -571,3 +572,108 @@ func (dev *NvidiaGPUDevices) AddResourceUsage(n *util.DeviceUsage, ctr *util.Con
...
+	prevnuma := -1
+	klog.InfoS("Allocating device for container request", "pod", klog.KObj(pod), "card request", k)
+	var tmpDevs map[string]util.ContainerDevices
+	tmpDevs = make(map[string]util.ContainerDevices)
+	reason := make(map[string]int)
+	for i := len(devices) - 1; i >= 0; i-- {
+		dev := devices[i]
+		klog.V(4).InfoS("scoring pod", "pod", klog.KObj(pod), "device", dev.ID, "Memreq", k.Memreq, "MemPercentagereq", k.MemPercentagereq, "Coresreq", k.Coresreq, "Nums", k.Nums, "device index", i)
+
+		found, numa := nv.CheckType(annos, *dev, k)
```
Shouren: The `CheckType` function should start with a lower case letter

pkg/device/nvidia/device.go
```diff
@@ -571,3 +572,108 @@ func (dev *NvidiaGPUDevices) AddResourceUsage(n *util.DeviceUsage, ctr *util.Con
...
+		if numa && prevnuma != dev.Numa {
+			if k.Nums != originReq {
+				reason[common.NumaNotFit] += len(tmpDevs)
+				klog.V(5).InfoS(common.NumaNotFit, "pod", klog.KObj(pod), "device", dev.ID, "k.nums", k.Nums, "numa", numa, "prevnuma", prevnuma, "device numa", dev.Numa)
+			}
+			k.Nums = originReq
+			prevnuma = dev.Numa
+			tmpDevs = make(map[string]util.ContainerDevices)
+		}
+		if !nv.CheckUUID(annos, *dev) {
```
Shouren: Same as `CheckType`, this `CheckUUID` function should also start with a lower case

pkg/device/nvidia/device.go
```diff
@@ -571,3 +572,108 @@ func (dev *NvidiaGPUDevices) AddResourceUsage(n *util.DeviceUsage, ctr *util.Con
...
+	tmpDevs = make(map[string]util.ContainerDevices)
+	reason := make(map[string]int)
+	for i := len(devices) - 1; i >= 0; i-- {
+		dev := devices[i]
+		klog.V(4).InfoS("scoring pod", "pod", klog.KObj(pod), "device", dev.ID, "Memreq", k.Memreq, "MemPercentagereq", k.MemPercentagereq, "Coresreq", k.Coresreq, "Nums", k.Nums, "device index", i)
+
+		found, numa := nv.CheckType(annos, *dev, k)
+		if !found {
+			reason[common.CardTypeMismatch]++
+			klog.V(5).InfoS(common.CardTypeMismatch, "pod", klog.KObj(pod), "device", dev.ID, dev.Type, k.Type)
```
lengrongfu, line 591: this logger level we can unified debug is 4. Other log levels are similar.

pkg/device/ascend/device.go
```diff
@@ -290,3 +287,99 @@ func (dev *Devices) AddResourceUsage(n *util.DeviceUsage, ctr *util.ContainerDev
 	n.Usedmem += ctr.Usedmem
 	return nil
 }
+
+func (npu *Devices) Fit(devices []*util.DeviceUsage, request util.ContainerDeviceRequest, annos map[string]string, pod *corev1.Pod, allocated *util.PodDevices) (bool, map[string]util.ContainerDevices, string) {
```
lengrongfu, line 291: This function l found every device type there is a similar logic, is there a better implementation of this?

ouyangluwei163: I feel the same way, Fine-grained interfaces are more friendly to subsequent scheduler expansion.

archlitchi:
> @archlitchi After a quick review of this pr, i find there are lots of duplicated code in the current implementation, especially in the `Fit` func of each device. I have not compared every implementation of device in detail, but i think it is better to create some helper functions in `pkg/util` or `pkg/device/util` to deal with common processing logic for each reasons.

yes, we need refine some logics here, for now, i just copy original fitInCertain method to every devices

pkg/device/ascend/device.go
archlitchi: that's our next move, we need to remove redundant logic in fit for each devices

pkg/device/nvidia/device.go
archlitchi: adopted

pkg/device/nvidia/device.go
archlitchi: adopted

PR #1117, opened 2025-06-12
feat(scheduler): make node lock timeout configurable

Kevinz857:
### Summary

This PR makes the HAMi scheduler's node lock timeout configurable, allowing operators to adjust the timeout duration based on their cluster characteristics and scheduling requirements.

### Changes Made

1. __Configuration Layer (`pkg/scheduler/config/config.go`)__

   - Added `NodeLockTimeout int` configuration parameter
   - Default value set to 1 minute to maintain backward compatibility

2. __Command Line Interface (`cmd/scheduler/main.go`)__

   - Added `--node-lock-timeout` flag to scheduler command line options
   - Flag accepts timeout value in minutes
   - Integrated timeout initialization in the `start()` function

3. __Runtime Behavior (`pkg/util/nodelock/nodelock.go`)__

   - The global `NodeLockTimeout` variable is now set from configuration
   - Maintains existing lock/unlock logic and timeout checking mechanism

### Motivation

Previously, the node lock timeout was hardcoded to 1 minute. In some cluster environments, this timeout may be too long, causing unnecessary delays in scheduling operations. By making this configurable, operators can:

- Reduce scheduling latency in fast clusters by setting shorter timeouts
- Increase timeout for slower clusters to avoid premature lock releases
- Fine-tune scheduler behavior based on workload characteristics

### Backward Compatibility

- Default timeout remains 1 minute, ensuring no behavior change for existing deployments
- Existing configurations continue to work without modification
- New flag is optional and uses sensible defaults

### Default timeout is 5 minute 
```shell
./scheduler
```

### Custom set 30s
```shell
./scheduler --node-lock-timeout=30s
```

### Custom set 2mins
```shell
./scheduler --node-lock-timeout=2m
```

### DIFF
--- a/cmd/scheduler/main.go
+++ b/cmd/scheduler/main.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"net/http"
 	"net/http/pprof"
+	"time"
 
 	"github.com/julienschmidt/httprouter"
 	"github.com/spf13/cobra"
@@ -32,6 +33,7 @@ import (
 	"github.com/Project-HAMi/HAMi/pkg/util"
 	"github.com/Project-HAMi/HAMi/pkg/util/client"
 	"github.com/Project-HAMi/HAMi/pkg/util/flag"
+	"github.com/Project-HAMi/HAMi/pkg/util/nodelock"
 	"github.com/Project-HAMi/HAMi/pkg/version"
 )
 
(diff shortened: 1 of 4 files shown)


archlitchi:
have you tested on your environment?

cmd/scheduler/main.go
```diff
@@ -72,6 +74,7 @@ func init() {
 	rootCmd.Flags().IntVar(&config.Burst, "kube-burst", client.DefaultBurst, "Burst to use while talking with kube-apiserver.")
 	rootCmd.Flags().IntVar(&config.Timeout, "kube-timeout", client.DefaultTimeout, "Timeout to use while talking with kube-apiserver.")
 	rootCmd.Flags().BoolVar(&enableProfiling, "profiling", false, "Enable pprof profiling via HTTP server")
+	rootCmd.Flags().IntVar(&config.NodeLockTimeout, "node-lock-timeout", 1, "timeout for node locks in minutes")
```
wawa0210: The default setting is 1 minute, which is consistent with the current setting. If you need to pay attention to this parameter, you can customize it.

pkg/scheduler/config/config.go
```diff
@@ -37,4 +37,7 @@ var (
 
 	// NodeLabelSelector is scheduler filter node by node label.
 	NodeLabelSelector map[string]string
+
+	// NodeLockTimeout is the timeout for node locks in minutes. default value is 1 minute.
+	NodeLockTimeout int
```
wawa0210: I suggest that the NodeLockTimeout type should be time.Duration. Also, the comment here should not have a description of the default value.

Kevinz857: Very good suggestion, already optimized and submitted

PR #1123, opened 2025-06-16
Support Metax sGPU Qos Policy

Kyrie336:
**What type of PR is this?**

/kind feature

Support Metax sGPU Qos Policy: user can specify qos policy for job

Fixes #

1. add `CustomInfo` filed in framework Device, vendor can provide other vendor custom info for device
2. interface `PatchAnnotations` `AddResourceUsage`  add args to pass Pod annotations

### DIFF
--- a/examples/metax/sgpu/allocate_specific_qos.yaml
+++ b/examples/metax/sgpu/allocate_specific_qos.yaml
@@ -3,7 +3,7 @@ kind: Pod
 metadata:
   name: gpu-pod
   annotations:
-    metax-tech.com/use-gpuuuid: "36beae85-c835-6b14-6ab2-02671837a59c" # allocate specific gpu
+    metax-tech.com/sgpu-qos-policy: "best-effort" # allocate specific qos sgpu
 spec:
   containers:
     - name: ubuntu-container
--- a/pkg/device/devices.go
+++ b/pkg/device/devices.go
@@ -51,9 +51,9 @@ type Devices interface {
 	LockNode(n *corev1.Node, p *corev1.Pod) error
 	ReleaseNodeLock(n *corev1.Node, p *corev1.Pod) error
 	GenerateResourceRequests(ctr *corev1.Container) util.ContainerDeviceRequest
-	PatchAnnotations(annoinput *map[string]string, pd util.PodDevices) map[string]string
+	PatchAnnotations(pod *corev1.Pod, annoinput *map[string]string, pd util.PodDevices) map[string]string
 	ScoreNode(node *corev1.Node, podDevices util.PodSingleDevice, previous []*util.DeviceUsage, policy string) float32
-	AddResourceUsage(n *util.DeviceUsage, ctr *util.ContainerDevice) error
+	AddResourceUsage(pod *corev1.Pod, n *util.DeviceUsage, ctr *util.ContainerDevice) error
 	Fit(devices []*util.DeviceUsage, request util.ContainerDeviceRequest, annos map[string]string, pod *corev1.Pod, allocated *util.PodDevices) (bool, map[string]util.ContainerDevices, string)
 	// This should not be associated with a specific device object
 	//ParseConfig(fs *flag.FlagSet)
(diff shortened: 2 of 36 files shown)


pkg/device/metax/sdevice.go
```diff
@@ -255,11 +272,23 @@ func (sdev *MetaxSDevices) ScoreNode(node *corev1.Node, podDevices util.PodSingl
...
 	n.Usedmem += ctr.Usedmem
 
+	if value, ok := n.CustomInfo["QosPolicy"]; ok {
+		if qos, ok := value.(string); ok {
+			expectedQos := annos[MetaxSGPUQosPolicy]
+			if ctr.Usedcores == 100 {
+				expectedQos = ""
+			}
+
+			n.CustomInfo["QosPolicy"] = expectedQos
```
archlitchi, line 287: so, the expectedQos is a per-GPU configuration which can be changed by pod annotations? how does it work?

archlitchi:
please modify website(https://github.com/Project-HAMi/website) to update document accordingly, if there are any modifications in user-guide and examples

pkg/device/metax/sdevice.go
Kyrie336: yes, you are right.  Currently, there are three kinds of Qos, which are `best-effort`, `fixed-share`, `burst-share`, and a GPU has only one Qos at a time. **When a GPU is unused, its Qos can be switched, that is modified by Pod annotations**

Kyrie336:
> i think pass the whole *corev1.Pod to the parameter instead of merely annotations is better, to avoid further modifications if need other field of pod

@archlitchi it's good advise, and i have updated the commit

examples/metax/sgpu/allocate_specific_qos.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: gpu-pod
+  annotations:
+    metax-tech.com/sgpu-qos-policy: "best-effort" # allocate specific qos sgpu
+spec:
+  containers:
+    - name: ubuntu-container
+      image: cr.metax-tech.com/public-ai-release/c500/colossalai:2.24.0.5-py38-ubuntu20.04-amd64
```
Shouren: Trying to pull this image, but here comes an error:
```bash
$ docker pull cr.metax-tech.com/public-ai-release/c500/colossalai:2.24.0.5-py38-ubuntu20.04-amd64
Error response from daemon: pull access denied for cr.metax-tech.com/public-ai-release/c500/colossalai, repository does not exist or may require 'docker login': denied: requested access to the resource is denied
```
If this repository requires `docker login`, please add a comment to tell how to get an account for login.

Kyrie336: yes, this image need account to pull. i will change the image to normal image like ubuntu

pkg/device/devices.go
```diff
@@ -51,9 +51,9 @@ type Devices interface {
 	LockNode(n *corev1.Node, p *corev1.Pod) error
 	ReleaseNodeLock(n *corev1.Node, p *corev1.Pod) error
 	GenerateResourceRequests(ctr *corev1.Container) util.ContainerDeviceRequest
-	PatchAnnotations(annoinput *map[string]string, pd util.PodDevices) map[string]string
+	PatchAnnotations(pod *corev1.Pod, annoinput *map[string]string, pd util.PodDevices) map[string]string
 	ScoreNode(node *corev1.Node, podDevices util.PodSingleDevice, previous []*util.DeviceUsage, policy string) float32
-	AddResourceUsage(n *util.DeviceUsage, ctr *util.ContainerDevice) error
+	AddResourceUsage(pod *corev1.Pod, n *util.DeviceUsage, ctr *util.ContainerDevice) error
```
Shouren, line 56: I prefer not to add `pod *corev1.Pod` as a new parameter for `AddResourceUsage` method. Current implementation of `Filter` in scheduler lacks a `context` mechanism that devices can store information in request's scope. 
So my suggestion is that we can design the `context` mechanism first, how do you think ? @Kyrie336 @archlitchi

Kyrie336: @Shouren @archlitchi **it's a good advice that is exactly what vendor developers need.** But due to project deadlines, add args is better for me

PR #1133, opened 2025-06-18
add global image for chart

calvin0327:
**What type of PR is this?**
/kind feature

add global image for chart

Fixes #

### DIFF
--- a/charts/hami/templates/_commons.tpl
+++ b/charts/hami/templates/_commons.tpl
@@ -1,0 +1,49 @@
+{{/*
+Return the proper image name
+{{ include "common.images.image" ( dict "imageRoot" .Values.path.to.the.image "global" $) }}
+*/}}
+{{- define "common.images.image" -}}
+{{- $registryName := .imageRoot.registry -}}
+{{- $repositoryName := .imageRoot.repository -}}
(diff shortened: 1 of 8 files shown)


calvin0327:
test：
image

charts/hami/templates/_commons.tpl
```diff
@@ -0,0 +1,50 @@
+{{/* vim: set filetype=mustache: */}}
```
Shouren: Why do we need to add Vim-related config here? I think it is not necessary for this file.

charts/hami/templates/_helpers.tpl
```diff
@@ -117,3 +110,74 @@ imagePullSecrets: {{ toYaml .Values.imagePullSecrets | nindent 2 }}
...
+
+{{/*
+Return the proper image name
+*/}}
+{{- define "hami.scheduler.patch.image" -}}
+{{ include "common.images.image" (dict "imageRoot" .Values.scheduler.patch.image "global" .Values.global) }}
+{{- end -}}
+
+{{/*
+Return the proper image name
```
Shouren: There are multiple lines of same comment above, please make it more meaningful or remove it.

Shouren (approved):
/lgtm

PR #1145, opened 2025-06-24
feat(helm): optionally disable admission webhook

Antvirf:
This simplifies the deployment considerably and makes HAMi less intrusive inclusters where only a minority of workloads actually require GPU scheduling.

**What type of PR is this?**

**Feature** that makes the helm chart more flexible.

/kind feature

- In clusters where only a minority of workloads will benefit from HAMi (e.g. a cluster with a lot of nodes where only a small subset is GPU-equipped), adding a webhook on all pod creations is needlessly intrusive and risky - any downtime of this webhook blocks scheduling of all pods.
- This flag simplifies the deployment manifests considerably, making it easier to install, maintain, troubleshoot and manage HAMi and its components.
- **When this flag is active, the operator/maintainer of the cluster is now responsible for setting `schedulerName` on relevant workload pods** - this can be done with generic helm charts, which is not a problem as in these kinds of mixed clusters, GPU workloads will often anyway need extra configs (e.g. tolerations/affinities).

Simple comparison of created resource counts with this option enabled (default) and with it disabled:

```
cd charts/hami/

# leaving parameter as default
❯ helm template . --values values.yaml  | grep -E "^kind: " | cut -d' ' -f2 | sort | uniq -c
      2 ClusterRole
      3 ClusterRoleBinding
      4 ConfigMap
      1 DaemonSet
      1 Deployment
      2 Job
      1 MutatingWebhookConfiguration
      1 Role
      1 RoleBinding
      2 Service
      3 ServiceAccount

# after, disabling the webhook
❯ helm template . --values values.yaml --set scheduler.admissionWebhook.enabled=false | grep -E "^kind: " | cut -d' ' -f2 | sort | uniq -c
      1 ClusterRole
      2 ClusterRoleBinding
      4 ConfigMap
      1 DaemonSet
      1 Deployment
      2 Service
      2 ServiceAccount

```

We remove several resources: `CR, CRB, Deployment, MutatingWebhookConfiguration, R, RB, SA 2x Jobs`.

If this is merged I will go ahead and do a similar docs update PR in `hami/website` to make a note where relevant that the webhook component can be disabled.

- [ ] To be confirmed whether the cert-manager resource is used for anything else except the mutating webhook. Given that it's an alternative for the `patch` approach, I presumed it is only used by the webhook and can therefore be disabled here. 

No, since the default value for this parameter is `enabled: true`, the behaviour of the chart does not change unless the user configures this parameter to false.

### DIFF
--- a/charts/hami/Chart.yaml
+++ b/charts/hami/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: hami
-version: 2.6.0
+version: 2.7.0
 kubeVersion: ">= 1.18.0-0"
 description: Heterogeneous AI Computing Virtualization Middleware
 keywords:
@@ -13,4 +13,3 @@ maintainers:
   - name: zhangxiao
     email: xiaozhang0210@hotmail.com
 appVersion: "2.6.0"
-
--- a/charts/hami/Chart.yaml
+++ b/charts/hami/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: hami
-version: 2.7.0
+version: 2.6.0
 kubeVersion: ">= 1.18.0-0"
 description: Heterogeneous AI Computing Virtualization Middleware
 keywords:
@@ -13,3 +13,4 @@ maintainers:
   - name: zhangxiao
     email: xiaozhang0210@hotmail.com
 appVersion: "2.6.0"
+
--- a/charts/hami/values.yaml
+++ b/charts/hami/values.yaml
@@ -106,6 +106,7 @@ scheduler:
   tolerations: []
   #serviceAccountName: "hami-vgpu-scheduler-sa"
   admissionWebhook:
+    enabled: true
     customURL:
       enabled: false
       # must be an endpoint using https.
--- a/charts/hami/templates/scheduler/certmanager.yaml
+++ b/charts/hami/templates/scheduler/certmanager.yaml
@@ -1,3 +1,4 @@
+{{- if .Values.scheduler.admissionWebhook.enabled }}
 {{- if .Values.scheduler.certManager.enabled }}
 apiVersion: cert-manager.io/v1
 kind: Certificate
@@ -26,4 +27,5 @@ metadata:
     {{- include "hami-vgpu.labels" . | nindent 4 }}
 spec:
   selfSigned: {}
+{{- end }}
 {{- end }}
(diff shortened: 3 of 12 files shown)


archlitchi:
hi, please resolve these conflicts and pass the CI check

charts/hami/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: hami
-version: 2.6.0
+version: 2.7.0
```
Shouren: Please do not update the version of Chart in this PR. It should be updated when a release branch created.

Shouren (changes requested):

charts/hami/Chart.yaml
Antvirf: Resolved, thank you!

Antvirf:
I'll reopen this PR - the workflow tries to find the original base commit which is no longer there since I force pushed updates.

Antvirf:
@archlitchi would you mind retriggering the CI - hopefully the reopened PR uses the proper base commit and should start the workflow.

Antvirf:
btw this will need a bit of work if https://github.com/Project-HAMi/HAMi/pull/1152 is merged first - happy to take care of that

charts/hami/values.yaml
```diff
@@ -106,6 +106,7 @@ scheduler:
   tolerations: []
   #serviceAccountName: "hami-vgpu-scheduler-sa"
   admissionWebhook:
+    enabled: true
```
Shouren, line 111: >* **When this flag is active, the operator/maintainer of the cluster is now responsible for setting `schedulerName` on relevant workload pods** - this can be done with generic helm charts, which is not a problem as in these kinds of mixed clusters, GPU workloads will often anyway need extra configs (e.g. tolerations/affinities).

@Antvirf There is an interface named `MutateAdmission` within each hami-supported `Devices`, the webhook will also call this method to mutate some fields of Pod. So the operator/maintainer of the cluster is also responsible for making those mutations, i think we should add a comment to this flag.

Antvirf: Good idea, will do

Antvirf: @Shouren added this in, let me know your thoughts

Will likely have GPU-enabled clusters at work shortly, once that comes around I'll work on updating the docs site about this option + examples on using it with nvidia devices at least.

Shouren (approved):
/lgtm

Antvirf:
@archlitchi @Shouren does this need a rebase or something? Github shows no conflicts, `tide` is complaining that PR can't be rebased

archlitchi:
> is need a rebase or something

let me see into it

Issue #1143, opened 2025-06-24
Narrowing down what privileges HAMi requires and why


Antvirf:
**Please provide an in-depth description of the question you have**:

Looking at the rendered helm chart, HAMi deploys several things that raise questions/concerns in terms of privileges/security:
- `DaemonSet/hami-device-plugin`:
  - Both the `device-plugin` and `vgpu-monitor` containers require `SYS_ADMIN` - [this is considered an overloaded capability](https://man7.org/linux/man-pages/man7/capabilities.7.html), do we have justification for this vs. specifying the actually required capabilities individually?
  - Given the capabilities we already define, are `privileged` and `allowPrivilegeEscalation` both still needed for the `device-plugin` container?
  - Needing `hostPID: true` makes sense, how come we also need `hostNetwork: true`?
- `ClusterRoleBinding/hami-scheduler`: wants `hami-scheduler` to be `cluster-admin`, why? A helm chart allows us to template out a properly scoped role, hard to see why this would be needed

**What do you think about this question?**:
Please don't mistake this for negativity, I think HAMi is a super impressive and very useful project! I believe that improving on these points in either the chart directly (wherever we can make changes to follow least-privilege) or a docs page explaining what these perms are needed for would make it easier for companies to adopt it 🙂

In addition, certain features/devices may require higher privileges (not sure?) - we should be clear in the docs and in the chart about what requires what, so that users only interested in one device type do not need to run higher-privileged workloads than necessary. I'm happy to submit PRs improving this stuff but raising an issue first for discussion.

**Environment**:
- HAMi version: 2.6.0 chart (latest release at time of writing)
- Kubernetes version: not relevant

archlitchi:
That's a very good question, the device-plugin and vGPUmonitor requires high privileges because
1. The dynamic-mig feature requires we enable/disable mig feature on GPU, which requires privilege
2. We need to copy libvgpu.so and relelated-file from image to hostpath '/usr/local/vgpu', so we need these corresponding permissions.
3. Patch Node annotations/ Pod annotations
For the scheduler part, it needs
1. List/Patch to Pod, nodes, and resource quota
2. all the necessary privileges used by scheduler-extender and mutatingwebhookconfiguration

We're very happy and welcome to these improvement. Since the majority of contributors are not very familiar and sensitive with security, there must be many reductant.  You can submit an issue about your plan and we can discuss it further.

Antvirf:
[...]

PR #1152, opened 2025-06-25
feat(scheduler-role): use a scoped-down role for scheduler

Antvirf:
**What type of PR is this?**

Improve the `hami` helm chart by scoping down the `ClusterRole` given to `hami-scheduler`.

- `cluster-admin` role for the `hami-scheduler` is overly broad. The purpose of this PR is to scope down the role following the principle of least-privilege. The permissions added were deduced from looking at the source code of `hami` and the types of calls it makes to the Kubernetes API.
- `cluster-admin` is a no-go in most organizations and is overly broad for `hami-scheduler`. A properly scope down role makes `hami` easier for orgs to adopt.
- We use the built-in system roles `kube-scheduler` and `volume-scheduler` to avoid needing to define these in fill ourselves. Custom `CR` and `R` are only used for very specific needs.

Partially addresses https://github.com/Project-HAMi/HAMi/issues/1143

More work is needed to close that ticket in terms of linux capabilities given to the container.

- [ ] To be confirmed how this affects the e2e tests, if at all
- [ ] Detailed review of the suggested perms is required here!

None planned.

### DIFF
--- a/charts/hami/templates/scheduler/clusterrole.yaml
+++ b/charts/hami/templates/scheduler/clusterrole.yaml
@@ -7,14 +7,15 @@ metadata:
     {{- include "hami-vgpu.labels" . | nindent 4 }}
 rules:
   - apiGroups: [""]
-    resources: ["pods"]
+    resources: ["pods", "configmaps"]
     verbs: ["get", "list", "watch", "patch", "update"]
   - apiGroups: [""]
     resources: ["pods/binding"]
     verbs: ["create"]
   - apiGroups: [""]
     resources: ["nodes"]
-    verbs: ["get", "list", "patch", "update"]
+    verbs: ["get", "list", "patch", "update", "watch"]
   - apiGroups: [""]
     resources: ["events"]
     verbs: ["create", "get", "list"]
+
--- a/charts/hami/templates/scheduler/clusterrole.yaml
+++ b/charts/hami/templates/scheduler/clusterrole.yaml
@@ -1,0 +1,20 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: {{ include "hami-vgpu.scheduler" . }}
+  labels:
+    app.kubernetes.io/component: "hami-scheduler"
+    {{- include "hami-vgpu.labels" . | nindent 4 }}
+rules:
+  - apiGroups: [""]
+    resources: ["pods"]
+    verbs: ["get", "list", "watch", "patch", "update"]
+  - apiGroups: [""]
+    resources: ["pods/binding"]
+    verbs: ["create"]
+  - apiGroups: [""]
+    resources: ["nodes"]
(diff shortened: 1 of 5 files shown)


archlitchi:
Hi, thanks for your contributing, please notify me if you're ready to integrate, we will start testing on MIG

Antvirf:
Hey @archlitchi - this PR would be good to test w/ MIG. I don't expect this to break that part yet as I didn't change any of the linux capabilities yet.

If ok for you, I'd say let's get this change in first and I'll start on the linux capabilities work in the meantime?

Antvirf:
@archlitchi I've made a couple of changes after more testing locally - I can't get past the pod pending stage since I don't have a GPU on this test cluster, but these changes cleared out any permissions errors encountered so far. Let's give the e2e test a go :slightly_smiling_face:

charts/hami/templates/scheduler/clusterrole.yaml
```diff
@@ -0,0 +1,21 @@
...
+kind: ClusterRole
+metadata:
+  name: {{ include "hami-vgpu.scheduler" . }}
+  labels:
+    app.kubernetes.io/component: "hami-scheduler"
+    {{- include "hami-vgpu.labels" . | nindent 4 }}
+rules:
+  - apiGroups: [""]
+    resources: ["pods", "configmaps"]
+    verbs: ["get", "list", "watch", "patch", "update"]
```
Shouren: @Antvirf After checking the code of HAMi scheduler, i think `update` can be removed. Can you try it in you test environment ?

Antvirf: Done, seems ok

charts/hami/templates/scheduler/clusterrole.yaml
```diff
@@ -0,0 +1,21 @@
...
+  - apiGroups: [""]
+    resources: ["pods/binding"]
+    verbs: ["create"]
+  - apiGroups: [""]
+    resources: ["nodes"]
+    verbs: ["get", "list", "patch", "update", "watch"]
```
Shouren, line 17: The `pkg/utils/nodelock` package is the only package within the HAMi scheduler that depends the `update` verb for `nodes` resource. Since we only update the annotations in node's metadata, i think we can refactor it to use `Patch`.
How do you think ? @Antvirf @archlitchi

archlitchi: yes, but we need to pass the e2e first

Shouren: @archlitchi I prefer to create a new issue about it with a `good first issue` label :yum:

Antvirf: Makes sense to me :+1:

Shouren (approved):
/lgtm

PR #1154, opened 2025-06-30
feat(scheduler): add node nouse gpuuuid function

ZhengW22:
**What type of PR is this?**

This PR adds the capability to disable GPUs at the node level by applying annotations to nodes. GPUs matching the specified UUIDs will no longer be allocated to any pods.

The implementation works by setting the used count of the corresponding node GPUs to their maximum capacity when calculating nodeUsage, effectively occupying those resources. This approach maintains compatibility with scheduling logic for different types of GPU cards.

No.

No.

No.


pkg/scheduler/scheduler.go
```diff
@@ -266,18 +268,33 @@ func (s *Scheduler) getNodesUsage(nodes *[]string, task *corev1.Pod) (*map[strin
 				userGPUPolicy = value
 			}
 		}
+		disableGPUUUIDMap := make(map[string]bool)
+		if nodeInfo.Node != nil {
+			if value, ok := nodeInfo.Node.Annotations[NodeNoUseUUID]; ok {
+				disableGPUUUIDList := strings.Split(value, ":")
```
Shouren: The type of device into consideration should be taken into consideration. Otherwise, there is a chance that different device plugins may generate same UUID.

ZhengW22: Ok, i will add vendor info.

ZhengW22: After checking code, I find I only can get vendor info from node.Devices variable. But its field DeviceVendor initializes in the https://github.com/Project-HAMi/HAMi/blob/bbe60088d975e5a57c60efa50fa78edbe4ac441e/pkg/device/devices.go#L122, this vendor name is written in the code not a conifg file. I think this is not user-friendly and the same uuid is very rarely.

Shouren: > After checking code, I find I only can get vendor info from node.Devices variable. But its field DeviceVendor initializes in the
> 
> https://github.com/Project-HAMi/HAMi/blob/bbe60088d975e5a57c60efa50fa78edbe4ac441e/pkg/device/devices.go#L122

So, i think it not the right place to the implement the feature. Perhaps you can check the `addNode` method of `nodeManager` in `pkg/scheduler/nodes.go`

ZhengW22:
Hello, I add vendor info when disable GPU by uuid on node.

ZhengW22:
@Shouren Hello, can you help me to review new commit that adds vendor info?

docs/config.md
```diff
@@ -63,6 +63,14 @@ In Kubernetes, in order for the API server to communicate with the webhook compo
 * `scheduler.certManager.enabled`:
   Boolean type, default value is false, if true, cert-manager will generate a self-signed certificate. **Note: This option requires cert-manager to be installed in your cluster first.** _See [cert-manager installation](https://cert-manager.io/docs/installation/kubernetes/) for more details._
 
+## Node configs: annotations
+
+* `nvidia.com/nouse-gpuuuid`:
```
Shouren, line 83: If the annotation is just for NVIDIA GPU, you can check the `filterdevices` field in`nodeconfig` in `charts/hami/templates/device-plugin/configmap.yaml`. That field will help you filter the GPUs.

Shouren (changes requested):
> So, i think it not the right place to the implement the feature. Perhaps you can check the `addNode` method of `nodeManager` in `pkg/scheduler/nodes.go`

As i mentioned in previous comment, i prefer that you implement this feature in the `addNode` method of `nodeManager` in `pkg/scheduler/nodes.go`. 

This benefits not only avoiding parsing the annotation every time iterating nodes in `getNodesUsage` method but also clarify the meaning of `nouse` which devices will not add into the nodeInfo

docs/config.md
ZhengW22: This feature is not just for NVIDIA GPU, but other vendor documents don't mention nouse annotation, although their code support this. Only NVIDIA document mentions nouse annotation. By the way, I also have tested this feature on metax.

ZhengW22:
@Shouren Hello, I moved this logical from scheduler.go to nodes.go. Can you help me to review new commit?

ZhengW22:
@Shouren @archlitchi Hello, can you take some time to review my PR?

PR #1158, opened 2025-07-03
Improve the impl of DevicePluginConfigs.Nodeconfig overwriting NvidiaConfig

### LINKED ISSUE #1150: Improve the impl of DevicePluginConfigs.Nodeconfig overwriting NvidiaConfig
Some fields of `NvidiaConfig` in `pkg/device/nvidia/device.go` may be overwritten by the value from `DevicePluginConfigs.Nodeconfig` by the following code:
``` Go
func readFromConfigFile(sConfig *nvidia.NvidiaConfig) (string, error) {
        ......
	for _, val := range deviceConfigs.Nodeconfig {
		if os.Getenv(util.NodeNameEnvName) == val.Name {
			klog.Infof("Reading config from file %s", val.Name)
			if val.Devicememoryscaling > 0 {
				sConfig.DeviceMemoryScaling = val.Devicememoryscaling
			}
			if val.Devicecorescaling > 0 {
				sConfig.DeviceCoreScaling = val.Devicecorescaling
			}
			if val.Devicesplitcount > 0 {
				sConfig.DeviceSplitCount = val.Devicesplitcount
			}
                        ......
		}
	}
	return mode, nil
}
```
However, the current definitions of 
```Go
type DevicePluginConfigs struct {
	Nodeconfig []struct {
		Name                string        `json:"name"`
		OperatingMode       string        `json:"operatingmode"`
		Devicememoryscaling float64       `json:"devicememoryscaling"`
		Devicecorescaling   float64       `json:"devicecorescaling"`
		Devicesplitcount    uint          `json:"devicesplitcount"`
		Migstrategy         string        `json:"migstrategy"`
		FilterDevice        *FilterDevice `json:"filterdevices"`
	} `json:"nodeconfig"`
}
```
And 
```Go
type NvidiaConfig struct {
        ......
	DeviceSplitCount             uint    `yaml:"deviceSplitCount"`
	DeviceMemoryScaling          float64 `yaml:"deviceMemoryScaling"`
	DeviceCoreScaling            float64 `yaml:"deviceCoreScaling"`
	// TODO Whether these should be removed
	DisableCoreLimit  bool                        `yaml:"disableCoreLimit"`
	MigGeometriesList []util.AllowedMigGeometries `yaml:"knownMigGeometries"`
	// GPUCorePolicy through webhook automatic injected to container env
	GPUCorePolicy GPUCoreUtilizationPolicy `yaml:"gpuCorePolicy"`
	// RuntimeClassName is the name of the runtime class to be added to pod.spec.runtimeClassName
	RuntimeClassName string `yaml:"runtimeClassName"`
	// LogLevel is LIBCUDA_LOG_LEVEL value
	LogLevel LibCudaLogLevel `yaml:"libCudaLogLevel"`
}
```
do not make this explicit. For example the `LogLevel` field in `NvidiaConfig` should have been a field in `Nodeconfig` but we failed to realize it during implementation and code review

So we should refactor those code and make it easier to maintain.

FouoF:
**What type of PR is this?**

Fixes #1150

The mergo has been used indirectly, so this pr does not import new third party package.

It changes the config structure of Nodeconfig and NvidiaConfig.

### DIFF
--- a/pkg/device/nvidia/device.go
+++ b/pkg/device/nvidia/device.go
@@ -99,18 +99,16 @@ const (
 )
 
 type NvidiaConfig struct {
-	ResourceCountName            string  `yaml:"resourceCountName"`
-	ResourceMemoryName           string  `yaml:"resourceMemoryName"`
-	ResourceCoreName             string  `yaml:"resourceCoreName"`
-	ResourceMemoryPercentageName string  `yaml:"resourceMemoryPercentageName"`
-	ResourcePriority             string  `yaml:"resourcePriorityName"`
-	OverwriteEnv                 bool    `yaml:"overwriteEnv"`
-	DefaultMemory                int32   `yaml:"defaultMemory"`
-	DefaultCores                 int32   `yaml:"defaultCores"`
-	DefaultGPUNum                int32   `yaml:"defaultGPUNum"`
-	DeviceSplitCount             uint    `yaml:"deviceSplitCount"`
-	DeviceMemoryScaling          float64 `yaml:"deviceMemoryScaling"`
-	DeviceCoreScaling            float64 `yaml:"deviceCoreScaling"`
+	NodeDefaultConfig            `yaml:",inline"`
+	ResourceCountName            string `yaml:"resourceCountName"`
+	ResourceMemoryName           string `yaml:"resourceMemoryName"`
+	ResourceCoreName             string `yaml:"resourceCoreName"`
+	ResourceMemoryPercentageName string `yaml:"resourceMemoryPercentageName"`
+	ResourcePriority             string `yaml:"resourcePriorityName"`
+	OverwriteEnv                 bool   `yaml:"overwriteEnv"`
+	DefaultMemory                int32  `yaml:"defaultMemory"`
+	DefaultCores                 int32  `yaml:"defaultCores"`
+	DefaultGPUNum                int32  `yaml:"defaultGPUNum"`
 	// TODO Whether these should be removed
 	DisableCoreLimit  bool                        `yaml:"disableCoreLimit"`
 	MigGeometriesList []util.AllowedMigGeometries `yaml:"knownMigGeometries"`
@@ -118,8 +116,15 @@ type NvidiaConfig struct {
 	GPUCorePolicy GPUCoreUtilizationPolicy `yaml:"gpuCorePolicy"`
 	// RuntimeClassName is the name of the runtime class to be added to pod.spec.runtimeClassName
 	RuntimeClassName string `yaml:"runtimeClassName"`
+}
+
+// These configs can be sepecified for each node by using Nodeconfig.
+type NodeDefaultConfig struct {
+	DeviceSplitCount    *uint    `yaml:"deviceSplitCount" json:"devicesplitcount"`
+	DeviceMemoryScaling *float64 `yaml:"deviceMemoryScaling" json:"devicememoryscaling"`
+	DeviceCoreScaling   *float64 `yaml:"deviceCoreScaling" json:"devicecorescaling"`
 	// LogLevel is LIBCUDA_LOG_LEVEL value
-	LogLevel LibCudaLogLevel `yaml:"libCudaLogLevel"`
+	LogLevel *LibCudaLogLevel `yaml:"libCudaLogLevel" json:"libCudaLogLevel"`
 }
 
 type FilterDevice struct {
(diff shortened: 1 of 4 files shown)


pkg/device/nvidia/device.go
```diff
@@ -131,13 +136,11 @@ type FilterDevice struct {
...
 type DevicePluginConfigs struct {
 	Nodeconfig []struct {
-		Name                string        `json:"name"`
-		OperatingMode       string        `json:"operatingmode"`
-		Devicememoryscaling float64       `json:"devicememoryscaling"`
-		Devicecorescaling   float64       `json:"devicecorescaling"`
-		Devicesplitcount    uint          `json:"devicesplitcount"`
-		Migstrategy         string        `json:"migstrategy"`
-		FilterDevice        *FilterDevice `json:"filterdevices"`
+		NvidiaConfig  NodeDefaultConfig `json:"nvidiaconfig"`
```
Shouren: 1. Please do not change the current field in this struct, which leads to a backward compatible issue. I think it is better to use `json:,inline` tag to keep it compatible with previous releases. 
2. Please leave a comment here to inform that this config is shared by `NvidiaConfig` struct

pkg/device/nvidia/device.go
```diff
@@ -99,27 +99,32 @@ const (
...
+}
+
+// These configs can be sepecified for each node by using Nodeconfig.
+type NodeDefaultConfig struct {
+	DeviceSplitCount    *uint    `yaml:"deviceSplitCount" json:"devicesplitcount"`
+	DeviceMemoryScaling *float64 `yaml:"deviceMemoryScaling" json:"devicememoryscaling"`
+	DeviceCoreScaling   *float64 `yaml:"deviceCoreScaling" json:"devicecorescaling"`
 	// LogLevel is LIBCUDA_LOG_LEVEL value
-	LogLevel LibCudaLogLevel `yaml:"libCudaLogLevel"`
+	LogLevel *LibCudaLogLevel `yaml:"libCudaLogLevel" json:"libCudaLogLevel"`
```
Shouren: Perhaps using `json:"libcudaloglevel"` for keeping the style of lower case

pkg/device/nvidia/device.go
```diff
@@ -99,27 +99,32 @@ const (
...
-	ResourceMemoryPercentageName string  `yaml:"resourceMemoryPercentageName"`
-	ResourcePriority             string  `yaml:"resourcePriorityName"`
-	OverwriteEnv                 bool    `yaml:"overwriteEnv"`
-	DefaultMemory                int32   `yaml:"defaultMemory"`
-	DefaultCores                 int32   `yaml:"defaultCores"`
-	DefaultGPUNum                int32   `yaml:"defaultGPUNum"`
-	DeviceSplitCount             uint    `yaml:"deviceSplitCount"`
-	DeviceMemoryScaling          float64 `yaml:"deviceMemoryScaling"`
-	DeviceCoreScaling            float64 `yaml:"deviceCoreScaling"`
+	NodeDefaultConfig            `yaml:",inline"`
```
Shouren, line 103: Please leave a comment here to inform that this config is shared by `Nodeconfig` struct

Shouren (changes requested):

pkg/device/nvidia/device.go
FouoF: Yes, I should not break the backward compatibility in this pr. But my original intention is to notice the user that this part of config can be override to reduce the confusion like "Why my config dont work". 
Is there anything like config template that I can put the notice instead?

Shouren: @FouoF This PR #1159 added some docs about the `Nodeconfig`, so if your original purpose is to notice the user, you can update that docs. And for developers who may miss it, some comments for this `inline` struct is enough i think.

FouoF: Thank you, I think #1159 is enough.

archlitchi:
CC @Shouren

pkg/device/nvidia/device.go
```diff
@@ -99,27 +99,33 @@ const (
...
 	MigGeometriesList []util.AllowedMigGeometries `yaml:"knownMigGeometries"`
 	// GPUCorePolicy through webhook automatic injected to container env
 	GPUCorePolicy GPUCoreUtilizationPolicy `yaml:"gpuCorePolicy"`
 	// RuntimeClassName is the name of the runtime class to be added to pod.spec.runtimeClassName
 	RuntimeClassName string `yaml:"runtimeClassName"`
+}
+
+// These configs can be sepecified for each node by using Nodeconfig.
+type NodeDefaultConfig struct {
+	DeviceSplitCount    *uint    `yaml:"deviceSplitCount" json:"devicesplitcount"`
```
Shouren, line 124: @FouoF Can we replace `*<type>` to `<type>` here in `NodeDefaultConfig` struct?

FouoF: Using pointer is necessary for nil pointer can represent that the field is unset so we should not override it. Otherwise values like number "0" or empty string will be arbitrary.

Shouren (approved):
/lgtm

archlitchi (approved):
/approve

PR #1165, opened 2025-07-07
Fix e2e CI

archlitchi:
/kind bug

e2e CI must be modified according to the Chart


archlitchi:
Because https://github.com/Project-HAMi/HAMi/pull/1163 is intend to fix another issue which related to e2e as well, so we have to integrate this PR in order for that e2e to work properly

PR #1169, opened 2025-07-08
remove redundant metrics for vgpu allocation

FouoF:
**What type of PR is this?**

/kind design

The metric vGPUPodsDeviceAllocated docs at https://project-hami.io/zh/docs/userguide/monitoring/device-allocation/ 
is using label to expose core allocation. But it is not a standard usage of metric and not friendly for dashboard like Grafana to display.
This pr removes and replaces it with two metrics that already exist.

This pr change the metrics name and value. But we can discuss how to migrate more user-friendly.


archlitchi:
CC @Shouren

Shouren (approved):
/lgtm

PR #1193, opened 2025-07-11
Support Metax sGPU topology aware

Kyrie336:
**What type of PR is this?**

/kind feature

when sgpu allocate 100%core, it behaves exactly like a exclusive device that can take advantage of the inter-GPU topology. we can obtain better scheduling results by aware the topology.

Fixes #

### DIFF
--- a/charts/hami/templates/scheduler/device-configmap.yaml
+++ b/charts/hami/templates/scheduler/device-configmap.yaml
@@ -97,6 +97,7 @@ data:
       resourceVCountName: {{ .Values.metaxResourceName }}
       resourceVMemoryName: {{ .Values.metaxResourceMem }}
       resourceVCoreName: {{ .Values.metaxResourceCore }}
+      sgpuTopologyAware: {{ .Values.metaxsGPUTopologyAware }}
     enflame:
       resourceCountName: "enflame.com/vgcu"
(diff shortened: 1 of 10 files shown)


archlitchi:
@Kyrie336 please resolve this conflict

archlitchi:
i have an idea: you could put pod.annotations into device.customInfo in the Fit method, then extract them from scoreNode, in that way, you don't need to modify the API anymore

Kyrie336:
> i have an idea: you could put pod.annotations into device.customInfo in the Fit method, then extract them from scoreNode, in that way, you don't need to modify the API anymore

@archlitchi  That's one way to do it, but it's a little strange.  

About the `scoreNode ` API change itself:
before: `ScoreNode(node *corev1.Node, podDevices util.PodSingleDevice, previous []*util.DeviceUsage, policy string) float32`
now: `ScoreNode(node *corev1.Node, podDevices util.PodSingleDevice, previous []*util.DeviceUsage, pod *corev1.Pod) float32`

The scheduler  score phase is evaluating how well a Pod matches a Node. so i think add Pod args is reasonable and scalable. What's your opinion and do you think it's reasonable?

Kyrie336:
@archlitchi i update the commit according to your advise. Please check, thanks

PR #1216, opened 2025-07-16
updated dri section to combine text for better readability

mpetason:
**What type of PR is this?**

/kind documentation

Updating the README so that the complete thought is made before showing the example, instead of splitting up the sentence between the example and the image showing the memory used.

This is just an opinion-based documentation update. The information was already there, just split up instead of combined. I originally thought the issue was that the text wasn't included in the comment in the resource yaml because it didn't make sense on its own. 

Only in that the documentation is updated for formatting to help with clarity.


archlitchi (approved):
/approve

PR #1244, opened 2025-07-31
feat:NodeLockTimeout set from env

miaobyte:
**What type of PR is this?**

/kind feature

This PR adds functionality to configure the node lock timeout duration via an environment variable (`HAMI_NODELOCK_EXPIRE`). This allows users to customize the timeout period according to their specific requirements without needing to modify and recompile the code.

Fixes #

The default timeout remains at 5 minutes when the environment variable is not set or cannot be parsed correctly. When parsing fails, an error log is generated, and the system falls back to the default value.

Yes. Users can now set the HAMI_NODELOCK_EXPIRE environment variable to customize the node lock timeout duration.


miaobyte:
@wawa0210  ok

miaobyte:
@wawa0210 done

charts/hami/values.yaml
```diff
@@ -148,6 +148,7 @@ scheduler:
     extraArgs:
       - --debug
       - -v=4
+  nodeLockExpire: "1m"
```
wawa0210: I recommend keeping the default value consistent with the code. The default value in the HAMi code is 5 minutes.

Custom modifications are possible if required

miaobyte: done

charts/hami/values.yaml
```diff
@@ -227,7 +228,7 @@ scheduler:
   service:
     type: NodePort  # Default type is NodePort, can be changed to ClusterIP
     httpPort: 443   # HTTP port
-    schedulerPort: 31998  # NodePort for HTTP
+    schedulerPort: 31998  31998  # NodePort for HTTP
```
wawa0210: Is this a mistake?

miaobyte: mistake

PR #1249, opened 2025-08-01
feat: add informer-based pod cache to reduce API server load

miaobyte:
# PR Description: Informer-based Pod Cache

**What type of PR is this?**

/kind feature

This PR adds an informer-based pod cache mechanism to the vGPUmonitor component to significantly reduce Kubernetes API server load. The key improvements include:

1. Implemented a local pod cache using Kubernetes Informers that only watches pods on the current node
2. Added event handlers to maintain cache consistency on pod lifecycle events (add/update/delete)
3. Optimized container list refresh with time-based throttling (default 30s interval)
4. Added configurable resync interval via the `HAMI_RESYNC_INTERVAL` environment variable
5. Improved container cleanup logic using cached pod information

**Real-world problem scenario**: When the cluster has more than 50 nodes, the number of vGPUmonitor instances equals the number of nodes, creating excessive pod list query pressure on the kube-apiserver. In our testing, API server CPU usage significantly increased in large clusters. This optimization can reduce this load by more than 90%.

Previously, the vGPUmonitor would frequently make direct API calls to the Kubernetes API server to verify pod existence, which could cause significant load in large clusters. This implementation reduces API calls by maintaining a local cache and only performing full reconciliation at configurable intervals.

Fixes #

The default configuration uses:
- 30-second container list refresh interval
- 5-minute informer resync period

Both values are configurable through environment variables, allowing for tuning based on cluster size and activity level. The pod cache is protected with read-write locks to ensure thread safety while maximizing concurrent access.

Key optimization points:
- Uses fieldSelector filter to only retrieve pods on the current node
- Event-driven cache updates reduce polling requirements
 
No. This is a performance optimization that doesn't change any user-visible behavior or interfaces. The improvement should be transparent to users while reducing load on the Kubernetes API server.

### DIFF
--- a/pkg/monitor/nvidia/cudevshr.go
+++ b/pkg/monitor/nvidia/cudevshr.go
@@ -17,239 +17,386 @@ limitations under the License.
 package nvidia
 
 import (
-	"context"
-	"errors"
-	"fmt"
-	"os"
-	"path/filepath"
-	"strings"
-	"sync"
-	"syscall"
-	"time"
-	"unsafe"
+    "errors"
+    "fmt"
+    "os"
+    "path/filepath"
+    "strings"
+    "sync"
+    "syscall"
+    "time"
+    "unsafe"
 
-	v0 "github.com/Project-HAMi/HAMi/pkg/monitor/nvidia/v0"
-	v1 "github.com/Project-HAMi/HAMi/pkg/monitor/nvidia/v1"
-	"github.com/Project-HAMi/HAMi/pkg/util"
+    v0 "github.com/Project-HAMi/HAMi/pkg/monitor/nvidia/v0"
+    v1 "github.com/Project-HAMi/HAMi/pkg/monitor/nvidia/v1"
+    "github.com/Project-HAMi/HAMi/pkg/util"
 
-	corev1 "k8s.io/api/core/v1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/client-go/kubernetes"
-	"k8s.io/client-go/tools/clientcmd"
-	"k8s.io/klog/v2"
+    corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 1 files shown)


wawa0210:
----    
# PR Description: Informer-based Pod Cache

**What type of PR is this?**

/kind feature

This PR adds an informer-based pod caching mechanism to the vGPUmonitor component, significantly reducing the request load to the Kubernetes API server. Key improvements include:

1. Implement a local cache system that only monitors pods on the current node
2. Added event handler to maintain cache consistency when pod lifecycle events (add/update/delete) occur
3. Optimized container list refresh with time-based throttling (default 30 second interval)
4. Add a configurable resync interval via the `HAMI_RESYNC_INTERVAL` environment variable
5. Improve container cleaning logic using cached pod information

**Practical problem scenario**: When the number of cluster nodes exceeds 50, the number of vGPUmonitors is equal to the number of nodes, and the pressure to query the pod list of kube-apiserver will be too high. In our tests, the CPU usage of API servers in large clusters increased significantly, and this optimization can reduce this load by more than 90%.

Previously, vGPUmonitor frequently called the Kubernetes API server directly to verify that the pod exists, which caused significant load in large clusters. This implementation reduces API calls by maintaining the local cache and performing full coordination only at configurable intervals.

** Which issue(s) this PR fixes**:
Fixes #

The default configuration uses:
- 30 seconds container list refresh interval
- 5 minutes informer resynchronization cycle

Both values can be configured via environment variables, allowing adjustments to be made based on cluster size and activity level. The pod cache is protected by read and write locks to ensure thread safety while maximizing concurrent access.

Key optimization points:
- Use the fieldSelector filter to get only the pods of this node
- Event-driven cache updates reduce polling requirements
 
no. This is a performance optimization that does not change any user-visible behavior or interface. This improvement should be transparent to users while reducing the load on the Kubernetes API server.

Issue #1156, opened 2025-07-02
Refactor the impl of pkg/util/nodelock to remove dependency on `update` verb


Shouren:
**Task description**:
As mentioned in pull request #1152
> The `pkg/utils/nodelock` package is the only package within the HAMi scheduler that depends the `update` verb for `nodes` resource. Since we only update the annotations in node's metadata, i think we can refactor it to use `Patch`.

**Solution**:

1. Please do some research to make sure that replacing `Update` with `Patch` does not change the behavior of current impl.
2. If it is possible to using `Patch`, then refactor the code in this package. 
3. Pass the test in this package.
4. Update the `charts/hami/templates/scheduler/clusterrole.yaml` file after that pull request merged.

**Who can join or take the task**:

The good first issue is intended for `first-time contributors` to get started on his/her contributor journey.

**How to join or take the task**:

Just reply on the issue with the message `/assign` in a separate line.

Then, the issue will be assigned to you.

**How to ask for help**:

If you need help or have questions, please feel free to ask on this issue.
The issue author or other members of the community will guide you through the contribution process.

mayooot:
@Shouren, can you take a look at this [PR](https://github.com/Project-HAMi/HAMi/pull/1192)?

Shouren:
@mayooot Thank you for your PR. I want to know if it is safe to replace `Update` with `Patch` ? For two clients sending a `Patch` request to API Server simultaneously, will the later request failed using `Patch` ?

mayooot:
> [@mayooot](https://github.com/mayooot) Thank you for your PR. I want to know if it is safe to replace `Update` with `Patch` ? For two clients sending a `Patch` request to API Server simultaneously, will the later request failed using `Patch` ?



I don't tink any failures will occur. I guess you're worried that it might fail when the Kubernetes object's `ResourceVersion` faills behind the latest `ResourceVersion`, but that won't happen.

I alway using Patch instead of Update in my work, you can refer this: https://zhuanlan.zhihu.com/p/721919153

PR #1252, opened 2025-08-01
refactor(util/nodelock): replace manual polling with k8s.io/client-go/util/retry

mayooot:
**What type of PR is this?**
/kind cleanup

I replaced manual polling with `k8s.io/client-go/util/retry`, with the purpose of simplifying and improving the process of setting node annotations over and over again.

 none

Especially, I use `retry.RetryOnConflict()` rather than `retry.OnError()`, because we need to handle conflic errors, which occur when the `resourceVersion` is not up to date.

none

### DIFF
--- a/pkg/util/nodelock/nodelock.go
+++ b/pkg/util/nodelock/nodelock.go
@@ -28,19 +28,27 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/util/retry"
 	"k8s.io/klog/v2"
 )
 
 const (
-	NodeLockKey  = "hami.io/mutex.lock"
-	MaxLockRetry = 5
-	NodeLockSep  = ","
+	NodeLockKey = "hami.io/mutex.lock"
+	NodeLockSep = ","
 )
 
 var (
 	lock sync.Mutex
 	// NodeLockTimeout is the global timeout for node locks.
 	NodeLockTimeout time.Duration = time.Minute * 5
+
+	DefaultStrategy = wait.Backoff{
+		Steps:    5,
+		Duration: 100 * time.Millisecond,
+		Factor:   1.0,
+		Jitter:   0.1,
+	}
 )
 
 func SetNodeLock(nodeName string, lockname string, pods *corev1.Pod) error {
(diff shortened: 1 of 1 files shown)


mayooot:
If this PR works well and is bug-free, I'll apply the same changes to other manual polling cases.

peachest:
> **What type of PR is this?这个 PR 是什么类型的？** /kind cleanup  /kind 清理
> 
> **What this PR does / why we need it**:这个 PR 做了什么 / 为什么我们需要它： I replaced manual polling with `k8s.io/client-go/util/retry`, with the purpose of simplifying and improving the process of setting node annotations over and over again.我将手动轮询替换为 `k8s.io/client-go/util/retry` ，目的是简化并改进反复设置节点注释的过程。
> 
> **Which issue(s) this PR fixes**:此 PR 修复了哪些问题： none  无
> 
> **Special notes for your reviewer**:给审查者的特别说明： Especially, I use `retry.RetryOnConflict()` rather than `retry.OnError()`, because we need to handle conflic errors, which occur when the `resourceVersion` is not up to date.特别是，我使用 `retry.RetryOnConflict()` 而不是 `retry.OnError()` ，因为我们需要处理冲突错误，这些错误发生在 `resourceVersion` 不是最新时。
> 
> **Does this PR introduce a user-facing change?**:这个 PR 是否引入了面向用户的变化？ none  无

The code you're wrapping with `retry.RetryOnConflict` uses a patch to update the resource, and the patch mechanism doesn't require the updated resource to have the latest resourceVersion.
As long as the content is legal, the patch will simply merge the patch content into the resource and increment the version number, so the patch will not fail due to a version conflict and will not return a Conflict error, making the use of `retry.RetryOnConflict` redundant.

mayooot:
@peachest No, It's necessary to include `ResourceVersion` when patching a node, It's used to acheive optimistic locking, see: https://github.com/Project-HAMi/HAMi/issues/1156

mayooot:
> > @peachest No, It's necessary to include `ResourceVersion` when patching a node, It's used to acheive optimistic locking, see: #1156
> 
> Sorry, I just realised that the resource version was used in the original patch data. But I still don't think the PR should pass.
> 
> A change in this PR will still **cause the code to behave differently** than before the change. Before the modification, the code will be retried directly regardless of any error returned by the patch, while after the modification, the code will be retried only when a Conflict error occurs. This is the core reason why I think this PR should not be passed.

I've already described this situation in the `Special notes for your reviewer` section.Now we need a maintainer to reviews this PR.

peachest:
> > > @peachest No, It's necessary to include `ResourceVersion` when patching a node, It's used to acheive optimistic locking, see: #1156
> > 
> > 
> > Sorry, I just realised that the resource version was used in the original patch data. But I still don't think the PR should pass.
> > A change in this PR will still **cause the code to behave differently** than before the change. Before the modification, the code will be retried directly regardless of any error returned by the patch, while after the modification, the code will be retried only when a Conflict error occurs. This is the core reason why I think this PR should not be passed.
> 
> I've already described this situation in the `Special notes for your reviewer` section.Now we need a maintainer to reviews this PR.

Yes, I now understand that you want to retry on Conflict Error with `retry.RetryOnFonflict`, but have you considered other kinds of errors? The review of [gemini-code-assist](https://github.com/apps/gemini-code-assist) already hints at this **twice**, and it's possible that the error that is eventually returned is not a Conflict Error. Please click on the following two links to be redirected to the corresponding reviews
- https://github.com/Project-HAMi/HAMi/pull/1252#discussion_r2247712954
- https://github.com/Project-HAMi/HAMi/pull/1252#discussion_r2247712951

I'm just raising my issue that there is no retry for errors other than the Conflict error returned by the patch (e.g. the error returned by Get, and other errors returned by the patch).

pkg/util/nodelock/nodelock.go
```diff
@@ -90,22 +103,27 @@ func ReleaseNodeLock(nodeName string, lockname string, pod *corev1.Pod, timeout
...
-	}
+		if err != nil {
+			klog.ErrorS(err, "Failed to patch node when retry to patch", "node", nodeName)
+			return err
+		}
+		return nil
+	})
 	if err != nil {
-		return fmt.Errorf("releaseNodeLock exceeds retry count %d", MaxLockRetry)
+		return fmt.Errorf("failed to release node lock on node %s: %w", nodeName, err)
```
Shouren: @mayooot I think it is better to record the information of `DefaultStrategy` in the error message.

archlitchi:
please rebase the commit to pass the CI

mayooot:
> please rebase the commit to pass the CI

OK, got it. Please check again.

Issue #1221, opened 2025-07-18
The app in the container uses more GPU memory than the allocated GPU memory, but the container does not crash


christu:
**What happened**:

**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

- The output of `nvidia-smi -a` on your host
- Your docker or containerd configuration file (e.g: `/etc/docker/daemon.json`)
- The hami-device-plugin container logs
- The hami-scheduler container logs
- The kubelet logs on the node (e.g: `sudo journalctl -r -u kubelet`)
- Any relevant kernel output lines from `dmesg`

**Environment**:
- HAMi version:
- nvidia driver or other AI device driver version:
- Docker version from `docker version`
- Docker command, image and tag used
- Kernel version from `uname -a`
- Others:

HAMi version 2.6.0
操作系统：Debian 12
glibc：2.36-9+deb12u10

容器内分配显存为800MB：

<img width="972" height="357" alt="Image" src="https://github.com/user-attachments/assets/3a41d7a7-9785-4234-bd4a-e45e6fdf39de" />

宿主机上查看，app实际使用显存为1096MB：

<img width="943" height="465" alt="Image" src="https://github.com/user-attachments/assets/34702eea-065c-4408-b557-339d71689686" />

容器未崩溃，且在容器内看到的显存使用量为0，是哪里出问题了？

hami版本  v2.6.0
nvidia driver : 570.133.07

<img width="691" height="251" alt="Image" src="https://github.com/user-attachments/assets/c3d86161-d08a-4378-a0cb-98f037e338ce" />

容器内环境变量：

<img width="1063" height="404" alt="Image" src="https://github.com/user-attachments/assets/68676a3d-a8d8-43d5-8dad-941323822b52" />

PR #1256, opened 2025-08-04
Update hami-core version to fix

archlitchi:
/kind bug

https://github.com/Project-HAMi/HAMi/issues/1221


archlitchi:
CC @Nimbus318

PR #1271, opened 2025-08-12
fix: release dangling node lock

### LINKED ISSUE #810: this pr will lead pod schedule hang very long time.  https://github.com/Project-HAMi/HAMi/pull/714
**What happened**:
title 中提到的pr会导致pod的调度很久都不能成功，大量的报错 bind 超时，即使吧extender scheduler的配置超时时间设置成5分钟也会超时。
正常的调度，bind不会超时。

我们的生产环境中待调度的pod 随时都保持在几千个左右。版本是master分之，因为看到pr从2.4.1到master最近合了合多的pr。导致调度慢，卡住，bind超时的原因是 nodelock.go中增加了sycn.mutex，在加减锁的时候吧这两个func都锁了，赶紧吧这个pr重新优化下，把锁去掉

**Environment**:
- HAMi version:
- Others:

peachest:
**What type of PR is this?**

/kind bug

Fixes #714 
Fixes #810 
Fixes #1244 

## **Background**

The background details of the above issues and the history of the fixes are summarised here:

The root of the problem originates from the poor design of the device-plugin framework provided by k8s. In the device-plugin framework, the device-plugin can report the extended resources on the node through the ListAndWatch interface, and handle the initialization of the device through the Allocate interface, as well as mounting the necessary contents for the container to run, including the device file, the control device file, the driver directory, the command line tools, the environment variables, etc. Kubelet will call the Allocate interface of the device-plugin after the Pod is bound to the node. After the Pod is bound to a node, Kubelet calls the Allocate interface of the device-plugin to set up and mount the response for the container based on the returned information.

If the device plugin only needs to manage physical hardware devices, and each device is exclusive to the container (i.e., a device can only be mounted to one container at a time), then all of the above works very well. Because in this case each device is stateless, each device corresponds to a unique extension resource reported by the device plugin via ListAndWatch and has a unique ID; when the Allocate interface is called, kubelet selects the appropriate resource among the unallocated resources and passes the corresponding ID List as a request parameter. The plugin simply selects the corresponding hardware devices based on the ID List for container.

Until the need to manage virtualised device resources arises, at which point each physical device is no longer stateless, using HAMi's nvidia share as an example. Eeach Nvidia device is divided into 100 for sharing. When scheduling Pods, the scheduler not only needs to determine the Node for Pod, but also needs to decide the specific physical device to be allocated to each container based on the remaining available resources of each device on the node, so the actual device that the container needs to mount is no longer determined by the kubelet or the device plugin, but rather by the scheduler's scheduling results. The scheduler can easily record the scheduling result to the corresponding Pod, e.g. by writing it to the Pod annotation during the bind phase, as implemented in HAMi.

On the one hand, as mentioned above, the interface of the device plugin framework is stateless from the very beginning of its design, and device plugins cannot get any information about the Pod or container that needs to be assigned a device through the parameters of the Allocate interface provided by the framework. On the other hand, the k8s scheduling mechanism divides Pod scheduling into a serial Filter phase and a parallel Bind phase, where different Pods may be bound to the same node at the same time, so the device plugin can't get the Pod info which is currently being allocated by kubelet by geting pods bound to the node.

HAMi chooses to use NodeLock to serialise Pod bindings and device Allocating, by adding a NodeLock to Node Annotations so that at most one Pod is bound to the node at a time and processed by the device plugin called by Kubelet. The NodeLock records the Pod, timestamp, and other information, and the device plugin parses it to get the Pod that is currently being processed and gets the scheduler's scheduling result through the Pod annotation to complete the allocation of the device. When the device plugin finishes allocating, it removes the NodeLock. So that next pod can be bound by scheduler and allocated by device plugin.

The initial NodeLock, while recording the Pod that set the NodeLock, did not restrict which pod can release the lock and thus could lead to Pod B accidentally releasing the nodeLock set by Pod A, and further lead to multiple Pods pending. The PR #714 adds an owner check for the lock. Also, it allows skipping the owner check and releasing directly in case the lock expires. So after the PR #714 fix, the only way Pod B can release a lock set by Pod A is if the lock expires.

Subsequently, Issue #810 proposed that PR #714 would cause Pod B to be pending for a long time waiting for Pod A's NodeLock to be expired and released. Therefore, in order to reduce the waiting time, PR #1244 proposes to allow the NodeLock timeout to be set via environment variables, and change the original 5min to user configurable.

## Problem Description

The NodeLock will become dangling due to the owner Pod being accidentally deleted between the time the scheduler sets a NodeLock with this Pod as the owner in the bind phase and the time device plugin finish allocating Pod and release nodelock.
All subsequent Pods that are scheduled to the same node are unable to release the dangling NodeLock by any means other than waiting for the NodeLock to expire.

## What this PR does

This PR fixes the problem described above.
There are two main fixes: 
1. add `goto ReleaseNodeLocks` to the PatchPodAnnotations step in the Bind function implementation to clear the NodeLock when an error occurs
2. allow removal of the dangling NodeLock even if not expired

NodeLock can now be deleted by non-holders (Pods) under timeout or dangling conditions.

## Problem Reproducing

Since PatchPodAnnotation does not release the NodeLock when an error occurs, if a pod is deleted immediately after the NodeLock is set at the time of binding, and a new pod needs to be bound to the same node within 5 minutes, the new pod will be pending, and the event `binding reject: node <nodename> has been locked within 5 minutes` will arise when inspecting pod via `kubectl describe`

The above problem can **be reproduced stably** using the following code. This code deletes the owner Pod as soon as a NodeLock is detected, so that the NodeLock will become dangling and persist on the node untile expire, and other Pods have to wait for its timeout.
```go
package main

import (
	"context"
	"flag"
	"fmt"
	"os"
	"strings"
	"time"

	"github.com/pkg/errors"
	"github.com/sirupsen/logrus"
	corev1 "k8s.io/api/core/v1"
	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
	"k8s.io/client-go/informers"
	"k8s.io/client-go/kubernetes"
	"k8s.io/client-go/tools/cache"
	ctrl "sigs.k8s.io/controller-runtime"
)

const (
	NodeLockKey  = "hami.io/mutex.lock"
	MaxLockRetry = 5
	NodeLockSep  = ","
)

// Using Global Variables to Simplify Code
var (
	// NodeName is the name of the node to watch for locks
	NodeName   string
	logger     *logrus.Logger
	kubeClient kubernetes.Interface
	ctx        context.Context
)

func main() {
	flag.StringVar(&NodeName, "node", "zzdev39", "The node name to handle")
	flag.Parse()

	fmt.Printf("Node lock watcher started for node: %s\n", NodeName)

	logger = logrus.New()
	SetupInformer()

	// Clean possible node lock before starting
	if err := releaseNodeLock(); err != nil {
		logger.WithError(err).Errorf("Failed to release node lock for node %s", NodeName)
		os.Exit(1)
	}

	// Block until CTRL-C
	<-ctx.Done()
	logger.Info("Shutting down gracefully...")

	// Release node lock before exiting
	if err := releaseNodeLock(); err != nil {
		logger.WithError(err).Errorf("Failed to release node lock for node %s", NodeName)
		os.Exit(1)
	}
	os.Exit(0)
}

func SetupInformer() {
	config := ctrl.GetConfigOrDie()
	kubeClient = kubernetes.NewForConfigOrDie(config)
	kubeFactory := informers.NewSharedInformerFactory(kubeClient, 10*time.Minute)
	nodeInformer := kubeFactory.Core().V1().Nodes()

	ctx = ctrl.SetupSignalHandler()

	// Listen to nodelock added
	nodeInformer.Informer().AddEventHandler(GetNodeLockUpdateHandler(ctx))

	kubeFactory.Start(ctx.Done())
	if !cache.WaitForCacheSync(
		ctx.Done(),
		nodeInformer.Informer().HasSynced,
	) {
		logger.Errorf("failed to sync informers")
		os.Exit(1)
	}
}

// GetNodeLockUpdateHandler returns a ResourceEventHandler that
// delete the pod when the node lock annotation added to trigger NodeLock 5min issue
func GetNodeLockUpdateHandler(ctx context.Context) cache.ResourceEventHandler {
	return cache.ResourceEventHandlerFuncs{
		UpdateFunc: func(oldObj, newObj any) {
			oldNode := oldObj.(*corev1.Node)
			newNode := newObj.(*corev1.Node)

			if oldNode.Name != NodeName {
				return // Ignore updates for other nodes
			}

			if oldNode.ResourceVersion == newNode.ResourceVersion {
				return // No change
			}

			var oldVal, newVal string
			var oldExists, newExists bool
			if oldNode.Annotations != nil {
				oldVal, oldExists = oldNode.Annotations[NodeLockKey]
			}

			if newNode.Annotations != nil {
				newVal, newExists = newNode.Annotations[NodeLockKey]
			}

			if oldVal == newVal {
				return // No change in lock annotation
			}
			logger.WithFields(logrus.Fields{
				"old_lock": oldVal,
				"new_lock": newVal,
				"node":     newNode.Name,
			}).Info("Node lock annotation changed")

			if !oldExists && newExists {
				// New lock added
				_, ns, name, err := ParseNodeLock(newVal)
				if err != nil {
					logger.WithError(err).WithField("node", newNode.Name).
						Error("Failed to parse new lock annotation")
					return
				}
				// Parse lock info and delete pod
				kubeClient.CoreV1().Pods(ns).Delete(ctx, name, metav1.DeleteOptions{})
			}
		},
	}
}

// Parse Node Lock like:
//
// hami.io/mutex.lock: 2025-07-24T10:12:33Z,hyx,ubuntu-deployment-86d59cb8d-7xpnm
//
// Refer to: https://github.com/Project-HAMi/HAMi/blob/master/pkg/util/nodelock/nodelock.go#L132-L155
func ParseNodeLock(value string) (lockTime time.Time, ns, name string, err error) {
	if !strings.Contains(value, NodeLockSep) {
		lockTime, err = time.Parse(time.RFC3339, value)
		return lockTime, "", "", err
	}
	s := strings.Split(value, NodeLockSep)
	if len(s) != 3 {
		lockTime, err = time.Parse(time.RFC3339, value)
		return lockTime, "", "", err
	}
	lockTime, err = time.Parse(time.RFC3339, s[0])
	return lockTime, s[1], s[2], err
}

// releaseNodeLock releases the lock on the node by removing the annotation
func releaseNodeLock() error {
	// Release node
	var node *corev1.Node
	var err = errors.New("")
	ctx := context.Background()
	for i := 0; i < MaxLockRetry && err != nil; i++ {
		node, err = kubeClient.CoreV1().Nodes().Get(ctx, NodeName, metav1.GetOptions{})
		if err != nil {
			logger.WithError(err).Errorf("Failed to get node %s", NodeName)
			time.Sleep(100 * time.Millisecond)
			continue
		}
		delete(node.Annotations, NodeLockKey)
		_, err = kubeClient.CoreV1().Nodes().Update(ctx, node, metav1.UpdateOptions{})
	}
	logger.Info("Node lock released successfully")
	return err
}
```

Usage：
```shell
go build -o test-node-lock main.go
# Use sudo to read KUBECONFIG
sudo ./test-node-lock --node=<your node>
```

Check the UUID of GPUs on the node:
```shell
nvidia-smi -L
# or
nvidia-smi --query-gpu=uuid --format=csv
```

Then deploy using Depolyment and limit the scheduling scope to the corresponding node with use-uuid:
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
    name: ubuntu-deployment
spec:
    replicas: 2
    selector:
        matchLabels:
            app: ubuntu
    template:
        metadata:
            labels:
                app: ubuntu
                nvidia.com/use-gpuuuid: <GPU1>,<GPU2>
        spec:
            containers:
              - name: ubuntu-cnt-0
                image: ubuntu:22.04
                imagePullPolicy: IfNotPresent
                command: [ "sh", "-c", "sleep infinity" ]
                resources:
                    limits:
                       nvidia.com/gpu: 1
```

第一个 Pod 被调度时，会在设置 NodeLock 时被删除，NodeLock 会保持在 Node 上没有被删除。然后 Deployment 自动创建出新的 Pod，这些新的 Pod 必定会处于 pending 状态，并且出现事件 `binding reject: node <nodename> has been locked within 5 minutes`

When the first pod is scheduled and nodeLock has been set, the code will delete the pod immediately, and new pods will be pending, and the event `binding reject: node <nodename> has been locked within 5 minutes` will arise when inspecting pod via `kubectl describe`

## Notes

This PR only eliminate the problem of NodeLock becoming dangling due to an accidental deletion of a Pod, resulting in other Pods having to wait for the NodeLock to timeout and remain in a pending state.

There are still other scenarios that can cause a Pod to be pending due to NodeLock, such as having a large number of Pods on the same node that need to be bound at the same time, each of which needs to wait for the previous Pod to be processed before it can enter its own binding process.

No, this PR doesn't change any user-facing nor function signature.
However, we do change the name of a single parameter from `timeout` to `skipNodeLockOwnerCheck` in function `ReleaseNodeLock` for better readability and comprehensibility


archlitchi:
CC @Shouren

archlitchi:
please fix the UT

Shouren (approved):
/lgtm

archlitchi:
please resolve this conflict, we're ready to go:)

peachest:
We've synchronised to the latest master branch and resolved the conflict.

PR #1273, opened 2025-08-14
enhance scheduler logging with dedicated API endpoint

Wangmin362:
**What type of PR is this?**

/kind feature

- Introduce a new HTTP API endpoint `/api/v1/namespaces/{namespace}/pods/{name}/schedulerlog` to expose detailed pod scheduling logs.
- Implement a thread-safe LRU cache (`SchedulerLogCache`) to store and manage scheduling diagnostics for pods, with configurable maximum pod count.
- Design a structured JSON response format to clearly report scheduling outcomes, including:
  - Overall scheduling status
  - Per-container and per-node evaluation results with failure reasons or scores
  - Filter and bind phase summaries
- Add support for real-time log updates during the scheduling process, allowing incremental writes to container and node results.
- Enable querying logs by pod name and namespace for precise troubleshooting.
- Significantly improve user experience by providing a clear, accessible way to diagnose why a pod failed to schedule, especially for users unfamiliar with K8s internals.

Fixes #

no


wawa0210:
----    
cc @S

docs/pod-scheduling-diagnostics-enhancement.md
```diff
@@ -0,0 +1,285 @@
...
+
+This is particularly problematic in production environments where users need immediate feedback on why their AI training or inference workloads failed to schedule.
+
+## Proposed Solution
+
+We propose introducing a dedicated API endpoint that provides structured scheduling diagnostics for Pods, with a focus on making failure reasons clear and actionable for end users.
+
+### API Endpoint
+```
+GET /api/v1/namespaces/{namespace}/pods/{name}/schedulerlog
```
Shouren, line 50: @Wangmin362 Is there a way to get scheduler log for all pods in a namespace ?

charts/hami/templates/scheduler/deployment.yaml
```diff
@@ -97,6 +97,8 @@ spec:
           command:
             - scheduler
             - --http_bind=0.0.0.0:443
+            - --scheduler-log-http_bind={{ .Values.scheduler.schedulerLogServerBindAddress }}
+            - --max-cached-pods={{ .Values.scheduler.maxCachePod }}
```
Shouren: @Wangmin362 I prefer to naming this option to `scheduler-log-max-cached-pods`

charts/hami/values.yaml
```diff
@@ -69,6 +69,17 @@ scheduler:
...
+  # Format: ":<port>" or "host:<port>". Use ":<port>" (e.g., ":9396") to bind to all network interfaces on the specified port.
+  # Default: ":9396"
+  schedulerLogServerBindAddress: ":9396"
+  # @param maxCachePod defines the maximum number of Pod scheduling logs to cache.
+  # The scheduler stores logs for recently scheduled Pods in memory to serve the
+  # /api/v1/pod/{namespace}/{name}/schedulerlog API. This parameter controls the
+  # upper limit of this in-memory cache. When the number of cached Pods exceeds
+  # this value, the oldest entries are automatically evicted (LRU policy).
+  # Default: 1000
+  maxCachePod: 1000
```
Shouren: @Wangmin362 Because this options is in the scope of scheduler log, i prefer to rename it to `schedulerLogMaxCachePod`

pkg/scheduler/scheduler.go
```diff
@@ -59,14 +59,18 @@ type Scheduler struct {
 	overviewstatus map[string]*NodeUsage
 
 	eventRecorder record.EventRecorder
+
+	// Scheduler log cache
+	schedulerLogCache *SchedulerLogCache
```
Shouren, line 64: @Wangmin362 Why we need to define `schedulerLogCache` as a point here ?

docs/pod-scheduling-diagnostics-enhancement.md
Wangmin362: We currently do not plan to provide an API for fetching logs of all Pods at once. We believe this endpoint is not essential, as there is generally no such requirement. The typical use case is to query the scheduling logs for a specific Pod, rather than retrieving logs for all Pods.

charts/hami/templates/scheduler/deployment.yaml
Wangmin362: Alright, I have modified it.

charts/hami/values.yaml
Wangmin362: Alright, I have modified it.

pkg/scheduler/scheduler.go
Wangmin362: Great question! The reason we define `schedulerLogCache` as a pointer (`*SchedulerLogCache`) is primarily due to **Go's method set rules and the principle of value vs. pointer receivers**.

Here's the detailed explanation:

1.  **The `SchedulerLogCache` has methods with pointer receivers**:
    The `SchedulerLogCache` struct is designed to be a stateful, concurrent-safe cache. Its core methods, such as `SetStatus`, `AddNodeResult`, `Remove`, etc., **modify the internal state** of the cache (e.g., updating the `items` map, manipulating the `oldest` list).
    ```go
    func (c *SchedulerLogCache) SetStatus(namespace, podName, status string) {
        c.mu.Lock() // Modifying c.mu
        defer c.mu.Unlock()
        // ... modifying c.items and c.oldest ...
    }
    ```
    Because these methods need to mutate the struct's fields, they must use a **pointer receiver** (`*SchedulerLogCache`).

2.  **Consistency in the Method Set**:
    In Go, if *any* method on a type uses a pointer receiver, it's a strong convention to use pointer receivers for *all* methods on that type for consistency.
    More importantly, **only the pointer to the type has the complete method set**. A value of the type only has methods with value receivers.

3.  **The `Scheduler` struct needs to call mutating methods**:
    The `Scheduler` will need to call methods like `schedulerLogCache.SetStatus(...)` during the scheduling process. Since `SetStatus` has a pointer receiver, you can only call it on a **pointer to** `SchedulerLogCache`.

Therefore, if `schedulerLogCache` were a value (`SchedulerLogCache`), the `Scheduler` would be holding a copy. To call the mutating methods, Go would need to take the address of that copy, which is not idiomatic and can lead to confusion. By defining it as a pointer from the start, we make it clear that `Scheduler` holds a reference to a single, shared `SchedulerLogCache` instance, and all operations on it are performed through its pointer-based API.

Wangmin362:
@archlitchi  can you help me rerun e2e-test? there are some error about Loading Docker image from image.tar.

wawa0210:
This PR has a conflict, please resolve it as soon as possible

@Wangmin362

Wangmin362:
> This PR has a conflict, please resolve it as soon as possible
> 
> @Wangmin362

i have already push another PR,  i will close this one

PR #1296, opened 2025-08-25
Optimize pkg/util.go and distribute logics to corresponding logics 

archlitchi:
/king cleanup

Right now, pkg/util contains logics from various packages and defines types for different packages.

It is not a good practice because such types should be defined in their own package and be self-sufficient. util package should be a toolbox and shouldn't contains any core logic. 

For example util.deviceUsage, util.deviceInfo should be defined in device package.

So i transfer struct and logics defined in pkg/util to corresponding packages,  to be more specific:

device specific logics and types in pkg/util ----------> pkg/device
config related logics in pkg/device ------> pkg/scheduler/config
cleanup grpc-related code defined in cmd/vGPUmonitor


archlitchi:
CC @Shouren

wawa0210:
----    
cc @S

pkg/device-plugin/nvidiadevice/nvinternal/plugin/server.go
```diff
@@ -57,10 +57,10 @@ import (
 	kubeletdevicepluginv1beta1 "k8s.io/kubelet/pkg/apis/deviceplugin/v1beta1"
 	cdiapi "tags.cncf.io/container-device-interface/pkg/cdi"
 
-	"github.com/Project-HAMi/HAMi/pkg/device"
 	"github.com/Project-HAMi/HAMi/pkg/device-plugin/nvidiadevice/nvinternal/cdi"
 	"github.com/Project-HAMi/HAMi/pkg/device-plugin/nvidiadevice/nvinternal/rm"
 	"github.com/Project-HAMi/HAMi/pkg/device/nvidia"
+	"github.com/Project-HAMi/HAMi/pkg/scheduler/config"
```
Shouren, line 63: @archlitchi I prefer to using import alias for `config` package here. I think we can pick `schconfig`, `schcfg` or something else for indicating that this config from scheduler

PR #1354, opened 2025-09-17
feat(nvidia): default gpucores=100 when memory is exclusive and cores…

xrwang8:
**Title**: nvidia: default gpucores=100 when memory is exclusive and cores unset

- When a container requests GPUs and the memory is exclusive (gpumem-percentage=100, or not set while defaultMem=0 so it falls back to 100%), and the user did not specify nvidia.com/gpucores, set gpucores to 100 automatically.
- Keeps defaultCores unchanged (0 by default), so non-exclusive memory cases (e.g., gpumem-percentage: 50) remain shareable and do NOT get cores=100.
- Aligns semantics and metrics: requesting only nvidia.com/gpu now implies core=100 in exclusive-memory scenarios, so GPUDeviceCoreAllocated/vGPUCoreAllocated show 100.

Docs: explain the conditional-default behavior in config.md and config_cn.md.

When this kicks in

  - Pods must request the GPU count resource (config.ResourceCountName, default nvidia.com/gpu) in limits or requests; otherwise the admission hook skips them.
  - The GPU core resource (config.ResourceCoreName, default nvidia.com/gpucores) must be absent from both limits and requests; explicit values are never overwritten.
  - Memory is considered exclusive if any of the following is true:
      1. config.ResourceMemoryPercentageName (default nvidia.com/gpumem-percentage) is present and equals 100.
      2. config.ResourceMemoryName (default nvidia.com/gpumem) is configured as an empty string.
      3. No memory resource is set and config.ResourceMemoryName is left at its default (HAMi falls back to 100% memory).

  If all three bullets above hold, MutateAdmission injects the core resource into Resources.Limits with the value 100. Otherwise the container spec is left untouched.

  Example specs

  1. Exclusive via percentage:

     resources:
       limits:
         nvidia.com/gpu: 1
         nvidia.com/gpumem-percentage: 100
     # gpucores gets injected as 100
  2. No explicit memory fields (defaults to 100%):

     resources:
       limits:
         nvidia.com/gpu: 1
     # gpucores defaults to 100
  3. Percentage only in requests:

     resources:
       requests:
         nvidia.com/gpu: 1
         nvidia.com/gpumem-percentage: 100
       limits:
         nvidia.com/gpu: 1
     # admission writes gpucores: 100 into limits
  4. Shareable memory (no change):

     resources:
       limits:
         nvidia.com/gpu: 1
         nvidia.com/gpumem-percentage: 50
     # gpucores remains unset
  5. Custom resource names:

     resources:
       limits:
         hami.io/gpu: 1
         hami.io/gpumem-percentage: 100
     # gpucores injected as 100 using the custom name

  Notes

  - Works with and without MIG; the logic only checks whether the request is effectively exclusive.
  - Explicit gpucores in either limits or requests takes precedence—no override happens.
  - The default is applied during admission; changes require recreating or reapplying the pod spec.

**What type of PR is this?**

/kind feature

Align core allocation semantics and metrics with effective exclusivity when users only request `nvidia.com/gpu`. If a container ends up with exclusive memory and did not set `nvidia.com/gpucores`, default `gpucores` to 100. This makes `GPUDeviceCoreAllocated` and `vGPUCoreAllocated` consistent with the allocation outcome while preserving shareability for non‑exclusive memory cases.

Fixes #

- Implementation:
  - `pkg/device/nvidia/device.go` (`GenerateResourceRequests`): Detect if `nvidia.com/gpucores` is explicitly set. If not and `gpumem-percentage` is 100 (or falls back to 100 when unset and `defaultMem==0`), set `corenum=100`. Otherwise, keep existing behavior.
  - Docs: add notes to `docs/config.md` and `docs/config_cn.md` describing the conditional default.
- Backward compatibility:
  - No change when `nvidia.com/gpucores` is explicitly set.
  - No change for non‑exclusive memory (e.g., `gpumem-percentage: 50`); such pods remain shareable and do not get `gpucores=100`.
  - `nvidia.defaultCores` remains unchanged (0 by default).
- Interactions:
  - MIG: Works as expected; core=100 applies only when memory is exclusive and cores are unset.
  - Memory oversubscription: “Exclusive” uses registered memory (after scaling), preserving consistent semantics.

Yes. When memory is exclusive and cores are unset, HAMi now defaults `nvidia.com/gpucores` to 100. Non‑exclusive memory requests remain shareable and unchanged. Documentation updated accordingly.

### DIFF
--- a/pkg/device/nvidia/device.go
+++ b/pkg/device/nvidia/device.go
@@ -320,6 +320,9 @@ func (dev *NvidiaGPUDevices) MutateAdmission(ctr *corev1.Container, p *corev1.Po
 	}
 
 	hasResource := dev.mutateContainerResource(ctr)
+	if dev.defaultExclusiveCoreIfNeeded(ctr) {
+		hasResource = true
+	}
 
 	if hasResource {
 		// Set runtime class name if it is not set by user and the runtime class name is configured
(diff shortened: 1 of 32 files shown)


Shouren (commented):
@xrwang8 As it is mentiond that 'HAMi will default `nvidia.com/gpucores` to 100 for that container', is it better to implement this feature in `pkg/scheduler/webhook.go` ?

Shouren:
> @xrwang8 As it is mentiond that 'HAMi will default `nvidia.com/gpucores` to 100 for that container', is it better to implement this feature in `pkg/scheduler/webhook.go` ?

@xrwang8 I should make it clear that i want you to use the webhook mechanism not implement this feature in `webhook.go` right in that file.
And there is a `MutateAdmission` interface for device to mutate the pod, so please migrate you code to `func (dev *NvidiaGPUDevices) MutateAdmission(ctr *corev1.Container, p *corev1.Pod) (bool, error)` of `pkg/device/nvidia/device.go`

xrwang8:
@Shouren   I’ve rebased on the latest upstream master and reworked the core defaulting exactly as requested: the logic now lives in NvidiaGPUDevices.MutateAdmission

wawa0210:
----    
cc @S

pkg/device/nvidia/device.go
```diff
@@ -356,6 +359,67 @@ func (dev *NvidiaGPUDevices) mutateContainerResource(ctr *corev1.Container) bool
...
+	}
+
+	if !exclusive {
+		return false
+	}
+
+	if ctr.Resources.Limits == nil {
+		ctr.Resources.Limits = corev1.ResourceList{}
+	}
+	ctr.Resources.Limits[coreName] = *resource.NewQuantity(100, resource.BinarySI)
```
Shouren: @xrwang8 I think is should be `resource.DecimalSI` rather than `resource.BinarySI` for cores

Shouren (commented):
@archlitchi Shall we add a switch in helm chart for this feature?

archlitchi:
> @archlitchi Shall we add a switch in helm chart for this feature?

i think no need, because i can't imagine a scenario which we shouldn't set core to 100

pkg/device/nvidia/device.go
xrwang8: Thanks for pointing that out — DecimalSI is indeed the right fit for a GPU core count. I’ve updated the change to use resource.NewQuantity(100, resource.DecimalSI)

archlitchi (approved):
/lgtm

Issue #1330, opened 2025-09-08
vGPU count on single GPU  is over the splitcount limit


climby:
###  **What happened**:

create 1vGPU request pod with  10 concurrents/sec.  
the vGPU distribution on each GPU is not balance.  Here is the metrics from hame-scheduler:31993/metrics
-----------------------

`

# HELP GPUDeviceSharedNum Number of containers sharing this GPU
# TYPE GPUDeviceSharedNum gauge
GPUDeviceSharedNum{deviceidx="0",deviceuuid="GPU-17a8e6e5-16c0-621d-e674-7ffc89498e4d",nodeid="gn-10-245-36-20",zone="vGPU"} 11
GPUDeviceSharedNum{deviceidx="0",deviceuuid="GPU-c780634e-8a90-c7d6-4da2-3f400a364ba8",nodeid="gn-10-245-36-34",zone="vGPU"} 9
GPUDeviceSharedNum{deviceidx="1",deviceuuid="GPU-191941e0-7b6d-34c4-4e69-72169b3ff7fe",nodeid="gn-10-245-36-20",zone="vGPU"} 9
GPUDeviceSharedNum{deviceidx="1",deviceuuid="GPU-94d5df55-a5f6-a0d9-3d52-f782ab39373a",nodeid="gn-10-245-36-34",zone="vGPU"} 10
GPUDeviceSharedNum{deviceidx="2",deviceuuid="GPU-34d69f5c-7b8e-87a8-5678-e00c9eecd4a5",nodeid="gn-10-245-36-20",zone="vGPU"} 8
GPUDeviceSharedNum{deviceidx="2",deviceuuid="GPU-feec1ca5-d9cf-4d8b-e87b-a7a3f3155d7f",nodeid="gn-10-245-36-34",zone="vGPU"} 8
GPUDeviceSharedNum{deviceidx="3",deviceuuid="GPU-0778ac95-7f1f-2150-a1ad-5bda2266f6a5",nodeid="gn-10-245-36-20",zone="vGPU"} 11
GPUDeviceSharedNum{deviceidx="3",deviceuuid="GPU-e1d81879-e979-7d5a-1c8b-c2582826dfd6",nodeid="gn-10-245-36-34",zone="vGPU"} 5
GPUDeviceSharedNum{deviceidx="4",deviceuuid="GPU-8e04e647-05f1-b1bf-c2f7-184b0eb41c11",nodeid="gn-10-245-36-34",zone="vGPU"} 9
[...]

PR #1359, opened 2025-09-22
feat: add NVIDIA Resourcequota

FouoF:
**What type of PR is this?**

/kind feature

This pr add resource quota support for Nvidia multi-device request and whole device request.

Fixes #

### DIFF
--- a/pkg/device/enflame/device.go
+++ b/pkg/device/enflame/device.go
@@ -327,3 +327,11 @@ func (enf *EnflameDevices) Fit(devices []*device.DeviceUsage, request device.Con
 	}
 	return false, tmpDevs, common.GenReason(reason, len(devices))
 }
+
+func (dev *EnflameDevices) GetResourceNames() device.ResoureNames {
+	return device.ResoureNames{
+		ResourceCountName:  EnflameResourceNameVGCU,
+		ResourceMemoryName: EnflameResourceNameVGCUPercentage,
+		ResourceCoreName:   "",
+	}
+}
--- a/pkg/device/enflame/device.go
+++ b/pkg/device/enflame/device.go
@@ -35,8 +35,8 @@ type EnflameDevices struct {
 }
 
 const (
-	EnflameGPUDevice     = "Enflame"
-	EnflameGPUCommonWord = "Enflame"
+	EnflameVGCUDevice     = "Enflame"
+	EnflameVGCUCommonWord = "Enflame"
 	// IluvatarUseUUID is user can use specify Iluvatar device for set Iluvatar UUID.
 	EnflameUseUUID = "enflame.com/use-gpuuuid"
 	// IluvatarNoUseUUID is user can not use specify Iluvatar device for set Iluvatar UUID.
@@ -54,14 +54,17 @@ const (
 func InitEnflameDevice(config EnflameConfig) *EnflameDevices {
 	EnflameResourceNameVGCU = config.ResourceNameVGCU
 	EnflameResourceNameVGCUPercentage = config.ResourceNameVGCUPercentage
-	device.SupportDevices[EnflameGPUDevice] = "hami.io/enflame-vgpu-devices-allocated"
+	_, ok := device.SupportDevices[EnflameVGCUDevice]
+	if !ok {
+		device.SupportDevices[EnflameVGCUDevice] = "hami.io/enflame-vgpu-devices-allocated"
+	}
 	return &EnflameDevices{
 		factor: 0,
 	}
 }
 
 func (dev *EnflameDevices) CommonWord() string {
-	return EnflameGPUCommonWord
+	return EnflameVGCUCommonWord
 }
 
 func (dev *EnflameDevices) MutateAdmission(ctr *corev1.Container, p *corev1.Pod) (bool, error) {
@@ -111,7 +114,7 @@ func (dev *EnflameDevices) GetNodeDevices(n corev1.Node) ([]*device.DeviceInfo, 
 			Count:   100,
 			Devmem:  100,
 			Devcore: 100,
-			Type:    EnflameGPUDevice,
+			Type:    EnflameVGCUDevice,
 			Numa:    0,
 			Health:  true,
 		})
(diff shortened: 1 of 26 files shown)


wawa0210:
----    
cc @archlitchi

pkg/device/enflame/device.go
```diff
@@ -327,3 +327,11 @@ func (enf *EnflameDevices) Fit(devices []*device.DeviceUsage, request device.Con
 	}
 	return false, tmpDevs, common.GenReason(reason, len(devices))
 }
+
+func (dev *EnflameDevices) GetResourceNames() device.ResoureNames {
+	return device.ResoureNames{
+		ResourceCountName:  EnflameResourceNameVGCU,
+		ResourceMemoryName: EnflameResourceNameVGCUPercentage,
```
Shouren, line 337: @FouoF Is it right to use  `EnflameResourceNameVGCUPercentage` as `ResourceMemoryName`?

pkg/scheduler/scheduler.go
```diff
@@ -122,6 +157,7 @@ func (s *Scheduler) Start() {
 	informerFactory := informers.NewSharedInformerFactoryWithOptions(s.kubeClient, time.Hour*1)
 	s.podLister = informerFactory.Core().V1().Pods().Lister()
 	s.nodeLister = informerFactory.Core().V1().Nodes().Lister()
+	s.quotaLister = informerFactory.Core().V1().ResourceQuotas().Lister()
```
Shouren, line 160: @FouoF I think we should update `charts/hami/templates/scheduler/clusterrole.yaml` to make this Lister works.

FouoF: Added it.

pkg/device/nvidia/device.go
```diff
@@ -659,6 +670,11 @@ func (nv *NvidiaGPUDevices) Fit(devices []*device.DeviceUsage, request device.Co
 			//This incurs an issue
 			memreq = dev.Totalmem * k.MemPercentagereq / 100
 		}
+		if !fitQuota(tmpDevs, pod.Namespace, int64(memreq), int64(k.Coresreq)) {
```
Shouren, line 676: @FouoF I doubt that if this is the right place to check quota. 

According to the [doc](https://github.com/kubernetes/design-proposals-archive/blob/main/resource-management/admission_control_resource_quota.md#admissioncontrol-plugin-resourcequota):
> The ResourceQuota plug-in introspects all incoming admission requests.

and it seem that the ResourceQuota admission plugin is enable by default, so that the check should be in Admission Webhook rather than the Filter phase of scheduler.

pkg/device/enflame/device.go
archlitchi: yes, solved

pkg/device/nvidia/device.go
archlitchi: because you don't know the exact number of specs(like gpumemory, gpucores) you are going to use in webhook.
for example, you can allocate a whole card by only assign 'nvidia.com/gpu:1', but you don't know which card you will be using before scheduling

archlitchi:
@FouoF I think it's better to add this quota usage to scheduler monitor

pkg/device/nvidia/device.go
Shouren: @archlitchi Yes, it's really hard to know the extract resource before scheduling.

FouoF:
> @FouoF I think it's better to add this quota usage to scheduler monitor

added

pkg/scheduler/scheduler.go
```diff
@@ -96,7 +102,9 @@ func (s *Scheduler) onAddPod(obj any) {
 		return
 	}
 	podDev, _ := device.DecodePodDevices(device.SupportDevices, pod.Annotations)
-	s.addPod(pod, nodeID, podDev)
+	if s.addPod(pod, nodeID, podDev) {
+		s.quotaManager.AddUsage(pod, podDev)
```
Shouren, line 106: @FouoF If this is the only place to call `AddUsage` of `QuotaManager`, there is a chance that ResourceQuota may not work well in a situation that multiple Pods calls the `FitQuota` of `QuotaManager` simultaneously.

archlitchi: both addPod and AddUsage are protected by their mutex lock, so i think it's thread-safe

Shouren: @archlitchi It is not about the mutex lock that make it thread-safe. It is about the timing when to add usage to the `QuotaManager`.

Consider the following situation that quota for target namespace remains resources fitting just one Pod and now Pod A and B call the `FitQuota` simultaneously in `Filter` phase, because `Used` counter inside QuotaManager has not been added up right after the first one calls, both will pass the quota check(let me know if i miss understand something).

Shouren: @FouoF @archlitchi I checked the [code](https://github.com/kubernetes/kubernetes/blob/0e014fc8dd5312a5a0344005257946692061ac49/staging/src/k8s.io/apiserver/pkg/admission/plugin/resourcequota/controller.go#L50) of `ResourceQuota` admission plugin in Kubernetes API Server, it used the workqueue which is indexed by namespace, perhaps we can take it as reference.

FouoF: Can a namespace scope mutex lock solve it? If this namespace quota is in use, directly deny it.

Shouren: @FouoF I think a namespace-scoped reentrantLock which can be acquired in `FitQuota` and released in `AddUsage` may solve it.

archlitchi: > @archlitchi It is not about the mutex lock that make it thread-safe. It is about the timing when to add usage to the `QuotaManager`.
> 
> Consider the following situation that quota for target namespace remains resources fitting just one Pod and now Pod A and B call the `FitQuota` simultaneously in `Filter` phase, because `Used` counter inside QuotaManager has not been added up right after the first one calls, both will pass the quota check(let me know if i miss understand something).

yes, you are right, but this problem is beyond the resourceQuota scope, it explains this issue: https://github.com/Project-HAMi/HAMi/issues/1330, when multiple tasks pass the Fit() and before AddResource(). I think we need to carefully design how to sync these on v2.8, and add it to RoadMap

Shouren: @archlitchi All right. I suggest adding a note about concurrency issue in doc of resourceQuota in this version to remind user that it may not work under high pressure.

archlitchi (approved):
/lgtm

PR #1395, opened 2025-10-10
feat: optimize nodelock with per-node fine-grained locking

### LINKED ISSUE #1342: Why isn't nodelock set to lock based on nodename
https://github.com/Project-HAMi/HAMi/blob/00ea0c71fb34d4099340dcc3e66af56ce8ddbdc1/pkg/util/nodelock/nodelock.go#L44

When the scheduler binds, it locks, and when the device plugin allocates, it unlocks. One global lock causes scheduling to be particularly slow. Why not set multiple locks based on the node name

xrwang8:
**What type of PR is this?**

  /kind feature

  This PR optimizes the node lock implementation to improve concurrent scheduling performance by replacing the global lock with per-node fine-grained locks.

  **Current Problem:**
  - The existing implementation uses a single global `sync.Mutex` for all node lock operations
  - When the scheduler binds pods to nodes and device plugins allocate resources, all operations are serialized even when they target different nodes
  - This creates a significant performance bottleneck in high-concurrency scenarios, especially in large clusters with many nodes

  **Solution:**
  - Introduced `nodeLockManager` that maintains a map of per-node locks
  - Each node now has its own mutex, allowing concurrent operations on different nodes
  - Operations on the same node remain properly synchronized with mutual exclusion
  - Added `cleanupLock` method to prevent unbounded memory growth in long-running processes

  **Performance Impact:**
  - **Before**: All node lock operations were serialized by a single global lock
  - **After**: Operations on different nodes can proceed concurrently
  - **Test Results**: Concurrent operations on 5 different nodes complete in ~0.5ms total time vs ~1.4ms accumulated time, demonstrating true parallelism

  Fixes #1342

### DIFF
--- a/pkg/util/nodelock/test_helpers.go
+++ b/pkg/util/nodelock/test_helpers.go
@@ -1,3 +1,19 @@
+/*
+Copyright 2024 The HAMi Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
 package nodelock
 
 import "sync"


pkg/util/nodelock/nodelock.go
```diff
@@ -53,6 +54,30 @@ var (
...
+	locks map[string]*sync.Mutex
+}
+
+// getLock returns the mutex for a specific node, creating it if necessary.
+// This method is thread-safe.
+func (m *nodeLockManager) getLock(nodeName string) *sync.Mutex {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	if m.locks == nil {
```
DSFans2014: it's better to initialize `locks` when define `nodeLocks` to avoid checking for nil in critical sections?

pkg/util/nodelock/nodelock.go
```diff
@@ -53,6 +54,33 @@ var (
...
 
+// nodeLockManager manages locks on a per-node basis to allow concurrent
+// operations on different nodes while maintaining mutual exclusion for
+// operations on the same node.
+type nodeLockManager struct {
+	mu    sync.Mutex
+	locks map[string]*sync.Mutex
+}
+
+// newNodeLockManager returns a nodeLockManager with its lock map initialized.
```
Shouren, line 65: @xrwang8 I want to know if there is a way to clean up the locks in nodeLockManager ?

xrwang8: @Shouren Thank you for your thoughtful review! Your question about lock cleanup is very important and shows great attention to production system requirements. Let me provide a comprehensive response.

## Understanding the Lock Architecture

First, let me clarify how our locking mechanism works:

1. **Memory mutex (`sync.Mutex`)**: Only serializes concurrent API calls to the same node
2. **Distributed lock (Node annotation)**: The actual lock state stored in etcd with 5-minute timeout
3. **Lock lifecycle**: Each operation acquires and releases the mutex within milliseconds

The memory mutex is just a guardian for API operations, not a state holder.

## Memory Impact Analysis

I've done detailed calculations on the memory footprint:

```
Per entry: ~64 bytes (including map overhead)
- 1,000 nodes = 64 KB
- 10,000 nodes = 640 KB  
- 100,000 nodes = 6.4 MB (extremely unlikely)

For context: A typical scheduler uses 200-500 MB for informer caches
```

## Node Failure Recovery Compatibility

Your concern made me think about node failure scenarios. I'm happy to confirm that the current implementation handles them correctly:

**Scenario: Node fails and recovers after 3 minutes**
```
T0: Node-A healthy, Pod-1 scheduled, lock set with 5-min timeout
T1: Node-A fails (network/kubelet issue)
T3: Node-A recovers (lock still valid, 2 minutes remaining)
T4: New Pod-2 tries to schedule to Node-A
    → System checks if Pod-1 still exists
    → If Pod-1 was evicted: releases lock, proceeds
    → If Pod-1 still exists: waits for timeout or completion
```

The fine-grained locking actually **improves** recovery performance because Node-A's recovery doesn't block operations on Node-B, Node-C, etc.

## Should We Add Cleanup?

Thanks for raising this. While the memory impact of keeping per-node mutexes is negligible, it's still better engineering to close the lifecycle gap. Below are the options with an updated recommendation:

### Option A (Preferred): Reference Counting + Explicit Cleanup
```go
type lockEntry struct {
    mu   sync.Mutex
    refs int
}

// getLock: create-or-get, refs++
// cleanupLock: refs--, delete when refs==0
// SetNodeLock/ReleaseNodeLock: defer unlock + cleanup
```
Pros:
- Zero external deps, precise lifecycle, minimal code
- Concurrent-safe: waiting goroutines keep a live reference to the mutex
- Directly resolves the “unbounded growth” concern

Cons:
- Requires disciplined pairing (getLock/cleanupLock)

### Option B: Keep Current Implementation (No Cleanup)
Pros: simplest code; memory impact tiny in practice

Cons: leaves an open question in review about lifecycle; theoretically unbounded

### Option C: LRU Cache
Pros: automatic bounds; familiar pattern

Cons: needs an eviction-safe pinning mechanism to avoid evicting in-use locks (which effectively reintroduces ref-counting); adds dependency

### Option D: Periodic Cleanup
Pros: no external deps

Cons: background goroutine + API calls; stale detection semantics; complexity

## My Recommendation (Updated)

Adopt **Option A: reference counting + explicit cleanup**. It keeps the implementation lightweight, eliminates the theoretical leak, and squarely addresses the review concern without added dependencies or background jobs.


```go
type lockEntry struct { mu sync.Mutex; refs int }

func (m *nodeLockManager) getLock(node string) *sync.Mutex {
    m.mu.Lock(); defer m.mu.Unlock()
    e := m.locks[node]
    if e == nil { e = &lockEntry{}; m.locks[node] = e }
    e.refs++
    return &e.mu
}

func (m *nodeLockManager) cleanupLock(node string) {
    m.mu.Lock(); defer m.mu.Unlock()
    if e := m.locks[node]; e != nil {
        if e.refs <= 0 { delete(m.locks, node); return }
        e.refs--
        if e.refs == 0 { delete(m.locks, node) }
    }
}

func SetNodeLock(node string, lockname string, pod *corev1.Pod) error {
    lk := nodeLocks.getLock(node); lk.Lock()
    defer func(){ lk.Unlock(); nodeLocks.cleanupLock(node) }()
    // ... business logic
}

func ReleaseNodeLock(node string, lockname string, pod *corev1.Pod, skip bool) error {
    lk := nodeLocks.getLock(node); lk.Lock()
    defer func(){ lk.Unlock(); nodeLocks.cleanupLock(node) }()
    // ... business logic
}
```

Shouren: @xrwang8 It is not about reference counting. You should consider that node may be added to or removed from cluster, especially for some node-level autoscaler. Current implementation does not consider the removing situation, which leads to a object leak in map.

xrwang8:
@Shouren

wawa0210:
----    
@S descendants

wawa0210:
----    
@欧阳路为163 @F even of

Issue #1363, opened 2025-09-24
hami-device-plugin fails to parse NUMA affinity "0-1" from "nvidia-smi topo -m“


RumiaLunnar:
**What happened**:
When installing HAMi on a server with a multi-NUMA node architecture, the hami-device-plugin pod runs, but it continuously prints errors to the log. The error message is "failed to get numa information" err="strconv.Atoi: parsing \"0-1\": invalid syntax".

This error causes the device plugin to fail to correctly register GPU NUMA information. As a result, any Pod that requests GPU resources gets stuck in the Pending state indefinitely because the scheduler cannot find available and correctly registered devices.
Here is a snippet of the logs from the hami-device-plugin container:
```
I0924 01:50:47.759225   47843 register.go:232] Successfully registered annotation. Next check in 30s seconds...
I0924 01:51:17.768948   47843 register.go:149] MemoryScaling= 1 registeredmem= 40960
E0924 01:51:18.377637   47843 register.go:165] "failed to get numa information" err="strconv.Atoi: parsing \"0-1\": invalid syntax" idx=2
I0924 01:51:18.377678   47843 register.go:178] nvml registered device id=2, memory=40960, type=NVIDIA A100-PCIE-40GB, numa=0
I0924 01:51:18.377805   47843 register.go:149] MemoryScaling= 1 registeredmem= 40960
E0924 01:51:18.712095   47843 register.go:165] "failed to get numa information" err="strconv.Atoi: parsing \"0-1\": invalid syntax" idx=3
I0924 01:51:18.712129   47843 register.go:178] nvml registered device id=3, memory=40960, type=NVIDIA A100-PCIE-40GB, numa=0
I0924 01:51:18.712213   47843 register.go:149] MemoryScaling= 1 registeredmem= 40960
E0924 01:51:18.981424   47843 register.go:165] "failed to get numa information" err="strconv.Atoi: parsing \"0-1\": invalid syntax" idx=0
I0924 01:51:18.981451   47843 register.go:178] nvml registered device id=0, memory=40960, type=NVIDIA A100-PCIE-40GB, numa=0
I0924 01:51:18.981560   47843 register.go:149] MemoryScaling= 1 registeredmem= 40960
E0924 01:51:20.762087   47843 register.go:165] "failed to get numa information" err="strconv.Atoi: parsing \"0-1\": invalid syntax" idx=1
[...]

PR #1401, opened 2025-10-13
Update device-numa acquisition logic

archlitchi:
/kind bug

https://github.com/Project-HAMi/HAMi/issues/1363

### DIFF
--- a/pkg/device-plugin/nvidiadevice/nvinternal/plugin/register.go
+++ b/pkg/device-plugin/nvidiadevice/nvinternal/plugin/register.go
@@ -141,10 +141,6 @@ func (plugin *NvidiaDevicePlugin) getAPIDevices() *[]*device.DeviceInfo {
 		ok, numa, err := GetNumaNode(ndev)
 		if !ok {
 			klog.ErrorS(err, "failed to get numa information from sysfs", "idx", idx)
-			numa, ret = ndev.GetNumaNodeId()
-			if ret != nvml.SUCCESS {
-				klog.ErrorS(err, "failed to get numa information from nvml", "idx", idx)
-			}
 		}
 		res = append(res, &device.DeviceInfo{
 			ID:      UUID,


pkg/device-plugin/nvidiadevice/nvinternal/plugin/register.go
```diff
@@ -49,66 +49,41 @@ import (
...
-				return 0, nil
-			}
-			result, err := strconv.Atoi(words[numaAffinityColumnIndex])
-			if err != nil {
-				return result, err
-			}
-			return result, nil
-		}
+// GetNumaNode returns the NUMA node associated with the GPU device
+func GetNumaNode(d nvml.Device) (bool, int, error) {
```
Shouren, line 68: @archlitchi I think we can get NumaNodeID from `nvml.Device` by `GetNumaNodeId` [func](https://github.com/NVIDIA/go-nvml/blob/c4406d3f035b24eeedc4a2d898d34699499790b5/pkg/nvml/device.go#L3027) first, if it failed, then try to get it from sysfs. How do you think ?

pkg/device-plugin/nvidiadevice/nvinternal/plugin/register.go
```diff
@@ -49,66 +49,41 @@ import (
...
+	info, ret := d.GetPciInfo()
+	if ret != nvml.SUCCESS {
+		return false, 0, fmt.Errorf("error getting PCI Bus Info of device: %v", ret)
+	}
+	// Discard leading zeros.
+	busID := strings.ToLower(strings.TrimPrefix(uint8Slice(info.BusId[:]).String(), "0000"))
+	b, err := os.ReadFile(fmt.Sprintf("/sys/bus/pci/devices/%s/numa_node", busID))
+	if err != nil {
+		return false, 0, nil
```
Shouren: @archlitchi We shall return the err instead if nil

pkg/device-plugin/nvidiadevice/nvinternal/plugin/register.go
archlitchi: i copy the code from here: https://github.com/NVIDIA/k8s-device-plugin/blob/cf6cbd929817cec3d386b7f7c594f4cfd7691580/internal/rm/nvml_devices.go#L151
looks like they get numa information directly from sysfs, so i take sysfs as first priority

pkg/device-plugin/nvidiadevice/nvinternal/plugin/register.go
archlitchi: adopted

pkg/device-plugin/nvidiadevice/nvinternal/plugin/register.go
Shouren: > i copy the code from here: https://github.com/NVIDIA/k8s-device-plugin/blob/cf6cbd929817cec3d386b7f7c594f4cfd7691580/internal/rm/nvml_devices.go#L151 looks like they get numa information directly from sysfs, so i take sysfs as first priority

This make sense.

PR #1418, opened 2025-10-17
Add the some missing scheduling failure events

luohua13:
**What type of PR is this?**
/kind bug

Missing two events `ResourceQuotaNotFit` and `CardNotHealth`

### DIFF
--- a/pkg/scheduler/score_test.go
+++ b/pkg/scheduler/score_test.go
@@ -24,7 +24,7 @@ import (
 	"gotest.tools/v3/assert"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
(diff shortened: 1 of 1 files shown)


pkg/scheduler/score_test.go
```diff
@@ -24,10 +24,11 @@ import (
 	"gotest.tools/v3/assert"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/mZZeta/v1"
```
Shouren: @luohua13 It seems a typo for metav1 package

Shouren (changes requested):

PR #1430, opened 2025-10-23
Fix: After removing the device plugin from the gpu node, it can still be scheduled to this node

luohua13:
**What type of PR is this?**
/kind bug

1. `CheckHealth` is always return `true, true` when removing the device plugin from the gpu node.
2. The scheduler component and plugin component continuously update different prefix values ​​for the `hami.io/node-handshake` key on the node annotations. The correct way is for the device plugin to update regularly and the scheduler component to check the status.


archlitchi:
please pass the CI check

PR #1432, opened 2025-10-23
Fix: Add leader check for metrics collect correctly

luohua13:
**What type of PR is this?**

/kind bug

Repeatedly collect metrics for multiple instances

Fixes #

### DIFF
--- a/cmd/scheduler/main.go
+++ b/cmd/scheduler/main.go
@@ -129,11 +129,6 @@ func start() error {
 
 	config.InitDevices()
 	sher = scheduler.NewScheduler()
-	sher.Start()
-	defer sher.Stop()
-
-	// start monitor metrics
-	go sher.RegisterFromNodeAnnotations()
 
 	if config.LeaderElect {
 		go func() {
@@ -143,8 +138,13 @@ func start() error {
 
 				if currentLeader && !lastLeader {
 					klog.Info("🔹 Became leader")
+					sher.Start()
+					// start monitor metrics
+					go sher.RegisterFromNodeAnnotations()
+					go initMetrics(config.MetricsBindAddress)
 				} else if !currentLeader && lastLeader {
 					klog.Info("🔸 Lost leadership, exiting to trigger restart...")
+					sher.Stop()
 					os.Exit(1)
 				}
 
@@ -153,8 +153,6 @@ func start() error {
 			}
 		}()
 	}
-
-	go initMetrics(config.MetricsBindAddress)
 
 	// start http server
 	router := httprouter.New()


cmd/scheduler/main.go
```diff
@@ -101,6 +106,12 @@ func injectProfilingRoute(router *httprouter.Router) {
 	})
 }
 
+func isLeader() bool {
+	lease, _ := client.GetClient().CoordinationV1().Leases(config.LeaderElectResourceNamespace).
+		Get(context.Background(), config.LeaderElectResourceName, metav1.GetOptions{})
+	return lease.Spec.HolderIdentity != nil && *lease.Spec.HolderIdentity == os.Getenv("HOSTNAME")
+}
+
 func start() error {
```
Shouren, line 120: @luohua13 As mentioned in the meeting, i think we can use the `RunOrDie` func from [leaderelection](https://pkg.go.dev/k8s.io/client-go/tools/leaderelection#RunOrDie) package, you can take [this](https://github.com/kubernetes/client-go/blob/master/examples/leader-election/main.go) as example.

luohua13: The `leaderelection.RunOrDie` function actually handles the leader election logic. The extender container shouldn't participate in the leader election; it only needs to know whether the current pod is the leader.

Shouren: @luohua13 I had intended to use the **Coordinated Leader Election feature** in [LeaderElectionConfig](https://pkg.go.dev/k8s.io/client-go/tools/leaderelection#LeaderElectionConfig), however i didn't notice that it is still in ALPHA stage. So i suggest that we can add a TODO comment here for future improvement.

cmd/scheduler/main.go
```diff
@@ -61,6 +64,8 @@ func init() {
 	rootCmd.Flags().StringVar(&config.HTTPBind, "http_bind", "127.0.0.1:8080", "http server bind address")
 	rootCmd.Flags().StringVar(&tlsCertFile, "cert_file", "", "tls cert file")
 	rootCmd.Flags().StringVar(&tlsKeyFile, "key_file", "", "tls key file")
+	rootCmd.Flags().StringVar(&config.LeaderElectResourceName, "leader-elect-resource-name", "", "The name of resource object that is used for leader election")
```
Shouren, line 68: @luohua13 We should add those two flags in deployment.yaml.

cmd/scheduler/main.go
```diff
@@ -118,6 +129,24 @@ func start() error {
...
+
+			if currentLeader && !lastLeader {
+				fmt.Println("🔹 Became leader")
+			} else if !currentLeader && lastLeader {
+				fmt.Println("🔸 Lost leadership, exiting to trigger restart...")
+				os.Exit(1)
+			}
+
+			lastLeader = currentLeader
+			time.Sleep(5 * time.Second)
```
Shouren: @luohua13 I think we set the duration of Sleep based on the `leaseDurationSeconds` in spec of Lease resource.

cmd/scheduler/main.go
```diff
@@ -101,6 +107,16 @@ func injectProfilingRoute(router *httprouter.Router) {
 	})
 }
 
+func isLeader() bool {
+	lease, err := client.GetClient().CoordinationV1().Leases(config.LeaderElectResourceNamespace).
+		Get(context.Background(), config.LeaderElectResourceName, metav1.GetOptions{})
+	if err != nil {
+		klog.ErrorS(err, "failed to get leader election lease", "namespace", config.LeaderElectResourceNamespace, "name", config.LeaderElectResourceName)
+		return false
+	}
```
DSFans2014, line 116: Is it necessary to check if the lease has expired?

PR #1456, opened 2025-10-30
Fix: After removing the device plugin from the gpu node, it can still…

luohua13:
What type of PR is this?
/kind bug

CheckHealth is always return true, true when removing the device plugin from the gpu node.
The scheduler component and plugin component continuously update different prefix values ​​for the hami.io/node-handshake key on the node annotations. The correct way is for the device plugin to update regularly and the scheduler component to check the status.


archlitchi (approved):
/lgtm

PR #1459, opened 2025-10-31
Synchronise podManager cache periodically in goroutine

### LINKED ISSUE #1458: GPU metric anomaly: Cache resources were not promptly cleared following pod deletion, resulting in resources continuing to be counted in the metric.
**What happened**:

HAMi version: v2.5.2

At a certain point, a NVIDIA GPU card was unoccupied by any Pods (nvidia-smi indicated no memory usage, and no Pod annotations showed binding to this card). However, the Prometheus metric `nodevGPUMemoryAllocated` reported by the HAMi scheduler showed the card's entire memory allocated. This persistent state prevented the actual idle memory resources from being properly allocated and resulted in erroneous monitoring metrics.

Screenshot of cluster HAMi metrics: 
Image

Screenshot of actual resource usage of the node:
Image

Upon deletion of a GPU Pod, the HAMi scheduler must accurately update the device resource usage status within its internal cache. Should a GPU card have no active Pods occupying it, the `nodevGpumemoryallocated` metric should correctly reflect either 0 or the actual usage amount. No resource allocation records pertaining to deleted Pods should remain.

After restarting the HAMi scheduler, the metric is correct. Screenshot as follows:
Image

Presumed to be related to the Informer event handling mechanism (see next subsection), difficult to reproduce.

Through source code analysis, the root cause of this issue may be as follows:

- The HAMi scheduler listens Pod Add/Update/Delete events via the Informer and calls `podManager.delPod()` to clear the cache upon receiving a Delete event.
- Should the eventHandler fail to complete processing of the Delete event (due to reasons including: 1. loss of the Delete event, or   2. the Informer unable to get the Pod's state and pass an object of `DeletedFinalStateUnknown` type into delete eventHandler), the resource information for the deleted Pod remains in the `podManager` cache.
- `getNodesUsage()` recalculates node resource utilization based on this cache during each scheduling or periodically update in `RegisterFromNodeAnnotations` goroutine, causing released GPU memory to remain counted in `Device.Usedmem` and ultimately reflected in Prometheus metrics.

Currently, v2.5.2, subsequent branch versions, and the master branch lack mechanisms for periodic full cache synchronisation or fallback validation.

**Interim Solution**: Restart the HAMi scheduler to force cache rebuilding from the current Pod state.  
**Long-Term Solution**: Implement a periodic full-sync mechanism (e.g., traversing all GPU Pods every hour to calibrate the `podManager` cache) and purge records of non-existent Pods.

peachest:
…t intervals

**What type of PR is this?**

/kind bug

Pod Delete EventHandler correctly handle the special objects linked to delete events during re-listing caused by the missing of object delete event when the watch connection between informer and apiserver disconnects. This prevents the podManager from persistently caching entries for pods that have been acutually deleted in cluster, ensuring that the node's remaining available resources consistent with reality. Consequently, it guarantees the availability of node resources during scheduling and the accuracy of metrics.

Fixes #1458 

No

### DIFF
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -136,6 +136,7 @@ func (s *Scheduler) onDelPod(obj any) {
 		}
 	default:
 		klog.Errorf("Received unknown object type on pod delete")
+		return
 	}
 
 	_, ok = pod.Annotations[util.AssignedNodeAnnotations]


pkg/scheduler/scheduler.go
```diff
@@ -214,6 +216,61 @@ func (s *Scheduler) Stop() {
...
+
+		// List pods scheduled by hami with hami.io/vgpu-node annotation
+		var podList []*corev1.Pod
+		if podList, err = s.podLister.List(vgpuNodeSelector); err != nil {
+			klog.ErrorS(err, "Failed to list scheduled pods")
+			continue
+		}
+
+		var cachedPods []*corev1.Pod
+		if cachedPods, err = s.podManager.ListPodsUID(); err != nil {
```
DSFans2014: Is there potential inconsistency between the data retrieved by `podLister.List` and `ListPodsUID`? Would it be preferable to implement auxiliary validation mechanisms, such as timestamps or version checks?

peachest: This PR addresses the inconsistency between the podManager cache and actual pods state in cluster by utilising `podLister.List` to retrieve pods still in the cluster that were scheduled by HAMi. This enables the purging of potentially inconsistent cached entries within the podManager.

I'm somewhat unclear on the specific aspects of potential inconsistency mentioned in your comment. Might you provide an illustrative scenario to clarify?

DSFans2014: I'm not sure if there will be any issues in some corner cases such as  a new pod created after `podLister.List` and before `ListPodsUID`.  Are the calls to these two functions atomic operations or do they fetch consistent data sets?

pkg/scheduler/scheduler.go
```diff
@@ -214,6 +216,64 @@ func (s *Scheduler) Stop() {
...
+		// However, as the podManager's deletion operation is inherently idempotent, this poses no issue.
+
+		// List pods scheduled by hami with hami.io/vgpu-node annotation
+		var podList []*corev1.Pod
+		if podList, err = s.podLister.List(vgpuNodeSelector); err != nil {
+			klog.ErrorS(err, "Failed to list scheduled pods")
+			continue
+		}
+
+		// Find out pods that are cached in pod manager but is not in cluster
```
DSFans2014: I'm not sure if there's a guarantee that the informer's callback functions will complete execution before podLister.List is called

DSFans2014: The core problem is acquiring a consistent state, akin to a database snapshot at a particular moment

pkg/scheduler/scheduler.go
```diff
@@ -121,11 +121,23 @@ func (s *Scheduler) onUpdatePod(_, newObj any) {
...
+		pod = t
+		klog.V(4).InfoS("Pod deleted, cleaning up cache", "pod", pod.Namespace+"/"+pod.Name)
+	case cache.DeletedFinalStateUnknown:
+		if pod, ok = t.Obj.(*corev1.Pod); ok {
+			klog.V(4).InfoS("Pod tombstone deleted, cleaning up cache", "pod", t.Key)
+		} else {
+			klog.Errorf("Received tombstone for non-pod object on pod delete")
+		}
+	default:
+		klog.Errorf("Received unknown object type on pod delete")
```
DSFans2014, line 138: Do we need to keep going if there's an error?

Shouren (approved):
/lgtm

krzysztof-magosa:
Hello @Shouren.
Any idea when there will be next stable release including this fix?

peachest:
> Hello @Shouren. Any idea when there will be next stable release including this fix?

I think it has been planned to release v2.8.0 this Friday

PR #1486, opened 2025-11-11
add ServiceMonitor Configuration

clcc2019:
**What type of PR is this?**
/kind feature

Fixes #


charts/hami/templates/device-plugin/servicemonitor.yaml
```diff
@@ -0,0 +1,33 @@
...
+kind: ServiceMonitor
+metadata:
+  {{- if .Values.devicePlugin.monitor.servicemonitor.annotations }}
+  annotations: 
+    {{ toYaml .Values.devicePlugin.monitor.servicemonitor.annotations | nindent 4 }}
+  {{- end }}
+  name: {{ include "hami-vgpu.device-plugin" . }}
+  namespace: {{ include "hami-vgpu.namespace" . }}
+  labels:
+    {{- include "hami-vgpu.labels" . | nindent 4 }}
```
FouoF, line 12: Considering add default label `release: prometheus` for prometheus to select this ServiceMonitor.

FouoF:
#1255 did the same work, can you try to merge it?

clcc2019:
> #1255 did the same work, can you try to merge it?

The `release` tag in `serviceMonitorSelector` is not required, and I don't think it's necessary to configure it by default.

clarifai-fmarceau (changes requested):
I don't think it makes sense to set interval in the ServiceMonitors by default. It should only be set if the user defines something. This chart should not be dictating what the default. Prometheus-operator prometheus CR already has a global default set which is what I'd expect to kick in if interval is not set.

HAMi's ServiceMonitors should also support setting other endpoints configs such as : scrapeTimeout & relabelings. As stated previously, if these are not defined by the user, they should not be set at all in favour of the Prometheus default for scrapeTimeout.

If you want to future proof this and don't necessarily want to support every possibility that the CRD offers, you should also add a generic `servicemonitor.extraEndpointProperties` (something like that, it could also be `servicemonitor.extraEndpointConfig`).

One last thing regarding the namespace where these get created, I think it makes sense by default to create them in the same namespace where HAMi is deployed however most of the community charts that I've seen over the past years generally offer the possibility to overwrite that and deploy them in a custom namespace.

FouoF:
out of date

PR #1496, opened 2025-11-17
fix: scheduler may start serving before the resource state synchronization is completed

### LINKED ISSUE #1495: The `scheduler` may start serving before the resource state synchronization is completed
The events registered in the `scheduler`'s `start` function return and start scheduling before the callbacks of event are completed. This may cause some resources and status to be incompletely initialized, leading to scheduling problems. For example, when there are a large number of Pods, it is possible that the status of the `quotaManager` may not be synchronized.

https://github.com/Project-HAMi/HAMi/blob/5e38dc3dea02d3a141908ae96d8017135afbc8a1/pkg/scheduler/scheduler.go#L194-L209

In the `WaitForCacheSync` function, for each informer, `cache.WaitForCacheSync(stopCh, informer.HasSynced)` is called. This means it waits until the `HasSynced` method of each informer returns true before returning.
```go
for informType, informer := range informers {
	res[informType] = cache.WaitForCacheSync(stopCh, informer.HasSynced)
}
```
In the function declaration of `HasSynced` in `SharedInformer` interface, we can see the following description:
```go
// HasSynced returns true if the shared informer's store has been
// informed by at least one full LIST of the authoritative state
// of the informer's object collection.  This is unrelated to "resync".
//
// Note that this doesn't tell you if an individual handler is synced!!
// For that, please call HasSynced on the handle returned by
// AddEventHandler.
HasSynced() bool
```
https://github.com/kubernetes/client-go/blob/release-1.34/tools/cache/shared_informer.go#L188-L195

This means that the completion of the `WaitForCacheSync` function does not guarantee that all event callbacks have finished.

DSFans2014:
**What type of PR is this?**

Fixes #1495 

When we call `AddEventHandler`, it returns a `ResourceEventHandlerRegistration`. By calling `HasSynced` on the `ResourceEventHandlerRegistration`, we can determine whether all the events have been synchronized.

```go
// Opaque interface representing the registration of ResourceEventHandler for
// a SharedInformer. Must be supplied back to the same SharedInformer's
// `RemoveEventHandler` to unregister the handlers.
//
// Also used to tell if the handler is synced (has had all items in the initial
// list delivered).
type ResourceEventHandlerRegistration interface {
	// HasSynced reports if both the parent has synced and all pre-sync
	// events have been delivered.
	HasSynced() bool
}
```
https://github.com/kubernetes/client-go/blob/1bb1ad283de66456c2557dea53d05dcf44b39f50/tools/cache/shared_informer.go#L239C1-L249C2

### DIFF
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -23,6 +23,7 @@ import (
 	"sort"
 	"strconv"
 	"strings"
+	"sync/atomic"
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
@@ -60,6 +61,7 @@ type Scheduler struct {
 	overviewstatus map[string]*NodeUsage
 	eventRecorder  record.EventRecorder
 	quotaManager   *device.QuotaManager
+	started        uint32 // 0 = false, 1 = true
 }
 
 func NewScheduler() *Scheduler {
@@ -68,6 +70,7 @@ func NewScheduler() *Scheduler {
 		stopCh:       make(chan struct{}),
 		cachedstatus: make(map[string]*NodeUsage),
 		nodeNotify:   make(chan struct{}, 1),
+		started:      0,
 	}
 	s.nodeManager = newNodeManager()
 	s.podManager = device.NewPodManager()
@@ -183,7 +186,7 @@ func (s *Scheduler) onDelQuota(obj interface{}) {
 	s.quotaManager.DelQuota(quota)
 }
 
-func (s *Scheduler) Start() {
+func (s *Scheduler) Start() error {
 	klog.InfoS("Starting HAMi scheduler components")
 	s.kubeClient = client.GetClient()
 	informerFactory := informers.NewSharedInformerFactoryWithOptions(s.kubeClient, time.Hour*1)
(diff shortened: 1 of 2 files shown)


pkg/scheduler/scheduler.go
```diff
@@ -183,31 +186,43 @@ func (s *Scheduler) onDelQuota(obj interface{}) {
...
 		DeleteFunc: s.onDelQuota,
 	})
+	if err != nil {
+		return fmt.Errorf("failed to register resource quota event handler: %v", err)
+	}
 	informerFactory.Start(s.stopCh)
 	informerFactory.WaitForCacheSync(s.stopCh)
+	cache.WaitForCacheSync(s.stopCh, podEventHandlerRegistration.HasSynced, nodeEventHandlerRegistration.HasSynced, resourceQuotaEventHandlerRegistration.HasSynced)
 	s.addAllEventHandlers()
+	atomic.StoreUint32(&s.started, 1)
```
dl239, line 224: call node.addusage

DSFans2014: In the `Filter` function, `getNodesUsage` will be called. So is it necessary to call `getNodesUsage` here? @archlitchi 
https://github.com/Project-HAMi/HAMi/blob/7bb4584c8106ae153a4f235f8225a9ae97732f49/pkg/scheduler/scheduler.go#L536

DSFans2014:
@archlitchi please review this pr again when you are free. Let me know if there are any problems

pkg/scheduler/scheduler.go
```diff
@@ -218,82 +234,89 @@ func (s *Scheduler) RegisterFromNodeAnnotations() {
...
-		for _, val := range rawNodes {
-			nodeNames = append(nodeNames, val.Name)
-			klog.V(5).InfoS("Processing node", "nodeName", val.Name)
+		s.RefreshNode()
+	}
+}
 
-			for devhandsk, devInstance := range device.GetDevices() {
-				klog.V(5).InfoS("Checking device health", "nodeName", val.Name, "deviceVendor", devhandsk)
+func (s *Scheduler) RefreshNode() {
```
Shouren: @DSFans2014 It seems that the `RefreshNode` func is only called inside the schduler package, so it it better to make it private by renaming it to `refreshNode`

pkg/scheduler/scheduler.go
```diff
@@ -183,31 +186,44 @@ func (s *Scheduler) onDelQuota(obj interface{}) {
...
 		DeleteFunc: s.onDelQuota,
 	})
+	if err != nil {
+		return fmt.Errorf("failed to register resource quota event handler: %v", err)
+	}
 	informerFactory.Start(s.stopCh)
 	informerFactory.WaitForCacheSync(s.stopCh)
+	cache.WaitForCacheSync(s.stopCh, podEventHandlerRegistration.HasSynced, nodeEventHandlerRegistration.HasSynced, resourceQuotaEventHandlerRegistration.HasSynced)
 	s.addAllEventHandlers()
+	s.RefreshNode()
```
Shouren: @DSFans2014 Why we need to refresh node here ?

DSFans2014: nodes are added to the `NodeManager` via `addNode`, which is called from `RegisterFromNodeAnnotations`
https://github.com/Project-HAMi/HAMi/blob/a5411abd7bd0d3e07385fea1803456e4707bafe8/pkg/scheduler/scheduler.go#L282
If not called at startup,  node registration is delayed until the scheduled task runs, which may result in scheduling problems during the delay.

Shouren: @DSFans2014 The AddFunc of node informer calls `doNodeNotify` which will trigger `refreshNode`, so i think this func should has already been called when `cache.WaitForCacheSync` returns.

DSFans2014: yep, I missed the `nodeNotify`, thank you

Issue #1368, opened 2025-09-25
HAMi Scheduler Not Trying to Schedule Previously Pending Workload for 5mins


CcccYxx:
**What happened**: There was one workload that used all the node resources, and another workload was in kueue's queue due to insufficient resources. Now, terminate the first workload, the second workload will be admitted by kueue, then it was handed over to hami sheduler to begin scheduling. However, this workload is pending for around 5mins until it gets scheduled by hami sheculder. It turned out when the first workload was terminating, the second workload failed filtering since the first workload is still terminating. But after the first worload was terminated, the second worload was still not retrying filtering for 5mins, leading to unnecessary wait. Below are proof that the filter endpoint was only invoked twice during all those 5mins (14m-9m)
```
Events:
  Type     Reason            Age                    From            Message
  ----     ------            ----                   ----            -------
  Warning  FailedScheduling  14m                    hami-scheduler  0/1 nodes are available: 1 NodeUnfitPod. preemption: 0/1 nodes are available: 1 No preemption victims found for incoming pod.
  Warning  FailedScheduling  14m                    hami-scheduler  0/1 nodes are available: 1 NodeUnfitPod. preemption: 0/1 nodes are available: 1 No preemption victims found for incoming pod.
  Normal   Scheduled         9m32s                  hami-scheduler  Successfully assigned test-project/test-ws-1-0 to ip-172-31-35-36
  Warning  FilteringFailed   14m (x2 over 14m)      hami-scheduler  no available node, 1 nodes do not meet
  Normal   FilteringSucceed  9m33s                  hami-scheduler  find fit node(ip-172-31-35-36), 0 nodes not fit, 1 nodes fit(ip-172-31-35-36:0.00)
```

**What you expected to happen**: The second workload should be scheduled immediately after the first workload is terminated.

**How to reproduce it (as minimally and precisely as possible)**:
1. Install both kueue and hami in an nvidia cluster (a single node server is sufficient)
2. Create a kueue clusterqueue and localqueue
3. Submit a gpu workload that uses as much resource as possible to the kueue queue.
4. Submit another gpu worload to kueue queue (this workload will be pending in the kueue queue)
5. Terminate the first gpu workload
6. The second workload will be admitted by kueue and handed over to hami
7. The second workload will sometimes be pending for around 5mins before it gets admitted
8. If the above steps does not reproduce the issue, repeat step 1-7.

**Environment**:
- HAMi version: v2.6.1
[...]

PR #1499, opened 2025-11-18
Refine Node Register logic

archlitchi:
/kind cleanup

1. Refine node device register logic, remove unnecessary handshake logic for nvidia device. It now updates nodeDevices only  when 'nodes.allocatable' changes
2. Convert array to map in struct deviceInfo for better supporting heterogeneous devices
3. Use json for node device register for future compatibility, only working on nvidia right now
4. Combine 'nvidiaGPUCommonword' and 'nvidiaGPUDevice' into one for nvidia devices

### DIFF
--- a/pkg/device/devices.go
+++ b/pkg/device/devices.go
@@ -267,7 +267,7 @@ func EncodeNodeDevices(dlist []*DeviceInfo) string {
 	return tmp
 }
 
-// MarshalNodeDevices will only marshal general information, customInfo is neglected
+// MarshalNodeDevices will only marshal general information, customInfo is neglected.
 func MarshalNodeDevices(dlist []*DeviceInfo) string {
 	devAnnos := []*DeviceInfo{}
 	for _, val := range dlist {


test/utils/node.go
```diff
@@ -42,6 +43,7 @@ func UpdateNode(clientSet *kubernetes.Clientset, node *v1.Node) (*v1.Node, error
 		return nil, err
 	}
 
+	time.Sleep(time.Second * 30)
```
Shouren, line 46: @archlitchi Why we need to sleep 30s here ?

pkg/device-plugin/nvidiadevice/nvinternal/plugin/register.go
```diff
@@ -167,7 +167,12 @@ func (plugin *NvidiaDevicePlugin) RegisterInAnnotation() error {
 		klog.Errorln("get node error", err.Error())
 		return err
 	}
-	encodeddevices := device.EncodeNodeDevices(*devices)
+	encodeddevices := device.MarshalNodeDevices(*devices)
+	if encodeddevices == plugin.deviceCache {
```
Shouren, line 171: @archlitchi The type of `devices` object is `[]*device.DeviceInfo` and the `DeviceInfo` has a `CustomInfo` field which is `map[string]any`, because the JSON output of Maps in Golang are Non-deterministic, so it may leads to a false condition even if the underlying struct are same. Although we do not set the `CustomInfo` in the current implementation of `getAPIDevices` func, but **there is a risk** here that someone may change the code between line 145-155, so i prefer to adding a comment in that line about this.

test/utils/node.go
archlitchi: because of this issue: https://github.com/Project-HAMi/HAMi/issues/1368

The pod is submitted before the dp initialized, leading to a pending state, it won't get re-scheduled even after the dp initialized.

we add sleep here is a temporary fix to allow it to pass the e2e check, and will be removed once this issue is solved

pkg/device-plugin/nvidiadevice/nvinternal/plugin/register.go
archlitchi: adopted

PR #1506, opened 2025-11-26
Optimize the issue where configmap hami-device-plugin is overwritten …

leolingg:
**What type of PR is this?**

/kind bug

Optimize the issue where configmap hami-device-plugin is overwritten with default values during upgrades.

Fixes #

When the configmap hami-device-plugin has been custom modified, using helm upgrade will overwrite it with the default configuration of the template.

This PR is to address this issue.

Add the devicePlugin.nodeConfiguration option, which includes 2 sub-items: config and configName:

1. Default value: The default value for both is empty, indicating the use of the default example configuration.

2. config: Specify a value for config.json (a JSON string) to override the default value.

3. externalConfigName: Specify an existing configmap to mount into the device-plugin (ensure the content is correct).

Priority of the three: externalConfigName > config > default config

NO


charts/hami/templates/device-plugin/configmap.yaml
```diff
@@ -8,6 +8,11 @@ metadata:
     app.kubernetes.io/component: hami-device-plugin
     {{- include "hami-vgpu.labels" . | nindent 4 }}
 data:
+  {{ $existing := (lookup "v1" "ConfigMap" (include "hami-vgpu.namespace" .) (include "hami-vgpu.device-plugin" .)) -}}
+  {{ if and $existing $existing.data (hasKey $existing.data "config.json") -}}
+  config.json: |
+{{ get $existing.data "config.json" | indent 4 }}
```
DSFans2014: If the `config.json` is modified in the future, it won't be updated during a normal upgrade?

archlitchi: yes, we may update the config in the future, so we need to add a switch(If override config) here with default value true

leolingg:
> This doesn't feel right to me - maybe I'm missing something so feel free to correct me, to address this I believe that 2 changes/ fixes are needed instead of the current proposed approach :
> 
> * Support setting an external configmap
> * Improve current configmap to support customizing it

---

Thank you for your suggestion.

> Support setting an external configmap

I don't quite understand what this means? 

> Improve current configmap to support customizing it

This is a good suggestion, and I will modify it according to this suggestion.

charts/hami/values.yaml
```diff
@@ -286,6 +286,14 @@ devicePlugin:
   deviceSplitCount: 10
   deviceMemoryScaling: 1
   deviceCoreScaling: 1
+  # Node configuration for device plugin, Priority: configName > config > default config
+  nodeConfiguration:
+    # If you want to use a custom config.json, you can set the content here.
+    # If this is set, it will override the default config.json.
+    config: {}
```
Shouren: @leolingg I think we can put the content of the current configMap here and render it. How do you think ?

charts/hami/values.yaml
```diff
@@ -286,6 +286,14 @@ devicePlugin:
...
+    # If you want to use an existing ConfigMap, you can set the name here.
+    # If this is set, the chart will not create the ConfigMap and will use the existing one.
+    configName: ""
```
Shouren: @leolingg This variable name confuse me that i think it is the name of ConfigMap for nodeConfiguration, so i think it is better to rename it to `configMapRef` or `externalConfigName`, how do you think ?

charts/hami/values.yaml
leolingg: @Shouren What does "content of the current configMap" refer to?

1. represents the value of the hami-device-plugin configmap in the current cluster? 
- As discussed earlier, it doesn't feel like a good suggestion (similar to inserting the current value in "template  charts/hami/templates/device-plugin/configmap.yaml"), or do you mean something else?
2. What if it represents the value of a configuration example?
- Currently, the charts/hami/templates/device-plugin/configmap.yaml template already has an example (when.Values.devicePlugin.nodeConfiguration.config is empty). 

```json
{
  "nodeconfig": [
    {
      "name": "m5-cloudinfra-online02",
      "operatingmode": "hami-core",
      "devicememoryscaling": 1.8,
      "devicesplitcount": 10,
      "migstrategy": "none",
      "filterdevices": {
        "uuid": [],
        "index": []
      }
    }
  ]
}
```

charts/hami/values.yaml
leolingg: done, i think "externalConfigName" is better.

charts/hami/values.yaml
Shouren: @leolingg Sorry for the ambiguity, i mean the latter. And rendering `nodeconfig` from the `config` can help simplify the code in `configmap.yaml` template by removing the check of `config`.

leolingg: > And rendering nodeconfig from the config can help simplify the code in configmap.yaml template by removing the check of config.

@Shouren done

Shouren:
@leolingg Please rebase this branch onto master so that the [hami-robot](https://github.com/apps/hami-robot) can merge this PR.

PR #1517, opened 2025-12-04
Update go version to v1.25.5

archlitchi:
/kind cleanup

In order to be compatible with https://github.com/Project-HAMi/HAMi/pull/1515 and https://github.com/Project-HAMi/HAMi/pull/1513


Shouren (approved):
/lgtm

PR #1534, opened 2025-12-11
feat: install `mock-device-plugin`

DSFans2014:
**What type of PR is this?**
feature

install the `mock-device-plugin` alongside `hami`

Fixes #

This PR depends on https://github.com/Project-HAMi/mock-device-plugin/pull/6 and should be merged afterwards.

### DIFF
--- a/charts/hami/templates/device-plugin/daemonsetmock.yaml
+++ b/charts/hami/templates/device-plugin/daemonsetmock.yaml
@@ -23,7 +23,7 @@ spec:
         operator: Exists
       containers:
       - image: {{ include "hami.mockDevicePlugin.image" . }}
-        imagePullPolicy: {{ .Values.devicePlugin.image.pullPolicy }}
+        imagePullPolicy: {{ .Values.mockDevicePlugin.image.pullPolicy }}
         name: hami-mock-dp-cntr
         env:
           - name: NODE_NAME
@@ -34,12 +34,6 @@ spec:
           - ./k8s-device-plugin
           - -v=5
           - --device-config-file=/device-config.yaml
-        securityContext:
-          privileged: true
-          allowPrivilegeEscalation: true
-          capabilities:
-            drop: ["ALL"]
-            add: ["SYS_ADMIN"]
         volumeMounts:
           - name: dp
             mountPath: /var/lib/kubelet/device-plugins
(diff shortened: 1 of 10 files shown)


charts/hami/values.yaml
```diff
@@ -344,6 +344,25 @@ devicePlugin:
 #      cpu: 100m
 #      memory: 100Mi
 
+mockDevicePlugin:
+  enabled: true
```
DSFans2014: Should we install the mock device plugin by default? @archlitchi

FouoF: I think the mock tool should not be installed by default. It is mostly for development.

DSFans2014: > I think the mock tool should not be installed by default. It is mostly for development.

`mock-device-plugin` is not a tool for development. it will patch the specific resource to `node.status.capacity`.  refer https://github.com/Project-HAMi/mock-device-plugin @FouoF

Shouren: @DSFans2014 I do not know if there is any side effect after patching resources to `node.status.capacity`, so i prefer not to installing the mock device plugin by default.

charts/hami/templates/device-plugin/daemonsetmock.yaml
```diff
@@ -0,0 +1,61 @@
...
+          - name: NODE_NAME
+            valueFrom:
+              fieldRef:
+                fieldPath: spec.nodeName
+        command:
+          - ./k8s-device-plugin
+          - -v=5
+          - --device-config-file=/device-config.yaml
+        securityContext:
+          privileged: true
```
Shouren: @DSFans2014 Why this device plugin requires privileged ?

charts/hami/templates/scheduler/role.yaml
```diff
@@ -10,3 +10,17 @@ rules:
...
+kind: ClusterRole
+metadata:
+  name: {{ include "hami-vgpu.mock-device-plugin" . }}
+rules:
+  - apiGroups: [""]
+    resources: ["nodes"]
+    verbs: ["get", "update", "list", "patch"]
+  - apiGroups: [""]
+    resources: ["pods"]
+    verbs: ["get", "list", "update", "watch", "patch"]
```
Shouren: @DSFans2014 Does this device plugin do `update` and `patch` to a Pod ?

PR #1549, opened 2025-12-20
The image’s startup command already includes the xpu-device-plugin command.

121812:
The image’s startup command already includes the xpu-device-plugin command.

**What type of PR is this?**
/kind documentation

The image’s startup command already includes the xpu-device-plugin command.

If it’s not removed, the startup command will become:
xpu-device-plugin xpu-device-plugin --memory-unit=MiB --resource-name=kunlunxin.com/vxpu -logtostderr ...
As a result, xpu-device-plugin will start using its default parameters and will not read the following arguments.


docs/kunlun-vxpu-support.md
```diff
@@ -82,7 +82,6 @@ spec:
               memory: 500Mi
               cpu: 500m
           args:
-            - xpu-device-plugin
```
DSFans2014, line 85: please sign-off your commit

PR #1550, opened 2025-12-21
chore: remove repeated var

googs1025:
**What type of PR is this?**
/kind cleanup

Fixes None

None
None


pkg/scheduler/policy/constant.go
```diff
@@ -1,45 +0,0 @@
-/*
```
googs1025, line 1: removed

PR #1553, opened 2025-12-22
Implementing leader-follower high availability using leader-election

peachest:
**What type of PR is this?**
/kind feature

The current HAMi implementation lacks sufficient support for multi-replica high availability.

Whilst the kube-scheduler has implemented master-slave election via Kubernetes' built-in leader election mechanism, the vgpu-scheduler-extender has not. Direct deployment of multiple replicas may encounter the following issues:
1. Multiple instances repeatedly initiating handshakes with device plugins on nodes
2. Multiple instances duplicating metric collection efforts #1432 

The `holderIdentity` field is used for identification within the Lease. The `kube-scheduler` constructs the `holderIdentity` using the hostname. Taking version 1.34.2 as an example:
```go
	// add a uniquifier so that two processes on the same host don't accidentally both become active
	id := hostname + "_" + string(uuid.NewUUID())
```
This PR determines whether the current `vgpu-scheduler-extender` container belongs to Leader instance by acquiring the lease and checking whether the current Pod's hostname prefix exists in `Lease.Spec.hoderIdentity`.

- Before executing `RegisterFromNodeAnnotation` each time, the scheduler must first determine whether it is the leader, ensuring only the leader instance initiates handshakes with nodes (and their device plugins).
- Notify the current instance via a channel that it has been elected leader, and immediately execute `RegisterFromNodeAnnotation` even if the scheduled interval has not yet elapsed.
- By defining the container's readinessProbe, ensure only the leader instance joins the hami-scheduler Service endpoints, preventing duplicate metric collection by other instances. This also affects webhooks, with only the primary instance's webhook now executing.

The hami-scheduler-newversion configuration employs SchedulerConfig to ensure that the kube-scheduler within each Pod exclusively accesses the local vgpu-scheduler-extender. Consequently, there is no concern regarding interference with calls to the `Filter/` and `Bind/` endpoints.

### DIFF
--- a/charts/hami/templates/scheduler/deployment.yaml
+++ b/charts/hami/templates/scheduler/deployment.yaml
@@ -191,10 +191,10 @@ spec:
           requiredDuringSchedulingIgnoredDuringExecution:
           - labelSelector:
               matchExpressions:
-              - key: app.kubernetes.io/name
+              - key: app.kubernetes.io/component
                 operator: In
                 values:
-                - {{ include "hami-vgpu.name" . }}
+                - hami-scheduler
             topologyKey: "kubernetes.io/hostname"
       {{- end }}
       volumes:


pkg/util/leaderelection/leaderelection.go
```diff
@@ -0,0 +1,183 @@
...
+)
+
+type LeaderManager interface {
+	IsLeader() bool
+	LeaderNotifyChan() <-chan struct{}
+
+	cache.ResourceEventHandler
+}
+
+var _ LeaderManager = &leaderManager{}
```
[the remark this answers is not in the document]
peachest: I'm sure that `cache.FilteringResourceEventHandler` implements `ResourceEventHandler` interface

charts/hami/templates/scheduler/deployment.yaml
```diff
@@ -165,6 +168,17 @@ spec:
...
+          {{- if .Values.scheduler.leaderElect }}
+          readinessProbe:
+            httpGet:
+              path: /readyz
+              port: 443
+              scheme: HTTPS
+            initialDelaySeconds: 10
+            periodSeconds: 10
+            failureThreshold: 3
+            timeoutSeconds: 5
```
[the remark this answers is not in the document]
peachest: fixed

pkg/scheduler/scheduler.go
```diff
@@ -75,6 +79,10 @@ func NewScheduler() *Scheduler {
 	s.nodeManager = newNodeManager()
 	s.podManager = device.NewPodManager()
 	s.quotaManager = device.NewQuotaManager()
+	s.leaderManager = leaderelection.NewDummyLeaderManager(true)
+	if config.LeaderElect {
+		s.leaderManager = leaderelection.NewLeaderManager(config.HostName, config.LeaderElectResourceNamespace, config.HostName)
```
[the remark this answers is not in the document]
peachest: fixed

cmd/scheduler/main.go
```diff
@@ -112,9 +117,16 @@ func start() error {
 	)
 
 	config.InitDevices()
+
+	var err error
+	config.HostName, err = os.Hostname()
```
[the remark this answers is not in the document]
peachest: fixed

pkg/scheduler/scheduler.go
```diff
@@ -266,6 +285,11 @@ func (s *Scheduler) RegisterFromNodeAnnotations() {
 			klog.InfoS("Received stop signal, exiting RegisterFromNodeAnnotations")
 			return
 		}
+		// TODO: may be we should lock node when we are doing register.
+		// Only do registration when we are leader
+		if !s.leaderManager.IsLeader() {
+			continue
```
Shouren, line 322: @peachest I think it is too late to do registration when we are leader. When the kube-scheduler in the same Pod becomes the leader and sends a filter request to this extender, there is no guarantee this the registration will be done before handling that request.

FouoF: @Shouren I think this check is for addressing 

> Multiple instances repeatedly initiating handshakes with device plugins on nodes

But #1499 refine the handshake logic, so I'm not sure if the new sync logic still have this problem. If not, we don't need this section any more.

And for the requests, the leader control is delegated to the k8s-scheduler. Only the leader k8s-scheduler will pass requests to extender. So the extender don't need to care about that. As the description:

> The holderIdentity field is used for identification within the Lease. The kube-scheduler constructs the holderIdentity using the hostname. Taking version 1.34.2 as an example:

	// add a uniquifier so that two processes on the same host don't accidentally both become active
	id := hostname + "_" + string(uuid.NewUUID())

> This PR determines whether the current vgpu-scheduler-extender container belongs to Leader instance by acquiring the lease and checking whether the current Pod's hostname prefix exists in Lease.Spec.hoderIdentity

peachest: I think what @Shouren means is that we should make sure finish doing `RegisterFromNodeAnnotation` at least one time before doing filter when the  `Filter/` endpoint is call. 

I am tring to implement this by adding a synced flag and just wait for syncing before filtering. It's reasonable that, when we finished `RegisterFromNodeAnnotation`, we set `synced` as true, and set `synced` to false when we lost leadership because we won't do any register from now on. 

Is there any better idea?

Shouren: @FouoF @peachest Let me explain it in detail, the check is achieved by following line
> `			health, needUpdate := devInstance.CheckHealth(devhandsk, val)`

and the implementation of `CheckHealth` differs for each devInstance, #1499 refactor the implementation of `CheckHealth` for `NvidiaGPUDevices` that it does not call `device.CheckHealth` any more. But there are four devInstances(kunlun, hygon, iluvatar & ascend) still calling `device.CheckHealth`, so we need this check.

However, the `RegisterFromNodeAnnotations` function will execute the following line
> `				s.addNode(val.Name, nodeInfo)`

to add information of node to `nodeManager` of `scheduler` when initializing scheduler or node needs to be updated. My concern is that current implementation skips this step before the kube-scheduler becomes the leader and it might be too late when a filter request comes but the nodes in that request has not been added to `nodeManager`.

Perhaps we need a new variable to indicate whether handshake is required, but i'm not sure if it will bring new issues.

Shouren: > Is there any better idea?

@peachest Maybe we can add a new parameter to `device.CheckHealth` to control the behavior of handshake, so that only the leader(Let's put the fencing issue aside for now) can patch the annotations of node and keep the rest of code works as before. But i have to check the code to see if there are any conflicts with this implementation.

pkg/util/leaderelection/leaderelection.go
```diff
@@ -0,0 +1,183 @@
...
+	m.leaseLock.RLock()
+	defer m.leaseLock.RUnlock()
+
+	if m.observedLease == nil {
+		return false
+	}
+
+	// TODO: should we check valid lease here?
+	return m.isHolderOf(m.observedLease)
+}
```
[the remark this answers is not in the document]
peachest: already add checking lease valid in IsLeader()

pkg/scheduler/scheduler.go
peachest: Currently I add a synced field in Scheduler and implement WaitForCacheSynced by polling to check whether the synced flag is set to true, which determines the cache has synchronised prior to executing Filter/ logic.

Two callback functions is used to notify whether we acquired or lost leadership

Synced is set to false when initializing. It's set to true after finishing register, and set to false again when the scheduler lost leadership.

pkg/util/leaderelection/leaderelection.go
```diff
@@ -0,0 +1,183 @@
...
+		return
+	}
+
+	m.leaseLock.Lock()
+	defer m.leaseLock.Unlock()
+
+	m.setObservedRecord(lease)
+	// Notify if we are the leader from the very begging
+	if m.isHolderOf(lease) {
+		m.leaderNotify <- struct{}{}
```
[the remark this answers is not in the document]
peachest: close because outdated

pkg/util/leaderelection/leaderelection.go
```diff
@@ -0,0 +1,183 @@
...
+	m.setObservedRecord(newLease)
+	// Notify if we have been elected to become the leader
+	if m.isHolderOf(newLease) {
+		oldLease, ok := oldObj.(*coordinationv1.Lease)
+		if !ok {
+			return
+		}
+
+		if !m.isHolderOf(oldLease) {
+			m.leaderNotify <- struct{}{}
```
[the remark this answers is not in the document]
peachest: close because outdated

pkg/scheduler/routes/route.go
```diff
@@ -56,6 +56,16 @@ func PredicateRoute(s *scheduler.Scheduler) httprouter.Handle {
...
 			}
 		} else {
+			synced := s.WaitForCacheSync(r.Context())
+			if !synced {
+				// Poll may return false when context is cancelled
+				err := fmt.Errorf("context cancelled")
+				klog.ErrorS(err, "Cache not synced, cannot proceed with filtering")
+				w.Header().Set("Content-Type", "application/json")
+				w.WriteHeader(http.StatusInternalServerError)
+				w.Write([]byte(err.Error()))
```
Shouren, line 66: @peachest The content we write here is not the `Content-Type` we set in header.

peachest: Do you have any better suggestions regarding how to configure the header and the content to be returned? My implementation only referenced the error handling for `resultBody, err := json.Marshal(extenderFilterResult)` below.

Shouren: @peachest Sorry, i didn't notice the code below, perhaps a new PR should fix the `Content-Type` instead of this one.

peachest:
> looks like we have a problem with e2e

Two problems cause the e2e failed previously
1. The absence of list and watch permissions for leased resources prevented the service from starting correctly.
2. Selecting an incorrect label to define podAntiAffinity resulted in the Job Pod launched by postStartHook failing to schedule (as E2E testing was conducted on a single node).

E2E test now passed, please do the final review

Shouren (approved):
/lgtm

PR #1555, opened 2025-12-23
fix: resolve multi-node GPU count tracking in nvidia device plugin

pepesi:
Change ReportedGPUNum from int64 to map[string]int64 to track GPU count per node instead of sharing state across all nodes. Add mutex protection for concurrent access.

**What type of PR is this?**

Fixes #

### DIFF
--- a/pkg/device/nvidia/device.go
+++ b/pkg/device/nvidia/device.go
@@ -23,6 +23,7 @@ import (
 	"slices"
 	"strconv"
 	"strings"
+	"sync"
 
 	spec "github.com/NVIDIA/k8s-device-plugin/api/config/v1"
 	corev1 "k8s.io/api/core/v1"
@@ -157,7 +158,8 @@ type DeviceConfig struct {
 
 type NvidiaGPUDevices struct {
 	config         NvidiaConfig
-	ReportedGPUNum int64
+	ReportedGPUNum map[string]int64 // key: nodeName, value: reported GPU count
+	mu             sync.Mutex       // protects concurrent access to ReportedGPUNum
 }
 
 func InitNvidiaDevice(nvconfig NvidiaConfig) *NvidiaGPUDevices {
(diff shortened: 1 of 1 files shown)


pepesi:
> @pepesi I understand this bug but I still think the mutex is currently unnecessary. The bug is caused by improper variable use and the change to map is enough to fix it.

Since the current function is executed within a single loop, I think this lock is optional! But for the purpose of defensive programming, it is still better to add it, because it is unknown whether concurrent access may occur in the future! Alternatively, you can change it to sync.Map!

pkg/device/nvidia/device.go
```diff
@@ -157,7 +158,8 @@ type DeviceConfig struct {
 
 type NvidiaGPUDevices struct {
 	config         NvidiaConfig
-	ReportedGPUNum int64
+	ReportedGPUNum map[string]int64 // key: nodeName, value: reported GPU count
+	mu             sync.RWMutex     // protects concurrent access to ReportedGPUNum
```
DSFans2014: Can a regular `sync.Mutex` be enough without using `RLock()` ?

pepesi: Yes, using sync.Mutex is enough

pkg/device/nvidia/device.go
```diff
@@ -221,21 +223,27 @@ func (dev *NvidiaGPUDevices) CheckHealth(devType string, n *corev1.Node) (bool,
 	if quantity != nil {
 		current = quantity.Value()
 	}
-	klog.Infoln("-=-=-=Current device count from allocatable", current, "Reported Devices=", dev.ReportedGPUNum)
+
+	dev.mu.Lock()
+	defer dev.mu.Unlock()
+
+	reported := dev.ReportedGPUNum[n.Name]
+	klog.Infoln("-=-=-=Current device count from allocatable", current, "Reported Devices=", reported, "Node=", n.Name)
```
archlitchi: please use klog.v3.infoS instead, and trim this log

archlitchi (approved):
/lgtm

PR #1560, opened 2025-12-27
fix(nodelock): improve parsing of lock annotation

Mirza-Samad-Ahmed-Baig:
This pull request fixes a bug in the ParseNodeLock function that caused confusing error messages when processing malformed lock annotations. The function is now more robust and will return a specific error if the annotation is not in the expected format, improving the node locking mechanism's correctness by catching and reporting such errors early.


archlitchi (approved):
/lgtm

PR #1570, opened 2026-01-07
feat: update mock-device-plugin version

DSFans2014:
**What type of PR is this?**

Fixes #

### DIFF
--- a/charts/hami/values.yaml
+++ b/charts/hami/values.yaml
@@ -379,7 +379,7 @@ mockDevicePlugin:
   image:
     registry: "docker.io"
     repository: "projecthami/mock-device-plugin"
-    tag: "0.1.0"
+    tag: "1.0.1"
     ## Specify a imagePullPolicy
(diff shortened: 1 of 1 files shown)


FouoF (approved):

charts/hami/values.yaml
```diff
@@ -379,7 +379,7 @@ mockDevicePlugin:
   image:
     registry: "docker.io"
     repository: "projecthami/mock-device-plugin"
-    tag: "0.1.0"
+    tag: "1.0.1"
```
archlitchi, line 382: looks like the tag version is 0.0.1 according to the dockerhub
https://hub.docker.com/r/projecthami/mock-device-plugin/tags

DSFans2014: 0.0.1 is a test tag. Should we release version 1.0.1 now, or wait to release it together with the new version of HAMi ?

archlitchi: i think we need to release now, for users to get started

archlitchi (approved):
/lgtm

PR #1573, opened 2026-01-08
Watch and hot reload the updated certificate

### LINKED ISSUE #1572: The HTTPS webhook server can not continue to serve when certificate is updated
**What happened**:

When use CertManager to manage certificates for the HAMi Webhook, whenever CertManager periodically renews certificates, the hami scheduler encounters the following exception:
1. The container vgpu-scheduler-extender generates the following log:
  ```
  http: TLS handshake error from <IP>:<Port>: remote error: tls: bad certificate
  ```

Image

3. The HAMi webhook is not functioning correctly, resulting in the pod's `schedulerName` not being modified. Consequently, the pod is handed over to the default scheduler for scheduling, triggering a `FailedScheduling` event with the message `Insufficient xxx`. 

Image

The issue was resolved after restarting the hami-scheduler pod.

The webhook continues to function normally after certificate is renewed. 

Deploy HAMi using the following values.yaml
```yaml
scheduler:
    certManager:
        enabled: true
    patch:
        enabled: false
```

Once deployment is complete and the `hami-scheduler` Pod is Ready, delete the secret `hami-scheduler-tls` to trigger CertManager Operator to recreate it, simulating a certificate update/renew. Deploying the workload again at this point will reproduce the aforementioned issue.

When managing certificates using CertManager, certificates for hami webhook are stored within the `hami-scheduler-tls` Secret and are automatically renewed 30 days in advance by default.
- The Secret did not exist during initial deployment; or the Secret resource was unexpectedly deleted during runtime. In both scenarios, CertManager will generate a new certificate Secret based on the Issuer and Certificate resources.
- Upon renewal, fields such as `notBefore` and `notAfter` within the certificate content will change,  therefore the certificate content is modified.

Hami starts the HTTPS server via `http.ListenAndServeTLS(config.HTTPBind, tlsCertFile, tlsKeyFile, router);`. The call chain is: `http.ListenAndServeTLS` => `server.ListenAndServeTLS` => `server.ServeTLS`. Within `server.ServeTLS`, the current implementation specifies `certFile` and `keyFile` without invoking `config.GetCertificate` or `config.GetConfigForClient`. Consequently, the certificate is loaded once during service initialisation via `tls.LoadX509KeyPair(certFile, keyFile)`.

The source code: 
```go
func (s *Server) ServeTLS(l net.Listener, certFile, keyFile string) error {
        //...

	configHasCert := len(config.Certificates) > 0 || config.GetCertificate != nil || config.GetConfigForClient != nil
	if !configHasCert || certFile != "" || keyFile != "" {
		var err error
		config.Certificates = make([]tls.Certificate, 1)
		config.Certificates[0], err = tls.LoadX509KeyPair(certFile, keyFile)
		if err != nil {
			return err
		}
	}

	tlsListener := tls.NewListener(l, config)
	return s.Serve(tlsListener)
}
```

- When deploying HAMI using kube-webhook-certgen to generate certificates (specifically the `hami-scheduler-tls` Secret), certificates are not updated mid-cycle and thus remain unaffected.
- For certificates managed by CertManager, HAMI implements self-signed certificates via Certificate+Issuer resources and renews the secret  periodically. Upon renewal, the certificate's date stamp changes.

When creating a Pod, the API server,  carrying the updated certificate, invokes the HAMI webhook; however, the actual webhook HTTPS server continues to utilise the outdated certificate loaded during server startup. Due to this certificate mismatch, the webhook returns the error `tls: bad certificate`.

As the hami webhook's default `failurePolicy` is set to `Ignore`, the Pod is still created after the error. However, the Pod's `schedulerName` remains `default-scheduler`, and its resource field includes resources managed by hami itself, such as `nvidia.com/gpucores`. Since these resources are unavailable in the cluster, so the default-scheduler will report an error: `Insufficient xxx`.

**Environment**:
- HAMi version: 2.7.1

peachest:
**What type of PR is this?**
/kind bug

Fixes #1572 

There is approximately a **one-minute interval** between the actual certificate update and the certwatcher detecting the change and loading the latest certificate content. If a Pod is created during this period, the API server will still report the error `tls: bad certificate` when invoking the webhook. By changing the webhook's `failurePolicy` from `Ignore` to `Fail`, one can prevent the creation of GPU Pods with a schedulerName not matching `hami-scheduler` due to webhook invocation failures.

Logs for `vgpu-scheduler-extender` container:
image

>[!note]
> Notice how the certwatcher detected current TLS certificate is updated and trigger CHMOD events to load the newest tls.crt and tls.key file content

Events for replicaSet which is deployed after certificate for hami-scheduler is updated
image

>[!note]
> The api-server denied the pod create request from replicaSet because the webhook server returned `tls: bad certificate` error and the `failurePolicy` of webhook is set to `Fail`.
> One minute later, certwatcher has completed certificate hot reloading, the webhook server is functioning normally, and replicaSet can create Pods without issue. Consequently, the creator/user of this workload will not perceive the certificate hot reloading occurring midway.

### DIFF
--- a/charts/hami/values.yaml
+++ b/charts/hami/values.yaml
@@ -170,7 +170,7 @@ scheduler:
       # - kube-system
       # - istio-system
     reinvocationPolicy: Never
-    failurePolicy: Fail
+    failurePolicy: Ignore
   ## TLS Certificate Option 1: Use cert-manager to generate self-signed certificate.
   ## If enabled, always takes precedence over options 2.
   certManager:


charts/hami/values.yaml
```diff
@@ -170,7 +170,7 @@ scheduler:
       # - kube-system
       # - istio-system
     reinvocationPolicy: Never
-    failurePolicy: Ignore
+    failurePolicy: Fail
```
FouoF: Set the `failurePolicy` to `Fail` is dangerous. It will block all pods creating even those use default scheduler when the webhook crash.
And currently the webhook also applied to itself so this change will make it hard to recover from misconfigures.

peachest: Given that setting it to `Fail` could potentially have a broader impact than `Ignore`, I have decided to revert to the original `Ignore` configuration. This way, even in extreme edge cases where a pod is not processed by the Hami webhook, the issue can be resolved simply by manually deleting and re-creating the pod. This avoids widespread pod creation failures, including for HAMi itself.

archlitchi (approved):
/lgtm

PR #1574, opened 2026-01-09
update Makefile for helm package

archlitchi:
/kind bug

We have added hami-dra to chart dependency, we need to resolve dependency before package to avoid release error


Shouren (approved):
/lgtm

PR #1591, opened 2026-01-15
check resource quota in webhook

DSFans2014:
**What type of PR is this?**

/kind feature

currently, resource quotas are validated in the hami-scheduler's `Fit` phase. If validation fails, the pod remains in `Pending` state. This behavior is inconsistent with Kubernetes' native validation.
However, some resources such as `nvidia.com/gpumem-percentage` can only be determined with specific values during the scheduling phase, and cannot be validated at the webhook stage

So we only check the fixed resource request such as `nvidia.com/gpumem` and `nvidia.com/gpucores`  in webhook

Fixes #


charts/hami/values.yaml
```diff
@@ -163,14 +163,26 @@ scheduler:
...
-      path: /webhook
+      path: /webhook/mutating
     whitelistNamespaces:
     # Specify the namespaces that the webhook will not be applied to.
       # - default
       # - kube-system
       # - istio-system
     reinvocationPolicy: Never
     failurePolicy: Ignore
+  validatingWebhook:
```
DSFans2014: reusing mutaing

PR #1604, opened 2026-01-20
fix: Reponse with correct content-type

Shouren:
**What type of PR is this?**
/kind bug

An incorrect `Content-Type` header can result in errors during response body parsing

Fixes #


archlitchi (approved):
/lgtm

PR #1605, opened 2026-01-20
feat: check resource quota in webhook

DSFans2014:
**What type of PR is this?**

/kind feature

currently, resource quotas are validated in the hami-scheduler's `Fit` phase. If validation fails, the pod remains in `Pending` state. This behavior is inconsistent with Kubernetes' native validation.
However, some resources such as `nvidia.com/gpumem-percentage` can only be determined with specific values during the scheduling phase, and cannot be validated at the webhook stage

So we only check the fixed resource request such as `nvidia.com/gpumem` and `nvidia.com/gpucores`  in webhook
Fixes #


pkg/scheduler/webhook.go
```diff
@@ -96,10 +97,62 @@ func (h *webhook) Handle(_ context.Context, req admission.Request) admission.Res
...
+
+func fitResourceQuota(pod *corev1.Pod) bool {
+	for deviceName, dev := range device.GetDevices() {
+		// Only supports NVIDIA
+		if deviceName != nvidia.NvidiaGPUDevice {
+			continue
+		}
+		memoryFactor := nvidia.MemoryFactor
+		resourceNames := dev.GetResourceNames()
+		resourceName := corev1.ResourceName(corev1.ResourceName(resourceNames.ResourceCountName))
```
FouoF: It looks like a typo here for casting `resourceNames.ResourceCountName` to `corev1.ResourceName` twice.

FouoF (approved):

PR #1612, opened 2026-01-21
add device type label in metrics

### LINKED ISSUE #1611: add metric with gpu type label
For example, i want to add gputype in GPUDeviceCoreAllocated and the others, because i think auto scale gpu instance need to know the gpu type when the cluster have different GPU type.
Image

Image

**What type of PR is this?**

/kind feature

Fixes #

xiyichan:
**What type of PR is this?**

Fixes #1611 


xiyichan:
> Nvidia DCGM uses label like `modelName="NVIDIA A100-SXM4-40GB"` for this field and as I known some other products such as Metax also uses the same label. So I think we should keep consistent with them for better compatibility.

Should i change label name gputype to modelname?

PR #1630, opened 2026-01-27
fix: protect scheduler status maps

hoteye:
What type of PR is this?
/kind bug

Fixes a data-race/panic risk in scheduler metrics by guarding overviewstatus/cachedstatus with a lock and returning a snapshot map to readers. Adds a unit test to verify InspectAllNodesUsage returns a snapshot map.

Which issue(s) this PR fixes:
N/A

AI assistance disclosure: I used AI (Codex/GPT) to help analyze the code and draft the change/test; I reviewed and validated the final fix.

No. This is an internal concurrency safety fix.

### DIFF
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -452,6 +452,7 @@ func (s *Scheduler) InspectAllNodesUsage() *map[string]*NodeUsage {
 	return &snapshot
 }
 
+// TODO(@Shouren): Move clone helpers to dedicated files or methods in their packages.
 func cloneNodeUsage(usage *NodeUsage) *NodeUsage {
 	if usage == nil {
 		return nil
@@ -480,6 +481,7 @@ func cloneNodeUsage(usage *NodeUsage) *NodeUsage {
 	return copied
 }
 
+// TODO(@Shouren): Move clone helpers to dedicated files or methods in their packages.
 func cloneDeviceUsage(src *device.DeviceUsage) *device.DeviceUsage {
 	if src == nil {
 		return nil
@@ -507,6 +509,7 @@ func cloneDeviceUsage(src *device.DeviceUsage) *device.DeviceUsage {
 	return &dst
 }
 
+// TODO(@Shouren): Move clone helpers to dedicated files or methods in their packages.
 func clonePodInfo(src *device.PodInfo) *device.PodInfo {
 	if src == nil {
 		return nil
@@ -522,6 +525,7 @@ func clonePodInfo(src *device.PodInfo) *device.PodInfo {
 	return dst
 }
 
+// TODO(@Shouren): Move clone helpers to dedicated files or methods in their packages.
 func clonePodDevices(src device.PodDevices) device.PodDevices {
 	if src == nil {
 		return nil


pkg/scheduler/scheduler.go
```diff
@@ -224,8 +225,8 @@ func (s *Scheduler) onDelNode(obj any) {
 // cleanupNodeUsage removes the node from overviewstatus and cachedstatus maps
 // to ensure metrics no longer report data for deleted nodes.
 func (s *Scheduler) cleanupNodeUsage(nodeID string) {
-	s.lock.Lock()
-	defer s.lock.Unlock()
+	s.statusLock.Lock()
```
FouoF, line 229: `getNodesUsage` also writes the `overviewstatus` and it is not protected. Can you check it?

pkg/scheduler/scheduler.go
```diff
@@ -439,8 +440,13 @@ func (s *Scheduler) WaitForCacheSync(ctx context.Context) bool {
 }
 
 // InspectAllNodesUsage is used by metrics monitor.
+// It returns a shallow snapshot of the map; callers must treat returned values as read-only.
 func (s *Scheduler) InspectAllNodesUsage() *map[string]*NodeUsage {
-	return &s.overviewstatus
+	s.statusLock.RLock()
+	defer s.statusLock.RUnlock()
+	snapshot := make(map[string]*NodeUsage, len(s.overviewstatus))
+	maps.Copy(snapshot, s.overviewstatus)
```
DSFans2014: `maps.Copy` is shallow copy. when other places use/modify `NodeUsage`, there may have the data race?

DSFans2014: now the `NodeUsage` is not readonly.  such as the below:
https://github.com/Project-HAMi/HAMi/blob/a217c72a3b5d7cbf4f0c2bf4a91f020bbf840a0c/pkg/scheduler/score.go#L69
so I think it's better to copy the `NodeUsage` before assign to `s.overviewstatus` or `cachenodeMap`
https://github.com/Project-HAMi/HAMi/blob/a217c72a3b5d7cbf4f0c2bf4a91f020bbf840a0c/pkg/scheduler/scheduler.go#L538-L549

DSFans2014: 1. I think data security and program stability are more important than little performance loss. That's also the reason for this PR.
2. we can address performance issue (if there is) through many approaches like cache, copy on write, or as you mentioned above and so on.

hoteye: Agree. NodeUsage is mutated in the scheduling path (e.g. sort / AddResourceUsage), so a shallow snapshot still shares pointers and isn’t fully safe. I’ll switch to separating the two: keep cachenodeMap as the mutable working set for scheduling, and publish a deep‑copied snapshot for overviewstatus (metrics). That gives data safety; if performance becomes an issue we can revisit with COW/DTO.

pkg/scheduler/scheduler.go
```diff
@@ -440,7 +442,106 @@ func (s *Scheduler) WaitForCacheSync(ctx context.Context) bool {
...
+	defer s.statusLock.RUnlock()
+	snapshot := make(map[string]*NodeUsage, len(s.overviewstatus))
+	//nolint:modernize // Deep copy is required; maps.Copy would keep shared pointers.
+	for nodeID, usage := range s.overviewstatus {
+		snapshot[nodeID] = cloneNodeUsage(usage)
+	}
+	return &snapshot
+}
+
+func cloneNodeUsage(usage *NodeUsage) *NodeUsage {
```
Shouren, line 456: @hoteye It seems to me that we have just implemented the `Clone` method for `NodeUsage`, `DeviceUsage`, `PodInfo` and `PodDevices` structs, can we make it a method for those structs or is there some reason that we must put it in `scheduler.go` ?

Shouren: @hoteye Sorry for the late reply. Let's add some TODO comments to those functions for future refactoring.

PR #1637, opened 2026-01-29
fix: handle GetMemoryInfo ERROR_NOT_SUPPORTED for unified memory GPUs

### LINKED ISSUE #1511: [Bug] Device plugin panics on NVIDIA GB10 (DGX Spark) - GetMemoryInfo returns "Not Supported"
## Description
HAMi device plugin v2.7.1 crashes with panic when running on NVIDIA DGX Spark with GB10 GPU. The GB10 GPU uses unified memory architecture and NVML's `GetMemoryInfo()` returns "Not Supported", causing the device plugin to panic instead of handling the error gracefully.

## Environment
- **HAMi Version**: v2.7.1
- **Kubernetes Version**: v1.32.9 (RKE2)
- **Node Architecture**: ARM64
- **GPU**: NVIDIA GB10 (DGX Spark)
- **GPU Architecture**: Blackwell
- **Driver Version**: 580.95.05
- **CUDA Version**: 13.0
- **OS**: Ubuntu 24.04

## GPU Information
Product Name : NVIDIA GB10
Product Brand : NVIDIA RTX
Product Architecture : Blackwell

## nvidia-smi output
The GB10 GPU shows "Not Supported" for Memory-Usage, which is expected for unified memory architecture:
+-----------------------------------------------------------------------------------------+
| NVIDIA-SMI 580.95.05 Driver Version: 580.95.05 CUDA Version: 13.0 |
+-----------------------------------------+------------------------+----------------------+
| GPU Name Persistence-M | Bus-Id Disp.A | Volatile Uncorr. ECC |
| Fan Temp Perf Pwr:Usage/Cap | Memory-Usage | GPU-Util Compute M. |
|=========================================+========================+======================|
| 0 NVIDIA GB10 On | 0000000F:01:00.0 Off | N/A |
| N/A 30C P8 4W / N/A | Not Supported | 0% Default |
+-----------------------------------------+------------------------+----------------------+
## Error Log
I1201 14:46:48.346963 1795965 register.go:200] Starting WatchAndRegister
I1201 14:46:48.346973 1795965 health.go:61] Check Health start Running
E1201 14:46:48.355522 1795965 register.go:114] nvml get memory error ret=Not Supported
panic: 0

goroutine 74 [running]:
github.com/Project-HAMi/HAMi/pkg/device-plugin/nvidiadevice/nvinternal/plugin.(*NvidiaDevicePlugin).getAPIDevices(0x4000430000)
/k8s-vgpu/pkg/device-plugin/nvidiadevice/nvinternal/plugin/register.go:115 +0x938

## Root Cause
The issue is in `pkg/device-plugin/nvidiadevice/nvinternal/plugin/register.go` at line 109-115:

```go
memory, ret := ndev.GetMemoryInfo()
if ret == nvml.SUCCESS {
    memoryTotal = int(memory.Total)
} else {
    klog.Error("nvml get memory error ret=", ret)
    panic(0)  // <-- This causes the crash
}

### PR DESCRIPTION
**Pull request by jsl9208**
**What type of PR is this?**

/kind bug

NVIDIA GB10 (DGX Spark) uses a unified memory architecture where CPU and GPU share the same physical memory. On these GPUs, `nvmlDeviceGetMemoryInfo()` returns `ERROR_NOT_SUPPORTED` instead of memory information. The current code treats any non-SUCCESS return as fatal and calls `panic(0)`, which crashes the entire device plugin daemonset and prevents the node from registering any GPU devices.

This PR handles `ERROR_NOT_SUPPORTED` gracefully by:
1. **Device registration** (`register.go`): Falls back to a new `defaultDeviceMemory` config value (in MiB). If not configured, the device is skipped with an error log instead of panicking.
2. **Metrics collection** (`metrics.go`): Skips memory metrics for unsupported devices instead of returning an error every scrape cycle.
3. **Config plumbing**: Adds `defaultDeviceMemory` field to `NvidiaConfig`, Helm chart configmap, and `values.yaml`.

Fixes #1511

AI assistance disclosure: This PR was developed with Claude Code assisting in code analysis and review. The fix was designed, implemented, and validated by a human on real GB10 hardware.

Tested on real hardware: NVIDIA DGX Spark (GB10, ARM64, Driver 580.95.05, CUDA 13.0, Ubuntu 24.04, K8s v1.34.1).

| Scenario | Result |
|----------|--------|
| Unpatched v2.8.0 on GB10 | `panic: 0` at `register.go:115` ❌ |
| Patched with `defaultDeviceMemory: 131072` | Device registered, pod scheduled, vGPU isolation works ✅ |
| Patched without `defaultDeviceMemory` | Device skipped gracefully, no crash ✅ |

Usage — for unified memory GPUs, set `defaultDeviceMemory` to the total GPU memory in MiB:

```yaml
# values.yaml
devicePlugin:
  defaultDeviceMemory: 131072  # 128 GiB for GB10
```

Without this config, the device will be skipped (not registered) but the plugin won't crash.

Yes. Adds a new optional Helm value `devicePlugin.defaultDeviceMemory` (default: `0`). Only needed for GPUs with unified memory architecture (e.g., NVIDIA GB10/DGX Spark) where `nvmlDeviceGetMemoryInfo()` is not supported. When set, the device plugin uses this value as the total device memory fallback instead of panicking.

### DIFF
--- a/charts/hami/templates/scheduler/device-configmap.yaml
+++ b/charts/hami/templates/scheduler/device-configmap.yaml
@@ -22,6 +22,7 @@ data:
       defaultMemory: 0
       defaultCores: 0
       defaultGPUNum: 1
+      preConfiguredDeviceMemory: {{ .Values.devicePlugin.preConfiguredDeviceMemory | default 0 }}
       memoryFactor: 1
       deviceSplitCount: {{ .Values.devicePlugin.deviceSplitCount }}
       deviceMemoryScaling: {{ .Values.devicePlugin.deviceMemoryScaling }}
--- a/pkg/device-plugin/nvidiadevice/nvinternal/plugin/register.go
+++ b/pkg/device-plugin/nvidiadevice/nvinternal/plugin/register.go
@@ -108,9 +108,23 @@ func (plugin *NvidiaDevicePlugin) getAPIDevices() *[]*device.DeviceInfo {
 		}
 		memoryTotal := 0
 		memory, ret := ndev.GetMemoryInfo()
-		if ret == nvml.SUCCESS {
+		switch ret {
+		case nvml.SUCCESS:
 			memoryTotal = int(memory.Total)
-		} else {
+		case nvml.ERROR_NOT_SUPPORTED:
+			// Unified memory architecture GPUs (e.g., NVIDIA GB10/DGX Spark) don't support
+			// traditional memory queries. Use PreConfiguredDeviceMemory from config as fallback.
+			if plugin.schedulerConfig.PreConfiguredDeviceMemory != nil && *plugin.schedulerConfig.PreConfiguredDeviceMemory > 0 {
+				memoryTotal = int(*plugin.schedulerConfig.PreConfiguredDeviceMemory) * 1024 * 1024
+				klog.Warningf("GetMemoryInfo not supported for device %s, using configured PreConfiguredDeviceMemory: %d MB",
+					UUID, *plugin.schedulerConfig.PreConfiguredDeviceMemory)
+			} else {
+				klog.Errorf("GetMemoryInfo not supported for device %s (unified memory architecture) "+
+					"and PreConfiguredDeviceMemory not configured. Skipping this device. "+
+					"Set 'preConfiguredDeviceMemory' in nvidia config to the total GPU memory in MB.", UUID)
+				continue
+			}
+		default:
 			klog.Error("nvml get memory error ret=", ret)
 			panic(0)
 		}
--- a/pkg/device/nvidia/device.go
+++ b/pkg/device/nvidia/device.go
@@ -124,9 +124,10 @@ type NvidiaConfig struct {
 
 // These configs can be specified for each node by using Nodeconfig.
 type NodeDefaultConfig struct {
-	DeviceSplitCount    *uint    `yaml:"deviceSplitCount" json:"devicesplitcount"`
-	DeviceMemoryScaling *float64 `yaml:"deviceMemoryScaling" json:"devicememoryscaling"`
-	DeviceCoreScaling   *float64 `yaml:"deviceCoreScaling" json:"devicecorescaling"`
+	DeviceSplitCount          *uint    `yaml:"deviceSplitCount" json:"devicesplitcount"`
+	DeviceMemoryScaling       *float64 `yaml:"deviceMemoryScaling" json:"devicememoryscaling"`
+	DeviceCoreScaling         *float64 `yaml:"deviceCoreScaling" json:"devicecorescaling"`
+	PreConfiguredDeviceMemory *int64   `yaml:"preConfiguredDeviceMemory" json:"preconfigureddevicememory"`
 	// LogLevel is LIBCUDA_LOG_LEVEL value
 	LogLevel *LibCudaLogLevel `yaml:"libCudaLogLevel" json:"libcudaloglevel"`
 }
--- a/charts/hami/values.yaml
+++ b/charts/hami/values.yaml
@@ -302,6 +302,10 @@ devicePlugin:
   deviceSplitCount: 10
   deviceMemoryScaling: 1
   deviceCoreScaling: 1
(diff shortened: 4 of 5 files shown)


pkg/device-plugin/nvidiadevice/nvinternal/plugin/register.go
```diff
@@ -108,9 +108,23 @@ func (plugin *NvidiaDevicePlugin) getAPIDevices() *[]*device.DeviceInfo {
...
+			} else {
+				klog.Errorf("GetMemoryInfo not supported for device %s (unified memory architecture) "+
+					"and DefaultDeviceMemory not configured. Skipping this device. "+
+					"Set 'defaultDeviceMemory' in nvidia config to the total GPU memory in MB.", UUID)
+				continue
+			}
+		default:
 			klog.Error("nvml get memory error ret=", ret)
 			panic(0)
 		}
```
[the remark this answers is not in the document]
jsl9208: That's pre-existing code outside this diff. Keeping this PR minimal to the bugfix scope.

pkg/device-plugin/nvidiadevice/nvinternal/plugin/register.go
```diff
@@ -108,9 +108,23 @@ func (plugin *NvidiaDevicePlugin) getAPIDevices() *[]*device.DeviceInfo {
...
+				klog.Warningf("GetMemoryInfo not supported for device %s, using configured DefaultDeviceMemory: %d MB",
+					UUID, plugin.schedulerConfig.DefaultDeviceMemory)
+			} else {
+				klog.Errorf("GetMemoryInfo not supported for device %s (unified memory architecture) "+
+					"and DefaultDeviceMemory not configured. Skipping this device. "+
+					"Set 'defaultDeviceMemory' in nvidia config to the total GPU memory in MB.", UUID)
+				continue
+			}
+		default:
 			klog.Error("nvml get memory error ret=", ret)
```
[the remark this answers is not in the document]
jsl9208: Same here, out of scope for this fix.

charts/hami/templates/scheduler/device-configmap.yaml
```diff
@@ -22,6 +22,7 @@ data:
       defaultMemory: 0
       defaultCores: 0
       defaultGPUNum: 1
+      defaultDeviceMemory: {{ .Values.devicePlugin.defaultDeviceMemory | default 0 }}
```
archlitchi: please rename this value to 'preConfiguredDeviceMemory' to better demonstrate this field.

jsl9208: > please rename this value to 'preConfiguredDeviceMemory' to better demonstrate this field.

Done.

pkg/device/nvidia/device.go
```diff
@@ -112,6 +112,7 @@ type NvidiaConfig struct {
 	DefaultMemory                int32  `yaml:"defaultMemory"`
 	DefaultCores                 int32  `yaml:"defaultCores"`
 	DefaultGPUNum                int32  `yaml:"defaultGPUNum"`
+	PreConfiguredDeviceMemory    int64  `yaml:"preConfiguredDeviceMemory"`
```
Shouren: @jsl9208 The `PreConfiguredDeviceMemory` in `NvidiaConfig` is a cluster level config, Is there a scenario where different nodes may have different `PreConfiguredDeviceMemory` ?

jsl9208: > @jsl9208 The `PreConfiguredDeviceMemory` in `NvidiaConfig` is a cluster level config, Is there a scenario where different nodes may have different `PreConfiguredDeviceMemory` ?

Yes, a mixed cluster with different unified-memory GPUs would need per-node values. The existing NodeDefaultConfig / nodeconfig mechanism could support this by moving the field there. 

Would you like me to add that in this PR, or is a follow-up OK?

Shouren: @jsl9208 I prefer to add it in this PR, i think it's not a big change to current implemention

jsl9208: > @jsl9208 I prefer to add it in this PR, i think it's not a big change to current implemention

Done. Added PreConfiguredDeviceMemory to NodeDefaultConfig struct to support per-node configuration.
Tested on GB10 with node-level config override, works as expected.

PR #1645, opened 2026-02-05
feat: add support for vastai device

DSFans2014:
**What type of PR is this?**

/kind feature

Fixes #

### DIFF
--- a/docs/vastai-support.md
+++ b/docs/vastai-support.md
@@ -12,7 +12,201 @@ We now support sharing `vastaitech.com/va` (Vastaitech) devices and provides the
 
 ### Enabling Vastai Device Sharing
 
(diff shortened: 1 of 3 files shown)


juice928:
👋 Hi, I'm an automated AI code review bot. I ran some checks on this PR and found **1 point that might be worth attention** (could be false positives, please use your judgment):

1. <strong>The memory allocation for Vastai devices appears to be currently fixed at 100%</strong>
   - [Location](https://github.com/Project-HAMi/HAMi/blob/f9898f4c669634bb45745976fb8a45e1cdc5886c/pkg/device/vastai/device.go#L152-L160): `pkg/device/vastai/device.go:L152-L160`
   - Impact: This prevents device sharing as the first container claims all memory, causing subsequent containers to fail the resource check.
   - Suggestion: Consider calculating `MemPercentagereq` dynamically based on the actual request ratio to enable effective time-slicing.

If you find these suggestions disruptive, you can reply "stop" , and I'll automatically skip this repository in the future.

PR #1650, opened 2026-02-11
fix: support device allocation for multi-container  with init containers

haitwang-cloud:
**What type of PR is this?**

/kind bug
This commit fixes the 'device request not found' error that occurs when
allocating devices to multi-container pods that contain init containers.

Problem:
- When a pod has both init containers and regular containers, the device
  allocation index mapping was incorrect
- The annotation format uses ';' to separate containers, but the code was
  not properly handling the index offset between init containers and
  regular containers
- Empty ContainerDevices entries were being skipped, breaking the index
  mapping

Fixes #

### DIFF
--- a/pkg/device-plugin/nvidiadevice/nvinternal/plugin/util_test.go
+++ b/pkg/device-plugin/nvidiadevice/nvinternal/plugin/util_test.go
@@ -99,7 +99,7 @@ func TestGenerateMigTemplate(t *testing.T) {
 			expectedPos:   1,
 			expectedReset: true,
 			expectedMig: map[string]int32{
-				"1g.5gb": 1,
+				"1g.5gb":  1,
 				"2g.10gb": 3,
 			},
 		},
--- a/pkg/device/devices_test.go
+++ b/pkg/device/devices_test.go
@@ -850,7 +850,7 @@ type mockDevices struct {
 	resourceRequest ContainerDeviceRequest
 }
 
-func (m *mockDevices) CommonWord() string                                  { return "mock" }
+func (m *mockDevices) CommonWord() string { return "mock" }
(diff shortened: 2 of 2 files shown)


pkg/device-plugin/nvidiadevice/nvinternal/plugin/util.go
```diff
@@ -61,9 +61,27 @@ func GetNextDeviceRequest(dtype string, p corev1.Pod) (corev1.Container, device.
 	if !ok {
 		return corev1.Container{}, res, errors.New("device request not found")
 	}
+	
+	// The annotation format follows the order: init containers first, then regular containers
+	// Index mapping:
+	//   0 to len(InitContainers)-1: init containers
+	//   len(InitContainers) to len(InitContainers)+len(Containers)-1: regular containers
+	initContainerCount := len(p.Spec.InitContainers)
```
Shouren, line 69: @haitwang-cloud AFAIK, HAMi does not support allocating devices for initContainers. So let me know if i miss something.

haitwang-cloud: @Shouren TnitContainers can seamlessly obtain vGPU from the Nvidia K8s-device-plugin, so I think that Hami should support the same functionality.

haitwang-cloud:
@DSFans2014 Done,PTAL again

pkg/device-plugin/nvidiadevice/nvinternal/plugin/util.go
Shouren: @haitwang-cloud Agree, can you create an issue to track this? And we can discuss this feature in the weekly meeting.

haitwang-cloud:
@Shouren @DSFans2014 @archlitchi PTAL, and let me know if u have any concerns

PR #1651, opened 2026-02-11
chore: remove deprecated scheduler policy configmap

haitwang-cloud:
**What type of PR is this?**

/kind cleanup

The scheduler policy configmap (config.json) is no longer needed as the project has moved to a different scheduler configuration approach. This file was used for the legacy Kubernetes scheduler extender policy configuration.

This change removes the deprecated configmap template that was conditionally created when scheduler.kubeScheduler.enabled was true and dra.enabled was false.

Fixes #

### DIFF
--- a/charts/hami/templates/scheduler/configmap.yaml
+++ b/charts/hami/templates/scheduler/configmap.yaml
@@ -38,7 +38,7 @@ data:
       weight: 1
       httpTimeout: 30s
       managedResources:
-      {{- include "hami-vgpu.scheduler.managedResources.yaml" . | nindent 6 }}
+      {{- include "hami-vgpu.scheduler.managedResources" . | nindent 6 }}
 {{- else }}
   {{/* K8s < 1.22: Use legacy Policy API */}}
   config.json: |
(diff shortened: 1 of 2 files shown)


haitwang-cloud:
@DSFans2014 Thank you very much for pointing that out. I've explored and adopted another, more sophisticated approach to deal with this matter. I've thoroughly verified that it functions correctly within our Kubernetes (K8s) cluster. Please take a look again.

charts/hami/templates/scheduler/configmap.yaml
```diff
@@ -30,113 +63,11 @@ data:
...
-                    },
-                    {
-                        "name": "{{ .Values.metaxResourceCore }}",
-                        "ignoredByScheduler": true
-                    },
-                    {
-                        "name": "{{ .Values.metaxResourceMem }}",
-                        "ignoredByScheduler": true
-                    }
+                    {{- include "hami-vgpu.scheduler.managedResources.json" . | nindent 20 }}
```
DSFans2014: Can `hami-vgpu.scheduler.managedResources.yaml` be directly converted to JSON here by `toJson` to avoid duplicate definitions?

haitwang-cloud: @DSFans2014 I think in that case it would increase the structural complexity and make it hard to maintain; the current design is already clear, simple, and easy to understand.

DSFans2014 (commented):
/lgtm

archlitchi (approved):
/lgtm

PR #1677, opened 2026-03-03
fix: failing readiness probe when replica > 1

### LINKED ISSUE #1617: Failing readiness probe on 2.8.0 when replicas>1
**What happened**:
After upgrading from `2.7.1` to `2.8.0` readiness probe is failing on 2 of 3 replicas of `hami-scheduler/vgpu-scheduler-extender`. Looks that only currently elected master is green, rest is yellow. It bothers ArgoCD and it's confusing for people maintaining environment. Otherwise HAMI functions properly.

All replicas should be green all the time.

Set such values for HELM chart:
```
scheduler:
  replicas: 3
  leaderElect: true
```

**Environment**:
- HAMi version: 2.8.0

Shouren:
**What type of PR is this?**
/kind bug

As described in #1617, the PR #1553 brings the readinessProbe check for vgpu-scheduler-extender container in hami-scheduler Pod, however it the current implementation don't work well with the rolling update strategy of deployment.

So in this PR, A new label is introduced to make sure that only the leader is selected by the service

Fixes #1617

**AI Assistant disclosure**:
I used kimi-cli to generate the testing code

### DIFF
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"maps"
+	"os"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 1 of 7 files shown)


pkg/scheduler/scheduler.go
```diff
@@ -353,11 +354,45 @@ func (s *Scheduler) register(labelSelector labels.Selector, printedLog map[strin
...
+			if err != nil {
+				klog.ErrorS(err, "Failed to patch leader label to pod", "namespace", os.Getenv("POD_NAMESPACE"), "pod", os.Getenv("POD_NAME"))
+			} else {
+				klog.V(4).InfoS("Successfully patched leader label to hami scheduler pod", "namespace", os.Getenv("POD_NAMESPACE"), "pod", os.Getenv("POD_NAME"))
+			}
+		}
+	} else {
+		// Current pod is not leader, add or change the role label to follower
+		if pod.Labels != nil && pod.Labels[util.HAMiRoleLabel] != util.HAMiRoleLabelValueFollower {
+			err := util.PatchPodLabels(
```
FouoF: The follower update their roles themself may not reliable. If the leader crash, the role label will not be updated and two leader will exist together (Though the crashed one will removed from the endpoint slice). Let the leader to update previous leader to follower may be more reliable.

Shouren: @FouoF Done.

krzysztof-magosa:
@Shouren hi, any idea when this will be released? thanks

Shouren:
> @Shouren hi, any idea when this will be released? thanks

@krzysztof-magosa It supposed to be released next month(maybe second week) in v2.9.0

PR #1715, opened 2026-03-26
fix vastai fit loop iteration direction

DSFans2014:
**What type of PR is this?**

- fix vastai fit loop iteration direction

Fixes #

### DIFF
--- a/pkg/device/vastai/device.go
+++ b/pkg/device/vastai/device.go
@@ -231,7 +231,7 @@ func (va *VastaiDevices) Fit(devices []*device.DeviceUsage, request device.Conta
 	tmpDevs := make(map[string]device.ContainerDevices)
 	reason := make(map[string]int)
 	dieMode := isDieMode(devices)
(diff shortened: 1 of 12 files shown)


pkg/device/vastai/device.go
```diff
@@ -231,7 +231,7 @@ func (va *VastaiDevices) Fit(devices []*device.DeviceUsage, request device.Conta
 	tmpDevs := make(map[string]device.ContainerDevices)
 	reason := make(map[string]int)
 	dieMode := isDieMode(devices)
-	for i := range len(devices) {
+	for i := len(devices) - 1; i >= 0; i-- {
```
Shouren: @DSFans2014 I think we can use `slices.Backward` as mentioned in [this](https://cs.opensource.google/go/go/+/master:src/slices/iter.go;l=26) package

Shouren (commented):
/lgtm

FouoF (approved):

PR #1743, opened 2026-04-10
fix(device-plugin): align kubelet allocation with scheduler annotations (#1741)

### LINKED ISSUE #1741: Device allocation mismatch: scheduler assigns 8 GPUs but Allocate only receives 5 unique devices
## Problem Description

When using HAMi with 8 H100 GPUs and a pod requesting 8 GPUs, the scheduler correctly assigns 8 GPUs (as seen in pod annotations), but the kubelet `Allocate` request only contains 5 unique physical GPU devices. This causes downstream tools like DCGM Exporter to only see 5 GPU-Pod associations instead of 8.

## Environment

- HAMi version: v2.7.1
- Kubernetes version: v1.32
- GPU: H100 x 8
- Container runtime: containerd 1.6.28

## Logs

Pod annotation shows 8 GPUs:
```
hami.io/vgpu-devices-allocated: "GPU-1cb8edae-82f3-6a6f-1c76-a25281b02033:1,2,8,17,12,6,13,10,NVIDIA,NVIDIA A100 80GB PCIe"
```

But kubelet Allocate request only has 5 unique physical GPUs (with duplicates):
```
Allocate req.DevicesIDs=[GPU-1cb8edae-82f3-6a6f-1c76-a25281b02033-17 GPU-1cb8edae-82f3-6a6f-1c76-a25281b02033-6 GPU-1cb8edae-82f3-6a6f-1c76-a25281b02033-2 GPU-f451c579-45ef-a275-4c26-9a0d074f3c27-1 GPU-34b65693-63f9-a612-8353-ede963df7e0e-1 GPU-1cb8edae-82f3-6a6f-1c76-a25281b02033-13 GPU-f451c579-45ef-a275-4c26-9a0d074f3c27-6 GPU-1cb8edae-82f3-6a6f-1c76-a25281b02033-10]
```

Note: GPU-1cb8edae... appears 4 times, GPU-f451c579... appears 2 times.

## Root Cause

1. `GetPluginDevices()` generates virtual devices (GPU-UUID-N format, 10 per physical GPU)
2. kubelet scheduler with binpack policy selects 8 virtual devices
3. These 8 virtual devices map to only 5 unique physical GPUs
4. `Allocate()` uses `req.DevicesIDs` (5 unique) instead of validating against annotation devices (8 GPUs)

## Suggested Fix

In `Allocate()`, validate that `req.DevicesIDs` covers all devices from pod annotations. If not, override with annotation devices to ensure all allocated GPUs are properly passed to containers.

## Related Code

- `pkg/device-plugin/nvidiadevice/nvinternal/plugin/server.go` - Allocate()
- `pkg/device-plugin/nvidiadevice/nvinternal/rm/devices.go` - GetPluginDevices()

xrwang8:
## Problem

When using HAMi with 8 H100 GPUs and a pod requesting 8 GPUs, the scheduler correctly assigns 8 GPUs (as seen in pod annotations), but the kubelet `Allocate` request only contains 5 unique physical GPU devices. This causes downstream tools like DCGM Exporter to only see 5 GPU-Pod associations instead of 8.

## Root Cause

1. `GetPluginDevices()` generates virtual devices (GPU-UUID-N format, 10 per physical GPU)
2. kubelet scheduler with binpack policy selects 8 virtual devices
3. These 8 virtual devices map to only 5 unique physical GPUs
4. `Allocate()` uses `req.DevicesIDs` (5 unique) instead of validating against annotation devices (8 GPUs)

## Solution

This fix:
- Enables `GetPreferredAllocation` to override kubelet's default device selection
- Adds `selectPreferredDeviceIDsFromAnnotatedDevices()` to align device selection with scheduler decisions from pod annotations
- Ensures requested GPU count matches allocated devices

## Changes

- `pkg/device-plugin/nvidiadevice/nvinternal/plugin/server.go`: Enable GetPreferredAllocation, add alignment logic
- `pkg/device-plugin/nvidiadevice/nvinternal/plugin/server_test.go`: Add unit tests

## Testing

```bash
go test -v ./pkg/device-plugin/nvidiadevice/nvinternal/plugin/...
```

All tests pass.

Fixes #1741


archlitchi:
CC @Shouren

wawa0210:
----    
cc @S descendants

pkg/device-plugin/nvidiadevice/nvinternal/plugin/server.go
```diff
@@ -458,27 +460,122 @@ func (plugin *NvidiaDevicePlugin) ListAndWatch(e *kubeletdevicepluginv1beta1.Emp
...
+		candidates := availableByPhysical[physicalID]
+		if len(candidates) == 0 {
+			return nil, fmt.Errorf("no available slice device found for annotated GPU %s", physicalID)
+		}
+		selected = append(selected, candidates[0])
+		availableByPhysical[physicalID] = candidates[1:]
+	}
+
+	if len(selected) != allocationSize {
+		return nil, fmt.Errorf("preferred allocation selected %d devices, want %d", len(selected), allocationSize)
```
Shouren, line 545: @xrwang8 @archlitchi I think this introduces a breaking change to the current device allocation mechanism. Should it be enforced strictly or treated as best-effort？How do you think ?

Update:
The `deviceManager` of `kubelet` seems to return an error if the `callGetPreferredAllocationIfAvailable` [function](https://github.com/kubernetes/kubernetes/blob/03779bbd00da25c7fcd03711ddfe466a3322e1d7/pkg/kubelet/cm/devicemanager/manager.go#L1030) returns an error in `devicesToAllocate` [function](https://github.com/kubernetes/kubernetes/blob/03779bbd00da25c7fcd03711ddfe466a3322e1d7/pkg/kubelet/cm/devicemanager/manager.go#L582)

xrwang8: I re-checked the kubelet control flow and HAMi's end-to-end allocation path.

  Because kubelet records the allocation result from `devicesToAllocate()`, fixing this only in `Allocate()` is not sufficient. The alignment needs to happen earlier, in `GetPreferredAllocation()`, so kubelet records device IDs aligned
  with the scheduler-selected devices.

  I also agree with your concern about error handling: if `GetPreferredAllocation()` returns an error, kubelet fails `devicesToAllocate()` directly instead of falling back.

  So the modification direction I am considering is:
  - keep the annotation-based alignment in `GetPreferredAllocation()` for the normal path
  - if pending pod context is unavailable, annotation cannot be decoded, or the annotation cannot be mapped safely to kubelet `AvailableDeviceIDs`, fall back to `plugin.rm.GetPreferredAllocation(...)`
  - avoid introducing a new kubelet-facing hard failure path in this PR

  This keeps the fix in the correct place while preserving the current allocation behavior in fallback cases.

Shouren: @xrwang8 Agreed with your direction. 
And i am also considering adding a Pod annotation so user can control the `GetPreferredAllocation` behavior; I think it is important to keep the monitoring tool happy in production.

xrwang8:
@Shouren  PTAL

Shouren (approved):
/lgtm

PR #1748, opened 2026-04-12
test: improve coverage for pkg/version/version.go

Yonsun-w:
**What type of PR is this?**

/kind cleanup

Adds unit coverage for `pkg/version/version.go` by testing:
- `Version()` field population
- `VersionCmd` command output

This improves the package coverage for the `#569` UT coverage improvement task.

Related to #569

This PR claims the `pkg/version/version.go` subtask from issue `#569`.

Validation performed:
- `go test ./pkg/version`
- `go test -coverprofile=/tmp/version.cover ./pkg/version`
- `bash hack/verify-license.sh`
- `bash hack/verify-import-aliases.sh`

`bash hack/verify-all.sh` currently fails in `verify-staticcheck.sh` because the installed `golangci-lint v2.8.0` binary is built with Go 1.25, while dependency analysis hits code requiring Go 1.26; this is an environment/tooling issue rather than a failure caused by this change.

No.

### DIFF
--- a/pkg/version/version_test.go
+++ b/pkg/version/version_test.go
@@ -17,9 +17,9 @@ limitations under the License.
 package version
 
 import (
+	"bytes"
 	"fmt"
 	"runtime"
-
 	"testing"
 )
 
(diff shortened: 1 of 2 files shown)


DSFans2014 (commented):
please signoff your commit

pkg/version/version_test.go
```diff
@@ -74,3 +75,57 @@ platform:         ` + fmt.Sprintf("%s/%s", runtime.GOOS, runtime.GOARCH),
...
+	}
+
+	wantPlatform := fmt.Sprintf("%s/%s", runtime.GOOS, runtime.GOARCH)
+	if got.Platform != wantPlatform {
+		t.Errorf("Version().Platform = %q, want %q", got.Platform, wantPlatform)
+	}
+}
+
+func TestVersionCmd(t *testing.T) {
+	oldStdout := os.Stdout
```
Shouren: @Yonsun-w I think it is better to replace `fmt.Println(Print())` with `cmd.Println(Print())` in the following code in `version.go` 
```Go
var (
        VersionCmd = &cobra.Command{
                Use:   "version",
                Short: "print version",
                Run: func(cmd *cobra.Command, _ []string) {
                        fmt.Println(Print())
                },
        }
)
```
Then you can use [SetOut](https://pkg.go.dev/github.com/spf13/cobra#Command.SetOut) function from `cobra.Command` which simplify the testing code. And i think this also solves the comment left by gemini-code-assist

archlitchi (approved):
/lgtm

PR #1759, opened 2026-04-15
fix(scheduler): add missing return in ondelpod

### LINKED ISSUE #1751: #scheduler.go onDelPod: Missing return in DeletedFinalStateUnknown else branch
**Please provide an in-depth description of the question you have**:
In scheduler.go, the onDelPod method handles cache.DeletedFinalStateUnknown tombstone objects like this:
https://github.com/Project-HAMi/HAMi/blob/df7e05e775cf0437b0fcb21e6d100ca0eb748ac5/pkg/scheduler/scheduler.go#L182 When the type assertion fails (the else branch), the code logs an error but does not return. This means execution continues past the switch block, where pod is still nil (its zero value from var pod *corev1.Pod). The subsequent pod.Annotations[util.AssignedNodeAnnotations] access would cause a nil pointer panic.

**What do you think about this question?**:
A one-line fix would be adding return after the klog.Errorf(...) in the else branch:
```
case cache.DeletedFinalStateUnknown:
	if pod, ok = t.Obj.(*corev1.Pod); ok {
		klog.V(4).InfoS("Pod tombstone deleted, cleaning up cache", "pod", t.Key)
	} else {
		klog.Errorf("Received tombstone for non-pod object on pod delete")
		return
	}
```

**Environment**:
- HAMi version: master (latest)
- Kubernetes version: N/A (code-level observation)
- Others: N/A

CFH2436:
**What type of PR is this?**
/kind bug

Added the missing return to ensure the function exits correctly after handling the deletion.

Fixes #1751

No

### DIFF
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -180,6 +180,7 @@ func (s *Scheduler) onDelPod(obj any) {
 			klog.V(4).InfoS("Pod tombstone deleted, cleaning up cache", "pod", t.Key)
 		} else {
 			klog.Errorf("Received tombstone for non-pod object on pod delete")
+			return
 		}
 	default:
 		klog.Errorf("Received unknown object type on pod delete")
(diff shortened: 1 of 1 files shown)


Shouren (commented):
@CFH2436 The update seems fine to me. Don't forget to sign off your commits

archlitchi (approved):
/lgtm

Shouren (approved):
/lgtm

PR #1761, opened 2026-04-15
docs: add v2.9.0 documentation audit report

mesutoezdil:
Updates the docs status document to reflect current master state (May 2026).

benchmark.md is fresh, updated with A100/vLLM/k8s-v1.35.4 data.
develop/tasklist.md removed from inventory, deleted from master.
New files added to inventory: ascend910-hami-vnpu-core-support.md, resource-quota.md, scheduling-policy.md.
Completed action items removed from the plan.

Two open questions remain at the top of the document for maintainers.
Nothing moves or renames existing files.

PR #1762, opened 2026-04-15
docs: mark benchmark.md as outdated

mesutoezdil:
/kind documentation

The benchmark.md document references Kubernetes v1.12.9 and Tesla V100 (2018).
Current GPU baselines are A100/H100 on Kubernetes v1.28+. Users evaluating HAMi
may rely on these numbers and form incorrect expectations.

This PR adds a visible warning at the top of the file. No content is removed.
A replacement benchmark with current hardware should be tracked as a follow-up issue.

Special notes for your reviewer: One file changed, 5 lines added. No existing content modified.

### DIFF
--- a/docs/benchmark.md
+++ b/docs/benchmark.md
@@ -1,3 +1,8 @@
+> [!WARNING]
+> **This document is outdated and no longer reflects current HAMi performance.**
+> It was written for Kubernetes v1.12.9 + Tesla V100 (2018). Current baselines are A100/H100 on Kubernetes v1.28+.
+> Do not use these numbers for evaluation. A replacement benchmark is tracked in [GitHub Issues](https://github.com/Project-HAMi/HAMi/issues).
+
 ## Benchmarks
 
 Three instances from ai-benchmark have been used to evaluate vGPU-device-plugin performance as follows:


rootsongjc:
Thanks for the update — makes sense to mark this as outdated.

One thing to clarify: we plan to migrate docs (including blogs/benchmarks) to the  website repo, which will become the source of truth.

Given that, I see two possible approaches:
1) Keep this minimal (just the warning) and avoid further edits here  
2) Or, if you plan to update content, it would be better to do it directly in the website repo and delete the docs here

Otherwise we may end up maintaining the same content in two places.

archlitchi:
I think we need to just replace this benchmark by using vllm

mesutoezdil:
Thanks @archlitchi and @rootsongjc!

. . .

And @rootsongjc: I believe the rest of your comment was referring to "https://github.com/Project-HAMi/HAMi/pull/1761" more as a suggestion for the website migration. Let’s see how that evolves.

archlitchi:
Hi, we've just updated the benchmark, could you check it out?

mesutoezdil:
@archlitchi thanks, the new vLLM/A100/v1.35.4 benchmark looks great and addresses the issue this PR was raising. Closing this since the deprecation warning is no longer needed.

Issue #1667, opened 2026-02-25
Device allocation issue for multi-container pods with init containers


haitwang-cloud:
**What happened**:

Based on PR [#1650](https://github.com/Project-HAMi/HAMi/pull/1650), device allocation did not properly support multi-container pods that include init containers. This PR attempts to fix device allocation behavior for such cases.

**What you expected to happen**:

Multi-container pods, including those with init containers, should be able to allocate devices correctly through HAMi, without misassignment or failures.

**How to reproduce it (as minimally and precisely as possible)**:

1. Create a pod spec with multiple containers, including at least one init container that requires device allocation.
2. Deploy to a HAMi-enabled cluster.
3. Observe device allocation results and logs.

**Anything else we need to know?**:

See https://github.com/Project-HAMi/HAMi/pull/1650 for related fixes and context.

- The output of `nvidia-smi -a` on your host
- Your docker or containerd configuration file (e.g: `/etc/docker/daemon.json`)
- The hami-device-plugin container logs
- The hami-scheduler container logs
- The kubelet logs on the node (e.g: `sudo journalctl -r -u kubelet`)

mesutoezdil:
**Note:** As of today, this issue/PR has been flagged under our new activity policy. If there is no response or update by **July 18, 2026**, it will be closed. It can be reopened at any time if still relevant.

maishivamhoo123:
> **Note:** As of today, this issue/PR has been flagged under our new activity policy. If there is no response or update by **July 18, 2026**, it will be closed. It can be reopened at any time if still relevant.


@mesutoezdil i am working on this issue.

mesutoezdil:
fixed by #1773, merged.

PR #1773, opened 2026-04-18
feat(scheduler): correct resource accounting for init containers

maishivamhoo123:
I fixed the resource accounting logic for pods with Init Containers: the scheduler now correctly calculates the effective request as $max(\sum \text{App Containers}, \text{max}(\text{Init Containers}))$. This prevents incorrect "node full" errors or quota denials for pods that reuse GPU resources during initialization.Technically, I modernized the codebase for Go 1.22+ by:Using the built-in max() function instead of manual if blocks.Implementing range over integers for cleaner loops.Removing redundant type declarations to satisfy staticcheck.Finally, I updated the unit tests in scheduler_test.go and score_test.go. Since the scoring logic now accurately reflects node utilization under the Binpack strategy, I adjusted the expected device UUIDs (e.g., device1 $\rightarrow$ device3) to match the scheduler's optimized selection. These changes ensure the HAMi scheduler is both logically sound and compliant with modern Go standards.
Fixes :- #1667 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved device scheduling, scoring, and allocation for pods with init containers, including safer usage accounting/rollback, correct slot handling, and clearer failure reporting.
  * Updated admission and resource quota logic to properly evaluate init-container resource requirements and reflect more accurate denial reasons.
  * Made generated reason output deterministic for consistent messaging.

* **Tests**
  * Expanded scheduler filtering/scoring and webhook quota tests for init-container scenarios (including insufficiency and tie-handling), with updated expected failure reasons.


maishivamhoo123:
@archlitchi , @DSFans2014  , @Shouren and @member can you please review this PR.

Shouren:
> @archlitchi , @DSFans2014 , @Shouren and @member can you please review this PR.

@maishivamhoo123 I am busy testing the recent updates for k8s-dra-driver, i will be back when the tests are finished.

maishivamhoo123:
@Shouren can you please check this [google docs](https://docs.google.com/document/d/1nX1FlJEiZ8lNKOAOuZHtiUs4OSwj5VJxVE1cU0y1LEU/edit?usp=sharing).

maishivamhoo123:
Screenshot 2026-04-27 021917


> @Shouren can you please check this [https://docs.google.com/document/d/1nX1FlJEiZ8lNKOAOuZHtiUs4OSwj5VJxVE1cU0y1LEU/edit?usp=sharing](Google Docs).

Shouren:
> @Shouren can you please check this [google docs](https://docs.google.com/document/d/1nX1FlJEiZ8lNKOAOuZHtiUs4OSwj5VJxVE1cU0y1LEU/edit?usp=sharing).

@maishivamhoo123 Can you show me the Pod Annotations after a successful scheduling?

pkg/scheduler/score.go
```diff
@@ -136,10 +138,36 @@ func (s *Scheduler) calcScore(nodes *map[string]*NodeUsage, resourceReqs device.
...
+
+			numInitContainers := len(task.Spec.InitContainers)
+
+			var initialNodeBytes []byte
+			var appContainersNode *NodeUsage
+
+			if numInitContainers > 0 {
+				var err error
+				// Snapshot the node's initial state so init containers don't drain the pool
+				initialNodeBytes, err = json.Marshal(node)
```
Shouren: @maishivamhoo123 I think we can add an optimization here that if the resource request from an initContainer can be satisfied by a resource request from an app container. We can skip the fit and reuse the allocation. How do you think?

maishivamhoo123: @Shouren  Thank you for review yes you are right .Since init containers run sequentially and exit before app containers start, the node only ever needs to hold max(initRequest, appRequest) free at any moment. So if the init container's request is already smaller than or equal to the app containers' sum, the app allocation implicitly covers it — we can skip the clone entirely and reuse the same device assignment.

maishivamhoo123:
@Shouren can you please review this PR?

archlitchi:
hi, since i just got this PR merged:https://github.com/Project-HAMi/HAMi/pull/1818, does this one needs to adjust?

maishivamhoo123:
> hi, since i just got this PR merged:#1818, does this one needs to adjust?

Yes i need to change the JSON serialization logic to deep copy function i will change  and let you know . Thank you!

maishivamhoo123:
@archlitchi , @Shouren i made all the required changes can you please review this PR?

pkg/scheduler/score.go
```diff
@@ -136,25 +184,60 @@ func (s *Scheduler) calcScore(nodes *map[string]*NodeUsage, resourceReqs device.
...
 
-				// container need no device and we have got certain deviceType
+				// implicitly satisfied by the app container allocation since
+				// init and app containers never run simultaneously — the node
+				// only needs max(init, app) capacity at any moment.
+				if ctrid < numInitContainers && !needsInitClone {
+					if deviceType != "" {
+						score.Devices[deviceType] = append(score.Devices[deviceType], device.ContainerDevices{})
+					}
+					continue
```
Shouren: @maishivamhoo123 So we skip the allocation of devices for init container here if it can be satisfied by app container, but how can we know the extra allocation for init container, i didn't see the codes that setting the devices from app container. Let me if i misunderstand it.

pkg/scheduler/score.go
```diff
@@ -102,6 +108,40 @@ func fitInDevices(node *NodeUsage, requests device.ContainerDeviceRequests, pod
...
+}
+
+// podAppContainerTotalRequest returns the sum of device requests across all
+// regular (non-init) containers, considering both GPU count AND memory.
+func podAppContainerTotalRequest(resourceReqs device.PodDeviceRequests, numInitContainers int) containerResourceSummary {
+	total := containerResourceSummary{}
+	for i := numInitContainers; i < len(resourceReqs); i++ {
+		for _, k := range resourceReqs[i] {
+			total.nums += int(k.Nums)
+			total.memreq += k.Memreq
```
Shouren: @maishivamhoo123 Consider the following scenario:
- One initContainer requests 1 GPU with 40GiB of memory.
- Four appContainers, each requesting 1 GPU with 20GiB of memory, totaling 80GiB.
Now suppose there is a node with 8 GPUs, but each GPU has less than 30GiB of memory available.

If we simply sum up the memory requirements, the total (40 + 80 = 120GiB) would appear to be schedulable on this node. However, since no single GPU has enough memory to satisfy the initContainer's 40GiB request, the actual allocation would fail.

What do you think?

pkg/scheduler/score.go
maishivamhoo123: Thank you @Shouren for your review ,please let me clear one thing  initContainers run SEQUENTIALLY, not in parallel with app containers. so init containers and app containers NEVER run at the same time , So node only needs to reserve capacity for the MAXIMUM of: The largest init container request, OR The sum of all app container requests
When needsInitClone = false (init request ≤ app request):
InitContainers will reuse the SAME devices already allocated for appContainers
We don't need separate device allocation for init containers.
The device slot is left empty because the same GPU will handle both.
Is this what you were asking about?

pkg/scheduler/score.go
maishivamhoo123: Oo What a excellent catch I completely missed this edge case.
you are right i am missing per-device constraints.
i want to fix this like this

 I'll add a canFitInitContainer() helper method to validate that at least ONE 
device on the node can individually satisfy the init container's requirements 
(memory + cores), then filter nodes early in calcScore() before allocation.
is it ok ?

maishivamhoo123: @Shouren i made the recommended changes. can you please Review this?

pkg/scheduler/score.go
Shouren: @maishivamhoo123 As mentioned above 
> When needsInitClone = false (init request ≤ app request):
InitContainers will reuse the SAME devices already allocated for appContainers

This is what i am concerned, we skip the allocation for init container here, but we have to update the `device.ContainerDevices{}` for init container later when allocation is done for app container. Otherwise the device plugin would not know which device to allocate for init container.

maishivamhoo123: Thank you @Shouren for pointing this out.  i made the required changes can you please Review it.
Thank you!

maishivamhoo123:
@Shouren can you please review this PR i made all the recommended changes?

Shouren:
> @Shouren can you please review this PR i made all the recommended changes?

@maishivamhoo123 I am busying working on ascend dra driver right now. I will come back next week.

maishivamhoo123:
@Shouren @archlitchi  @FouoF  and @team can you please review this PR ?

pkg/scheduler/score.go
```diff
@@ -164,6 +272,7 @@ func (s *Scheduler) calcScore(nodes *map[string]*NodeUsage, resourceReqs device.
 						score.Devices[idx] = append(emptyPodSingleDevice, score.Devices[idx]...)
 					}
```
mesutoezdil: The cores check here does nothing. `availableCores >= 0` is always true, so a device with zero free cores still passes this check.

`containerResourceSummary` only has `nums` and `memreq` fields. Core requirements are never tracked or checked in `canFitInitContainer`. A node where every device has 0 free cores will not be filtered out, even if the init container needs them.

Fix: add a `coresreq int32` field to `containerResourceSummary`, populate it in `podInitContainerMaxRequest`, and change this line to:
```go
if availableMem >= maxInitReq.memreq && availableCores >= maxInitReq.coresreq {
```

pkg/scheduler/score.go
```diff
@@ -136,25 +224,45 @@ func (s *Scheduler) calcScore(nodes *map[string]*NodeUsage, resourceReqs device.
...
 				}
+
 				klog.V(5).InfoS("fitInDevices", "pod", klog.KObj(task), "node", nodeID)
-				fit, reason := fitInDevices(node, n, task, nodeInfo, &score.Devices)
-				// found certain deviceType, fill missing empty allocation for containers before this
+
+				var workingNode *NodeUsage
+				if needsInitClone && ctrid < numInitContainers {
+					workingNode = node.DeepCopy()
+				} else {
```
mesutoezdil: When `node` is nil, returning `true` means the node passes the init container check. That is the wrong answer. If the node is nil there is no capacity at all. Return `false` here, or remove this branch if nil nodes are never passed to this function.

pkg/scheduler/score.go
```diff
@@ -177,6 +286,32 @@ func (s *Scheduler) calcScore(nodes *map[string]*NodeUsage, resourceReqs device.
...
+					// We need to fill empty init slots with app container allocations
+
+					if len(containerList) > numInitContainers {
+						firstAppAllocation := containerList[numInitContainers]
+
+						// Updated to modern Go range-over-int syntax
+						for initIdx := range numInitContainers {
+							if initIdx < len(containerList) {
+								containerList[initIdx] = firstAppAllocation
+								klog.V(4).InfoS(
```
mesutoezdil: This iterates a single-entry map just to get one string key. Write it directly:
```go
failureReason["InsufficientInitDeviceMemory"] = append(failureReason["InsufficientInitDeviceMemory"], nodeID)
```

mesutoezdil:
@maishivamhoo123 a few issues found after reading the diff.

1. score.go, `canFitInitContainer` cores check (line 273): `availableCores >= 0` is always true. `containerResourceSummary` has no `coresreq` field so core requirements are never checked. A node with zero free cores still passes. Why was this written as `>= 0` instead of comparing against the actual core request? Add `coresreq int32` to the struct, populate it in `podInitContainerMaxRequest`, and fix the check to `availableCores >= maxInitReq.coresreq`.

2. score.go, `canFitInitContainer` nil guard (line 260): when node is nil the function returns `true`, meaning the node passes the check. Why does a nil node mean "can fit"? It should return `false`.

3. score.go, map literal loop (line 302): one iteration over a single-entry map just to get one string key. Write it as a direct assignment.

4. scheduler_test.go, spread policy test now accepts device3 or device4: the original test expected exactly device3. What changed in the scoring that makes both valid now? A test that accepts multiple outputs does not catch regressions.

5. Pod annotation screenshot: @Shouren asked for this two months ago. Please share the pod annotations after a successful scheduling.

maishivamhoo123:
> @maishivamhoo123 a few issues found after reading the diff.
> 
> 1. score.go, `canFitInitContainer` cores check (line 273): `availableCores >= 0` is always true. `containerResourceSummary` has no `coresreq` field so core requirements are never checked. A node with zero free cores still passes. Why was this written as `>= 0` instead of comparing against the actual core request? Add `coresreq int32` to the struct, populate it in `podInitContainerMaxRequest`, and fix the check to `availableCores >= maxInitReq.coresreq`.
> 2. score.go, `canFitInitContainer` nil guard (line 260): when node is nil the function returns `true`, meaning the node passes the check. Why does a nil node mean "can fit"? It should return `false`.
> 3. score.go, map literal loop (line 302): one iteration over a single-entry map just to get one string key. Write it as a direct assignment.
> 4. scheduler_test.go, spread policy test now accepts device3 or device4: the original test expected exactly device3. What changed in the scoring that makes both valid now? A test that accepts multiple outputs does not catch regressions.
> 5. Pod annotation screenshot: @Shouren asked for this two months ago. Please share the pod annotations after a successful scheduling.

@mesutoezdil  Thank you for your review 
I will make all the Recommended changes , Currently i am testing this on my local environment .
I will update it and share all the information.
Thank you.

maishivamhoo123:
###  Pod with Init ≤ App Resources (`test-init-basic`)

**Purpose:** Verify basic scheduling where init requests ≤ app requests.

**YAML:**
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: test-init-basic
spec:
  initContainers:
  - name: init-small
    image: busybox
    command: ["sh", "-c", "echo init done && sleep 5"]
    env:
    - name: CUDA_DISABLE_CONTROL
      value: "true"
    resources:
      limits:
        nvidia.com/gpu: "1"
        nvidia.com/gpumem: "5000"
        nvidia.com/gpucores: "30"
  containers:
  - name: app-large
    image: busybox
    command: ["sleep", "3600"]
    env:
    - name: CUDA_DISABLE_CONTROL
      value: "true"
    resources:
      limits:
        nvidia.com/gpu: "1"
        nvidia.com/gpumem: "8000"
        nvidia.com/gpucores: "40"
```

**Observation:**

```bash
# Pod startup (init → app)
kubectl get pods -w
```
```
test-init-basic   0/1   Init:0/1          0   3s
test-init-basic   0/1   PodInitializing   0   9s
test-init-basic   1/1   Running           0   12s
```

```bash
# Execution order (init before app)
kubectl describe pod test-init-basic | tail -20
```
```
Normal  Started  50s  spec.initContainers{init-small}: Container started
Normal  Started  42s  spec.containers{app-large}: Container started
```

```bash
# Score
kubectl describe pod test-init-basic | grep FilteringSucceed
```
```
Normal  FilteringSucceed  53s  hami-scheduler  find fit node(nvml-mock-test-control-plane), 0 nodes not fit, 1 nodes fit(nvml-mock-test-control-plane:0.00)
```

```bash
# Device allocation
kubectl get pod test-init-basic -o jsonpath='{.metadata.annotations.hami\.io/vgpu-devices-allocated}'
```
```
GPU-12345678-1234-1234-1234-123456780003,NVIDIA,5000,30:;GPU-12345678-1234-1234-1234-123456780002,NVIDIA,8000,40:;
```

✅ Init receives `5000 MiB / 30 cores`; app receives `8000 MiB / 40 cores`.  
✅ Init container started **before** app container.  
✅ Score `0.00` – correct for an empty node.

---

### 3. Pod with Init > App Resources – Multi‑GPU (`test-init-larger`)

**Purpose:** Validate the `needsInitClone` path when init containers demand more GPUs than the app.

**YAML:**
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: test-init-larger
spec:
  initContainers:
  - name: init-heavy
    image: busybox
    command: ["sh", "-c", "echo init heavy && sleep 5"]
    env:
    - name: CUDA_DISABLE_CONTROL
      value: "true"
    resources:
      limits:
        nvidia.com/gpu: "2"
        nvidia.com/gpumem: "15000"
        nvidia.com/gpucores: "50"
  containers:
  - name: app-light
    image: busybox
    command: ["sleep", "3600"]
    env:
    - name: CUDA_DISABLE_CONTROL
      value: "true"
    resources:
      limits:
        nvidia.com/gpu: "1"
        nvidia.com/gpumem: "5000"
        nvidia.com/gpucores: "10"
```

**Observation:**

```bash
kubectl describe pod test-init-larger | grep FilteringSucceed
```
```
Normal  FilteringSucceed  44s (x2 over 46s)  hami-scheduler  find fit node(nvml-mock-test-control-plane), 0 nodes not fit, 1 nodes fit(nvml-mock-test-control-plane:1.52)
```

```bash
kubectl get pod test-init-larger -o jsonpath='{.metadata.annotations.hami\.io/vgpu-devices-allocated}'
```
```
GPU-12345678-1234-1234-1234-123456780001,NVIDIA,15000,50:GPU-12345678-1234-1234-1234-123456780000,NVIDIA,15000,50:;GPU-12345678-1234-1234-1234-123456780007,NVIDIA,5000,10:;
```

✅ Init receives **exactly two GPUs** (15000 MiB / 50 cores each).  
✅ App receives one GPU (5000 MiB / 10 cores).  
✅ Score `1.52` – higher due to increased node utilisation.  
✅ `needsInitClone` correctly isolates init and app reservations.

---

### 4. Mixed Containers – Empty Slices Are Non‑nil (`test-mixed`)

**Purpose:** Prove the normalisation fix replaces `nil` with empty `ContainerDevices{}`.

**YAML:**
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: test-mixed
spec:
  containers:
  - name: no-gpu
    image: busybox
    command: ["sleep", "3600"]
  - name: with-gpu
    image: busybox
    command: ["sleep", "3600"]
    env:
    - name: CUDA_DISABLE_CONTROL
      value: "true"
    resources:
      limits:
        nvidia.com/gpu: "1"
        nvidia.com/gpumem: "2000"
        nvidia.com/gpucores: "10"
```

**Observation:**

```bash
kubectl get pod test-mixed -o jsonpath='{.metadata.annotations.hami\.io/vgpu-devices-allocated}'
```
```
;GPU-12345678-1234-1234-1234-123456780005,NVIDIA,2000,10:;
```

✅ Leading `;` = empty `{}` slice for container `no-gpu`.  
✅ No `nil` values – the fix works.

---

### 5. Normal Pod Without Init Containers (`test-noinit`)

**Purpose:** Regression test to ensure normal pods remain unaffected.

**YAML:**
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: test-noinit
spec:
  containers:
  - name: sleep
    image: busybox
    command: ["sleep", "3600"]
    env:
    - name: CUDA_DISABLE_CONTROL
      value: "true"
    resources:
      limits:
        nvidia.com/gpu: "1"
```

**Observation:**

```bash
kubectl get pod test-noinit -o jsonpath='{.metadata.annotations.hami\.io/vgpu-devices-allocated}'
```
```
GPU-12345678-1234-1234-1234-123456780005,NVIDIA,40960,100:;
```

✅ Clean single‑device allocation; no extra padding.

---

### 6. Node Assignment

```bash
kubectl get pod -o wide | grep test-init
```
```
test-init-basic   1/1   Running   ...   nvml-mock-test-control-plane
test-init-larger  1/1   Running   ...   nvml-mock-test-control-plane
```

✅ All containers (init + app) always run on the same node.

---

### Summary

| Requirement                                         | Test case            | Result |
|-----------------------------------------------------|----------------------|--------|
| Effective request = max(init, app)                  | basic, larger        | ✅ |
| Init container runs before app container            | basic events         | ✅ |
| `needsInitClone` when init > app (multi‑GPU)        | larger               | ✅ |
| Non‑nil empty slices after normalisation            | mixed                | ✅ |
| No regression for normal pods                       | noinit               | ✅ |
| Unit tests pass                                     | `make test`          | ✅ |

All critical scheduling guarantees are met. **Ready for final review.**
```

**Comment by archlitchi** [comment]
CC @DSFans2014

**Comment by wawa0210** [comment]
----    
CC @DSFans2014

**Review by DSFans2014 — COMMENTED** [review]
@maishivamhoo123 It looks like the PR title doesn't correspond to the changes being made.

File: pkg/scheduler/scheduler.go [hunk 7]
```diff
@@ -738,12 +738,32 @@ ReleaseNodeLocks:
...
+	// 2. Calculate the SUM of requests among Regular Containers
+	for i := numInitContainers; i < len(resourceReqs); i++ {
+		for _, k := range resourceReqs[i] {
+			appReqTotal += int(k.Nums)
 		}
 	}
+
+	// 3. The effective total request is the MAX of (InitContainers, RegularContainers)
+	resourceReqTotal := appReqTotal
+	resourceReqTotal = max(resourceReqTotal, initReqTotal)
```
DSFans2014: `resourceReqTotal` is used to determine whether the pod requests resources managed by hami. Could we just use a boolean here instead of such complex calculation?

pkg/scheduler/score.go
```diff
@@ -102,6 +109,70 @@ func fitInDevices(node *NodeUsage, requests device.ContainerDeviceRequests, pod
...
+		device := deviceList.Device
+		if device == nil {
+			continue
+		}
+
+		availableMem := device.Totalmem - device.Usedmem
+		availableCores := device.Totalcore - device.Usedcores
+
+		// Check if this device individually satisfies init requirements
+		if availableMem >= maxInitReq.memreq && availableCores >= maxInitReq.coresreq {
```
DSFans2014: Since `DeviceLists` includes all device resources, do we need to verify that the device types match?

pkg/scheduler/score.go
```diff
@@ -102,6 +109,70 @@ func fitInDevices(node *NodeUsage, requests device.ContainerDeviceRequests, pod
...
+}
+
+// podAppContainerTotalRequest returns the sum of device requests across all
+// regular (non-init) containers, considering both GPU count AND memory.
+func podAppContainerTotalRequest(resourceReqs device.PodDeviceRequests, numInitContainers int) containerResourceSummary {
+	total := containerResourceSummary{}
+	for i := numInitContainers; i < len(resourceReqs); i++ {
+		for _, k := range resourceReqs[i] {
+			total.nums += int(k.Nums)
+			total.memreq += k.Memreq
```
DSFans2014: Why do we add resources from different devices together?

DSFans2014 (commented):
Should we need add special handling before `AddPod`? Otherwise, resources from both initContainers and containers will be counted.
https://github.com/Project-HAMi/HAMi/blob/430b458c75c37092b2ea48c8b17bd6d1cfcf45f4/pkg/scheduler/scheduler.go#L798-L805

pkg/scheduler/webhook.go
```diff
@@ -133,24 +152,47 @@ func fitResourceQuota(pod *corev1.Pod) bool {
...
+		}
+
+		// 2. Calculate the SUM of requests among Regular Containers
 		for _, ctr := range pod.Spec.Containers {
-			req, ok := getRequest(&ctr, resourceName)
+			_, ok := getRequest(&ctr, resourceName)
 			if ok {
 				if memReq, ok := getRequest(&ctr, memResourceName); ok {
-					memoryReq += memReq * req
+					appMemoryReq += memReq
```
DSFans2014: why delete the req

pkg/scheduler/score.go
```diff
@@ -42,66 +49,227 @@ func viewStatus(usage NodeUsage) {
...
+			current.memreq += req.Memreq
+			current.coresreq += req.Coresreq
+			totals[devType] = current
+		}
+	}
+	return totals
+}
+
+// stripInitContainerAliasSlots removes the device allocations of init containers from the
+// PodDevices structure, keeping only regular (non-init) container allocations.
```
DSFans2014: Here we exclude init containers when calculating resources. If an init container requests more resources than the regular containers, it might cause the quota to be exceeded.

pkg/scheduler/webhook.go
maishivamhoo123: @DSFans2014 Fixed – the regular container loop now multiplies memory/core requests by the GPU count (req). The init container loop also uses max(..., memReq * req). The test TestFitResourceQuota confirms correct accounting for multi‑GPU pods.

pkg/scheduler/score.go
maishivamhoo123: @DSFans2014 The stripping is safe because the admission quota check already uses max(initReq, appReq), so the worst‑case resource demand is covered. Stripping only happens after all init containers have terminated, so ongoing usage correctly reflects only the running regular containers. The scheduler also validates both stages independently, preventing any oversubscription.

mesutoezdil: @maishivamhoo123 reminder: pls adhere to the new rules regarding PRs, descriptions, responses, and codes.

https://github.com/Project-HAMi/HAMi/issues/1998#issuecomment-4888877868

DSFans2014: Currently, `AddUsage` and `AddPod` only accounts for resource usage from regular containers. Consider a scenario where the quota is set to 30G. An init container requests 20G, and the regular containers request 10G, but we only count the 10G. When the init container is running, if another pod requests 20G, would that exceed the quota?

```
usageDevices := stripInitContainerAliasSlots(args.Pod, resourceReqs, m.Devices)

added := s.podManager.AddPod(args.Pod, m.NodeID, usageDevices)
if added {
	s.quotaManager.AddUsage(args.Pod, usageDevices)
}
```

DSFans2014: > Stripping only happens after all init containers have terminated

How do we ensure this works?

maishivamhoo123: Thank you @DSFans2014 for raising this issue i am really sorry. I didn't get this thought , 
i generated this issue on my  fake gpu environment ya it is there i will see it and let you know ?

```
apiVersion: v1
kind: ResourceQuota
metadata:
  name: test-quota
  namespace: default
spec:
  hard:
    nvidia.com/gpumem: "30000"
```
`resourcequota/test-quota created`

```
apiVersion: v1
kind: Pod
metadata:
  name: test-init-quota-pod1
spec:
  initContainers:
  - name: init-heavy
    image: busybox
    command: ["sleep", "120"]   # keep init running long enough
    env:
    - name: CUDA_DISABLE_CONTROL
      value: "true"
    resources:
      limits:
        nvidia.com/gpu: 1
        nvidia.com/gpumem: "20000"
  containers:
  - name: app
    image: busybox
    command: ["sleep", "3600"]
    env:
    - name: CUDA_DISABLE_CONTROL
      value: "true"
    resources:
      limits:
        nvidia.com/gpu: 1
        nvidia.com/gpumem: "10000"
  restartPolicy: Never

```
`pod/test-init-quota-pod1 created`

```
apiVersion: v1
kind: Pod
metadata:
  name: test-init-quota-pod2
spec:
  initContainers:
  - name: init-heavy
    image: busybox
    command: ["sleep", "120"]
    env:
    - name: CUDA_DISABLE_CONTROL
      value: "true"
    resources:
      limits:
        nvidia.com/gpu: 1
        nvidia.com/gpumem: "20000"
  containers:
  - name: app
    image: busybox
    command: ["sleep", "3600"]
    env:
    - name: CUDA_DISABLE_CONTROL
      value: "true"
    resources:
      limits:
        nvidia.com/gpu: 1
        nvidia.com/gpumem: "10000"
  restartPolicy: Never
```

shivam_kumar@LAPTOP-NMKA92EB:~/HAMi$ kubectl get pods test-init-quota-pod1 test-init-quota-pod2
kubectl describe pod test-init-quota-pod2 | grep -A5 Events
NAME                   READY   STATUS     RESTARTS   AGE
test-init-quota-pod1   0/1     Init:0/1   0          18s
test-init-quota-pod2   0/1     Init:0/1   0          7s
Events:
  Type    Reason            Age   From            Message
  ----    ------            ----  ----            -------
  Normal  Scheduled         7s    hami-scheduler  Successfully assigned default/test-init-quota-pod2 to nvml-mock-test-control-plane
  Normal  FilteringSucceed  8s    hami-scheduler  find fit node(nvml-mock-test-control-plane), 0 nodes not fit, 1 nodes fit(nvml-mock-test-control-plane:1.17)
  Normal  BindingSucceed    8s    hami-scheduler  Successfully binding node [nvml-mock-test-control-plane] to default/test-init-quota-pod2
shivam_kumar@LAPTOP-NMKA92EB:~/HAMi$


i will fix it and let you know?
Thank you!

saiyam1814 (commented):
Thanks for driving this — the max-vs-sum split for init containers is the right core idea, and it fixes a real over-reservation problem (#1942). One design gap and one PR-hygiene note, neither raised in the existing threads:

**1. Restartable init containers (sidecars) break the max-pool assumption.**

Since Kubernetes 1.28 (stable in 1.33), an init container with `restartPolicy: Always` is a sidecar: it keeps running for the pod's entire lifetime, concurrently with app containers. Kubernetes' own effective-request accounting therefore *sums* sidecars with app containers instead of max-pooling them with regular init containers.

This PR treats all entries of `pod.Spec.InitContainers` uniformly, which breaks in two places:

- `podInitContainerMaxRequest` max-pools a GPU-requesting sidecar with the other init containers. A sidecar holds its GPU *while the app containers run*, so its request must be added to the app-container sum. As written, a pod with a GPU sidecar under-reserves and can overcommit the node — the same class of bug this PR is fixing for the opposite direction.
- `stripInitContainerAliasSlots` is gated on every `InitContainerStatuses` entry being `Terminated` with exit code 0. A sidecar never terminates, so for any pod containing one, the strip never fires — the max-pool reservation for the *other* (terminated) init containers is held for the pod's lifetime.

Suggestion: given sidecar + GPU is likely rare today, the minimal correct v1 is to detect `restartPolicy: Always` on GPU-requesting init containers and treat them as app containers in both the accounting (`podAppContainerTotalRequest`) and the strip logic (never strip their slots, and exclude them from the all-terminated gate). Alternatively, explicitly reject GPU requests in restartable init containers at admission with a clear error and a TODO — silent mis-accounting is the only option that shouldn't ship.

Either way, a test case with `restartPolicy: Always` would lock the chosen semantics in.

**2. Consider splitting the unrelated cleanups out of this PR.**

The comment deletions in `NewScheduler`, the `klog.Warning` → `klog.ErrorS` change, and the removal of `recordScheduleFilterResultEvent` for pods with no HAMi resources are all reasonable on their own, but they're invisible inside a 1.7k-line feature diff — and the last one is a user-visible behavior change (pods no longer get a `FilteringFailed` event) that deserves its own review. A small separate PR would merge quickly and keep this one focused on the init-container semantics.

mesutoezdil:
> Thanks for driving this — the max-vs-sum split for init containers is the right core idea, and it fixes a real over-reservation problem (#1942). One design gap and one PR-hygiene note, neither raised in the existing threads:
> 
> **1. Restartable init containers (sidecars) break the max-pool assumption.**
> 
> Since Kubernetes 1.28 (stable in 1.33), an init container with `restartPolicy: Always` is a sidecar: it keeps running for the pod's entire lifetime, concurrently with app containers. Kubernetes' own effective-request accounting therefore _sums_ sidecars with app containers instead of max-pooling them with regular init containers.
> 
> This PR treats all entries of `pod.Spec.InitContainers` uniformly, which breaks in two places:
> 
> * `podInitContainerMaxRequest` max-pools a GPU-requesting sidecar with the other init containers. A sidecar holds its GPU _while the app containers run_, so its request must be added to the app-container sum. As written, a pod with a GPU sidecar under-reserves and can overcommit the node — the same class of bug this PR is fixing for the opposite direction.
> * `stripInitContainerAliasSlots` is gated on every `InitContainerStatuses` entry being `Terminated` with exit code 0. A sidecar never terminates, so for any pod containing one, the strip never fires — the max-pool reservation for the _other_ (terminated) init containers is held for the pod's lifetime.
> 
> Suggestion: given sidecar + GPU is likely rare today, the minimal correct v1 is to detect `restartPolicy: Always` on GPU-requesting init containers and treat them as app containers in both the accounting (`podAppContainerTotalRequest`) and the strip logic (never strip their slots, and exclude them from the all-terminated gate). Alternatively, explicitly reject GPU requests in restartable init containers at admission with a clear error and a TODO — silent mis-accounting is the only option that shouldn't ship.
> 
> Either way, a test case with `restartPolicy: Always` would lock the chosen semantics in.
> 
> **2. Consider splitting the unrelated cleanups out of this PR.**
> 
> The comment deletions in `NewScheduler`, the `klog.Warning` → `klog.ErrorS` change, and the removal of `recordScheduleFilterResultEvent` for pods with no HAMi resources are all reasonable on their own, but they're invisible inside a 1.7k-line feature diff — and the last one is a user-visible behavior change (pods no longer get a `FilteringFailed` event) that deserves its own review. A small separate PR would merge quickly and keep this one focused on the init-container semantics.

@saiyam1814 reminder: pls adhere to the new rules regarding responses (and codes, prs, descriptions).

https://github.com/Project-HAMi/HAMi/issues/1998#issuecomment-4888877868

maishivamhoo123:
> Thanks for driving this — the max-vs-sum split for init containers is the right core idea, and it fixes a real over-reservation problem (#1942). One design gap and one PR-hygiene note, neither raised in the existing threads:
> 
> **1. Restartable init containers (sidecars) break the max-pool assumption.**
> 
> Since Kubernetes 1.28 (stable in 1.33), an init container with `restartPolicy: Always` is a sidecar: it keeps running for the pod's entire lifetime, concurrently with app containers. Kubernetes' own effective-request accounting therefore _sums_ sidecars with app containers instead of max-pooling them with regular init containers.
> 
> This PR treats all entries of `pod.Spec.InitContainers` uniformly, which breaks in two places:
> 
> * `podInitContainerMaxRequest` max-pools a GPU-requesting sidecar with the other init containers. A sidecar holds its GPU _while the app containers run_, so its request must be added to the app-container sum. As written, a pod with a GPU sidecar under-reserves and can overcommit the node — the same class of bug this PR is fixing for the opposite direction.
> * `stripInitContainerAliasSlots` is gated on every `InitContainerStatuses` entry being `Terminated` with exit code 0. A sidecar never terminates, so for any pod containing one, the strip never fires — the max-pool reservation for the _other_ (terminated) init containers is held for the pod's lifetime.
> 
> Suggestion: given sidecar + GPU is likely rare today, the minimal correct v1 is to detect `restartPolicy: Always` on GPU-requesting init containers and treat them as app containers in both the accounting (`podAppContainerTotalRequest`) and the strip logic (never strip their slots, and exclude them from the all-terminated gate). Alternatively, explicitly reject GPU requests in restartable init containers at admission with a clear error and a TODO — silent mis-accounting is the only option that shouldn't ship.
> 
> Either way, a test case with `restartPolicy: Always` would lock the chosen semantics in.
> 
> **2. Consider splitting the unrelated cleanups out of this PR.**
> 
> The comment deletions in `NewScheduler`, the `klog.Warning` → `klog.ErrorS` change, and the removal of `recordScheduleFilterResultEvent` for pods with no HAMi resources are all reasonable on their own, but they're invisible inside a 1.7k-line feature diff — and the last one is a user-visible behavior change (pods no longer get a `FilteringFailed` event) that deserves its own review. A small separate PR would merge quickly and keep this one focused on the init-container semantics.

@saiyam1814 Thanks for the review! I have one question regarding the scope. As far as I know, HAMi currently does not support sidecars containers . This PR is intended to add support for regular init containers only. Since supporting  sidecars containers requires different resource-accounting semantics,  i think we can add this as a different issue/feature HAMi supporting sidecarcontainer?
Thank you!

pkg/scheduler/score.go
maishivamhoo123: @DSFans2014  i fixed this and added a test TestFitResourceQuota_InitContainerPeakSequence  , can you please see it  once . thank you.

pkg/scheduler/score.go
```diff
@@ -42,66 +49,224 @@ func viewStatus(usage NodeUsage) {
...
+func stripInitContainerAliasSlots(pod *corev1.Pod, resourceReqs device.PodDeviceRequests, devices device.PodDevices) device.PodDevices {
```
DSFans2014: This function is never called anywhere now

DSFans2014:
> Should we need add special handling before `AddPod`? Otherwise, resources from both initContainers and containers will be counted.
> 
> https://github.com/Project-HAMi/HAMi/blob/430b458c75c37092b2ea48c8b17bd6d1cfcf45f4/pkg/scheduler/scheduler.go#L798-L805

hi @maishivamhoo123, Is this issue still not resolved. I noticed that it was reverted in https://github.com/Project-HAMi/HAMi/pull/1773/changes/5c78cd7af607455024879680a64a803c5e142740

mesutoezdil:

maishivamhoo123:
> **Note:** As of today, this issue/PR has been flagged under our new activity policy. If there is no response or update by **July 18, 2026**, it will be closed. It can be reopened at any time if still relevant.

Haii @mesutoezdil I am currently working on this issue.

maishivamhoo123:
> > Should we need add special handling before `AddPod`? Otherwise, resources from both initContainers and containers will be counted.
> > https://github.com/Project-HAMi/HAMi/blob/430b458c75c37092b2ea48c8b17bd6d1cfcf45f4/pkg/scheduler/scheduler.go#L798-L805
> 
> hi @maishivamhoo123, Is this issue still not resolved. I noticed that it was reverted in [5c78cd7](https://github.com/Project-HAMi/HAMi/commit/5c78cd7af607455024879680a64a803c5e142740)

Hi @DSFans2014, thank you for following up , you're right, this is still not resolved.
The probelm in early version was :-  the version I had before stripInitContainerAliasSlots tried to remove init-container usage as soon as the pod was scheduled, before the init container had even run. So it cause a problem which made the tracker think less GPU was in use than actually was, which could let two pods over-subscribe the same GPU at the same time. That's a more serious problem than double-counting, so I reverted it in  https://github.com/Project-HAMi/HAMi/commit/5c78cd7af607455024879680a64a803c5e142740  .But you're right that reverting it just brings back the original double-counting problem.
The issue:- m.Devices holds one entry per container  init containers and app containers both  because that's what PatchAnnotations needs (so the device plugin knows which container gets which GPU). But AddPod/AddUsage also gets handed that same full list, and just adds everything up with no idea that init containers and app containers never actually run at the same time. So a pod's usage is counted as init + app, when in reality it should be max(app total, biggest init container).
I have two ideas for how to fix this properly, and I'd like your input on which direction you'd prefer before I implement it:
Option 1 :- Reserve the peak number once, at scheduling time, and never touch it again.
Before calling AddPod/AddUsage, convert m.Devices into a smaller usage-only view: for each GPU, take max(app containers' total, biggest init container) instead of adding them together. Store that instead of the raw per-container list. Annotations stay exactly as they are — only what we track for accounting changes.
Option 2 :-  Same as Option 1, but shrink the reservation once the init container actually finishes.
Watch for InitContainerStatuses turning Terminated, and at that point recompute usage as "app containers only," freeing up the difference.
 Which of these would you prefer, or is there a different approach you had in mind?
cc : @rootsongjc , @archlitchi @Shouren

DSFans2014:
> > > Should we need add special handling before `AddPod`? Otherwise, resources from both initContainers and containers will be counted.
> > > https://github.com/Project-HAMi/HAMi/blob/430b458c75c37092b2ea48c8b17bd6d1cfcf45f4/pkg/scheduler/scheduler.go#L798-L805
> > 
> > 
> > hi @maishivamhoo123, Is this issue still not resolved. I noticed that it was reverted in [5c78cd7](https://github.com/Project-HAMi/HAMi/commit/5c78cd7af607455024879680a64a803c5e142740)
> 
> Hi @DSFans2014, thank you for following up , you're right, this is still not resolved. The probelm in early version was :- the version I had before stripInitContainerAliasSlots tried to remove init-container usage as soon as the pod was scheduled, before the init container had even run. So it cause a problem which made the tracker think less GPU was in use than actually was, which could let two pods over-subscribe the same GPU at the same time. That's a more serious problem than double-counting, so I reverted it in [5c78cd7](https://github.com/Project-HAMi/HAMi/commit/5c78cd7af607455024879680a64a803c5e142740) .But you're right that reverting it just brings back the original double-counting problem. The issue:- m.Devices holds one entry per container init containers and app containers both because that's what PatchAnnotations needs (so the device plugin knows which container gets which GPU). But AddPod/AddUsage also gets handed that same full list, and just adds everything up with no idea that init containers and app containers never actually run at the same time. So a pod's usage is counted as init + app, when in reality it should be max(app total, biggest init container). I have two ideas for how to fix this properly, and I'd like your input on which direction you'd prefer before I implement it: Option 1 :- Reserve the peak number once, at scheduling time, and never touch it again. Before calling AddPod/AddUsage, convert m.Devices into a smaller usage-only view: for each GPU, take max(app containers' total, biggest init container) instead of adding them together. Store that instead of the raw per-container list. Annotations stay exactly as they are — only what we track for accounting changes. Option 2 :- Same as Option 1, but shrink the reservation once the init container actually finishes. Watch for InitContainerStatuses turning Terminated, and at that point recompute usage as "app containers only," freeing up the difference. Which of these would you prefer, or is there a different approach you had in mind? cc : @rootsongjc , @archlitchi @Shouren

I prefer Option 2. In most cases, initContainers do not execute for an extended period, so it would be optimal to free up the resources as soon as they terminate.  What do you think about this? @Shouren @archlitchi

archlitchi:
> > > > Should we need add special handling before `AddPod`? Otherwise, resources from both initContainers and containers will be counted.
> > > > https://github.com/Project-HAMi/HAMi/blob/430b458c75c37092b2ea48c8b17bd6d1cfcf45f4/pkg/scheduler/scheduler.go#L798-L805
> > > 
> > > 
> > > hi @maishivamhoo123, Is this issue still not resolved. I noticed that it was reverted in [5c78cd7](https://github.com/Project-HAMi/HAMi/commit/5c78cd7af607455024879680a64a803c5e142740)
> > 
> > 
> > Hi @DSFans2014, thank you for following up , you're right, this is still not resolved. The probelm in early version was :- the version I had before stripInitContainerAliasSlots tried to remove init-container usage as soon as the pod was scheduled, before the init container had even run. So it cause a problem which made the tracker think less GPU was in use than actually was, which could let two pods over-subscribe the same GPU at the same time. That's a more serious problem than double-counting, so I reverted it in [5c78cd7](https://github.com/Project-HAMi/HAMi/commit/5c78cd7af607455024879680a64a803c5e142740) .But you're right that reverting it just brings back the original double-counting problem. The issue:- m.Devices holds one entry per container init containers and app containers both because that's what PatchAnnotations needs (so the device plugin knows which container gets which GPU). But AddPod/AddUsage also gets handed that same full list, and just adds everything up with no idea that init containers and app containers never actually run at the same time. So a pod's usage is counted as init + app, when in reality it should be max(app total, biggest init container). I have two ideas for how to fix this properly, and I'd like your input on which direction you'd prefer before I implement it: Option 1 :- Reserve the peak number once, at scheduling time, and never touch it again. Before calling AddPod/AddUsage, convert m.Devices into a smaller usage-only view: for each GPU, take max(app containers' total, biggest init container) instead of adding them together. Store that instead of the raw per-container list. Annotations stay exactly as they are — only what we track for accounting changes. Option 2 :- Same as Option 1, but shrink the reservation once the init container actually finishes. Watch for InitContainerStatuses turning Terminated, and at that point recompute usage as "app containers only," freeing up the difference. Which of these would you prefer, or is there a different approach you had in mind? cc : @rootsongjc , @archlitchi @Shouren
> 
> I prefer Option 2. In most cases, initContainers do not execute for an extended period, so it would be optimal to free up the resources as soon as they terminate. What do you think about this? @Shouren @archlitchi

yeah, Option 1 doesn't solve the problem entirely, but such a major adjustment in resource counting needs a design document, @maishivamhoo123 can you add a design document to the docs/develop folder describing the change to the workflow or architecture and let us review before you actually get started on coding?

maishivamhoo123:
> > > > > Should we need add special handling before `AddPod`? Otherwise, resources from both initContainers and containers will be counted.
> > > > > https://github.com/Project-HAMi/HAMi/blob/430b458c75c37092b2ea48c8b17bd6d1cfcf45f4/pkg/scheduler/scheduler.go#L798-L805
> > > > 
> > > > 
> > > > hi @maishivamhoo123, Is this issue still not resolved. I noticed that it was reverted in [5c78cd7](https://github.com/Project-HAMi/HAMi/commit/5c78cd7af607455024879680a64a803c5e142740)
> > > 
> > > 
> > > Hi @DSFans2014, thank you for following up , you're right, this is still not resolved. The probelm in early version was :- the version I had before stripInitContainerAliasSlots tried to remove init-container usage as soon as the pod was scheduled, before the init container had even run. So it cause a problem which made the tracker think less GPU was in use than actually was, which could let two pods over-subscribe the same GPU at the same time. That's a more serious problem than double-counting, so I reverted it in [5c78cd7](https://github.com/Project-HAMi/HAMi/commit/5c78cd7af607455024879680a64a803c5e142740) .But you're right that reverting it just brings back the original double-counting problem. The issue:- m.Devices holds one entry per container init containers and app containers both because that's what PatchAnnotations needs (so the device plugin knows which container gets which GPU). But AddPod/AddUsage also gets handed that same full list, and just adds everything up with no idea that init containers and app containers never actually run at the same time. So a pod's usage is counted as init + app, when in reality it should be max(app total, biggest init container). I have two ideas for how to fix this properly, and I'd like your input on which direction you'd prefer before I implement it: Option 1 :- Reserve the peak number once, at scheduling time, and never touch it again. Before calling AddPod/AddUsage, convert m.Devices into a smaller usage-only view: for each GPU, take max(app containers' total, biggest init container) instead of adding them together. Store that instead of the raw per-container list. Annotations stay exactly as they are — only what we track for accounting changes. Option 2 :- Same as Option 1, but shrink the reservation once the init container actually finishes. Watch for InitContainerStatuses turning Terminated, and at that point recompute usage as "app containers only," freeing up the difference. Which of these would you prefer, or is there a different approach you had in mind? cc : @rootsongjc , @archlitchi @Shouren
> > 
> > I prefer Option 2. In most cases, initContainers do not execute for an extended period, so it would be optimal to free up the resources as soon as they terminate. What do you think about this? @Shouren @archlitchi
> 
> yeah, Option 1 doesn't solve the problem entirely, but such a major adjustment in resource counting needs a design document, @maishivamhoo123 can you add a design document to the docs/develop folder describing the change to the workflow or architecture and let us review before you actually get started on coding?

Yeah I will do it and let you know

maishivamhoo123:
@DSFans2014 i revised the pr can you please review it once .

pkg/scheduler/scheduler.go
```diff
@@ -157,18 +167,72 @@ func (s *Scheduler) onAddPod(obj any) {
...
+			klog.ErrorS(err, "failed to decode pod devices during shrink", "pod", klog.KObj(newPod))
+			return
+		}
+
+		appOnlyDevices := device.AppContainersOnlyDeviceUsage(newPod, rawDevices)
+
+		oldDevices, ok := s.podManager.ShrinkUsage(newPod, appOnlyDevices)
+		if ok {
+			s.quotaManager.RmUsage(newPod, oldDevices)
+			s.quotaManager.AddUsage(newPod, appOnlyDevices)
```
DSFans2014: `RmUsage` and `AddUsage` must be atomic operations to ensure consistency when updating device usage.

pkg/scheduler/scheduler.go
```diff
@@ -135,6 +133,18 @@ func (s *Scheduler) doNodeNotify() {
 	}
 }
 
+func allInitContainersSucceeded(pod *corev1.Pod) bool {
```
DSFans2014: move the helper function to pkg/util/util.go

pkg/scheduler/scheduler.go
```diff
@@ -157,18 +167,72 @@ func (s *Scheduler) onAddPod(obj any) {
...
+	if !pi.InitContainerResourceReleased && allInitContainersSucceeded(newPod) {
+		rawDevices, err := device.DecodePodDevices(device.SupportDevices, newPod.Annotations)
+		if err != nil {
+			klog.ErrorS(err, "failed to decode pod devices during shrink", "pod", klog.KObj(newPod))
+			return
+		}
+
+		appOnlyDevices := device.AppContainersOnlyDeviceUsage(newPod, rawDevices)
+
+		oldDevices, ok := s.podManager.ShrinkUsage(newPod, appOnlyDevices)
```
DSFans2014: Usage is out of `PodManager`'s scope. I think it's better to rename the function to        `UpdatePodDevice` or others and set `InitContainerResourceReleased` in `ok` block

pkg/scheduler/webhook.go
```diff
@@ -71,6 +71,27 @@ func (h *webhook) Handle(_ context.Context, req admission.Request) admission.Res
 	klog.V(5).Infof(template, pod.Namespace, pod.Name, pod.UID)
 	privilegedName, hasPrivileged := privilegedContainerName(pod)
 	hasResource := false
+
+	// 1. Process InitContainers
+	for idx, ctr := range pod.Spec.InitContainers {
+		c := &pod.Spec.InitContainers[idx]
+		if ctr.SecurityContext != nil {
+			if ctr.SecurityContext.Privileged != nil && *ctr.SecurityContext.Privileged {
```
DSFans2014: The privileged check has been modified in #2139. Please update the related code

pkg/scheduler/scheduler.go
maishivamhoo123: > `RmUsage` and `AddUsage` must be atomic operations to ensure consistency when updating device usage.


@DSFans2014 I added ReplaceUsage(old, new) to QuotaManager that removes old usage and adds new usage under a single mutex lock. The shrink path now calls this one method instead of separate RmUsage and AddUsage.

maishivamhoo123:
@DSFans2014 i resolved all the suggested changes like added  ReplaceUsage for atomic quota swap, moved AllInitContainersSucceeded to util, renamed ShrinkUsage to UpdatePodDevice with flag inside, and used the privilege check to isPrivilegedContainer.
can you please look it once?
Thank you!

pkg/scheduler/webhook.go
```diff
@@ -71,6 +71,25 @@ func (h *webhook) Handle(_ context.Context, req admission.Request) admission.Res
 	klog.V(5).Infof(template, pod.Namespace, pod.Name, pod.UID)
 	privilegedName, hasPrivileged := privilegedContainerName(pod)
 	hasResource := false
+
+	// 1. Process InitContainers
+	for idx := range pod.Spec.InitContainers {
+		c := &pod.Spec.InitContainers[idx]
+		if isPrivilegedContainer(c) {
```
DSFans2014: privileged has already been checked in `privilegedContainerName`, so it does not need to be checked here again.

pkg/scheduler/score.go
```diff
@@ -42,61 +42,158 @@ func viewStatus(usage NodeUsage) {
 func getNodeResources(list NodeUsage, t string) []*device.DeviceUsage {
 	l := []*device.DeviceUsage{}
 	for _, val := range list.Devices.DeviceLists {
-		if strings.Contains(val.Device.Type, t) {
+		if val.Device == nil {
+			continue
+		}
+		if getDeviceBaseType(val.Device.Type) == t {
```
DSFans2014: why call `getDeviceBaseType` here

pkg/scheduler/score.go
```diff
@@ -42,61 +42,158 @@ func viewStatus(usage NodeUsage) {
...
+	return bestName
+}
+
+// nodeDeviceBaseTypes returns the set of base device types physically present on the node, computed once so that callers don't repeatedly re-derive it (which, combined with any residual
+// ambiguity in getDeviceBaseType, previously risked producing different results across calls within the same scheduling attempt).
+func nodeDeviceBaseTypes(list policy.DeviceUsageList) map[string]struct{} {
+	types := make(map[string]struct{})
+	for _, dl := range list.DeviceLists {
+		if dl.Device != nil {
+			types[getDeviceBaseType(dl.Device.Type)] = struct{}{}
```
DSFans2014: Why not just use `Device.Type`

pkg/scheduler/score.go
```diff
@@ -42,61 +42,158 @@ func viewStatus(usage NodeUsage) {
...
+		d := node.Devices.DeviceLists[i].Device
+		saved[i] = devSnapshot{
+			dev:       d,
+			used:      d.Used,
+			usedcores: d.Usedcores,
+			usedmem:   d.Usedmem,
+		}
+	}
+	rollbackAll := func() {
+		for _, s := range saved {
```
DSFans2014: `saved`, `rollbackAll` and `modified` seems unnecessary

pkg/scheduler/score.go
```diff
@@ -121,85 +218,228 @@ func (s *Scheduler) calcScoreWithOptions(nodes *map[string]*NodeUsage, resourceR
...
+							initAllocs[typ] = append(initAllocs[typ], device.ContainerDevices{})
+						}
+					}
+					if len(req) == 0 {
+						for typ := range baseTypes {
+							initAllocs[typ] = append(initAllocs[typ], device.ContainerDevices{})
+						}
+						continue
+					}
+					nodeCopy := node.DeepCopy()
```
DSFans2014: Should we consider calculating a score for the device?

pkg/scheduler/score.go
```diff
@@ -121,85 +218,228 @@ func (s *Scheduler) calcScoreWithOptions(nodes *map[string]*NodeUsage, resourceR
...
+					if !fit {
+						klog.V(4).InfoS("Init container does not fit",
+							"pod", klog.KObj(task), "node", nodeID, "containerIndex", i, "reason", reason)
+						failedNodesMutex.Lock()
+						failedNodes[nodeID] = reason
+						for reasonType := range common.ParseReason(reason) {
+							failureReason[reasonType] = append(failureReason[reasonType], nodeID)
+						}
+						failedNodesMutex.Unlock()
+						initFit = false
```
DSFans2014: call `return` here

pkg/scheduler/score.go
```diff
@@ -121,85 +218,228 @@ func (s *Scheduler) calcScoreWithOptions(nodes *map[string]*NodeUsage, resourceR
...
+				}
 			}
 
-			if ctrfit {
-				fitNodesMutex.Lock()
-				res.NodeList = append(res.NodeList, &score)
-				fitNodesMutex.Unlock()
-				score.OverrideScore(snapshot, userNodePolicy)
-				klog.V(4).InfoS(common.NodeFitPod, "pod", klog.KObj(task), "node", nodeID, "score", score.Score)
+			for _, dl := range node.Devices.DeviceLists {
```
DSFans2014: This function is too long and does too many things. I think it would be better to split it into several smaller, focused functions to improve readability and maintainability

pkg/scheduler/score.go
maishivamhoo123: @DSFans2014 I think we should not add because bin‑pack score already captures perdevice utilisation by comparing the snapshot of device states before and after placement.so additional perdevice score would just recalculate the same information .

pkg/scheduler/webhook.go
```diff
@@ -153,24 +166,49 @@ func fitResourceQuota(pod *corev1.Pod) bool {
...
-					coresReq += coreReq * req
+					appCoresReq += coreReq * req
 				}
 			}
 		}
+
+		memoryReq := max(appMemoryReq, initMemoryReq)
+		coresReq := max(appCoresReq, initCoresReq)
+		effectiveCountReq := max(appCountReq, initCountReq)
+		_ = effectiveCountReq
```
DSFans2014: `effectiveCountReq` seems to have no use

pkg/scheduler/score.go
```diff
@@ -42,64 +42,336 @@ func viewStatus(usage NodeUsage) {
 func getNodeResources(list NodeUsage, t string) []*device.DeviceUsage {
 	l := []*device.DeviceUsage{}
 	for _, val := range list.Devices.DeviceLists {
-		if strings.Contains(val.Device.Type, t) {
+		if val.Device == nil {
+			continue
+		}
+		if getDeviceBaseType(val.Device.Type) == t {
```
DSFans2014: What's the problem with the original logic?

pkg/scheduler/score.go
```diff
@@ -110,90 +382,41 @@ func (s *Scheduler) calcScoreWithOptions(nodes *map[string]*NodeUsage, resourceR
...
 	close(errCh)
 
-	// only pod scheduler failure will record failure event
-	if recordEvents && len(res.NodeList) == 0 {
-		for reasonType, failureNodes := range failureReason {
-			sort.Strings(failureNodes)
-			reason := fmt.Errorf("%d nodes %s(%s)", len(failureNodes), reasonType, strings.Join(failureNodes, ","))
-			s.recordScheduleFilterResultEvent(task, EventReasonFilteringFailed, "", reason)
-		}
+	if len(res.NodeList) == 0 {
```
DSFans2014: why delete `recordEvents` from the if condition

maishivamhoo123:
@DSFans2014 i made all the required changes can you please review it.

pkg/scheduler/score.go
```diff
@@ -42,64 +42,336 @@ func viewStatus(usage NodeUsage) {
...
+}
+
+func allocateInitContainers(node *NodeUsage, nodeID string, resourceReqs device.PodDeviceRequests, task *corev1.Pod, nodeInfo *device.NodeInfo, baseTypes map[string]struct{}, numInitContainers int, peakUsage map[string]peakUsageSnapshot) (device.PodDevices, bool, string) {
+	initAllocs := make(device.PodDevices)
+
+	for i, req := range resourceReqs {
+		if i >= numInitContainers {
+			break
+		}
+		for typ := range baseTypes {
```
DSFans2014: The empty ContainerDevice has already been added after `updatePeakUsage`, so this part is no longer needed?

pkg/scheduler/webhook.go
maishivamhoo123: Yaa sorry , i left it by mistake .

pkg/scheduler/score.go
maishivamhoo123: Both are same i am reverting it back to original one.

pkg/scheduler/score.go
maishivamhoo123: @DSFans2014  i guess we need it because updatePeakUsage only tracks the maximum resource values across init containers, so it doesn’t touch the container index alignment in initAllocs. The empty ContainerDevice placeholder is there for a different reason one is when a container requests some device types but not others, the unrequested types would miss their slot for that container index. Without that alignment, merging init and app allocations later may breaks. So even after updatePeakUsage, we still need to add the empty entry for every type to keep all slices the same length as per container position.

pkg/scheduler/score.go
maishivamhoo123: reverted .

maishivamhoo123:
@DSFans2014 i made all the required changes can you please review it once. thank you.

pkg/scheduler/score.go
```diff
@@ -49,57 +49,326 @@ func getNodeResources(list NodeUsage, t string) []*device.DeviceUsage {
 	return l
 }
 
+func getDeviceBaseType(model string) string {
```
DSFans2014: please remove the unused function

pkg/scheduler/score.go
DSFans2014: empty ContainerDevices are added at both Line 233 and Line 252. Since it's already added at Line 252, why do we still need to check and add it again at Line 233?

maishivamhoo123:
@DSFans2014 removed the getDeviceBaseType and cleaned the  allocateInitContainers and allocateAppContainers.

maishivamhoo123:
@DSFans2014  i made all the changes and tested the feature and made a report for this all the test are passing. can you please once review this .
 https://docs.google.com/document/d/1b74hiPPflSlXMFdnbkNpg7X5zJpv4e8aqTgWM0vNw48/edit?usp=sharing 
thank you!

DSFans2014:
please resolve conflicts

maishivamhoo123:
> please resolve conflicts

@DSFans2014 i resolved the merge conflict please look in this once. Thank you!

pkg/device/nvidia/device.go
```diff
@@ -728,27 +728,6 @@ func (dev *NvidiaGPUDevices) AddResourceUsage(pod *corev1.Pod, n *device.DeviceU
 	return nil
 }
 
-func fitQuota(tmpDevs map[string]device.ContainerDevices, allocated *device.PodDevices, ns string, memreq int64, coresreq int64) bool {
```
DSFans2014, line 731: It looks like there are many unrelated changes

maishivamhoo123: @DSFans2014 This isn't an unrelated change.

```go
if !fitQuota(tmpDevs, allocated, pod.Namespace, int64(memreq), int64(k.Coresreq)) {
    reason[common.ResourceQuotaNotFit]++
    klog.V(3).InfoS(common.ResourceQuotaNotFit, "pod", pod.Name, "memreq", memreq, "coresreq", k.Coresreq)
    continue
}
```

was causing the overcounting issue during the test because fitQuota was being checked again. After removing this duplicate check, the overcounting problem was resolved.the remaining fitQuota check became redundant since the quota validation is already handled in quota.go. So the removal wasn't an unrelated change. it as removing redundant logic that existed after fixing the overcounting issue.
i hope now it is clear if not please let me know?

DSFans2014: The webhook stage can only verify fixed allocations. It cannot validate proportional allocations like `nvidia.com/gpumem-percentage`.
if there is overcounting issue, we should resolve it rather than skip the validation

maishivamhoo123:
@DSFans2014 i made the changes and tested everything and added  in this docs  https://docs.google.com/document/d/1b74hiPPflSlXMFdnbkNpg7X5zJpv4e8aqTgWM0vNw48/edit?tab=t.0#heading=h.98d1o3w5xw19 can you please review.

pkg/scheduler/webhook.go
```diff
@@ -139,15 +154,30 @@ func fitResourceQuota(pod *corev1.Pod) bool {
...
+			req := dev.GenerateResourceRequests(&pod.Spec.InitContainers[i])
+			if req.Nums == 0 {
+				continue
+			}
+			initMemoryReq = max(initMemoryReq, int64(req.Memreq)*int64(req.Nums))
+			initCoresReq = max(initCoresReq, int64(req.Coresreq)*int64(req.Nums))
 		}
+
+		memoryReq := max(appMemoryReq, initMemoryReq)
+		coresReq := max(appCoresReq, initCoresReq)
```
DSFans2014, line 180: see the case as below. After deploying `test-init-basic1`, a quota error occurs when deploying `test-init-basic2`.

ScreenShot_2026-08-10_104958_393


```
apiVersion: v1
kind: ResourceQuota
metadata:
  name: gpu-quota
  namespace: default
spec:
  hard:
    requests.nvidia.com/gpumem: "10000"
```

```
apiVersion: v1
kind: Pod
metadata:
  name: test-init-basic1
spec:
  initContainers:
  - name: init-small
    image: busybox
    command: ["sh", "-c", "echo init done && sleep 5"]
    env:
    - name: CUDA_DISABLE_CONTROL
      value: "true"
    resources:
      limits:
        nvidia.com/gpu: "1"
        nvidia.com/gpumem: "8000"
        nvidia.com/gpucores: "30"
  containers:
  - name: app-large
    image:  busybox
    command: ["sleep", "3600"]
    env:
    - name: CUDA_DISABLE_CONTROL
      value: "true"
    resources:
      limits:
        nvidia.com/gpu: "1"
        nvidia.com/gpumem: "5000"
        nvidia.com/gpucores: "40"

```

```
apiVersion: v1
kind: Pod
metadata:
  name: test-init-basic2
spec:
  initContainers:
  - name: init-small
    image: busybox
    command: ["sh", "-c", "echo init done && sleep 5"]
    env:
    - name: CUDA_DISABLE_CONTROL
      value: "true"
    resources:
      limits:
        nvidia.com/gpu: "1"
        nvidia.com/gpumem: "2000"
        nvidia.com/gpucores: "30"
  containers:
  - name: app-large
    image:  busybox
    command: ["sleep", "3600"]
    env:
    - name: CUDA_DISABLE_CONTROL
      value: "true"
    resources:
      limits:
        nvidia.com/gpu: "1"
        nvidia.com/gpumem: "5000"
        nvidia.com/gpucores: "40"
```

maishivamhoo123: @DSFans2014 thank you for your review. I reproduced this edge case in my machine, but this does not appear to be caused by this PR or HAMi.
Screenshot 2026-08-10 095109
I think this rejection comes from Kubernetes built-in ResourceQuota admission controller. which calculates the effective resource request for init containers using max(sum(appContainers), max(initContainers).
https://kubernetes.io/docs/tasks/administer-cluster/manage-resources/quota-memory-cpu-namespace/
In this case, the first pod consumes `8k` because of its init container, so the second pod's `5k` request exceeds the `10k` quota. The pod is rejected by the kube-apiserver before HAMi or the scheduler sees it.
So, i think this is upstream Kubernetes ResourceQuota behavior, not an issue introduced by this PR.
is it right or i am missing something?

DSFans2014: It would be better to document the inconsistency in resource calculation between quota-enabled and non-quota scenarios. cc @archlitchi

DSFans2014 (commented):
/lgtm

archlitchi (approved):
/lgtm

pkg/scheduler/webhook.go
archlitchi: yeah, i think so

archlitchi (approved):
/lgtm

PR #1789, opened 2026-04-24
Disable host network for device plugin

luohua13:
## Summary
- Disable hostNetwork for the NVIDIA device plugin DaemonSet.

## Motivation
In security-isolated environments, the host port exposure range is strictly controlled. Running the device plugin pod with hostNetwork makes the monitor endpoint bind to the host network, so ServiceMonitor cannot collect metrics from the expected pod/service network path.

## Testing
- Not run. The change only flips the hard-coded hostNetwork value in the Helm template.

### DIFF
--- a/charts/hami/templates/device-plugin/daemonsetnvidia.yaml
+++ b/charts/hami/templates/device-plugin/daemonsetnvidia.yaml
@@ -41,7 +41,7 @@ spec:
       serviceAccountName: {{ include "hami-vgpu.device-plugin" . }}
       priorityClassName: system-node-critical
       hostPID: true
-      hostNetwork: true
+      hostNetwork: false
       {{- include "hami.devicePlugin.imagePullSecrets" . | nindent 6 }}
       {{- if .Values.devicePlugin.gpuOperatorToolkitReady.enabled }}
       initContainers:


archlitchi:
Could this change lead to 31992 unusable?

luohua13:
> Could this change lead to 31992 unusable?


Absolutely not. Hostnetwork will directly cause container ports such as 80 and 443 to become host ports. Port 31992 is a service nodeport that will be forwarded to the container port, so it has no impact whatsoever.

spencercjh:
**Regarding metrics endpoint accessibility after disabling hostNetwork:**

Before this change (`hostNetwork: true`), the vgpu-monitor metrics endpoint was directly accessible via `<node-ip>:9394` since the pod shared the host network namespace. After setting `hostNetwork: false`, this direct node IP access no longer works — the pod now binds `:9394` on its own network namespace (pod IP).

However, **this does not affect Prometheus scraping** because:

1. The chart already includes a **Service** (`hami-device-plugin-monitor`, NodePort 31992 → targetPort 9394) whose selector matches the DaemonSet pods.
2. The chart already includes a **ServiceMonitor** (`hami-device-plugin-metrics-monitor`) that references this Service.
3. PrometheusOperator uses the Service to discover all pod IPs via Endpoints and generates scrape targets pointing directly to `<pod-ip>:9394` — traffic goes pod-to-pod, never through the Service VIP or node IP.

So as long as Prometheus is configured via ServiceMonitor (the default setup in this chart), metrics collection continues to work without any changes. The only scenario that would break is a manual scrape config targeting `node-ip:9394` directly, which is not the recommended approach.

luohua13:
> **Regarding metrics endpoint accessibility after disabling hostNetwork:**
> 
> Before this change (`hostNetwork: true`), the vgpu-monitor metrics endpoint was directly accessible via `<node-ip>:9394` since the pod shared the host network namespace. After setting `hostNetwork: false`, this direct node IP access no longer works — the pod now binds `:9394` on its own network namespace (pod IP).
> 
> However, **this does not affect Prometheus scraping** because:
> 
> 1. The chart already includes a **Service** (`hami-device-plugin-monitor`, NodePort 31992 → targetPort 9394) whose selector matches the DaemonSet pods.
> 2. The chart already includes a **ServiceMonitor** (`hami-device-plugin-metrics-monitor`) that references this Service.
> 3. PrometheusOperator uses the Service to discover all pod IPs via Endpoints and generates scrape targets pointing directly to `<pod-ip>:9394` — traffic goes pod-to-pod, never through the Service VIP or node IP.
> 
> So as long as Prometheus is configured via ServiceMonitor (the default setup in this chart), metrics collection continues to work without any changes. The only scenario that would break is a manual scrape config targeting `node-ip:9394` directly, which is not the recommended approach.



> **Regarding metrics endpoint accessibility after disabling hostNetwork:**
> 
> Before this change (`hostNetwork: true`), the vgpu-monitor metrics endpoint was directly accessible via `<node-ip>:9394` since the pod shared the host network namespace. After setting `hostNetwork: false`, this direct node IP access no longer works — the pod now binds `:9394` on its own network namespace (pod IP).
> 
> However, **this does not affect Prometheus scraping** because:
> 
> 1. The chart already includes a **Service** (`hami-device-plugin-monitor`, NodePort 31992 → targetPort 9394) whose selector matches the DaemonSet pods.
> 2. The chart already includes a **ServiceMonitor** (`hami-device-plugin-metrics-monitor`) that references this Service.
> 3. PrometheusOperator uses the Service to discover all pod IPs via Endpoints and generates scrape targets pointing directly to `<pod-ip>:9394` — traffic goes pod-to-pod, never through the Service VIP or node IP.
> 
> So as long as Prometheus is configured via ServiceMonitor (the default setup in this chart), metrics collection continues to work without any changes. The only scenario that would break is a manual scrape config targeting `node-ip:9394` directly, which is not the recommended approach.

There's no need to use the `hostnetwork` mode because when it's set to `true`, the container port becomes the host port, which can easily conflict with service ports on other hosts. Furthermore, the range of open host ports in a security-hardened environment is limited, so this approach shouldn't be used. If you want to access via the host port, simply use `nodeport`.

archlitchi:
@luohua13 i think the gemini review make sense: 

Hardcoding hostNetwork: false limits the flexibility of the Helm chart. While disabling host networking is a good security practice and resolves the monitoring issues described, it is better to make this configurable via values.yaml. This allows users with specific networking requirements to override the default if necessary.
      hostNetwork: {{ .Values.devicePlugin.hostNetwork | default false }}

we should add a switch to this

luohua13:
> @luohua13 i think the gemini review make sense:
> 
> Hardcoding hostNetwork: false limits the flexibility of the Helm chart. While disabling host networking is a good security practice and resolves the monitoring issues described, it is better to make this configurable via values.yaml. This allows users with specific networking requirements to override the default if necessary. hostNetwork: {{ .Values.devicePlugin.hostNetwork | default false }}
> 
> we should add a switch to this

If you want to access it via the host port, you can use a NodePort type service, so there's really no need to use the hostnetwork mode, and it would only increase maintenance complexity.

archlitchi:
@luohua13 i know, however, this changes the default settings of device metrics, some users are getting metrics from host node directly. They will be affected by this change when upgrading. I think a better approach is to add this configuration to values.yaml

luohua13:
> @luohua13 i know, however, this changes the default settings of device metrics, some users are getting metrics from host node directly. They will be affected by this change when upgrading. I think a better approach is to add this configuration to values.yaml


The upgrade will not affect port 31992. As long as clients do not directly access port 9394 through the host, there will be no problem. Nobody will do that.

archlitchi (approved):
/lgtm

PR #1808, opened 2026-04-27
fix: fix gpu scheduler policy sort bug(#1806)

### LINKED ISSUE #1806: gpu scheduler policy sort bug
**What happened**:

When using the `spread` or `binpack` GPU scheduler policy, HAMi always prioritizes GPUs on a fixed NUMA node regardless of actual workload load. This causes suboptimal allocation — GPUs are forced onto a busy NUMA node even when another NUMA node has plenty of idle capacity.

### Root cause

The `Less()` function in `pkg/scheduler/policy/gpu_policy.go` uses NUMA as the **primary** sort key, not Score:

https://github.com/Project-HAMi/HAMi/blob/e394008558ce037f1b64ce2f3dad580671cc12d0/pkg/scheduler/policy/gpu_policy.go#L45-L57

For both policies, NUMA is compared **before** Score. Devices on different NUMA nodes are never interleaved by Score — the sort always groups all devices from one NUMA before the other.

### Simulated scenario

Consider a node with 4 NVIDIA A100 GPUs across 2 NUMA nodes:

```
NUMA 0                          NUMA 1
┌──────────────────┐            ┌──────────────────┐
│ gpu-0: Score 0.9 │            │ gpu-2: Score 0.3 │
│ gpu-1: Score 0.7 │            │ gpu-3: Score 0.1 │
└──────────────────┘            └──────────────────┘
   (busy)                         (idle)
```

A new Pod requests **1 GPU** with `hami.io/gpu-scheduler-policy: spread`.

**Step 1 — `Less()` sorts by NUMA first (lower NUMA first for spread):**

```
Sorted: [gpu-0(NUMA0, 0.9), gpu-1(NUMA0, 0.7), gpu-2(NUMA1, 0.3), gpu-3(NUMA1, 0.1)]
         ──────── NUMA 0 group ────────   ──────── NUMA 1 group ────────
```

**Step 2 — `Fit()` in `pkg/device/nvidia/device.go` iterates from the tail:**

```go
for i := len(devices) - 1; i >= 0; i-- {
    // ... picks gpu-3 (NUMA 1, Score 0.1)
}
```

**Result:** The Pod gets `gpu-3` (NUMA 1). This is correct — `gpu-3` is the least loaded device.

**Now consider a different workload state:**

```
NUMA 0                          NUMA 1
┌──────────────────┐            ┌──────────────────┐
│ gpu-0: Score 0.9 │            │ gpu-2: Score 0.5 │
│ gpu-1: Score 0.1 │            │ gpu-3: Score 0.5 │
└──────────────────┘            └──────────────────┘
  (1 busy, 1 idle)                (2 idle)
```

A new Pod requests **2 GPUs** with `spread` policy.

**Step 1 — `Less()` sorts by NUMA first:**

```
Sorted: [gpu-0(NUMA0, 0.9), gpu-1(NUMA0, 0.1), gpu-2(NUMA1, 0.5), gpu-3(NUMA1, 0.5)]
```

**Step 2 — `Fit()` iterates from the tail:**

```
picks gpu-3 (NUMA 1, Score 0.5)
picks gpu-2 (NUMA 1, Score 0.5)
```

**Result:** The Pod gets `gpu-2` + `gpu-3`, both from NUMA 1. Both GPUs have Score 0.5, while `gpu-1` on NUMA 0 has a lower Score of 0.1 and would have been a better choice for spread. The scheduler **ignored** the more idle device because NUMA grouping took priority.

For `spread` and `binpack` policies, Score should be the **primary** sort key to respect the policy's intent (spread prefers idle devices, binpack prefers busy devices). NUMA should only serve as a **secondary** tiebreaker when scores are equal.

With the same scenario:

```
NUMA 0                          NUMA 1
┌──────────────────┐            ┌──────────────────┐
│ gpu-0: Score 0.9 │            │ gpu-2: Score 0.5 │
│ gpu-1: Score 0.1 │            │ gpu-3: Score 0.5 │
└──────────────────┘            └──────────────────┘
```

Pod requests 2 GPUs with `spread` policy.

**Expected `Less()` result — Score as primary, NUMA as tiebreaker:**

```
Sorted: [gpu-0(NUMA0, 0.9), gpu-2(NUMA1, 0.5), gpu-3(NUMA1, 0.5), gpu-1(NUMA0, 0.1)]
         ─── Score desc ───      ─── same Score, grouped by NUMA ───
```

**Expected allocation:** `Fit()` picks `gpu-1(NUMA0, 0.1)` + `gpu-3(NUMA1, 0.5)` — picks the most idle device first, consistent with spread's intent.

This would change the `Less()` logic to:

```go
func (l DeviceUsageList) Less(i, j int) bool {
	if l.Policy == util.GPUSchedulerPolicyBinpack.String() {
		if l.DeviceLists[i].Score != l.DeviceLists[j].Score {
			return l.DeviceLists[i].Score < l.DeviceLists[j].Score
		}
		return l.DeviceLists[i].Device.Numa < l.DeviceLists[j].Device.Numa
	}
	// default policy is spread
	if l.DeviceLists[i].Score != l.DeviceLists[j].Score {
		return l.DeviceLists[i].Score > l.DeviceLists[j].Score
	}
	return l.DeviceLists[i].Device.Numa < l.DeviceLists[j].Device.Numa
}
```

This can be verified directly by examining the existing unit test `TestDeviceUsageList_Less` in `pkg/scheduler/policy/gpu_policy_test.go`.
https://github.com/Project-HAMi/HAMi/blob/e394008558ce037f1b64ce2f3dad580671cc12d0/pkg/scheduler/policy/gpu_policy_test.go#L121-L197

The test has 3 binpack cases with different NUMA:

```go
// Case 1: same NUMA
{Numa: 0, Used: 10, Score: 10} vs {Numa: 0, Used: 20, Score: 20}
expectedLess: true   // Score 10 < 20, correct

// Case 2: different NUMA, Numa=1 vs Numa=0
{Numa: 1, Used: 10, Score: 10} vs {Numa: 0, Used: 20, Score: 20}
expectedLess: true   // Numa 1 > 0, so true — Score is ignored!

// Case 3: different NUMA, Numa=0 vs Numa=1
{Numa: 0, Used: 10, Score: 10} vs {Numa: 1, Used: 20, Score: 20}
expectedLess: false  // Numa 0 < 1, so false — Score is ignored!
```

Cases 2 and 3 have **identical Used and Score values** — the only difference is the NUMA node. If `Less()` were comparing by Score first (as `binpack` should), both cases should produce the **same** `expectedLess` result (true, since Score 10 < 20). Instead, they produce opposite results because NUMA is the primary sort key.

The same problem exists for the 3 spread cases:

```go
// Case 1: same NUMA
{Numa: 0, Used: 10, Score: 10} vs {Numa: 0, Used: 20, Score: 20}
expectedLess: false  // Score 10 < 20, so false — correct

// Case 2: different NUMA, Numa=1 vs Numa=0
{Numa: 1, Used: 10, Score: 10} vs {Numa: 0, Used: 20, Score: 20}
expectedLess: false  // Numa 1 > 0, so false — Score is ignored!

// Case 3: different NUMA, Numa=0 vs Numa=1
{Numa: 0, Used: 10, Score: 10} vs {Numa: 1, Used: 20, Score: 20}
expectedLess: true   // Numa 0 < 1, so true — Score is ignored!
```

Again, cases 2 and 3 should have the same `expectedLess` for spread (false, since Score 10 < 20 means device i is NOT "less than" device j in descending order). The fact that they differ proves NUMA overrides Score.

**Environment**:
- HAMi version:
- Others:

kaiiyvwu:
**What type of PR is this?**
/kind bug

Update the gpu policy Less function. If the gpu scheduler policy is binpack or spread, it will not compare numa as sort key

Fixes #1806

### DIFF
--- a/pkg/scheduler/policy/gpu_policy.go
+++ b/pkg/scheduler/policy/gpu_policy.go
@@ -19,6 +19,8 @@ package policy
 import (
 	"github.com/Project-HAMi/HAMi/pkg/device"
 	"github.com/Project-HAMi/HAMi/pkg/util"
+
+	"k8s.io/klog/v2"
 )
 
 type DeviceListsScore struct {
--- a/pkg/scheduler/policy/gpu_policy.go
+++ b/pkg/scheduler/policy/gpu_policy.go
@@ -19,8 +19,6 @@ package policy
 import (
 	"github.com/Project-HAMi/HAMi/pkg/device"
 	"github.com/Project-HAMi/HAMi/pkg/util"
-
-	"k8s.io/klog/v2"
 )
 
 type DeviceListsScore struct {
@@ -43,17 +41,17 @@ func (l DeviceUsageList) Swap(i, j int) {
 }
 
 func (l DeviceUsageList) Less(i, j int) bool {
-	if l.Policy == util.GPUSchedulerPolicyBinpack.String() {
-		if l.DeviceLists[i].Device.Numa == l.DeviceLists[j].Device.Numa {
-			return l.DeviceLists[i].Score < l.DeviceLists[j].Score
-		}
-		return l.DeviceLists[i].Device.Numa > l.DeviceLists[j].Device.Numa
-	}
-	// default policy is spread
-	if l.DeviceLists[i].Device.Numa == l.DeviceLists[j].Device.Numa {
+	switch l.Policy {
+	case util.GPUSchedulerPolicyBinpack.String():
+		return l.DeviceLists[i].Score < l.DeviceLists[j].Score
+	case util.GPUSchedulerPolicySpread.String():
 		return l.DeviceLists[i].Score > l.DeviceLists[j].Score
+	default:
+		if l.DeviceLists[i].Device.Numa == l.DeviceLists[j].Device.Numa {
+			return l.DeviceLists[i].Score > l.DeviceLists[j].Score
+		}
+		return l.DeviceLists[i].Device.Numa < l.DeviceLists[j].Device.Numa
 	}
-	return l.DeviceLists[i].Device.Numa < l.DeviceLists[j].Device.Numa
 }
 
 func (ds *DeviceListsScore) ComputeScore(requests device.ContainerDeviceRequests) {
--- a/pkg/scheduler/policy/gpu_policy_test.go
+++ b/pkg/scheduler/policy/gpu_policy_test.go
@@ -150,7 +150,7 @@ func TestDeviceUsageList_Less(t *testing.T) {
 				{Device: &device.DeviceUsage{Numa: 0, Used: 10}, Score: 10},
 				{Device: &device.DeviceUsage{Numa: 1, Used: 20}, Score: 20},
 			},
-			expectedLess: false,
+			expectedLess: true,
 		},
 		{
 			name:   "Spread policy with same Numa",
@@ -177,7 +177,7 @@ func TestDeviceUsageList_Less(t *testing.T) {
 				{Device: &device.DeviceUsage{Numa: 0, Used: 10}, Score: 10},
 				{Device: &device.DeviceUsage{Numa: 1, Used: 20}, Score: 20},
 			},
-			expectedLess: true,
+			expectedLess: false,
 		},
 	}
 


pkg/scheduler/policy/gpu_policy.go
```diff
@@ -43,17 +43,17 @@ func (l DeviceUsageList) Swap(i, j int) {
 }
 
 func (l DeviceUsageList) Less(i, j int) bool {
-	if l.Policy == util.GPUSchedulerPolicyBinpack.String() {
+	switch l.Policy {
+	case util.GPUSchedulerPolicyBinpack.String():
+		return l.DeviceLists[i].Score < l.DeviceLists[j].Score
+	case util.GPUSchedulerPolicySpread.String():
+		return l.DeviceLists[i].Score > l.DeviceLists[j].Score
+	default:
```
mesutoezdil, line 51: the `default` branch has no test. add a case with a non-binpack, non-spread policy (e.g. empty string or `"topology"`) and devices on different numa nodes to cover the two return paths here.

mesutoezdil (commented):
1 gap: the `default` case has no test. 
codecov flags 3 lines missing. 
the default path is the numa-aware branch and its behavior is unchanged from the original code, but it should still be covered. a test with policy="" or an unknown policy and two devices on different numa nodes would cover both lines in the default case.

mesutoezdil:

PR #1810, opened 2026-04-28
Fix: Handle Kernel 6.17 handshake edge cases in NVIDIA health checks

### LINKED ISSUE #1800: Node unregistered by HAMi scheduler on Linux kernel 6.17 - missing hami.io/node-handshake annotation
## Environment
- HAMi version: v2.8.1
- Kubernetes: v1.34.4
- Linux kernel: 6.17.0-22-generic
- NVIDIA driver: 570.211.01 / 535.288.01
- CUDA: 12.8 / 12.2
- GPU: NVIDIA RTX A4500
- OS: Ubuntu 24.04.4 LTS
- containerd: 2.2.3

## Problem
On Linux kernel 6.17, HAMi scheduler marks node as "unregistered".
The `hami.io/node-handshake` annotation is missing on the node.

Error:
0/3 nodes are available: 1 node unregistered

## Root cause
- hami.io/node-handshake annotation is NOT created on kernel 6.17
- hami-device-plugin logs show successful GPU registration
- Downgrading to kernel 6.8 resolves the issue immediately

## Steps to reproduce
1. Install HAMi v2.8.1
2. Worker node with Linux kernel 6.17.x
3. Try to schedule GPU pod

## Workaround
Downgrade kernel to 6.8.x

## Related issue
#272

maishivamhoo123:
Description:
This PR fixes #1800  an edge case caused by a Kernel 6.17 bug where the NVIDIA device health check fails or hangs when device counts are current=0 and reported=0, but a handshake is either pending or expired.

Changes Made:

Updated device.go to properly bypass or resolve the handshake pending/expired states without breaking normal device tracking.

Added specific unit tests for the Kernel 6.17 edge cases to prevent future regressions.

### DIFF
--- a/pkg/device/nvidia/device.go
+++ b/pkg/device/nvidia/device.go
@@ -231,12 +231,14 @@ func (dev *NvidiaGPUDevices) CheckHealth(devType string, n *corev1.Node) (bool, 
 	reported := dev.ReportedGPUNum[n.Name]
 	klog.V(3).InfoS("checking device health for node", "nodeName", n.Name, "deviceType", devType, "currentDevices", current, "reportedDevices", reported)
 
+	handshakeHealthy, handshakeChanged := device.CheckHealth(devType, n)
+
 	if current == 0 {
 		if reported == 0 {
-			return true, false
+			return handshakeHealthy, handshakeChanged
 		}
 		dev.ReportedGPUNum[n.Name] = current
-		return false, false
+		return false, handshakeChanged
 	}
 
 	if reported != current {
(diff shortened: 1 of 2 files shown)


archlitchi:
i don't quite get the problem: if reported==0, it means the NVIDIA GPU is not healthy, so we shouldn't report that node to scheduler, nor should we patch any related-annotations

maishivamhoo123:
> i don't quite get the problem: if reported==0, it means the NVIDIA GPU is not healthy, so we shouldn't report that node to scheduler, nor should we patch any related-annotations

Hi @archlitchi, you're correct in the normal case. But on Kernel 6.17, the device plugin registers the GPU successfully, yet the hami.io/node-handshake annotation is never written to the node due to a kernel bug. This leaves the node permanently stuck at current=0, reported=0 even though the GPU is healthy.
So this PR checks the handshake state in that case:

Handshake pending → node still initializing → healthy=true
Handshake expired → something is stuck → healthy=false
No handshake → new node → original behavior preserved

Without this, users on Kernel 6.17 must downgrade the kernel to use HAMi despite having a working GPU.

archlitchi (approved):
/lgtm

PR #1812, opened 2026-04-28
feat: support node filtering based on hami-vnpu-core annotation

ashergaga:
**What type of PR is this?**
/kind feature

This PR enhances the `hami-scheduler` filtering strategy by enabling it to perceive the state of `hami-vnpu-core` on a node-by-node basis.

**Core Logic:**
During the Predicate (Filter) phase, the scheduler now retrieves and inspects the `hami-vnpu-core` annotation on the Node. If a Pod's annotation declares `huawei.com/vnpu-mod: hami-core`, the scheduler will strictly match nodes that have enabled the `hami-vnpu-core` annotation.


pkg/device/ascend/device.go
```diff
@@ -430,16 +431,42 @@ func (npu *Devices) Fit(devices []*device.DeviceUsage, request device.ContainerD
...
+	reason := make(map[string]int)
 
 	vnpuMode := ""
 	if pod != nil && pod.Annotations != nil {
 		vnpuMode = pod.Annotations[VNPUModeAnnotation]
 	}
+
+	isHAMiCore := (vnpuMode == VNPUModeHamiCore)
+
+	// Verify whether the Node supports hami vnpu core
```
DSFans2014, line 445: In what situations is it not supported?  Does the device plugin require special configuration or deployment compared with other nodes?

ashergaga: **In what situations is it not supported?**
It is not supported if the Ascend devices on a node do not have device share mode enabled. We use the `hami-device-node-config` ConfigMap to explicitly declare which nodes support `hami-vnpu-core`, and the device plugin will add the `hami-vnpu-core` annotation to these supported nodes. Furthermore, each declared node must exclusively choose either hard partitioning or soft partitioning—simultaneous use of both modes is not allowed.

**Does the device plugin require special configuration or deployment compared with other nodes?**
 Yes. Compared to standard nodes, it requires:
* Deploying the `hami-device-node-config` ConfigMap to designate `hami-vnpu-core`-capable nodes.
* Enabling device sharing on the Ascend hardware to allow container sharing.

DSFans2014: what's the reason to enable/disable the `hami-vnpu-core` at the node level?

archlitchi: like dynamic-mig and HAMi-core for nvidia, we need to support a heterogeneous NPU cluster, when some of these use 'vnpu-core' while others use legacy Ascend vNPU

archlitchi (approved):
/lgtm

PR #1817, opened 2026-04-29
fix: filter device does not work

DSFans2014:
**What type of PR is this?**

/kind bug

device filter does not work well since #1541
Fixes # 


Shouren (approved):
/lgtm

PR #1819, opened 2026-04-29
fix: mig does not work

DSFans2014:
**What type of PR is this?**

/kind bug

mig does not work since #1743. It will rename GPU ID like GPU-xxxx[0-0]  -> GPU-xxxx， causing the `GetContainerDeviceStrArray` function not to generate the MIG template

Fixes #


archlitchi (approved):
/lgtm

PR #1820, opened 2026-04-29
fix(scheduler): guard against zero-value division in ComputeScore (#1…

### LINKED ISSUE #1780: vgpu-scheduler-extender panics in ComputeScore when no node can satisfy pod GPU request — cluster-wide GPU scheduling goes down
**What happened**

The HAMi vgpu-scheduler-extender panics inside `ComputeScore` when it is asked to score a pod whose GPU requirements cannot be satisfied by any node in the cluster. The panic propagates as container exit code 2, which puts the scheduler into `CrashLoopBackOff`. While the scheduler is down, **all GPU scheduling cluster-wide stops** — not just scheduling of the offending pod.

This appears related to #1327 (CLOSED, "hami scheduler runtime panic") but that report did not include a stack trace or reproducer, so I'm filing a fresh report with concrete detail.

**What you expected to happen**

The scheduler should return an *unschedulable* verdict for pods whose GPU requests cannot be satisfied, and keep serving other scheduling decisions. A malformed / oversized pod spec should not be able to take out cluster-wide GPU scheduling.

**How to reproduce it (as minimally and precisely as possible)**

1. Deploy any pod that requests a GPU quantity or `nvidia.com/gpumem` amount larger than what any single node in the cluster can provide. Example trigger pods in our case:
   - Pod A requested 4 GPU devices (`nvidia.com/gpu: 4`) on a cluster where the largest node exposes 2 GPUs.
   - Pod B requested gpumem > the largest node's total GPU memory (\"no available GPU resources on all N nodes\" in scheduler events).
2. Leave the pods Pending.
3. Scheduler pod exits 2 with a panic the next time it tries to score nodes for one of these pods, and enters `CrashLoopBackOff` (restarts ~every 20 s).
4. Deleting / scaling those pods to 0 replicas makes the scheduler recover immediately to steady state.

**Stack trace**

```
github.com/Project-HAMi/HAMi/pkg/scheduler/policy.(*DeviceListsScore).ComputeScore
  /k8s-vgpu/pkg/scheduler/policy/gpu_policy.go:70 +0x2d7
github.com/Project-HAMi/HAMi/pkg/scheduler.fitInDevices
  /k8s-vgpu/pkg/scheduler/score.go:74 +0x70
github.com/Project-HAMi/HAMi/pkg/scheduler.(*Scheduler).calcScore.func1
  /k8s-vgpu/pkg/scheduler/score.go:173 +0x845
```

The crash is a runtime panic (nil deref / index out-of-range is the likely shape given the file/line) inside `ComputeScore`, reached via `fitInDevices` → `calcScore`. The offset `+0x2d7` in `gpu_policy.go:70` suggests the score function is indexing into or dereferencing a device list that is empty after the fit step has eliminated all candidate devices.

**Suggested guard (conceptual, not a PR yet)**

In `pkg/scheduler/policy/gpu_policy.go` `ComputeScore`, return a zero / very-low score (and let the scheduler surface an unschedulable verdict upstream) when the candidate device list is empty, rather than indexing into it. Same guard would likely belong one level up in `fitInDevices` so the empty-fit case is handled exactly once.

- Pod survived the unschedulable condition for ~2+ days before we noticed the scheduler was down
- Remediation was entirely manual (scale offending pods to 0, recreate scheduler pod); there is no self-recovery path for the scheduler
- Blast radius: every GPU-consuming workload in the cluster stopped scheduling, including unrelated pods, because the single scheduler replica was the crashing process

**Environment**

- HAMi version: `v2.6.14` (via the Olares downstream build `beclab/hami:v2.6.14`, which tracks this upstream)
- Kubernetes: `v1.33.3+k3s1`
- Cluster: 8-node k3s (1 control-plane, 7 GPU workers spanning RTX 2080 / 3090 / 4070 Ti / 4090 / 5080 / 5090 / RTX PRO 6000)
- Container runtime: containerd 2.0.5-k3s2 (workers) / 1.7.28 (control plane)
- nvidia driver: mix of 590.x and 580.x across nodes
- `nvidia.com/gpumem` is used cluster-wide for memory-sliced oversubscription

Happy to attach scheduler/device-plugin logs (and a minimal reproducer manifest) if that would help.

lin121291:
**What type of PR is this?**

bug

`ComputeScore` (gpu_policy.go:78-80) and `ComputeDefaultScore`
(node_policy.go:90-92) divide by device capacity fields (`Count`,
`Totalcore`, `Totalmem`) without zero-checks. When any field is 0,
float32 division produces NaN/Inf, which breaks `sort.Sort`'s
total-ordering requirement — causing an index-out-of-range panic
that kills the entire scheduler extender process.

Since the offending pod stays Pending, kube-scheduler retries it on
every restart, creating a CrashLoopBackOff loop that halts all GPU
scheduling cluster-wide.

This PR adds zero-value guards to both functions so they return
score 0 instead of panicking, letting the scheduler gracefully
mark devices/nodes as unschedulable.

Note: `fitInDevices` already guards against "request more GPUs than
the node has" at score.go:65, but that check runs **after**
`ComputeScore`. The trigger is a device reporting zero-value
capacity, not an empty device list.

Fixes #1780

- `ComputeDefaultScore` in node_policy.go has the same unguarded
  division pattern — fixed in this PR as well
- Unit tests added for both zero-capacity scenarios
- This PR was written with assistance from Claude Code

Scheduler no longer panics when a device reports zero-value capacity
fields. Pods that cannot be satisfied are correctly marked as
unschedulable instead of crashing the scheduler.

### DIFF
--- a/pkg/scheduler/policy/gpu_policy.go
+++ b/pkg/scheduler/policy/gpu_policy.go
@@ -57,6 +57,10 @@ func (l DeviceUsageList) Less(i, j int) bool {
 }
 
 func (ds *DeviceListsScore) ComputeScore(requests device.ContainerDeviceRequests) {
+	if ds.Device.Count == 0 || ds.Device.Totalcore == 0 || ds.Device.Totalmem == 0 {
+		ds.Score = 0
+		return
+	}
 	request, core, mem := int32(0), int32(0), int32(0)
 	// Here we are required to use the same type device
 	for _, container := range requests {
--- a/pkg/scheduler/policy/gpu_policy.go
+++ b/pkg/scheduler/policy/gpu_policy.go
@@ -57,7 +57,7 @@ func (l DeviceUsageList) Less(i, j int) bool {
 }
 
 func (ds *DeviceListsScore) ComputeScore(requests device.ContainerDeviceRequests) {
-	if ds.Device.Count == 0 || ds.Device.Totalcore == 0 || ds.Device.Totalmem == 0 {
+	if ds.Device == nil || ds.Device.Count == 0 || ds.Device.Totalcore == 0 || ds.Device.Totalmem == 0 {
 		ds.Score = 0
 		return
 	}
--- a/pkg/scheduler/policy/gpu_policy_test.go
+++ b/pkg/scheduler/policy/gpu_policy_test.go
@@ -204,6 +204,30 @@ func TestComputeScore(t *testing.T) {
 		expectedScore float32
 	}{
 		{
+			name: "Zero capacity device returns score 0 without panic",
+			device: &device.DeviceUsage{
+				ID:        "test-device",
+				Type:      "type1",
+				Count:     0,
+				Totalcore: 0,
+				Totalmem:  0,
+			},
+			requests:      make(device.ContainerDeviceRequests),
+			expectedScore: 0,
+		},
+		{
+			name: "Partial zero capacity (Count=0) returns score 0 without panic",
+			device: &device.DeviceUsage{
+				ID:        "test-device",
+				Type:      "type1",
+				Count:     0,
+				Totalcore: 8,
+				Totalmem:  4096,
+			},
+			requests:      make(device.ContainerDeviceRequests),
+			expectedScore: 0,
+		},
+		{
 			name: "ContainerDeviceRequests has no data",
 			device: &device.DeviceUsage{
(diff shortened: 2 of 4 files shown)


archlitchi:
please resolve the AI comments

Shouren (approved):
/lgtm

PR #1824, opened 2026-04-30
feat: add enableGetPreferredAllocation flag

DSFans2014:
**What type of PR is this?**

/kind feature

#1743 will enable `GetPreferredAllocation` by default and rename DeviceIDs. 
We discussed this feature in the weekly meeting, and the final conclusion was to add a switch that is disabled by default.

Additionally, using the default fallback may cause the following error, so  it has been changed to return an empty response in this pr.

ScreenShot_2026-04-30_130603_447

Fixes #


Shouren (approved):
/lgtm

PR #1825, opened 2026-04-30
refactor: filter device at nvml-manager

DSFans2014:
**What type of PR is this?**

/kind design

#1817 add filter in server.go before register.  This solution filters before registration. However, if the filtered-out devices are faulty cards, this may cause additional issues.  So we filter at the `nvml_manager` during device identification.

Fixes #


Shouren (approved):
/lgtm

PR #1831, opened 2026-05-02
fix(ascend): check unmarshal error before iterating node devices

mesutoezdil:
/kind bug

The GetNodeDevices function in the Ascend device implementation was iterating over nodeDevices before checking the error returned by UnMarshalNodeDevices. If the unmarshal fails and returns a partial result alongside an error, the loop would mutate that partial data before the error is caught and returned.

The NVIDIA and Hygon implementations both check the error first. This aligns the Ascend implementation with the same pattern.


mesutoezdil:
@fouof can you take a look? moves the vendor loop after the unmarshal error check in pkg/device/ascend/device.go to match nvidia and hygon.

mesutoezdil:
@archlitchi can you take a look? moves the vendor loop after the unmarshal error check in pkg/device/ascend/device.go to match nvidia and hygon.

maishivamhoo123:
e2e job seems to be cancelled due to runner timeout, likely infra-related.
/lgtm

Shouren (approved):
/lgtm

PR #1837, opened 2026-05-06
feat: Supports multi-device requests with hami-vnpu-core enabled.

ashergaga:
**What type of PR is this?**
/kind feature

This PR enables multi-device request support when `hami-vnpu-core` is enabled. Previously, vNPU allocation was restricted to single-device requests. This enhancement allows users to request multiple virtual NPUs per workload, specifically addressing the hardware requirement for **paired module allocation** on **Ascend 910C** devices.


archlitchi (approved):
/lgtm

PR #1816, opened 2026-04-28
fix(device): parse handshake annotation timestamp in local timezone

mesutoezdil:
Fixes a timezone mismatch in `pkg/device/devices.go` `CheckHealth`.

The handshake annotation is written with `time.Now().Format(time.DateTime)` (local time),
but was read back with `time.Parse(time.DateTime, ...)` which returns UTC. On non-UTC hosts
this caused expired `Requesting_` annotations to appear still valid, so an unhealthy node
would pass the health check.

**Fix:** use `time.ParseInLocation(time.DateTime, str, time.Local)` to match the write side.

Found this while rebasing - the `TestCheckHealth` test added by #1810 fails consistently
on UTC+2 hosts because of this mismatch.

Tested locally with `go test ./pkg/device/nvidia/... -count=1`, all pass.

PR #1843, opened 2026-05-07
fix: recover scheduling on nodes with stale Deleted_ handshake

### LINKED ISSUE #1839: Scheduler leaves `hami.io/node-handshake: Deleted_<ts>` blocking all subsequent GPU scheduling until manually stripped
**What happened**:

After a node-level event (kubelet restart, node reboot, transient health failure, or a single failed schedule), the hami-scheduler can leave `hami.io/node-handshake: Deleted_<timestamp>` set on the node annotation. Once the annotation is in this state, every subsequent GPU pod fails to schedule on that node with:

```
0/N nodes are available: N node unregistered. no new claims to deallocate, preemption: 0/N nodes are available: N No preemption victims found for incoming pod.
```

even though, in parallel:
- `hami-device-plugin` is `Running` and healthy
- The node's `Allocatable` shows `nvidia.com/gpu: <N>`
- Scheduler logs show device health checks succeeding every 15s, e.g.: `device.go:232 "checking device health for node" nodeName="spark-5385" deviceType="NVIDIA" currentDevices=2 reportedDevices=2`

The only way to recover scheduling is to manually strip the annotation:

```
kubectl annotate node <name> hami.io/node-handshake- --overwrite
```

Within ~30s the device-plugin re-stamps `Requesting_<timestamp>` and pending GPU pods schedule normally on the next attempt.

Once `CheckHealth` reports the device(s) healthy again after a transient failure, the scheduler (or the device-plugin reconcile) should reset the handshake to `Requesting_`, allowing scheduling to resume. `Deleted_` should not be a sticky terminal state when the underlying device is back to healthy.

1. Have a single-GPU-worker cluster with HAMi installed and at least one GPU pod previously scheduled on the worker.
2. Reboot the worker (or otherwise cause kubelet/device-plugin to drop and reconnect).
3. While the device-plugin is restarting, the scheduler observes `health=false` for the node, calls `NodeCleanUp` → `MarkAnnotationsToDelete` writes `Deleted_<timestamp>` to `hami.io/node-handshake` (`pkg/util/util.go:225`), and removes the node from its cache via `s.rmNodeDevices(...)` (`pkg/scheduler/scheduler.go:407-414`).
4. Device-plugin comes back, GPU is re-registered (`nvidia.com/gpu: <N>` reappears in node `Allocatable`).
5. Apply a new GPU pod (`nvidia.com/gpu: 1`).
6. Observe pod `Pending` with `FailedScheduling: 0/N nodes are available: N node unregistered`. Scheduler logs show `device.go:208 "Device info unchanged, skipping annotation update"` — the device-plugin doesn't re-stamp `Requesting_` because nothing about the device changed from its perspective.
7. Pod stays `Pending` indefinitely until the annotation is manually stripped.

Pointers in the code (master HEAD `31a2792`):

- `pkg/scheduler/scheduler.go:407` is the call that triggers cleanup when a single health-check fails.
- `pkg/util/util.go:223-232` (`MarkAnnotationsToDelete`) is what writes the `Deleted_<timestamp>` value.
- `pkg/device/devices.go:443` is where `Requesting_` would be written, but only inside the device-plugin's positive registration path which short-circuits on "device info unchanged".

Possible fixes (in increasing scope):

1. In `MarkAnnotationsToDelete`, only write `Deleted_` if the previous value was `Requesting_`. If the previous value was already `Deleted_`, do not refresh the timestamp — and crucially, do not write `Deleted_` if the device-plugin has heart-beat the node within the last N seconds.
2. On the device-plugin's periodic reconcile, always re-stamp `Requesting_` if the current value starts with `Deleted_`, regardless of whether device info changed. (One-line change near `register.go:208`.)
3. Treat `Deleted_<old timestamp>` as a recoverable state in the scheduler's `CheckHealth` — if the timestamp is older than some threshold and devices currently report healthy, reset to `Requesting_` instead of skipping.

Workaround used in production: a `CronJob` in the cluster that strips `Deleted_*` handshake annotations every minute. Functional, but feels like the responsibility belongs upstream.

**Environment**:

- HAMi version: master (HEAD `31a2792`) and `v2.8.2`
- nvidia driver: `580.95.05`
- Kernel: `6.14.0-1015-nvidia`
- Hardware: NVIDIA DGX Spark (GB10), `aarch64`
- vCluster v0.33.1 hosting two Tenant Clusters; same behaviour observed on the host cluster directly.

---

_AI assistance disclosure: this issue was drafted with Claude Code based on hands-on reproduction during keynote demo prep; reviewed and submitted by me._

saiyam1814:
**What type of PR is this?**

/kind bug

`CheckHealth` in `pkg/device/devices.go` previously returned `(true, false)` for any handshake annotation containing `Deleted`, which is what the scheduler's reconcile loop interprets as *"node is healthy AND no cache update needed"* — so it skips re-adding the node to its cache. After `NodeCleanUp` writes `Deleted_<ts>` and `s.rmNodeDevices(...)` removes the node from `s.nodes` (e.g. after a transient health failure during a node reboot), the node never gets re-added even after the device-plugin reports devices healthy again. Every subsequent GPU pod fails to schedule with:

```
0/N nodes are available: N node unregistered. no new claims to deallocate
```

This PR mirrors the timestamp logic that the `Requesting` branch already uses: if `Deleted_<ts>` is older than the same 60s cooldown window, stamp `Requesting_<now>` and return `(true, true)` so the scheduler re-adds the node's devices on its next reconcile. Inside the cooldown, behaviour is unchanged. Bare `"Deleted"` without a timestamp (used in some unit tests) and any unparsable timestamp also keep `(true, false)` so we never accidentally recover from a malformed value.

This is a single-file change in shared code, so it covers every vendor that uses `CheckHealth` (NVIDIA, Hygon, Ascend, Vastai, Kunlun, Iluvatar, Enflame, Mthreads, Metax) — all of them are stuck unconditionally today once the annotation is in `Deleted_<ts>` state.

Fixes #1839

The shape of the fix follows the discussion on #1839, where @mesutoezdil pointed out that:

- Constraining `MarkAnnotationsToDelete` only prevents *new* stuck states.
- Re-stamping `Requesting_` from the device-plugin doesn't help: scheduler still takes the `(true, false)` exit (within the 60s window) and skips cache rebuild. Recovery requires `(true, true)` to fire the re-add.
- Mirroring the `Requesting_` time-window logic for `Deleted_<ts>` in `CheckHealth` is the right shape — single-file change, no device-plugin changes, covers all vendors.

The implementation also follows the maintainer's note that *"bare `Deleted` without a timestamp appears in some tests, make sure the parse-error path stays at `return true, false` rather than accidentally recovering."* Every parse and patch failure in the new branch returns the conservative `(true, false)`.

Local verification:

- `go test ./pkg/device/ -run Test_CheckHealth -v` — all 7 cases pass (4 existing + 3 new)
- `gofmt` clean, `go vet` clean
- The new test "Deleted state stale" returns `(true, false)` in unit-test env because `util.GetNode` fails outside a real cluster — matching the existing `"Unknown state"` case which exercises the same path.

The diff is small (`+95/-1` across `pkg/device/devices.go` and `pkg/device/devices_test.go`) and intentionally does not refactor the empty-annotation `else` branch even though the new branch is structurally similar — keeping the change minimal. Happy to factor out a shared helper in a follow-up if maintainers prefer.

**Marking as draft** while we get formal direction from an OWNERS approver on #1839; the code is ready to review but the discussion thread should ideally land first.

```release-note
Fix scheduling on GPU nodes that recover after a transient health
failure: the scheduler now re-adds nodes whose `hami.io/node-handshake`
annotation is `Deleted_<ts>` older than the 60s cooldown, instead of
leaving them permanently outside the scheduler cache.
```

---

_AI assistance disclosure: this PR was drafted with Claude Code, based on the maintainer-led discussion in #1839 and on direct verification of `pkg/device/devices.go` and the existing `Test_CheckHealth` patterns. Code authored by me; tests pass locally before submission._

### DIFF
--- a/pkg/device/devices.go
+++ b/pkg/device/devices.go
@@ -510,7 +510,7 @@ func CheckHealth(devType string, node *corev1.Node) (bool, bool) {
 		if len(parts) < 2 {
 			return true, false
 		}
-		formerTime, err := time.Parse(time.DateTime, parts[1])
+		formerTime, err := time.ParseInLocation(time.DateTime, parts[1], time.Local)
 		if err != nil {
 			return true, false
 		}
--- a/pkg/device/devices.go
+++ b/pkg/device/devices.go
@@ -502,6 +502,10 @@ func CheckHealth(devType string, node *corev1.Node) (bool, bool) {
 		// Bare "Deleted" without a timestamp (used in some unit tests) and
 		// any unparsable timestamp must keep the conservative (true, false)
 		// path so we never recover from a malformed value.
+		annoKey, ok := util.HandshakeAnnos[devType]
+		if !ok {
+			return true, false
+		}
 		parts := strings.SplitN(handshake, "_", 2)
 		if len(parts) < 2 {
 			return true, false
@@ -510,12 +514,17 @@ func CheckHealth(devType string, node *corev1.Node) (bool, bool) {
 		if err != nil {
 			return true, false
 		}
-		if time.Now().Before(formerTime.Add(time.Second * 60)) {
+		now := time.Now()
+		if now.Before(formerTime.Add(time.Second * 60)) {
 			return true, false
 		}
-		tmppat := make(map[string]string)
-		tmppat[util.HandshakeAnnos[devType]] = "Requesting_" + time.Now().Format(time.DateTime)
-		klog.V(5).InfoS("Recovering stale Deleted_ handshake", "nodeName", node.Name, "annotationKey", util.HandshakeAnnos[devType], "annotationValue", tmppat[util.HandshakeAnnos[devType]])
+		newHandshake := "Requesting_" + now.Format(time.DateTime)
+		tmppat := map[string]string{annoKey: newHandshake}
+		klog.V(5).InfoS("Recovering stale Deleted_ handshake", "nodeName", node.Name, "annotationKey", annoKey, "annotationValue", newHandshake)
+		// Mirror the empty-annotation branch: GetNode also acts as a guard
+		// for an empty node.Name and an uninitialised client (PatchNodeAnnotations
+		// panics in unit tests without it). Worth a follow-up to dedupe with
+		// the else branch into a helper.
 		n, err := util.GetNode(node.Name)
 		if err != nil {
 			klog.ErrorS(err, "Failed to get node", "nodeName", node.Name)
--- a/pkg/device/devices.go
+++ b/pkg/device/devices.go
@@ -492,7 +492,39 @@ func CheckHealth(devType string, node *corev1.Node) (bool, bool) {
 		formertime, _ := time.ParseInLocation(time.DateTime, strings.Split(handshake, "_")[1], time.Local)
 		return time.Now().Before(formertime.Add(time.Second * 60)), false
 	} else if strings.Contains(handshake, "Deleted") {
-		return true, false
+		// Mirror the timestamp logic used by the Requesting branch: a
+		// Deleted_<ts> older than 60s on a node whose devices are otherwise
+		// reporting healthy means the previous cleanup is stale and the
+		// scheduler should bring the node back into its cache. Stamp
+		// Requesting_<now> and return (true, true) so the caller re-adds
+		// node devices on the next reconcile.
+		//
+		// Bare "Deleted" without a timestamp (used in some unit tests) and
+		// any unparsable timestamp must keep the conservative (true, false)
+		// path so we never recover from a malformed value.
+		parts := strings.SplitN(handshake, "_", 2)
+		if len(parts) < 2 {
+			return true, false
+		}
+		formerTime, err := time.Parse(time.DateTime, parts[1])
+		if err != nil {
+			return true, false
+		}
+		if time.Now().Before(formerTime.Add(time.Second * 60)) {
+			return true, false
+		}
+		tmppat := make(map[string]string)
+		tmppat[util.HandshakeAnnos[devType]] = "Requesting_" + time.Now().Format(time.DateTime)
+		klog.V(5).InfoS("Recovering stale Deleted_ handshake", "nodeName", node.Name, "annotationKey", util.HandshakeAnnos[devType], "annotationValue", tmppat[util.HandshakeAnnos[devType]])
+		n, err := util.GetNode(node.Name)
+		if err != nil {
+			klog.ErrorS(err, "Failed to get node", "nodeName", node.Name)
+			return true, false
+		}
+		if err := util.PatchNodeAnnotations(n, tmppat); err != nil {
+			klog.ErrorS(err, "Failed to patch node annotations", "nodeName", node.Name)
+		}
+		return true, true
 	} else {
 		_, ok := util.HandshakeAnnos[devType]
 		if ok {
--- a/pkg/device/devices_test.go
+++ b/pkg/device/devices_test.go
@@ -622,6 +622,68 @@ func Test_CheckHealth(t *testing.T) {
 			want2: false,
 		},
 		{
+			// Inside the 60s cooldown window — keep the conservative path,
+			// scheduler should not yet re-add the node to its cache.
+			name: "Deleted state within cooldown",
+			args: struct {
+				devType string
+				n       corev1.Node
+			}{
+				devType: "huawei.com/Ascend910",
+				n: corev1.Node{
+					ObjectMeta: metav1.ObjectMeta{
+						Annotations: map[string]string{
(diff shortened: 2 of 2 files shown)


saiyam1814:
Thanks @gemini-code-assist — applied two of the three suggestions in d32e875:

1. **Annotation key validation** — added `annoKey, ok := util.HandshakeAnnos[devType]; if !ok { return true, false }`, matching the existing `else` branch's pattern. Good catch.
2. **Single `time.Now()`** — consolidated to one `now := time.Now()` for both the cooldown check and the new `Requesting_` stamp.

3. **Dropping `util.GetNode`** — kept it. You're right that `PatchNodeAnnotations` only reads `node.Name`, but `GetNode` is doing implicit safety work the existing `else` branch quietly relies on: it returns early if `node.Name` is empty or if `client.GetClient()` is nil. Without it, `PatchNodeAnnotations` immediately calls `.CoreV1()` on a nil client and panics — which is what blew up the unit tests when I tried this version locally:

   ```
   panic: ...
       util.PatchNodeAnnotations(...) at util.go:149
       device.CheckHealth(...) at devices.go:524
       ...
   FAIL  github.com/Project-HAMi/HAMi/pkg/device  1.241s
   ```

   So the redundancy is real *at runtime*, but the safety it provides is real too. Reverted that part to keep parity with the existing `else` branch. Left a `// follow-up to dedupe` comment near the call site — extracting a shared helper that does the validation + patch explicitly would be the right cleanup, but felt larger than this minimal bug-fix PR should carry.

All `Test_CheckHealth` cases pass after the update; lint and vet are clean.

saiyam1814:
@mesutoezdil

wawa0210:
----    
@mesutoezdil

mesutoezdil:
@saiyam1814 @wawa0210 

thx! logic is correct and matches what i described. scheduler iterates rawnodes from the lister so the cache-evicted node is still visible, getnodedevices reads annotations so it succeeds post-recovery, the 60s cooldown prevents premature re-add.                                                             

but 1 issue: "deleted state stale" test does not actually exercise the recovery path. util.getnode fails in unit-test env so the function falls back to (true, false), which is what the test expects. 
the actual fix (returning (true, true)) has no test coverage. not a blocker if there's no clean way to mock getnode without refactoring, but worth a note.

everything else looks good. lgtm pending that acknowledgement @wawa0210

mesutoezdil:
@wawa0210 1 more thing: the new code uses time.Parse(time.DateTime, parts[1]) which returns UTC, but the annotation is written with time.Now().Format(time.DateTime) (local time). on non-UTC hosts this makes stale Deleted_<ts> values appear still within the 60s cooldown window, delaying recovery by hours. 
same bug #1816 fixes for the Requesting_ branch. should be time.ParseInLocation(time.DateTime, parts[1], time.Local) here too (@archlitchi)

saiyam1814:
Should #1816 be merged first then?

mesutoezdil:
@saiyam1814 yes, either merge #1816 first and rebase, or apply the same fix here directly:                                       
                                                                                                                                   
  pkg/device/devices.go, in the new Deleted_ branch:                                                                               
    time.Parse(time.DateTime, parts[1])                                                                                            
  → time.ParseInLocation(time.DateTime, parts[1], time.Local)                                                                      
                                                                                                                                   
  without this, the cooldown check is wrong on any non-UTC host
  stale Deleted_<ts> values look fresh, recovery is delayed by     
  hours. same root cause as #1816.                                                                                                 
                                                                                                                                   
@archlitchi @wawa0210 can one of you approve #1816 so saiyam can rebase? it's a one-line change with a passing test.

archlitchi:
> #1816

approved and merged

saiyam1814:
Rebased onto current `master` after #1816 merged earlier today.

#1816 fixed the timezone parsing on the `Requesting` branch (`time.Parse` → `time.ParseInLocation` with `time.Local`). The new `Deleted_` recovery branch in this PR parses the same annotation format, so I aligned it with the same approach in af06e53 — otherwise the 60s cooldown check would be off by the local-vs-UTC offset on non-UTC hosts and we'd be re-introducing the bug #1816 just fixed.

No content conflicts during the rebase (different parts of `CheckHealth`); just one consistency tweak. Tests still pass (`go test ./pkg/device/ -run Test_CheckHealth`), `gofmt` and `go vet` clean.

saiyam1814:
@mesutoezdil

wawa0210:
----    
@mesutoezdil

mesutoezdil:
thx @saiyam1814 @wawa0210 @archlitchi

wawa0210:
----    
thx @saiyam1814 @wawa0210 @archlitchi

archlitchi (approved):
/lgtm

PR #1855, opened 2026-05-11
fix: propagate parse errors in DecodeNodeDevices

mesutoezdil:
DecodeNodeDevices was ignoring parse errors from strconv.ParseInt, strconv.ParseBool, and strconv.Atoi. A bad annotation value would silently become zero, leading to incorrect device allocation.

Each field now returns a wrapped error on parse failure. The safecast calls were removed because ParseInt with bitSize 32 already covers the int32 range check. Generic error messages are replaced with field-specific context. The test is updated to match.

### DIFF
--- a/pkg/device/devices.go
+++ b/pkg/device/devices.go
@@ -18,7 +18,6 @@ package device
 
 import (
 	"encoding/json"
-	"errors"
 	"fmt"
 	"maps"
 	"slices"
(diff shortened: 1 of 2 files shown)


pkg/device/devices.go
```diff
@@ -240,42 +239,48 @@ func GetDevices() map[string]Devices {
...
+						return nil, fmt.Errorf("invalid index field: %w", err)
+					}
+					mode = items[8]
 				}
 				i := DeviceInfo{
 					ID:      items[0],
-					Count:   count32,
-					Devmem:  devmem32,
-					Devcore: devcore32,
+					Count:   int32(count),
```
DSFans2014: why delete the `safecast.Convert` ?

mesutoezdil: @DSFans2014 parseint already does the range check for you when you pass bitsize=32
if the value doesn't fit in an int32, the parse itself fails. 
so by the time you reach the cast, the number is guaranteed to be safe. 
the safecast.convert call was just doing the same check twice and could never actually catch anything. 
I removed it to keep the code clean.

DSFans2014 (commented):
/lgtm

FouoF (approved):

PR #1857, opened 2026-05-11
fix: return error from DecodePodDevices on decode failure

mesutoezdil:
DecodePodDevices was returning nil error when DecodeContainerDevices failed, hiding the problem from all callers. onAddPod in the scheduler was discarding the error with blank identifier, so failures were never visible in logs.

DecodePodDevices now passes back the error. onAddPod logs it and continues tracking the pod with an empty device map.

### DIFF
--- a/pkg/device/devices.go
+++ b/pkg/device/devices.go
@@ -444,7 +444,7 @@ func DecodePodDevices(checklist map[string]string, annos map[string]string) (Pod
 		for s := range strings.SplitSeq(str, OnePodMultiContainerSplitSymbol) {
 			cd, err := DecodeContainerDevices(s)
 			if err != nil {
-				return PodDevices{}, nil
+				return PodDevices{}, err
 			}
 			// IMPORTANT: Do NOT skip empty ContainerDevices!
 			// We must preserve the index mapping between annotation entries and pod containers.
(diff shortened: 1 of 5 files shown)


pkg/device/devices.go
```diff
@@ -444,7 +444,7 @@ func DecodePodDevices(checklist map[string]string, annos map[string]string) (Pod
 		for s := range strings.SplitSeq(str, OnePodMultiContainerSplitSymbol) {
 			cd, err := DecodeContainerDevices(s)
 			if err != nil {
-				return PodDevices{}, nil
+				return PodDevices{}, err
```
DSFans2014, line 447: Returning an error may change the behavior of the caller.  Consider the following example:
https://github.com/Project-HAMi/HAMi/blob/491e32967abc1ff6b7cecfea1eb5cd6e4e63666a/cmd/vGPUmonitor/metrics.go#L584-L589

mesutoezdil: The vGPUmonitor call site now logs the error and uses `continue` to skip the bad pod instead of returning an error that would abort the entire per-node metrics loop. 
That way a single malformed annotation does not silence metrics for all other pods on the node.

DSFans2014 (commented):
/lgtm

Shouren (approved):
/lgtm

PR #1870, opened 2026-05-15
fix(nvidia): add explicit check for gpumem when determining exclusive…

### LINKED ISSUE #1863: gpucores request is automatically set as 100 when only gpumem is requested
**What happened**:
- Hami v2.9.0
- k8s v1.34

When we request only `nvidia.com/gpumem` for gpu fraction as follows, `nvidia.com/gpucores: 100` request is automatically added. This situation always occurs regardless of the value of `nvidia.com/gpumem`.
It does not occur only when partial `nvidia.com/gpumem-percentage` is requested.

```
    Limits:
      cpu:                           4
      memory:                        100Gi
      nvidia.com/gpu:                1
      nvidia.com/gpucores:           50 <-- added
      nvidia.com/gpumem:  20480
    Requests:
      cpu:                           4
      memory:                        10Gi
      nvidia.com/gpu:                1
      nvidia.com/gpucores:           50 <-- added
      nvidia.com/gpumem:  20480

```

It seems that the main cause is PR #1354 . Its purpose was requesting gpucores=100 only when exclusive memory is requested, but gpumem is omitted from the condition.

maishivamhoo123:
Fixes #1863 
## Problem
`nvidia.com/gpucores` is incorrectly auto-set to 100 when only `nvidia.com/gpumem` 
is requested with a specific value.

## Solution
Add explicit check in defaultExclusiveCoreIfNeeded() if gpumem is present 
with a specific value, treat as shareable memory (not exclusive).

### DIFF
--- a/pkg/device/nvidia/device.go
+++ b/pkg/device/nvidia/device.go
@@ -416,11 +416,17 @@ func (dev *NvidiaGPUDevices) defaultExclusiveCoreIfNeeded(ctr *corev1.Container)
 	}
 
 	exclusive := false
+	// gpumem-percentage explicitly set
 	if pct, ok := resourceValue(ctr, corev1.ResourceName(dev.config.ResourceMemoryPercentageName)); ok {
 		exclusive = pct == 100
+	} else if _, ok := resourceValue(ctr, corev1.ResourceName(dev.config.ResourceMemoryName)); ok {
+		// gpumem is explicitly set to a specific value (NOT exclusive)
+		exclusive = false
 	} else if dev.config.ResourceMemoryName == "" {
+		// ResourceMemoryName is disabled in config (treat as exclusive)
 		exclusive = true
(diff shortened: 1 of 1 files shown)


maishivamhoo123:
@archlitchi can you please review this PR?

archlitchi:
actually, i don't get how thing went wrong, we default to false, only when 'resourceMemoryName' is absent, we reset this value to true. how can a pod with 'deviceMemory' field bypass this check?

maishivamhoo123:
> actually, i don't get how thing went wrong, we default to false, only when 'resourceMemoryName' is absent, we reset this value to true. how can a pod with 'deviceMemory' field bypass this check?

@archlitchi Thank you review 
The change adds an explicit check for when gpumem is present. Instead of relying on the default value to handle the gpumem case, we now explicitly set `exclusive = false`. This makes the intent clear and the logic easier to follow.

PR #1880, opened 2026-05-18
fix(quota): clamp Used to zero in RmUsage to prevent negative tracking

mesutoezdil:
What type of PR is this?

/kind bug

Quota.Used can go negative when RmUsage is called for a pod that has no matching AddUsage entry.

FitQuota checks Used + alloc > Limit. When Used is negative the check passes even if the real usage would exceed the limit, so the scheduler allows more pods than the namespace quota permits.

PR #1400 added a second RmUsage call site in the terminated-pod path of onAddPod. In edge cases both onAddPod and onDelPod can attempt to remove the same pod from quota tracking, making the underflow reachable.

The fix uses max(0, Used-val) in RmUsage so Used never goes below zero. A test covers the duplicate removal case.

Which issue(s) this PR fixes:

Related: #1400

No behavior change for the normal path. The fix only prevents Used from going below zero.

Does this PR introduce a user-facing change?

No.

### DIFF
--- a/pkg/device/pod_test.go
+++ b/pkg/device/pod_test.go
@@ -542,3 +542,21 @@ func TestContainerDeviceDeepCopy(t *testing.T) {
 	_, exists := original.CustomInfo["key2"]
 	assert.False(t, exists, "original CustomInfo should not have key2")
 }
+
+func TestTakeAndDeletePodIsAtomic(t *testing.T) {
+	pm := NewPodManager()
+	pod := &corev1.Pod{ObjectMeta: metav1.ObjectMeta{
+		Name:      "p",
+		Namespace: "ns",
+		UID:       k8stypes.UID("uid-1"),
+	}}
+	pm.AddPod(pod, "node-1", PodDevices{})
+
+	pi1, ok1 := pm.TakeAndDeletePod(pod)
+	pi2, ok2 := pm.TakeAndDeletePod(pod)
+
+	assert.True(t, ok1)
+	assert.NotNil(t, pi1)
+	assert.False(t, ok2)
+	assert.Nil(t, pi2)
+}
--- a/pkg/device/pods.go
+++ b/pkg/device/pods.go
@@ -119,6 +119,18 @@ func (m *PodManager) GetPod(pod *corev1.Pod) (*PodInfo, bool) {
 	return pi, ok
 }
 
+func (m *PodManager) TakeAndDeletePod(pod *corev1.Pod) (*PodInfo, bool) {
+	m.mutex.Lock()
+	defer m.mutex.Unlock()
+
+	pi, ok := m.pods[pod.UID]
+	if ok {
+		delete(m.pods, pod.UID)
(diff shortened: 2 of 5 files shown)


DSFans2014 (commented):
> In edge cases both onAddPod and onDelPod can attempt to remove the same pod from quota tracking, making the underflow reachable.

There is the lock to protect the `Used` field in `RmUsage` function currently. Could you describe in detail why the `underflow` happened in this case?

mesutoezdil:
> > In edge cases both onAddPod and onDelPod can attempt to remove the same pod from quota tracking, making the underflow reachable.
> 
> There is the lock to protect the `Used` field in `RmUsage` function currently. Could you describe in detail why the `underflow` happened in this case?

the mutex prevents data corruption, not a double-call. onAddPod (terminated path, line 150) and onDelPod (line 196) run in separate goroutines. 
both can pass GetPod before either calls DelPod, so RmUsage executes twice sequentially
and the mutex serializes them but both subtractions happen. 
max(0, Used-val) makes the second call a no-op.

DSFans2014:
> > > In edge cases both onAddPod and onDelPod can attempt to remove the same pod from quota tracking, making the underflow reachable.
> > 
> > 
> > There is the lock to protect the `Used` field in `RmUsage` function currently. Could you describe in detail why the `underflow` happened in this case?
> 
> the mutex prevents data corruption, not a double-call. onAddPod (terminated path, line 150) and onDelPod (line 196) run in separate goroutines. both can pass GetPod before either calls DelPod, so RmUsage executes twice sequentially and the mutex serializes them but both subtractions happen. max(0, Used-val) makes the second call a no-op.

thanks for your reply. I think we should guarantee the value of `Used` is right in this situations rather than just non-negative.

mesutoezdil:
> > > > In edge cases both onAddPod and onDelPod can attempt to remove the same pod from quota tracking, making the underflow reachable.
> > > 
> > > 
> > > There is the lock to protect the `Used` field in `RmUsage` function currently. Could you describe in detail why the `underflow` happened in this case?
> > 
> > 
> > the mutex prevents data corruption, not a double-call. onAddPod (terminated path, line 150) and onDelPod (line 196) run in separate goroutines. both can pass GetPod before either calls DelPod, so RmUsage executes twice sequentially and the mutex serializes them but both subtractions happen. max(0, Used-val) makes the second call a no-op.
> 
> thanks for your reply. I think we should guarantee the value of `Used` is right in this situations rather than just non-negative.

pushed a new commit. instead of clamping the result, TakeAndDeletePod makes the get+delete atomic under a single lock so only 1 of onAddPod/onDelPod can win the race. RmUsage is now called exactly once, Used stays accurate.

DSFans2014 (commented):
please resolve the lint action error

mesutoezdil:
> please resolve the lint action error

all 3 prs ready to merge

DSFans2014 (commented):
/lgtm

archlitchi (approved):
/lgtm

PR #1881, opened 2026-05-18
fix(vGPUmonitor): prevent index out of bounds panic for priority above 1

mesutoezdil:
What type of PR is this?

/kind bug

utSwitchOn[uuid] was always initialized as a fixed two-element slice.

A container with CUDA_TASK_PRIORITY set to 2 or higher triggers a panic in 3 places.

Observe() writes utSwitchOn[uuid][priority] without growing the slice first.

CheckBlocking() and CheckPriority() loop up to the container priority and access utSwitchOn[uuid][p] without checking the slice length.

The bug was introduced in commit 54dc0de which replaced explicit for loops with range-over-int but kept the fixed two-element initialization.

The fix grows the slice in Observe() to fit any priority value and adds length guards in CheckBlocking() and CheckPriority().

### DIFF
--- a/cmd/vGPUmonitor/feedback.go
+++ b/cmd/vGPUmonitor/feedback.go
@@ -95,7 +95,7 @@ func CheckBlocking(utSwitchOn map[string]UtilizationPerDevice, p int, c *nvidia.
 		uuid := c.Info.DeviceUUID(i)
 		_, ok := utSwitchOn[uuid]
 		if ok {
-			for i := range p {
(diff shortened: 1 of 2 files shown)


archlitchi:
please re-sync with master to pass the e2e CI

mesutoezdil:
> please re-sync with master to pass the e2e CI

all 3 prs ready to merge

archlitchi (approved):
/lgtm

PR #1887, opened 2026-05-19
feat(helm): make device plugin security context configurable via values

### LINKED ISSUE #1888: [Feature] Make device plugin security context configurable via Helm values
## Motivation

As raised in #1143, not all security privileges HAMi requests are required for every use case or environment. Through testing we found that settings such as `privileged: true` and `hostPID: true` on the device plugin DaemonSet are not universally necessary — the exact requirements depend on the runtime, node OS, and which HAMi features are in use.

Users deploying HAMi in environments with strict Pod Security Admission policies or a least-privilege posture currently have no way to adjust these settings without patching the chart templates directly.

## Proposed solution

Expose the following hardcoded fields in `values.yaml` so they can be overridden per-deployment, while keeping the current settings as defaults (non-breaking):

- `devicePlugin.hostPID` (default: `true`)
- `devicePlugin.hostNetwork` (default: `false`)
- `devicePlugin.securityContext` — main `device-plugin` container (default: `privileged: true`, `allowPrivilegeEscalation: true`, `SYS_ADMIN`)
- `devicePlugin.monitor.securityContext` — `vgpu-monitor` container (default: `allowPrivilegeEscalation: false`, `SYS_ADMIN`)
- `devicePlugin.gpuOperatorToolkitReady.securityContext` — toolkit-validation init container (default: `privileged: true`, `runAsUser: 0`)

This is a first step toward the broader least-privilege improvements discussed in #1143, allowing operators to dial down privileges where their environment permits without waiting for upstream changes to the defaults.

ilia-medvedev:
## Summary

- `hostPID` and `hostNetwork` on the device plugin DaemonSet pod spec are now values-configurable (defaults: `true` / `false` — unchanged from current behaviour)
- `securityContext` for the main `device-plugin` container is now values-configurable (default: `privileged: true`, `allowPrivilegeEscalation: true`, `SYS_ADMIN`)
- `securityContext` for the `vgpu-monitor` container is now values-configurable (default: `allowPrivilegeEscalation: false`, `SYS_ADMIN`)
- `gpuOperatorToolkitReady.securityContext` for the toolkit-validation init container is now values-configurable (default: `privileged: true`, `runAsUser: 0`)

All defaults preserve the existing hardcoded values — this is a non-breaking change.

## Motivation

Users deploying HAMi in environments with custom Pod Security Admission policies or restricted node configurations need to be able to tune these settings without patching the chart templates.

Through testing we found that not all security defaults — such as `privileged: true` and `hostPID: true` — are required for every use case. Making these configurable allows users to apply a least-privilege posture where their environment permits.

Closes #1888
Related: #1143

> AI assistance used in drafting this PR.

## Summary by CodeRabbit

* **New Features**
  * Added new configuration options to control whether the device plugin uses host PID and host networking.
  * Introduced configurable security settings for the device plugin, its monitoring component, and the toolkit validation step.

* **Bug Fixes**
  * Replaced previously hardcoded privilege and container security settings with deployment-driven controls, improving compatibility and flexibility across environments.

### DIFF
--- a/charts/hami/templates/device-plugin/daemonsetnvidia.yaml
+++ b/charts/hami/templates/device-plugin/daemonsetnvidia.yaml
@@ -40,7 +40,7 @@ spec:
       {{- end }}
       serviceAccountName: {{ include "hami-vgpu.device-plugin" . }}
       priorityClassName: system-node-critical
-      hostPID: {{ .Values.devicePlugin.hostPID | default true }}
+      hostPID: {{ .Values.devicePlugin.hostPID }}
       hostNetwork: {{ .Values.devicePlugin.hostNetwork | default false }}
       {{- include "hami.devicePlugin.imagePullSecrets" . | nindent 6 }}
       {{- if .Values.devicePlugin.gpuOperatorToolkitReady.enabled }}
--- a/charts/hami/templates/device-plugin/daemonsetnvidia.yaml
+++ b/charts/hami/templates/device-plugin/daemonsetnvidia.yaml
@@ -40,17 +40,18 @@ spec:
       {{- end }}
       serviceAccountName: {{ include "hami-vgpu.device-plugin" . }}
       priorityClassName: system-node-critical
-      hostPID: true
-      hostNetwork: false
+      hostPID: {{ .Values.devicePlugin.hostPID }}
+      hostNetwork: {{ .Values.devicePlugin.hostNetwork }}
       {{- include "hami.devicePlugin.imagePullSecrets" . | nindent 6 }}
       {{- if .Values.devicePlugin.gpuOperatorToolkitReady.enabled }}
       initContainers:
         - name: toolkit-validation
           image: {{ include "hami.devicePlugin.image" . }}
           imagePullPolicy: {{ .Values.devicePlugin.image.pullPolicy }}
+          {{- with .Values.devicePlugin.initContainerSecurityContext }}
           securityContext:
-            privileged: true
-            runAsUser: 0
+            {{- toYaml . | nindent 12 }}
+          {{- end }}
           command: ["sh", "-c"]
           args:
             - |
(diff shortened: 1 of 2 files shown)


mesutoezdil (commented):
defaults are preserved correctly, all ci passes, and the {{- with }} pattern for securityContext is the right call. one issue with hostPID and hostNetwork.

charts/hami/templates/device-plugin/daemonsetnvidia.yaml
```diff
@@ -40,17 +40,18 @@ spec:
       {{- end }}
       serviceAccountName: {{ include "hami-vgpu.device-plugin" . }}
       priorityClassName: system-node-critical
```
mesutoezdil, line 42: `hostPID: {{ .Values.devicePlugin.hostPID }}` renders as `hostPID: ` (invalid yaml) if a user explicitly sets `devicePlugin.hostPID: null` in their values override. same for hostNetwork. the securityContext blocks use `{{- with }}` which handles null cleanly. hostPID and hostNetwork should do the same:

```yaml
hostPID: {{ .Values.devicePlugin.hostPID | default true }}
hostNetwork: {{ .Values.devicePlugin.hostNetwork | default false }}
```

charts/hami/templates/device-plugin/daemonsetnvidia.yaml
```diff
@@ -40,17 +40,18 @@ spec:
...
-      hostPID: true
-      hostNetwork: false
+      hostPID: {{ .Values.devicePlugin.hostPID | default true }}
```
mesutoezdil: `false | default true` = `true` in Sprig. Setting `devicePlugin.hostPID: false` always renders `hostPID: true`, the override does nothing. Drop the `| default true`, values.yaml already sets the default.

mesutoezdil:
The PR says Closes #1888 but 2 things from that issue are not working:

- Issue specifies `devicePlugin.initContainerSecurityContext` but the PR puts it under `devicePlugin.gpuOperatorToolkitReady.securityContext`. 
The path does not match.
- `hostPID: false` override does not work due to the boolean default bug mentioned above.

Both need to be fixed before this can actually close #1888.

ilia-medvedev:
@mesutoezdil I've fixed the hostPID . as for the initContainer- it's cleaner to use the already existing parent value for it so using `gpuOperatorToolkitReady.securityContext` is the correct choice. I've adjusted the issue and the PR comment accordingly.

archlitchi (approved):
/lgtm

Issue #1896, opened 2026-05-22
scheduler: quota usage may not be rolled back when PatchPodAnnotations fails in Filter


bolin-dai:
**What happened**:

I noticed a possible local state consistency issue in `Scheduler.Filter()`.

In `pkg/scheduler/scheduler.go`, `Filter()` adds the pod allocation into `podManager` and increases quota usage before patching pod annotations.

Current flow is roughly:

```go
if s.podManager.AddPod(args.Pod, m.NodeID, m.Devices) {
    s.quotaManager.AddUsage(args.Pod, m.Devices)
}

err = util.PatchPodAnnotations(args.Pod, annotations)
if err != nil {
    s.recordScheduleFilterResultEvent(args.Pod, EventReasonFilteringFailed, "", err)
    s.podManager.DelPod(args.Pod)
    return nil, err
}
```

If `util.PatchPodAnnotations(args.Pod, annotations)` fails, HAMi removes the pod from `podManager` by calling:

```go
s.podManager.DelPod(args.Pod)
```

However, the quota usage added by:

```go
s.quotaManager.AddUsage(args.Pod, m.Devices)
```

does not seem to be rolled back in this error path.

This may cause `quotaManager` to keep stale usage even though the allocation result was not successfully persisted to the Kubernetes API Server through pod annotations.

As a result, HAMi's local quota usage may become larger than the real allocation state, and later pods in the same namespace may be incorrectly rejected by quota checks.

**What you expected to happen**:

If quota usage has been added before `PatchPodAnnotations()` fails, the quota usage should be rolled back together with the `podManager` cleanup.

A possible fix could be:

```go
added := s.podManager.AddPod(args.Pod, m.NodeID, m.Devices)
if added {
    s.quotaManager.AddUsage(args.Pod, m.Devices)
}

err = util.PatchPodAnnotations(args.Pod, annotations)
if err != nil {
    s.recordScheduleFilterResultEvent(args.Pod, EventReasonFilteringFailed, "", err)
    if added {
        s.quotaManager.RmUsage(args.Pod, m.Devices)
    }
    s.podManager.DelPod(args.Pod)
    return nil, err
}
```

The `added` flag is used because `AddUsage()` is only called when `podManager.AddPod()` returns true. Therefore, `RmUsage()` should also only be called when `AddUsage()` has actually been executed.

This also seems consistent with other cleanup paths, such as pod deletion handling, where quota usage is removed before deleting the pod from `podManager`.

[...]

PR #1898, opened 2026-05-22
fix: rollback quota usage when filter annotation patch fails

bolin-dai:
**What type of PR is this?**

/kind bug

This PR attempts to fix a possible local quota usage consistency issue in `Scheduler.Filter()`.

Currently, `Filter()` calls `podManager.AddPod()` and `quotaManager.AddUsage()` before patching pod annotations. If `PatchPodAnnotations()` fails, HAMi removes the pod from `podManager`, but the quota usage added before the patch is not rolled back.

This PR stores the return value of `podManager.AddPod()` in an `added` variable. If `added` is true and `PatchPodAnnotations()` fails, it calls `quotaManager.RmUsage()` before deleting the pod from `podManager`.

This keeps `AddUsage()` and `RmUsage()` symmetric in the patch failure path.

Related to #1896

The `added` flag is used because `quotaManager.AddUsage()` is only called when `podManager.AddPod()` returns true. Therefore, `quotaManager.RmUsage()` should also only be called when `AddUsage()` has actually been executed.

I am not sure whether there is another later reconciliation path that corrects this quota usage automatically. If such a mechanism exists, please correct me.

```release-note
NONE

### DIFF
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -780,14 +780,19 @@ func (s *Scheduler) Filter(args extenderv1.ExtenderArgs) (*extenderv1.ExtenderFi
 		val.PatchAnnotations(args.Pod, &annotations, m.Devices)
 	}
 
-	if s.podManager.AddPod(args.Pod, m.NodeID, m.Devices) {
-		s.quotaManager.AddUsage(args.Pod, m.Devices)
+	added := s.podManager.AddPod(args.Pod, m.NodeID, m.Devices)
+	if added {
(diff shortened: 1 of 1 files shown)


DSFans2014 (commented):
/lgtm

archlitchi:
please fix the CI-lint

bolin-dai:
Hi maintainers, I have fixed the gofmt issue in the latest commit.
Could you please approve the pending workflow run when you have time? Thanks.

Shouren (approved):
/lgtm

PR #1900, opened 2026-05-24
fix: prevent scheduler from dropping GPU nodes after 60s handshake piry

### LINKED ISSUE #1899: Scheduler health check incorrectly marks GPU nodes as unhealthy after 60s, removing them from cache
**What happened**:

When running HAMi scheduler with multiple replicas, GPU nodes are incorrectly marked as unhealthy after ~60 seconds. The scheduler's CheckHealth() function in pkg/device/devices.go uses the hami.io/node-handshake annotation with a 60-second expiry window. Once the annotation timestamp expires, the scheduler removes nodes from its internal cache via rmNodeDevices(), which causes:

GPU nodes removed from scheduling (no new GPU pods can be scheduled)
hami_gpu_memory_limit_bytes and other node-level metrics drop to 0
The overviewstatus map becomes empty because addNode() is never called after initial registration
The root cause is in the register() loop in pkg/scheduler/scheduler.go. After the initial Requesting_<timestamp> is written and 60s passes, CheckHealth() returns (true, false) (healthy but needUpdate=false) if Allocatable has GPU resources, so addNode() is skipped. Eventually the node cache entry gets stale or cleaned, and since addNode() is never called again, the node disappears from metrics and scheduling.

GPU nodes should remain healthy and registered in the scheduler cache indefinitely, as long as the device-plugin is running and reporting devices. The health check mechanism should not create a 60-second TTL that causes nodes to be silently removed.

Deploy HAMi 2.9.0 with scheduler replicas: 3 and leaderElect: true
Have GPU nodes with NVIDIA devices registered via device-plugin
Wait > 60 seconds after scheduler starts
Observe scheduler logs: no "Node device added" messages after initial registration
Check metrics endpoint: curl http://<scheduler-svc>:9395/metrics | grep hami_gpu_memory_limit_bytes returns 0 results
In logs you may see "Total memory for device" totalMemory=0 for all devices
Workaround: A CronJob running every minute that sets hami.io/node-handshake=Refresh on GPU nodes. This forces the else branch in CheckHealth() which returns (true, true), triggering addNode() and re-populating the cache.

The issue is in the interaction between CheckHealth() and register():
```shell
//pkg/device/devices.go - CheckHealth()
handshake := node.Annotations[util.HandshakeAnnos[devType]]
if strings.Contains(handshake, "Requesting") {
    formertime, _ := time.ParseInLocation(time.DateTime, strings.Split(handshake, "_")[1], time.Local)
    if time.Now().Before(formertime.Add(time.Second * 60)) {
        return true, false  // healthy, NO update needed
    }
    qty := node.Status.Allocatable[corev1.ResourceName(resourceCountName)]
    if qty.Value() > 0 {
        return true, false  // healthy, NO update needed <-- THIS PATH
    }
    return false, false  // unhealthy
}
``` 
```shell
// pkg/scheduler/scheduler.go - register()
health, needUpdate := devInstance.CheckHealth(devhandsk, val)
if !health { s.rmNodeDevices(val.Name, devhandsk); continue }
if !needUpdate { continue }  // <-- skips addNode() forever after 60s
s.addNode(val.Name, nodeInfo)  // never reached again
``` 

Once 60s passes and Allocatable has GPU resources, the function always returns (true, false) → needUpdate=false → addNode() is never called again → cache eventually becomes empty → metrics stop.

**Environment**:

HAMi version: 2.9.0
nvidia driver version: 570.x
Kubernetes version: 1.31
Scheduler replicas: 3 (with leaderElect: true)
GPU: NVIDIA L40 (46068 MiB) × 4 per node, 4 nodes
Others: Issue also affects single-replica deployments but is more visible with multiple replicas

yxxhero:
When the Requesting_<ts> handshake annotation expires (>60s old) but the node still has allocatable GPU resources, CheckHealth() now refreshes the handshake and returns (true, true) instead of (true, false). This ensures the scheduler's register() loop calls addNode() periodically to keep the node cache fresh.

Previously, after 60s, CheckHealth() returned (healthy, needUpdate=false) causing addNode() to be skipped forever. The cache went stale, metrics dropped to zero, and no new GPU pods could be scheduled.

Also refactored for clarity:
- Extracted refreshHandshake() helper (deduplicating GetNode+PatchNodeAnnotations)
- Extracted handshakeExpiry constant
- Fixed formerTime naming inconsistency

Fixes: #1899

**What type of PR is this?**
/kind bug

Fixes #1899

### DIFF
--- a/pkg/device/devices.go
+++ b/pkg/device/devices.go
@@ -162,6 +162,8 @@ const (
 
 	// OnePodMultiContainerSplitSymbol this is when one pod having multi container and more than one container use device, use ; symbol to join device info.
 	OnePodMultiContainerSplitSymbol = ";"
+
+	handshakeExpiry = time.Second * 60
 )
 
 var (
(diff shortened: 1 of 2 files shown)


archlitchi:
This issue should been fixed in https://github.com/Project-HAMi/HAMi/pull/1879, are you still encountering this issue?

PR #1904, opened 2026-05-25
fix(quota): enforce explicit zero limit in ResourceQuota

iasthc:
**What type of PR is this?**

/kind bug

QuotaManager used Limit != 0 to decide whether a quota was configured. As a result, setting limits.nvidia.com/gpumem: "0" (or gpucores: "0") was silently treated as "no limit" and every pod was allowed through, instead of blocking GPU usage in the namespace as the docs describe.

This PR adds a LimitSet flag on Quota to distinguish an explicitly-set zero from an auto-initialized entry. FitQuota now gates on LimitSet instead of Limit != 0, so an explicit zero is honored as a hard block.

Fixes #

- AddQuota sets LimitSet = true; DelQuota clears it.
- Entries auto-created by AddUsage keep LimitSet = false, so behavior when no quota is configured is unchanged.
- Existing NVIDIA-only / nvidia.com/gpu == 1 limitations in the webhook are untouched.
- New tests: TestFitQuotaZeroLimitEnforced, TestAddQuotaExplicitZero. Existing fixtures updated.

Honor an explicit zero in `limits.nvidia.com/gpumem` / `limits.nvidia.com/gpucores` ResourceQuota as a hard block instead of silently treating it as "no limit".

### DIFF
--- a/cmd/scheduler/metrics.go
+++ b/cmd/scheduler/metrics.go
@@ -298,18 +298,22 @@ func (cc ClusterManagerCollector) Collect(ch chan<- prometheus.Metric) {
 	)
 	for ns, val := range sher.GetQuotaManager().GetResourceQuota() {
 		for quotaname, q := range *val {
+			limitLabel := "unlimited"
+			if q.LimitSet {
+				limitLabel = fmt.Sprint(q.Limit)
+			}
 			ch <- prometheus.MustNewConstMetric(
 				quotaUsedDesc,
 				prometheus.GaugeValue,
 				float64(q.Used),
-				ns, quotaname, fmt.Sprint(q.Limit),
+				ns, quotaname, limitLabel,
 			)
 			if legacy {
 				ch <- prometheus.MustNewConstMetric(
 					legacyQuotaUsed,
 					prometheus.GaugeValue,
 					float64(q.Used),
-					ns, quotaname, fmt.Sprint(q.Limit),
+					ns, quotaname, limitLabel,
 				)
 			}
 		}
--- a/pkg/device/quota.go
+++ b/pkg/device/quota.go
@@ -25,8 +25,9 @@ import (
 )
 
 type Quota struct {
-	Used  int64
-	Limit int64
+	Used     int64
+	Limit    int64
+	LimitSet bool
 }
 
 type DeviceQuota map[string]*Quota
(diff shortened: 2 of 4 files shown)


iasthc:
Hi @DSFans2014, this fixes a bug where `limits.nvidia.com/gpumem: "0"` in ResourceQuota was silently ignored. Would appreciate your review when you have time, thanks!

pkg/device/quota.go
```diff
@@ -25,8 +25,9 @@ import (
 )
 
 type Quota struct {
-	Used  int64
-	Limit int64
+	Used     int64
+	Limit    int64
+	LimitSet bool
```
DSFans2014, line 30: I think we can set the default value of limit to MaxInt64 to avoid introducing a new variable. The information from ResourceQuota will be used in metrics as below.
https://github.com/Project-HAMi/HAMi/blob/f852784dc1970119b04333d8019b81f77827d29f/cmd/scheduler/metrics.go#L299-L315

iasthc: Thanks for the suggestion. I think `LimitSet` is still the safer choice here:

1. **`FitQuota` would overflow.** [pkg/device/quota.go#L75-L78](https://github.com/Project-HAMi/HAMi/blob/master/pkg/device/quota.go#L75-L78) multiplies `Limit` by `memoryFactor`. With `Limit == MaxInt64` and `memoryFactor > 1` this overflows to a negative number, and legitimate requests get rejected. Guarding against it requires an `if Limit == MaxInt64` check — another sentinel flag in disguise.

2. **Metric label.** Fair point that the rendering needs work, but `limit="9223372036854775807"` isn't friendlier than `"0"`. I've pushed a follow-up that uses `LimitSet` to render `"unlimited"`. This case is real today: `AddUsage` creates `LimitSet=false` entries for namespaces with usage but no `ResourceQuota`, and #1898 / #1905 exercise those paths more often now.


3. **Sentinel semantics would spread.** `AddUsage` and `DelQuota` would both need to use MaxInt64, and every reader has to remember `0` and `MaxInt64` carry special meaning. A bool is one byte and explicit.

The metrics commit is on top of the quota fixes, rebased onto current master — PTAL.

cmd/scheduler/metrics.go
```diff
@@ -298,18 +298,22 @@ func (cc ClusterManagerCollector) Collect(ch chan<- prometheus.Metric) {
 	)
 	for ns, val := range sher.GetQuotaManager().GetResourceQuota() {
 		for quotaname, q := range *val {
+			limitLabel := "unlimited"
```
mesutoezdil, line 301: this changes the prometheus label value from "0" to "unlimited" for namespaces that have usage but no configured quota (LimitSet=false). existing dashboards or alerts that filter on limit="0" for the no-quota case will break silently. the new label is semantically correct, but the change should be called out explicitly for operators.

mesutoezdil (commented):
1 more thing worth flagging: 

the metrics label changes from a numeric string to "unlimited" for entries where no quota is configured (LimitSet=false)

and before this pr those entries showed limit="0" in prometheus. and after this pr they show limit="unlimited". 

any existing dashboard or alert filtering on limit="0" for the unset case will silently break.. this is technically a more correct label, but it is a label value change and it affects cardinality. 
worth adding a note in the pr description or release notes so operators know to update their dashboards.

mesutoezdil:

PR #1921, opened 2026-06-03
fix: replace --build-arg TARGET_ARCH with --platform in docker build

haitwang-cloud:
**What type of PR is this?**

/kind bug
TARGET_ARCH was passed as a build argument but never declared in any Dockerfile, making it a no-op. Replace with --platform so Docker actually respects the target architecture.

Also update TARGET_ARCH from 'amd64' to 'linux/amd64' to match the os/arch format required by --platform.
Fixes #


Makefile
```diff
@@ -33,8 +33,8 @@ docker:
 dockerwithlib:
 	docker build \
 	--no-cache \
+	--platform ${TARGET_OS}/${TARGET_ARCH} \
```
Shouren: @haitwang-cloud Can you replace `${TARGET_OS}/${TARGET_ARCH}` with `TARGET_PLATFORMS`? The `--platform` flag accepts options like `linux/amd64,linux/arm64`, however we can't setting the flags with `${TARGET_OS}/${TARGET_ARCH}` in Makefile

haitwang-cloud:
Done, @Shouren PTAL again, Users can now override at build time with `make docker TARGET_PLATFORMS=linux/amd64,linux/arm64 `for multi-arch builds.

Shouren (approved):
/lgtm

PR #1927, opened 2026-06-04
fix(scheduler): clean up podManager on Bind failures to prevent resource leak

peachest:
**What type of PR is this?**

/kind bug

When the HAMi scheduler acts as a kube-scheduler extender, the `Filter` function adds the pod's resource allocation to `podManager`. Later, `Bind` is called by the scheduler framework to perform the actual binding. However, if `Bind` fails (e.g., the pod was deleted externally between `Filter` and `Bind`), the old code returned immediately without cleaning up the `podManager` entry, causing a **permanent resource leak**.

The leaked entry makes `getNodesUsage()` permanently overcount allocated resources on the affected node, which blocks future pods from being scheduled there — even though the pod itself no longer exists.

This fix ensures every error path in `Bind` cleans up the `podManager` entry:

1. **`podLister.Get` failure**: the pod no longer exists in the API server → delete from `podManager` using `args.PodUID` (the UID key that `Filter` wrote).
2. **`nodeLister.Get` failure**: the target node was removed or not found → delete from `podManager`.
3. **`ReleaseNodeLocks` (catch-all)**: covers `LockNode` failure, `PatchPodAnnotations` failure, and `Bind` API failure — all now pass through a unified cleanup label that deletes from `podManager`.

N/A — no issue filed yet. The bug was identified from production log analysis.

original log 

The bug was identified from production log analysis of a `csm-instance` pod that experienced three consecutive scheduling failures due to this exact leak. A detailed diagnosis report is available at . 

[诊断报告-csm-instance资源泄漏.html](https://github.com/user-attachments/files/28591286/-csm-instance.html)

Key logs after cleaning
[focus.log](https://github.com/user-attachments/files/28592680/focus.log)

Logs after restarting the scheduler
[new-focus.log](https://github.com/user-attachments/files/28592698/new-focus.log)

The root cause involves two interacting factors:
- **Control flow bug (primary)**: `Bind` returned on failure without `podManager` cleanup — a clear missing `defer`/cleanup pattern.
- **Informer unreliability (secondary)**: Even if `onDelPod` had fired (it didn't in this case because the Pod was never persisted in the API Server), a concurrent informer `onDelPod` + extender `Filter/addPod` race can still produce a window where the leak occurs. Therefore the fix in `Bind` is a necessary safety net regardless.

Two test cases are added to `scheduler_test.go` to cover the Get-pod-failure and Get-node-failure error paths, including proper `eventRecorder` initialization (which `NewScheduler()` does not set by default).

No. This is a bug fix — it does not add or change any user-facing API, configuration, or behavior. It prevents an internal resource leak that could otherwise manifest as pods stuck in `Pending` state with `no enough resource` errors on nodes that still have free capacity.

### DIFF
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -667,26 +667,37 @@ func (s *Scheduler) Bind(args extenderv1.ExtenderBindingArgs) (*extenderv1.Exten
 		Target:     corev1.ObjectReference{Kind: "Node", Name: args.Node},
 	}
 
+	// Clean up podManager and quotaManager entries added by Filter, on Bind failure.
+	// The pod may have been deleted between Filter and Bind; without
+	// this cleanup both internal bookkeepers leak permanently.
+	var podToDelete *corev1.Pod
+	defer func() {
+		if podToDelete == nil {
+			return
+		}
+		if pi, ok := s.podManager.TakeAndDeletePod(podToDelete); ok {
+			if len(pi.Devices) > 0 {
+				s.quotaManager.RmUsage(podToDelete, pi.Devices)
+			}
+		}
+	}()
+
 	current, err := s.podLister.Pods(args.PodNamespace).Get(args.PodName)
 	if err != nil {
 		klog.ErrorS(err, "Failed to get pod from cache", "pod", args.PodName, "namespace", args.PodNamespace)
-		// Clean up podManager and quotaManager entries added by Filter.
-		// The pod may have been deleted between Filter and Bind; without
-		// this cleanup both internal bookkeepers leak permanently.
-		podToDelete := &corev1.Pod{
+		// current is nil, construct a minimal pod object from args for
+		// TakeAndDeletePod to look up the podManager entry by UID.
+		podToDelete = &corev1.Pod{
 			ObjectMeta: metav1.ObjectMeta{
 				UID:       args.PodUID,
 				Name:      args.PodName,
 				Namespace: args.PodNamespace,
 			},
 		}
-		if podInfosToDelete, ok := s.podManager.TakeAndDeletePod(podToDelete); ok {
-			if len(podInfosToDelete.Devices) > 0 {
-				s.quotaManager.RmUsage(podToDelete, podInfosToDelete.Devices)
-			}
-		}
 		return &extenderv1.ExtenderBindingResult{Error: err.Error()}, err
 	}
+
+	podToDelete = current
 
 	klog.InfoS("Trying to get the target node for pod", "pod", args.PodName, "namespace", args.PodNamespace, "node", args.Node)
 
@@ -694,11 +705,6 @@ func (s *Scheduler) Bind(args extenderv1.ExtenderBindingArgs) (*extenderv1.Exten
 	if err != nil {
 		klog.ErrorS(err, "Failed to get node from cache", "node", args.Node)
 		s.recordScheduleBindingResultEvent(current, EventReasonBindingFailed, []string{}, fmt.Errorf("failed to get node %s", args.Node))
-		if podInfosToDelete, ok := s.podManager.TakeAndDeletePod(current); ok {
-			if len(podInfosToDelete.Devices) > 0 {
-				s.quotaManager.RmUsage(current, podInfosToDelete.Devices)
-			}
-		}
 		res = &extenderv1.ExtenderBindingResult{Error: err.Error()}
 		return res, nil
 	}
@@ -728,6 +734,7 @@ func (s *Scheduler) Bind(args extenderv1.ExtenderBindingArgs) (*extenderv1.Exten
 		goto ReleaseNodeLocks
 	}
 
+	podToDelete = nil
 	s.recordScheduleBindingResultEvent(current, EventReasonBindingSucceed, []string{args.Node}, nil)
 	klog.InfoS("Successfully bound pod to node", "pod", args.PodName, "namespace", args.PodNamespace, "node", args.Node)
 	return &extenderv1.ExtenderBindingResult{Error: ""}, nil
@@ -736,11 +743,6 @@ ReleaseNodeLocks:
 	klog.InfoS("Release node locks", "node", args.Node)
 	for _, val := range device.GetDevices() {
 		val.ReleaseNodeLock(node, current)
-	}
-	if podInfosToDelete, ok := s.podManager.TakeAndDeletePod(current); ok {
-		if len(podInfosToDelete.Devices) > 0 {
-			s.quotaManager.RmUsage(current, podInfosToDelete.Devices)
-		}
 	}
 	s.recordScheduleBindingResultEvent(current, EventReasonBindingFailed, []string{}, err)
 	return &extenderv1.ExtenderBindingResult{Error: err.Error()}, nil
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -658,6 +658,12 @@ func (s *Scheduler) getPodUsage() (map[string]device.PodUseDeviceStat, error) {
 	return podUsageStat, nil
 }
 
+func cleanupStalePodAllocation(s *Scheduler, pod *corev1.Pod) {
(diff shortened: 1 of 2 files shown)


DSFans2014 (commented):
please take a look at the gemini's comments

pkg/scheduler/scheduler.go
```diff
@@ -717,6 +725,7 @@ ReleaseNodeLocks:
 	for _, val := range device.GetDevices() {
 		val.ReleaseNodeLock(node, current)
 	}
+	s.podManager.DelPod(current)
```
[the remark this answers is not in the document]
peachest: fixed

pkg/scheduler/scheduler.go
```diff
@@ -670,6 +670,13 @@ func (s *Scheduler) Bind(args extenderv1.ExtenderBindingArgs) (*extenderv1.Exten
 	current, err := s.podLister.Pods(args.PodNamespace).Get(args.PodName)
 	if err != nil {
 		klog.ErrorS(err, "Failed to get pod from cache", "pod", args.PodName, "namespace", args.PodNamespace)
+		// Clean up podManager entry added by Filter. The pod may have been
+		// deleted externally between Filter and Bind. Without this cleanup,
+		// the leaked entry causes getNodesUsage() to permanently overcount
+		// allocated resources, blocking future pod scheduling on the node.
+		s.podManager.DelPod(&corev1.Pod{
+			ObjectMeta: metav1.ObjectMeta{UID: args.PodUID},
+		})
```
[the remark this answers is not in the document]
peachest: fixed

pkg/scheduler/scheduler.go
```diff
@@ -679,6 +686,7 @@ func (s *Scheduler) Bind(args extenderv1.ExtenderBindingArgs) (*extenderv1.Exten
 	if err != nil {
 		klog.ErrorS(err, "Failed to get node from cache", "node", args.Node)
 		s.recordScheduleBindingResultEvent(current, EventReasonBindingFailed, []string{}, fmt.Errorf("failed to get node %s", args.Node))
+		s.podManager.DelPod(current)
```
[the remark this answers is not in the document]
peachest: fixed

pkg/scheduler/scheduler.go
```diff
@@ -679,6 +694,11 @@ func (s *Scheduler) Bind(args extenderv1.ExtenderBindingArgs) (*extenderv1.Exten
 	if err != nil {
 		klog.ErrorS(err, "Failed to get node from cache", "node", args.Node)
 		s.recordScheduleBindingResultEvent(current, EventReasonBindingFailed, []string{}, fmt.Errorf("failed to get node %s", args.Node))
+		if podInfosToDelete, ok := s.podManager.TakeAndDeletePod(current); ok {
```
dl239: it's better to clean the resource in the defer function?

peachest: please check my new implementation with defer. I try my best to keep code clean but I think this may probably make things more difficult to understand. Maybe we can discuss about this further.

DSFans2014: Using `err != nil` to decide whether to execute cleanup looks more readable than using `podToDelete != nil` ?
Moreover, I think the `ReleaseNodeLocks` block can also be moved to the defer function together as a refactor task to do later.

pkg/scheduler/scheduler.go
```diff
@@ -667,12 +667,38 @@ func (s *Scheduler) Bind(args extenderv1.ExtenderBindingArgs) (*extenderv1.Exten
...
+			if len(pi.Devices) > 0 {
+				s.quotaManager.RmUsage(podToDelete, pi.Devices)
+			}
+		}
+	}()
+
 	current, err := s.podLister.Pods(args.PodNamespace).Get(args.PodName)
 	if err != nil {
 		klog.ErrorS(err, "Failed to get pod from cache", "pod", args.PodName, "namespace", args.PodNamespace)
+		// current is nil, construct a minimal pod object from args for
```
archlitchi: please remove these AI-comments

pkg/scheduler/scheduler_test.go
```diff
@@ -1490,3 +1493,120 @@ func Test_onAddPod_BadDeviceAnnotation(t *testing.T) {
...
+	podUID := types.UID("test-uid-1")
+	pod := &corev1.Pod{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "test-pod",
+			Namespace: "default",
+			UID:       podUID,
+		},
+	}
+
+	// Create scheduler with minimal setup — no device initialization needed
```
archlitchi: please remove these AI-comments, and similar comments in this file

pkg/scheduler/scheduler_test.go
```diff
@@ -1490,3 +1493,120 @@ func Test_onAddPod_BadDeviceAnnotation(t *testing.T) {
...
+	// Verify podManager entry was cleaned up
+	podsAfter, _ := s.podManager.ListPodsUID()
+	require.Len(t, podsAfter, 0,
+		"Bind should clean up podManager entry when Get(pod) fails, "+
+			"otherwise Filter's addPod leaks resources permanently")
+}
+
+// Test_Bind_DelPodOnGetNodeFailure verifies that when Bind's nodeLister.Get
+// returns "not found", the podManager entry added by Filter is cleaned up.
+// This ensures every error path in Bind cleans up its podManager entry.
```
archlitchi: If you're adding a comment here, you should also add one in method 'Test_Bind_DelPodOnGetPodFailure'

peachest:
The previous implementation (defer-based approach) cleaned up `podManager` and `quotaManager` entries on **every** Bind failure path. This turned out to be too aggressive — it breaks the resource reservation semantics that the original code intentionally implements.

## The Resource Reservation Logic

After Filter succeeds, `podManager.AddPod` records the allocated device information in `podManager`, while `quotaManager.AddUsage` tracks namespace-level quota consumption. This provides resource reservation:

- `getNodesUsage` iterates over `podManager` entries and deducts allocated devices from the node's available resources
- When other pods are being scheduled, these resources are already "in use" and won't be double-allocated

If Bind fails for a **retryable** reason (LockNode contention, annotation patch failure, Bind API timeout), kube-scheduler will retry scheduling the pod. Filter will handle cleanup of the old reservation via `TakeAndDeletePod`, then re-reserve on the new node via `AddPod`. Cleaning up early in Bind would create a window with no reservation, allowing other pods to steal the resources.

## The Correct Strategy

Clean up only when **the resource reservation target itself no longer exists**:

| Failure path                | Clean up? | Rationale                                                                                                                                                                                                                 |
| --------------------------- | --------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| `podLister.Get` fails       | **Yes**   | Pod was deleted externally between Filter and Bind. The informer's `onDelPod` won't fire (the pod was never persisted). Without cleanup this is a permanent leak.                                                         |
| `nodeLister.Get` fails      | **Yes**   | Target node no longer exists. `PodInfo.NodeID` is stale. |
| `LockNode` fails            | **No**    | Both pod and node still exist. kube-scheduler will retry, and Filter's `TakeAndDeletePod` will naturally clean the old reservation. Resource reservation semantics preserved.                                             |
| `PatchPodAnnotations` fails | **No**    | Same as above.                                                                                                                                                                                                            |
| `Bind` API fails            | **No**    | Same as above.                                                                                                                                                                                                            |

## Implementation

- Removed the global `defer` + `podToDelete` flag approach
- Added a `cleanupStalePodAllocation` helper function, called only at the two specific failure points
- Success path no longer sets `podToDelete = nil` (removed entirely)

@dl239  @archlitchi @DSFans2014 @claude Please check the new implementation

pkg/scheduler/scheduler_test.go
peachest: removed

pkg/scheduler/scheduler_test.go
peachest: removed

pkg/scheduler/scheduler.go
peachest: removed

archlitchi (approved):
/lgtm

Issue #1929, opened 2026-06-06
Crash when `nvidia-mig-parted` is missing, even with `--mig-strategy=none`


maishivamhoo123:
What happened?
When deploying HAMi in an environment that does not have the nvidia-mig-parted binary (such as a mock‑GPU cluster using nvml-mock), the device plugin crashes immediately with:
This happens even though --mig-strategy is set to "none"
<img width="1423" height="332" alt="Image" src="https://github.com/user-attachments/assets/8b395d89-4b9a-4b11-a0c6-47bf21c6d9f8" />
The Crashing :- 

<img width="1877" height="768" alt="Image" src="https://github.com/user-attachments/assets/d2368af4-5058-4c78-8bf4-41fe5e06ab69" />

What did you expect to happen?
The plugin should start successfully without calling nvidia-mig-parted when MIG is explicitly disabled. It should discover GPUs and register them as usual.
Why does this happen?
In the Start() function of the device plugin (pkg/device-plugin/nvidiadevice/nvinternal/plugin/server.go), the code unconditionally runs:

```
if deviceSupportMig {
    cmd := exec.Command("nvidia-mig-parted", "export")
    ...
}
```
And later:

```
if deviceSupportMig {
    plugin.ApplyMigTemplate()
}
```
deviceSupportMig becomes true if any detected GPU model matches a built‑in MIG template (e.g., A100, H100).
Even with --mig-strategy=none, the binary is called, and if it is missing (or fails), the plugin crashes.

Proposed Fix
In server.go, wrap the nvidia-mig-parted execution and the ApplyMigTemplate() call with a check on the migStrategy flag.

pkg/device-plugin/nvidiadevice/nvinternal/plugin/server.go
Change 1 – Guard the nvidia-mig-parted execution:
Inside the if deviceSupportMig { … } block, add a condition:
```
if deviceSupportMig {
    migEnabled := plugin.config.Flags.MigStrategy != nil && *plugin.config.Flags.MigStrategy != "none"
    if migEnabled {
        cmd := exec.Command("nvidia-mig-parted", "export")
        // ... existing code ...
    }

    // Build migCurrent as before (always needed)
    if plugin.operatingMode == "mig" {
        // ...
    } else {
        // ...
        klog.Infoln("Close Mig export", plugin.migCurrent)
    }
}
```
Change 2 – Guard ApplyMigTemplate():
```
if deviceSupportMig {
[...]

PR #1930, opened 2026-06-06
Fix: Skip nvidia-mig-parted when MIG is disabled

maishivamhoo123:
Close :- #1929 
### What this PR does
Prevents the device plugin from calling `nvidia-mig-parted` and
`ApplyMigTemplate()` when `--mig-strategy=none` (the default).

### Why it's needed
The plugin unconditionally executes the MIG partitioner binary whenever a
MIG‑capable GPU model (A100, H100, …) is detected, even with MIG explicitly
disabled. In environments without the binary – such as NVIDIA's `nvml-mock`
– this causes a fatal crash:

### DIFF
--- a/pkg/device-plugin/nvidiadevice/nvinternal/plugin/server.go
+++ b/pkg/device-plugin/nvidiadevice/nvinternal/plugin/server.go
@@ -1,18 +1,34 @@
 /*
-Copyright 2024 The HAMi Authors.
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The HAMi Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
 
-Licensed under the Apache License, Version 2.0 (the "License");
-you may not use this file except in compliance with the License.
-You may obtain a copy of the License at
+/*
+ * Licensed to NVIDIA CORPORATION under one or more contributor
+ * license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright
+ * ownership. NVIDIA CORPORATION licenses this file to you under
+ * the Apache License, Version 2.0 (the "License"); you may
+ * not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
 
-    http://www.apache.org/licenses/LICENSE-2.0
-
-Unless required by applicable law or agreed to in writing, software
-distributed under the License is distributed on an "AS IS" BASIS,
-WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-See the License for the specific language governing permissions and
-limitations under the License.
-*/
+/*
+ * Modifications Copyright The HAMi Authors. See
+ * GitHub history for details.
+ */
 
 package plugin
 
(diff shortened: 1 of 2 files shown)


maishivamhoo123:
image


Img 2 :-----  

Screenshot 2026-06-06 153106
Img 3 :--- 

image

maishivamhoo123:
@archlitchi  @FouoF  and @Shouren  and @team can you please review this PR?

archlitchi:
Actually, the mig-strategy is always 'none' regardless of whether this node uses 'mig' or 'hamicore', we should get 'mode' by using configs

maishivamhoo123:
> Actually, the mig-strategy is always 'none' regardless of whether this node uses 'mig' or 'hamicore', we should get 'mode' by using configs

@archlitchi  Thank you for your Review !   i have a small question Should i add both mig-strategy  and Operating Mode (Config file ) or just   Operating Mode (Config file )  because if some user try to do it using mig-strategy then it may fail
Because it worked in my case
image

archlitchi:
@maishivamhoo123  i think using Operating mode is sufficient, because mig-strategy for 'dynamic-mig' is 'none' as well.

pkg/device-plugin/nvidiadevice/nvinternal/plugin/server.go
```diff
@@ -234,21 +234,103 @@ func (plugin *NvidiaDevicePlugin) Devices() rm.Devices {
...
+		}
+		deviceMigSupportMap[i] = isSupported
+		if isSupported {
+			nodeHasMigCapableGPU = true
+		}
+	}
+	return deviceMigSupportMap, nodeHasMigCapableGPU
+}
+
+// buildHybridMigConfig - properly handles mixed MIG/non-MIG devices
```
archlitchi: We shouldn't consider the scenario which a node consists of both 'mig-support' devices and 'mig-non-support' devices, very hard to validate because we're hard to reach such an environment

archlitchi: 'mig' should be 'all' or 'none' for a GPU node. If some GPU are not 'mig-supported' GPU, we should return an error

pkg/device-plugin/nvidiadevice/nvinternal/plugin/server.go
```diff
@@ -234,21 +234,103 @@ func (plugin *NvidiaDevicePlugin) Devices() rm.Devices {
...
+		return fmt.Errorf("failed to get device numbers: %w", err)
 	}
 
 	deviceNames, err := GetDeviceNames()
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to get device names: %w", err)
+	}
+
+	// 2. Lock Management (before Serve and Register)
```
archlitchi: Why create these Locks for 'non-mig' nodes?

pkg/device-plugin/nvidiadevice/nvinternal/plugin/server.go
```diff
@@ -234,21 +234,103 @@ func (plugin *NvidiaDevicePlugin) Devices() rm.Devices {
...
+	if err = CreateMigApplyLockDir(); err != nil {
+		klog.Fatalf("CreateMIGLockSubDir failed: %v", err)
+	}
+	if err = RemoveMigApplyLock(); err != nil {
+		klog.Fatalf("RemoveMigApplyLock failed: %v", err)
 	}
 
+	// 3. Start serving
```
archlitchi: please remove these 'AI-generated' comments

pkg/device-plugin/nvidiadevice/nvinternal/plugin/server.go
maishivamhoo123: @archlitchi  Thank you for review.  Got it! I will remove the hybrid/mixed-device handling logic completely. The plugin now uses an all-or-nothing approach: if any device on the node doesn't support MIG geometries, the entire node safely falls back to a uniform non-MIG configuration.

pkg/device-plugin/nvidiadevice/nvinternal/plugin/server.go
maishivamhoo123: I initially included these because i was thinking it would safely handle baseline setup, but I realize now it's completely redundant for non-MIG nodes. I've moved the lock creation and removal inside the MIG-specific block. Thanks for pointing this out .

maishivamhoo123:
@archlitchi  i made all the suggested changes can you look it again.
Thank you!

archlitchi (approved):
/lgtm

PR #1934, opened 2026-06-08
fix(charts): change scheduler anti-affinity to preferred for rolling …

### LINKED ISSUE #1933: [Bug] Scheduler rolling update stuck in Pending on single-node clusters due to hard podAntiAffinity
**What happened**:
After deploying HAMi via Helm on a single-node Kubernetes cluster (or a cluster with only one schedulable node), when i edit scheduler yaml，the scheduler Deployment cannot complete a rolling update. The new scheduler pod remains in `Pending` state indefinitely because the existing pod already occupies the node, and the hard `podAntiAffinity` rule prevents co-location on the same node.

The detailed scheduling logs of the new pod are as follows:
Warning FailedScheduling 2m10s default-scheduler 0/1 nodes are available: 1 node(s) didn't match pod anti-affinity rules. preemption: 0/1 nodes are available: 1 node(s) didn't match pod anti-affinity rules.

Scheduler rolling update works correctly.

1. Install HAMi via Helm on a single-node Kubernetes cluster
2. Trigger a rolling update of the scheduler Deployment (e.g., upgrade Helm chart, modify image tag, or run `kubectl rollout restart deployment/...`)
3. Observe the new scheduler pod status

lvkaili:
## PR type
/kind/bug

## Problem
In single-node clusters, the scheduler's `podAntiAffinity` is set to `requiredDuringSchedulingIgnoredDuringExecution`, which prevents rolling updates. When updating the scheduler deployment, the new pod cannot be scheduled because the existing pod already occupies the node, causing a deadlock.

## Solution
Change the scheduler affinity configuration in Helm charts from hard anti-affinity to preferred (or remove the hard constraint), allowing the scheduler pod to be rescheduled on the same node during rolling updates.

## Changes
- Modified `charts/hami/templates/scheduler/deployment.yaml` (or relevant affinity config)

## Testing
- Verified rolling update succeeds on single-node cluster
- Multi-node clusters still benefit from preferred spread

Fixes #1933

### DIFF
--- a/charts/hami/templates/scheduler/deployment.yaml
+++ b/charts/hami/templates/scheduler/deployment.yaml
@@ -195,7 +195,7 @@ spec:
       {{- if .Values.scheduler.leaderElect }}
       affinity:
         podAntiAffinity:
-          requiredDuringSchedulingIgnoredDuringExecution:
+          preferredDuringSchedulingIgnoredDuringExecution:
           - labelSelector:
               matchExpressions:
               - key: app.kubernetes.io/component
--- a/charts/hami/templates/scheduler/deployment.yaml
+++ b/charts/hami/templates/scheduler/deployment.yaml
@@ -196,13 +196,15 @@ spec:
       affinity:
         podAntiAffinity:
           preferredDuringSchedulingIgnoredDuringExecution:
-          - labelSelector:
-              matchExpressions:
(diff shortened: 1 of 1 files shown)


chaunceyjiang (approved):
LGTM

Shouren (approved):
/lgtm

PR #1947, opened 2026-06-12
examples: add vLLM cross-vGPU tensor parallel deployment

haitwang-cloud:
**What type of PR is this?**

**Example includes**:
- Request 2 vGPUs with gpumem-percentage and gpucores resource limits
- --disable-custom-all-reduce to bypass CUDA IPC
- PVC for HuggingFace cache, /dev/shm via emptyDir, Service,
  health probes, and GPU toleration

**Problem**:

  Worker_TP0/TP1 (pid=642/643) CUDA graph profiling phase
    → custom_all_reduce.cuh:455 'invalid argument'        ← CUDA IPC failure
    → HAMI-core ERROR: cuCtxSetCurrent111 failed res=4    ← HAMi intercepts ctx switch
    → Worker proc VllmWorker-0 died unexpectedly
    → EngineCore failed to start
    → RuntimeError: Engine core initialization failed

  **Root cause**: -`-tensor-parallel-size `2 makes vLLM's two worker processes
  communicate via custom all-reduce (based on CUDA IPC shared memory).
  HAMi vGPU virtualizes GPU devices and intercepts cuCtxSetCurrent calls,
  causing the IPC mechanism to fail entirely.

Fixes #

### DIFF
--- a/examples/nvidia/vllm_cross_vgpu.yaml
+++ b/examples/nvidia/vllm_cross_vgpu.yaml
@@ -1,0 +1,102 @@
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: vllm-cross-vgpu-cache
+spec:
+  accessModes:
+    - ReadWriteOnce
+  resources:
+    requests:
+      storage: 60Gi
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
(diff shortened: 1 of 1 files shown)


archlitchi (approved):
/lgtm

PR #1957, opened 2026-06-17
fix_leaderNotify Block

### LINKED ISSUE #1954: hami-scheduler multi-copy deployment problem
**What happened**:
按官方发布的内容，2.8版本的hami-scheduler已支持多副本，实际验证时，在helm的values里将hami-scheduler的副本数改为2时，hami-scheduler不能调度了，报错如下。手动重启hami-scheduler后恢复正常，但验证hami-scheduler主备切换时，备pod升主后，仍然会复现该问题。

Image

hami-scheduler备pod升主后，可以正常调度AI pod。

**Environment**:
- HAMi version:v2.9.0
- Others:

archlitchi:
/kind bug

fix: https://github.com/Project-HAMi/HAMi/issues/1954

When --leader-elect is enabled and multiple scheduler replicas are running, the OnStartedLeading callback used a blocking channel send (s.leaderNotify <- struct{}{}). If the RegisterFromNodeAnnotations goroutine was busy executing register() at the moment of a leader transition, the channel buffer would be full and the send would block indefinitely, hanging the informer event handler goroutine.

With the informer goroutine stuck, no further Lease events could be processed. As a result, IsLeader() would eventually return false (due to isLeaseValid expiring without new events), causing every subsequent call to register() to be skipped. The synced flag would never be set back to true, making WaitForCacheSync block all incoming /filter requests until the kube-scheduler extender timeout fired. The only recovery was a manual pod restart.

微信图片_20260512155027_686_380
微信图片_20260512155506_689_380

Fix:

Changed OnStartedLeading to use a non-blocking send (the same select/default pattern already used by doNodeNotify). If the channel already holds a pending notification — meaning register() is about to run anyway — the new notification is safely dropped without blocking the informer goroutine.

### DIFF
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -100,7 +100,10 @@ func NewScheduler() *Scheduler {
 	if config.LeaderElect {
 		callbacks := leaderelection.LeaderCallbacks{
 			OnStartedLeading: func() {
-				s.leaderNotify <- struct{}{}
+				select {
+				case s.leaderNotify <- struct{}{}:
+				default:
+				}
 			},
 			OnStoppedLeading: func() {
 				s.lock.Lock()
--- a/pkg/util/leaderelection/leaderelection.go
+++ b/pkg/util/leaderelection/leaderelection.go
@@ -112,12 +112,17 @@ func (m *leaderManager) onAdd(obj any) {
 	}
 
 	m.leaseLock.Lock()
-	defer m.leaseLock.Unlock()
-
 	m.setObservedRecord(lease)
-	// Notify if we are the leader from the very begging
-	if m.isHolderOf(lease) && m.callbacks.OnStartedLeading != nil {
-		m.callbacks.OnStartedLeading()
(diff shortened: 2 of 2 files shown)


archlitchi:
CC @Shouren

wawa0210:
----    
cc @S descendants

pkg/scheduler/scheduler.go
```diff
@@ -100,7 +100,10 @@ func NewScheduler() *Scheduler {
 	if config.LeaderElect {
 		callbacks := leaderelection.LeaderCallbacks{
 			OnStartedLeading: func() {
-				s.leaderNotify <- struct{}{}
+				select {
+				case s.leaderNotify <- struct{}{}:
+				default:
```
Shouren, line 105: @archlitchi We can log some message here

Shouren (approved):
/lgtm

pkg/scheduler/scheduler.go
archlitchi: i'll add once i confirm this bug has been solved

PR #1965, opened 2026-06-18
fix(device): CheckUUID ignores noUseKey when useKey is also set

mesutoezdil:
when a pod sets both `use-uuid` and `nouse-uuid`, the `nouse-uuid` list is never checked.

the old code returns as soon as `use-uuid` matches, so `nouse-uuid` has no effect. a user who sets `use-uuid=A,B` and `nouse-uuid=B` still gets device B.

fix: check both annotations without early return. if the device is in the use list, continue and check the nouse list too.

two test cases added: device in both lists (should be rejected), device only in use list (should be allowed).


DSFans2014 (commented):
Under what circumstances would a card be set to both use and nouse?

Shouren (approved):
/lgtm

PR #1966, opened 2026-06-18
fix(quota): clamp Used to zero in RmUsage to prevent negative tracking

mesutoezdil:
`RmUsage` subtracts without clamping, so `Used` can go below zero.

`Used < 0` is invariably wrong: a negative counter makes `FitQuota` see more capacity than actually exists, allowing pods to be scheduled past the namespace quota limit.

all current `RmUsage` call sites are gated by `TakeAndDeletePod` or the `added` flag, so this cannot be triggered today. the clamp is a defensive invariant that prevents future regressions if an ungated call site is introduced later.

fix: clamp `Used` to zero after each subtraction.

test: call `RmUsage` twice on counters already at zero, verify `Used` stays at 0.

### DIFF
--- a/pkg/device/quota.go
+++ b/pkg/device/quota.go
@@ -157,6 +157,10 @@ func (q *QuotaManager) RmUsage(pod *corev1.Pod, podDev PodDevices) {
 	for idx, val := range usage {
 		if qInfo, ok := (*dp)[idx]; ok && qInfo != nil {
 			qInfo.Used -= val
+			if qInfo.Used < 0 {
+				klog.V(4).InfoS("RmUsage: clamping negative Used to zero", "quota", idx, "val", val)
+				qInfo.Used = 0
+			}
 		}
 	}
 	if klog.V(4).Enabled() {
--- a/pkg/device/quota_test.go
+++ b/pkg/device/quota_test.go
@@ -198,6 +198,15 @@ func TestAddUsageAndRmUsage(t *testing.T) {
 	if (*qm.Quotas[ns])[coreName].Used != 0 {
 		t.Errorf("RmUsage: expected Used core 0, got %d", (*qm.Quotas[ns])[coreName].Used)
 	}
+
+	// remove more than tracked: Used must not go below zero
+	qm.RmUsage(pod, podDev)
+	if (*qm.Quotas[ns])[memName].Used < 0 {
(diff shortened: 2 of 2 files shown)


DSFans2014 (commented):
> pods that ran before the restart still trigger RmUsage on cleanup, but their usage was never added to the new instance.

hi @mesutoezdil Have you encountered this situation before?

mesutoezdil:
> > pods that ran before the restart still trigger RmUsage on cleanup, but their usage was never added to the new instance.
> 
> hi @mesutoezdil Have you encountered this situation before?

i found this through code review, not in a prod cluster. the fix is defensive: Used < 0 is always wrong so I clamp it. Happy to reword the PR description if the restart framing is misleading.

DSFans2014:
> > > pods that ran before the restart still trigger RmUsage on cleanup, but their usage was never added to the new instance.
> > 
> > 
> > hi @mesutoezdil Have you encountered this situation before?
> 
> i found this through code review, not in a prod cluster. the fix is defensive: Used < 0 is always wrong so I clamp it. Happy to reword the PR description if the restart framing is misleading.

As far as I know, during scheduler startup, the cache is synchronized and all pods are added to it. If there are any other special cases that are not synchronized, please let me know
https://github.com/Project-HAMi/HAMi/blob/bf7faa2c6b1c70a6f7d124ae2752b107ecdbcd15/pkg/scheduler/scheduler.go#L306

mesutoezdil:
you're right: i've updated the PR description. the clamp is still valid as a defensive invariant
`Used < 0` is always wrong and prevents regressions if an ungated call site is added later.

pkg/device/quota.go
```diff
@@ -157,6 +157,9 @@ func (q *QuotaManager) RmUsage(pod *corev1.Pod, podDev PodDevices) {
 	for idx, val := range usage {
 		if qInfo, ok := (*dp)[idx]; ok && qInfo != nil {
 			qInfo.Used -= val
+			if qInfo.Used < 0 {
+				qInfo.Used = 0
```
DSFans2014, line 162: It's better to log the error here so users can see that something went wrong with the quota

mesutoezdil: done

DSFans2014 (commented):
/lgtm

archlitchi (approved):
/lgtm

PR #1967, opened 2026-06-23
fix: return snapshot in `InspectAllNodesUsage` to avoid data race

DSFans2014:
**What type of PR is this?**

/kind bug

- The metrics system reads the data returned by `InspectAllNodesUsage`. Since the underlying data may be concurrently modified, this could lead to a data race.
- Remove the unused member `cachedstatus` from scheduler

Fixes #


mesutoezdil:
second commit (Jun 24) fixes the race: `getNodesUsage` no longer writes `s.overviewstatus` directly. `Filter` discards the return, `register` assigns under `s.lock.Lock()`. `RLock` in `InspectAllNodesUsage` is sufficient.

archlitchi (approved):
/lgtm

PR #1977, opened 2026-06-25
docs: fix Spread score description in scheduler-policy.md

mesutoezdil:
**What type of PR is this?**

/kind documentation

Both Spread sections said "the less it is used, the higher the score" which is wrong. The score formula is identical for Binpack and Spread; what differs is selection: Spread picks the node/GPU with the **lowest** score. Verified in `pkg/scheduler/policy/gpu_policy.go` (`Less` returns `score[i] > score[j]` for Spread, so sort picks the lowest).

NONE


archlitchi (approved):
/lgtm

PR #1990, opened 2026-06-30
fix(chart): declare metrics port on vGPU monitor container

ntheanh201:
**What type of PR is this?**

/kind bug

The `vgpu-monitor` container in the device-plugin DaemonSet serves Prometheus
metrics on `:9394` (the `--metrics-bind-address` default in `cmd/vGPUmonitor/main.go`),
and the `hami-device-plugin-monitor` Service forwards to that port. However, the
port was never declared in the container's `ports:` spec.

This means the metrics endpoint is not discoverable from the Pod spec (e.g. for
`PodMonitor`/`ServiceMonitor` selectors that reference a named port).

This PR declares the named `metrics` containerPort (9394) on the `vgpu-monitor`
container, matching the convention already used by the scheduler Deployment.

The monitor Service is intentionally left unchanged: it keeps a **numeric**
`targetPort: 9394`. A numeric targetPort routes to any selected pod serving on
that port, whereas a named targetPort would require every pod the Service
selects to declare a container port of that name. Keeping it numeric makes this
change fully non-breaking.

This replaces #1085, whose branch had drifted too far from `master` to rebase.

Fixes #

Verified with `helm template`; the container now renders a `metrics` port and the
Service `targetPort` remains the numeric `9394` (unchanged from `master`).

```release-note
NONE
```

---

AI assistance disclosure: this change was prepared with the help of Claude Code (per CONTRIBUTING.md).

## Summary by CodeRabbit

* **New Features**
  * Exposed a metrics port for the vGPU monitor container, making its metrics endpoint explicitly available on TCP port 9394.

### DIFF
--- a/charts/hami/templates/device-plugin/daemonsetnvidia.yaml
+++ b/charts/hami/templates/device-plugin/daemonsetnvidia.yaml
@@ -191,6 +191,10 @@ spec:
             {{- with .Values.devicePlugin.monitor.extraEnvs }}
             {{- . | toYaml | nindent 12 }}
             {{- end }}
+          ports:
+            - name: metrics
+              containerPort: 9394
+              protocol: TCP
           resources:
           {{- toYaml .Values.devicePlugin.monitor.resources | nindent 12 }}
           volumeMounts:


charts/hami/templates/device-plugin/daemonsetnvidia.yaml
```diff
@@ -191,6 +191,10 @@ spec:
             {{- with .Values.devicePlugin.monitor.extraEnvs }}
             {{- . | toYaml | nindent 12 }}
             {{- end }}
+          ports:
+            - name: metrics
+              containerPort: 9394
```
mesutoezdil, line 196: the pr description says the service was updated to use the named port, but the second commit reverted that. 
the description should be updated to reflect the current scope.
thanks!

archlitchi (approved):
/lgtm

PR #1997, opened 2026-07-02
fix(nvidia): validate gpumem-percentage range to prevent silent unschedulability

### LINKED ISSUE #1781: Scheduler extender returns MemPercentagereq:101 overflow sentinel when nvidia.com/gpu is requested without nvidia.com/gpumem — pod stays unschedulable
**What happened**

When a pod requests \`nvidia.com/gpu\` **without** a sibling \`nvidia.com/gpumem\`, the HAMi scheduler extender emits \`MemPercentagereq:101\` in its filter response. \`101\` is out of range for a percentage (valid 0–100) — it is clearly an overflow sentinel rather than a legitimate computed value. The extender then rejects every candidate node and the pod stays Pending indefinitely, even on nodes with plenty of free GPU memory.

There is no user-visible reason string — the pod just shows \`FailedScheduling\` / \"no available GPU resources on all N nodes,\" and only scheduler extender logs expose the \`101\` sentinel. So the silent-unschedulability surface is exactly the \"malformed pod spec that looks legal to k8s but is rejected by HAMi\" case.

I'm filing this alongside Project-HAMi/HAMi#1780 (\`vgpu-scheduler-extender\` panic in \`ComputeScore\` on unschedulable pods) — we hit both on the same cluster during triage.

**What you expected to happen**

One of:

1. HAMi should treat a missing \`nvidia.com/gpumem\` as \"request whole-card memory\" (current K8s API contract allows \`nvidia.com/gpu\` without \`gpumem\`), or
2. HAMi should surface a human-readable \"\`nvidia.com/gpumem\` required when requesting \`nvidia.com/gpu\`\" reject reason rather than a numeric overflow sentinel that silently prevents scheduling.

Either of those is fine. What should not happen: the pod stays Pending forever with no actionable signal.

**How to reproduce it (as minimally and precisely as possible)**

1. On a HAMi-managed cluster, deploy any pod that requests \`nvidia.com/gpu\` without a \`nvidia.com/gpumem\` sibling field. Minimal manifest:

   ```yaml
   apiVersion: v1
   kind: Pod
   metadata:
     name: hami-gpumem-missing-repro
   spec:
     runtimeClassName: nvidia
     containers:
       - name: cuda
         image: nvidia/cuda:12.4.0-base-ubuntu22.04
         command: [\"sleep\", \"infinity\"]
         resources:
           limits:
             nvidia.com/gpu: \"1\"
           requests:
             nvidia.com/gpu: \"1\"
   ```

2. Observe: pod stays Pending with \`FailedScheduling\` events. Scheduler extender filter response for every node contains \`MemPercentagereq:101\`.

3. Add \`nvidia.com/gpumem: \"4000\"\` (or any valid MiB value ≤ card capacity) to both \`requests\` and \`limits\`, re-apply, and the pod schedules normally on the first suitable node.

**Stack-level symptom (from extender logs during a real occurrence)**

\`MemPercentagereq:101\` appears in the extender's per-node filter response. \`101\` = \`100 + 1\` → suggests either a uint8-style clamp-and-overflow in the memory percentage computation, or a division-by-zero that falls through to a sentinel constant. The file most likely involved is \`pkg/scheduler/policy/gpu_policy.go\` (same area as #1780 but a different code path — score vs. filter).

Root cause hypothesis: when \`nvidia.com/gpumem\` is absent the request normalization probably ends up computing something like \`requestedMem * 100 / totalMem\` with \`requestedMem == 0\` or with an uninitialized default that overflows. 101 is suspicious as a clamped value.

- The failure is silent from a K8s-API perspective — \`kubectl describe pod\` just shows \`0/N nodes are available\` without mentioning the missing \`gpumem\` field.
- Workaround is trivial (always set \`gpumem\` alongside \`gpu\`) but users who don't yet know about HAMi's memory semantics hit it as a scheduling mystery.
- We're also adding a repo-side policy and (follow-up) CI lint to prevent recurrence in our infrastructure manifests.

**Environment**

- HAMi version: \`v2.6.14\` (via Olares downstream build \`beclab/hami:v2.6.14\`, which tracks this upstream)
- Kubernetes: \`v1.33.3+k3s1\`
- Cluster: 8-node k3s, mixed GPU fleet (RTX 2080 / 3090 / 4070 Ti / 4090 / 5080 / 5090 / RTX PRO 6000)
- Container runtime: containerd 2.0.5-k3s2 (workers) / 1.7.28 (control plane)

Happy to attach full scheduler extender logs showing the \`MemPercentagereq:101\` line and a manifest-as-applied diff if that helps.

saiyam1814:
**What type of PR is this?**

/kind bug

`nvidia.com/gpumem-percentage` is never range-validated. A pod requesting a value > 100 passes admission, but the scheduler then computes `memreq = Totalmem * pct / 100` — larger than any card — so the pod stays **Pending forever** with a misleading `CardInsufficientMemory` reason. A value of exactly `101` silently collides with the internal "percentage not set" sentinel in `GenerateResourceRequests`.

This PR:
- Rejects percentages outside `[0, 100]` at admission time in `MutateAdmission` with a clear error message (same pattern as Ascend's validation).
- Clamps out-of-range percentages to `100` in `GenerateResourceRequests` (validated on the raw int64 before the int32 cast to avoid wraparound), mirroring the existing `Coresreq > 100` clamp in `Fit`. This also covers init containers and pods that bypass the webhook — admission-time validation for init containers was dropped per review, as it needs more careful design.
- Adds regression tests for both paths, including an int32-overflow case.

Valid `0–100` requests are unchanged; only invalid specs (negative, 101, >100) change behavior — from silent forever-Pending to an immediate, actionable rejection.

Fixes #1781

The scenario as originally reported (missing `gpumem` on v2.6.14) is largely resolved on current releases: whole-card default exists since v2.3.2 and human-readable filter reasons landed in v2.7.0 (#1097). This PR closes the remaining gap that still reproduces the silent-unschedulability on master: out-of-range percentage values.

Known pre-existing limitation (not introduced here): pods bypassing the webhook don't get the admission-time exclusive-core mutation from `defaultExclusiveCoreIfNeeded`; the clamp keeps scheduling sane for them regardless.

```release-note
Pods requesting an out-of-range nvidia.com/gpumem-percentage (outside 0-100) are now rejected at admission with a clear error message instead of remaining permanently unschedulable.
```

### DIFF
--- a/pkg/device/nvidia/device.go
+++ b/pkg/device/nvidia/device.go
@@ -571,11 +571,11 @@ func (dev *NvidiaGPUDevices) GenerateResourceRequests(ctr *corev1.Container) dev
 			if ok {
 				mempnums, ok := mem.AsInt64()
 				if ok {
-					mempnum = int32(mempnums)
-					if mempnum < 0 || mempnum > 100 {
-						klog.ErrorS(nil, "memory percentage request out of range, clamping to 100", "container", ctr.Name, "requested", mempnum)
-						mempnum = 100
+					if mempnums < 0 || mempnums > 100 {
+						klog.ErrorS(nil, "memory percentage request out of range, clamping to 100", "container", ctr.Name, "requested", mempnums)
+						mempnums = 100
 					}
+					mempnum = int32(mempnums)
 				}
 			}
 			if mempnum == 101 && memnum == 0 {
--- a/pkg/device/nvidia/device.go
+++ b/pkg/device/nvidia/device.go
@@ -344,6 +344,16 @@ func (dev *NvidiaGPUDevices) GetNodeDevices(n corev1.Node) ([]*device.DeviceInfo
 
 func (dev *NvidiaGPUDevices) MutateAdmission(ctr *corev1.Container, p *corev1.Pod) (bool, error) {
 	/*gpu related */
+	if err := dev.validateMemoryPercentage(ctr); err != nil {
+		return false, err
+	}
+	// Init containers are scheduled by Resourcereqs but are not passed to
+	// MutateAdmission by the webhook, so validate them here as well.
+	for idx := range p.Spec.InitContainers {
+		if err := dev.validateMemoryPercentage(&p.Spec.InitContainers[idx]); err != nil {
+			return false, err
+		}
+	}
 	priority, ok := ctr.Resources.Limits[corev1.ResourceName(dev.config.ResourcePriority)]
 	if ok {
 		ctr.Env = append(ctr.Env, corev1.EnvVar{
@@ -379,6 +389,15 @@ func (dev *NvidiaGPUDevices) MutateAdmission(ctr *corev1.Container, p *corev1.Po
 		})
 	}
 	return hasResource, nil
+}
+
+func (dev *NvidiaGPUDevices) validateMemoryPercentage(ctr *corev1.Container) error {
+	if pct, ok := resourceValue(ctr, corev1.ResourceName(dev.config.ResourceMemoryPercentageName)); ok {
+		if pct < 0 || pct > 100 {
+			return fmt.Errorf("invalid %s value %d in container %s: must be an integer between 0 and 100", dev.config.ResourceMemoryPercentageName, pct, ctr.Name)
+		}
+	}
+	return nil
 }
 
 func (dev *NvidiaGPUDevices) mutateContainerResource(ctr *corev1.Container) bool {
@@ -553,6 +572,10 @@ func (dev *NvidiaGPUDevices) GenerateResourceRequests(ctr *corev1.Container) dev
 				mempnums, ok := mem.AsInt64()
 				if ok {
 					mempnum = int32(mempnums)
+					if mempnum < 0 || mempnum > 100 {
+						klog.ErrorS(nil, "memory percentage request out of range, clamping to 100", "container", ctr.Name, "requested", mempnum)
+						mempnum = 100
+					}
 				}
 			}
 			if mempnum == 101 && memnum == 0 {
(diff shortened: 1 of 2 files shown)


pkg/device/nvidia/device.go
```diff
@@ -344,6 +344,16 @@ func (dev *NvidiaGPUDevices) GetNodeDevices(n corev1.Node) ([]*device.DeviceInfo
 
 func (dev *NvidiaGPUDevices) MutateAdmission(ctr *corev1.Container, p *corev1.Pod) (bool, error) {
 	/*gpu related */
+	if err := dev.validateMemoryPercentage(ctr); err != nil {
+		return false, err
+	}
+	// Init containers are scheduled by Resourcereqs but are not passed to
```
archlitchi: i think we can neglect initContainers here, this part need more carefully design, it's not a very good practice to iterate all initContainers here

saiyam1814:
@archlitchi Thanks for the review — removed the init-container loop in d5a1fe3. Regular containers are still validated at admission, and init containers remain covered by the out-of-range clamp in GenerateResourceRequests on the scheduling path, so they can't produce permanently-Pending pods. Happy to explore a more holistic init-container admission design in a follow-up if there's interest.

pkg/device/nvidia/device.go
```diff
@@ -344,6 +344,9 @@ func (dev *NvidiaGPUDevices) GetNodeDevices(n corev1.Node) ([]*device.DeviceInfo
 
 func (dev *NvidiaGPUDevices) MutateAdmission(ctr *corev1.Container, p *corev1.Pod) (bool, error) {
 	/*gpu related */
+	if err := dev.validateMemoryPercentage(ctr); err != nil {
```
mesutoezdil, line 347: The PR description still says "Init containers are validated too", but the last commit dropped that loop. Please update the description to match the current scope.

saiyam1814: This is fixed.

archlitchi (approved):
/lgtm

Issue #1950, opened 2026-06-15
Dynamic MIG on H200 NVL: MIG devices are discovered but immediately marked unhealthy — `health.go:98 Could not determine device placement ... unable to parse UUID as MIG device`


allen84205:
### What happened

On a node with 2× **NVIDIA H200 NVL**, I enabled HAMi dynamic MIG (`operatingmode: mig`). After creating `1g.18gb` GPU instances, the HAMi device plugin **does discover and advertise them** (`nvidia.com/mig-1g.18gb: 7` appears in the node's allocatable), but **immediately marks every MIG device unhealthy**, so they are unusable and `vgpu-mode: mig` pods can never be scheduled (`CardTypeMismatch`).

### Device-plugin logs (the key lines)

```
util.go:337] Applying data= version: v1
mig-configs:
   mig-enabled: false
   mig-devices: {}
util.go:348] Mig apply MIG configuration applied successfully
health.go:98] Could not determine device placement for MIG-30d00c09-8a98-59b8-8c1a-1d64b4ec3ad2: unable to parse UUID as MIG device; Marking it unhealthy.
server.go:456] 'nvidia.com/mig-1g.18gb' device marked unhealthy: MIG-30d00c09-8a98-59b8-8c1a-1d64b4ec3ad2
... (repeated for all 7 MIG-<uuid> devices)
server.go:266] Registered device plugin for 'nvidia.com/gpu' with Kubelet
```

`hami.io/node-nvidia-register` still reports both GPUs as whole devices, not MIG:

```
[{"id":"GPU-...","count":10,"devmem":143771,"devcore":100,"type":"NVIDIA H200 NVL","mode":"hami-core","health":true}, {"id":"GPU-...","type":"NVIDIA H200 NVL","mode":"hami-core","health":true}]
```

### Scheduler (vgpu-scheduler-extender) result for a `vgpu-mode: mig` pod

```
FilteringFailed  hami-scheduler  1 nodes CardTypeMismatch(<node>)
FailedScheduling hami-scheduler  0/1 nodes are available: 1 NodeUnfitPod
```

### Environment

- HAMi: **v2.9.0** (Helm chart `hami-2.9.0`)
- GPU: **2× NVIDIA H200 NVL** (machine ASUS ESC8000A-E13P)
- Driver: **580.95.05**, CUDA **13.0**, NVIDIA-SMI 580.95.05
[...]

PR #1999, opened 2026-07-02
fix(device-plugin): surface NVML error when MIG device placement cannot be determined

saiyam1814:
**What type of PR is this?**

/kind bug

When the device-plugin health check cannot resolve a MIG device's placement, the NVML error from `DeviceGetHandleByUUID` is silently discarded and the code falls back to parsing the legacy MIG UUID format (`MIG-GPU-<parent-uuid>/<gi>/<ci>`). Modern drivers assign **opaque** MIG UUIDs (`MIG-<uuid>`) that carry no placement information, so for them the fallback is guaranteed to fail — and the only error the user ever sees is the misleading `unable to parse UUID as MIG device`, masking the actual NVML failure (e.g. `ERROR_NO_PERMISSION`, `ERROR_NOT_FOUND`).

This is exactly what makes reports like #1950 (H200 NVL, all MIG devices marked unhealthy) impossible to diagnose from the logs.

This PR:
- Propagates the NVML return code alongside the parse failure, so the health-check log shows the real reason, e.g. `failed to get MIG device handle for MIG-30d00c09-...: ERROR_NO_PERMISSION; unable to parse UUID as MIG device`.
- Adds unit tests for `parseMigDeviceUUID` and `getMigDeviceParts` (previously uncovered) using the go-nvml mock, covering the NVML success path, the legacy-UUID fallback path, the error-propagation path, and full-device rejection.

No behavior change for devices whose placement resolves successfully; only the error message for the already-failing path improves.

Related to #1950 (makes the reported failure diagnosable; the underlying NVML failure on that node still needs the improved log output to identify)

The `go vet` lock-by-value warnings in `devices.go` are pre-existing on master and untouched by this PR.

```release-note
Device-plugin health check now reports the underlying NVML error when a MIG device's placement cannot be determined, instead of the generic "unable to parse UUID as MIG device".
```

### DIFF
--- a/pkg/device-plugin/nvidiadevice/nvinternal/rm/health.go
+++ b/pkg/device-plugin/nvidiadevice/nvinternal/rm/health.go
@@ -338,7 +338,15 @@ func (r *nvmlResourceManager) getMigDeviceParts(d *Device) (string, uint32, uint
 		//nolint:gosec  // We know that the values returned from Get*InstanceId are within the valid uint32 range.
 		return parentUUID, uint32(gi), uint32(ci), nil
 	}
-	return parseMigDeviceUUID(uuid)
+	// Fall back to parsing the legacy MIG UUID format (MIG-GPU-<parent-uuid>/<gi>/<ci>).
+	// Modern drivers assign opaque MIG UUIDs that carry no placement information,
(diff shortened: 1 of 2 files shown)


archlitchi (approved):
/lgtm

Issue #1998, opened 2026-07-02
Strengthen the AI assistance policy in CONTRIBUTING.md with contribution gates


mesutoezdil:
We already have an AI Assistance Notice in \`CONTRIBUTING.md\`. The problem is that it's a norm with no gate: more PRs land where the diff is model-generated and the author can't explain it, and review threads get answered with generated text that misses the point, so the first-pass review falls on maintainers.

Below is a small set of enforceable additions on top of what we have, drawn from kagent, agent-substrate, NVIDIA OpenShell, and upstream Kubernetes. It keeps disclosure and does not ban AI.

## Proposed additions

Keep the existing AI Assistance Notice. Add the following.

1. **Author-understanding gate with a stated consequence.** If a maintainer asks how a change works and the author cannot explain it, the PR is closed. Repeated cases lead to interaction limits. (OpenShell, Kubernetes)

2. **Tests must be run on real GPU hardware before submitting.** For any change affecting scheduling decisions, device allocation, or in-container isolation, unit tests and mocked devices are not sufficient. The author validates the change on real GPU hardware and records in the PR what was tested and on which device and driver. A PR that touches device behavior with no evidence of real-hardware testing is not accepted. (HAMi-specific; aligns with the Kubernetes rule to verify before you submit)

3. **Cap on model-generated scope.** Large AI-generated PRs are not accepted. Anything beyond a small fix starts as an issue and is split into reviewable commits. AI-generated commit messages are not accepted; write your own. (Kubernetes)

4. **Review replies in your own words.** AI may help you understand a comment. The reply you post must be written by you and must address the specific point. Verbatim or canned AI replies, or replies that do not engage the comment, lead to the PR being closed. (Kubernetes)

5. **Commit trailer hygiene.** No AI listed as co-author, no \`assisted-by\`, \`co-developed\`, or similar trailers, no AI co-signing. Disclosure belongs in the PR description. (Kubernetes)

6. **Entry step for brand-new accounts.** Before a first large PR, a first-time contributor opens a short Discussion describing, in their own words, what they intend to change and why. Good first issues are exempt. (OpenShell)

7. **Issue-first for non-trivial changes.** Features, refactors, and any change to scheduler, webhook, or device-plugin behavior start as an issue with a short design and maintainer agreement before a full PR. Draft PRs are welcome for feedback. (kagent, agent-substrate)

[...]

PR #2001, opened 2026-07-03
fix(webhook): count the resource to fit quota in webhook when reqnum > 1

DSFans2014:
**What type of PR is this?**

/kind bug

count the resource to fit quota in webhook if reqnum > 1

Fixes #

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved GPU quota checks so memory and CPU requests are correctly counted when a workload requests multiple GPUs.
  * Multi-GPU jobs with matching memory requests are now more likely to pass resource fit validation as expected.
  * Added coverage for a multi-GPU scheduling scenario to help prevent regressions.

### DIFF
--- a/pkg/scheduler/webhook.go
+++ b/pkg/scheduler/webhook.go
@@ -135,12 +135,12 @@ func fitResourceQuota(pod *corev1.Pod) bool {
 		}
 		for _, ctr := range pod.Spec.Containers {
 			req, ok := getRequest(&ctr, resourceName)
-			if ok && req == 1 {
+			if ok {
 				if memReq, ok := getRequest(&ctr, memResourceName); ok {
-					memoryReq += memReq
+					memoryReq += memReq * req
 				}
 				if coreReq, ok := getRequest(&ctr, coreResourceName); ok {
-					coresReq += coreReq
+					coresReq += coreReq * req
 				}
 			}
 		}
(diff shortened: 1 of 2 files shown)


mesutoezdil:
> the old code assumed reqnum was always 1 — so if a container requested 2 GPUs with 400MB memory each, it would only count 400MB against quota instead of 800MB. multiplying by req fixes that.
> 
> same issue existed for cores. good catch.

Why are you rewriting the existing PR definition and code as a comment? Is this some kind of motivational support?

DSFans2014:
> > the old code assumed reqnum was always 1 — so if a container requested 2 GPUs with 400MB memory each, it would only count 400MB against quota instead of 800MB. multiplying by req fixes that.
> > same issue existed for cores. good catch.
> 
> Why are you rewriting the existing PR definition and code as a comment? Is this some kind of motivational support?

@mesutoezdil Where does your quoted content come from?

archlitchi (approved):
/lgtm

mesutoezdil:
> > > the old code assumed reqnum was always 1 — so if a container requested 2 GPUs with 400MB memory each, it would only count 400MB against quota instead of 800MB. multiplying by req fixes that.
> > > same issue existed for cores. good catch.
> > 
> > 
> > Why are you rewriting the existing PR definition and code as a comment? Is this some kind of motivational support?
> 
> @mesutoezdil Where does your quoted content come from?

As usual, someone had pasted a nonsensical LLM output; after i’d replied, he deleted the post.

We are working on it: https://github.com/Project-HAMi/HAMi/issues/1998#issuecomment-4888877868

Answers should be written by human.

PR #1610, opened 2026-01-20
feat: support module-pair allocation for Ascend 910C devices in SuperPod environments

ashergaga:
This PR enhances the HAMi scheduler to properly handle Ascend 910C devices in Huawei SuperPod clusters by enforcing module-pair allocation:

- Ascend 910C hardware consists of **2 NPUs per physical module**, and valid allocations must respect this topology (i.e., allocate either 1 NPU — auto-rounded to 2 — or an even number of NPUs).
- The scheduler now **rejects odd-numbered requests > 1** (e.g., 3, 5) with a clear error message.
- For valid requests, devices are allocated **only in complete module pairs**.

PR #2005, opened 2026-07-03
feat(ascend): support 910C vNPU templates vir05_1c_16g and vir10_3c_32g

### LINKED ISSUE #2004: feat(ascend): support 910C vNPU templates vir05_1c_16g and vir10_3c_32g (Ascend HDK 26.0.RC1)
## Background

Starting with Ascend HDK 26.0.RC1, the Ascend 910C device resource huawei.com/Ascend910C supports NPU virtualization, also known as vNPU.

With vNPU support, a single Ascend 910C NPU can be sliced into smaller compute instances.

The following vNPU templates are provided:

| Template | Memory | aiCore | aiCPU |
|---|---:|---:|---:|
| vir05_1c_16g | 16 GB | 5 | 1 |
| vir10_3c_32g | 32 GB | 10 | 3 |

## Current Behavior

In HAMi, the Ascend 910C is currently registered without any vNPU templates.

As a result, the 910C can only be allocated as a whole card. Fractional requests based on memory or core resources cannot be scheduled on 910C, even though Ascend HDK 26.0.RC1 already supports vNPU for this device.

## Expected Behavior

HAMi should recognize the following vNPU templates for Ascend 910C:

- vir05_1c_16g
- vir10_3c_32g

This allows fractional Ascend 910C requests to be scheduled properly.

Whole-card allocation and the existing Ascend 910C topology behavior should remain unaffected.

## Environment

- Ascend HDK: 26.0.RC1
- Device: Ascend 910C
- Device resource: huawei.com/Ascend910C

ouyangluwei163:
Ascend HDK 26.0.RC1 enables NPU virtualization (vNPU) on the 910C. The 910C was previously registered without any vNPU templates, so it could only be allocated as a whole card and fractional (memory/core) requests could not be scheduled.

Add the vir05_1c_16g (16GB/5 aiCore/1 aiCPU) and vir10_3c_32g (32GB/10 aiCore/3 aiCPU) templates to the scheduler device config, and add examples/ascend/job-910C.yaml demonstrating a vNPU request.

**What type of PR is this?**
/kind feature

Fixes #2004 

## Summary by CodeRabbit

* **New Features**
  * Added new VNPU template profiles for Ascend910C/Ascend910 in scheduler configuration.
  * Introduced a new example Pod manifest demonstrating template-specific Ascend910C resource requests.
* **Behavior Changes**
  * Enhanced Ascend910C request handling: allocation rounding and related validation now apply only when “SuperPod” mode is enabled.
* **Tests**
  * Updated and expanded admission logic tests to cover SuperPod and vNPU split behavior.

### DIFF
--- a/examples/ascend/job-910C.yaml
+++ b/examples/ascend/job-910C.yaml
@@ -1,0 +1,16 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: ascend910c-job
+spec:
+  containers:
+    - name: ubuntu-container
+      image: ascendhub.huawei.com/public-ascendhub/ascend-mindspore:23.0.RC3-centos7
+      command: ["bash", "-c", "sleep 86400"]
+      resources:
+        limits:
+          huawei.com/Ascend910C: 1 # requesting 1 NPU
+          # requesting device memory; the value maps to a vNPU template (Ascend HDK 26.0.RC1):
+          #   16384 -> vir05_1c_16g (16GB, 5 aiCore, 1 aiCPU)
+          #   32768 -> vir10_3c_32g (32GB, 10 aiCore, 3 aiCPU)
+          huawei.com/Ascend910C-memory: 16384
--- a/charts/hami/templates/scheduler/device-configmap.yaml
+++ b/charts/hami/templates/scheduler/device-configmap.yaml
@@ -443,5 +443,14 @@ data:
         aiCPU: 7
         runtimeClassName: "{{ .Values.devices.ascend.runtimeClassName }}"
         overwriteEnv: {{ .Values.scheduler.overwriteEnv | default "false" }}
+        templates:
+          - name: vir05_1c_16g
+            memory: 16384
+            aiCore: 5
+            aiCPU: 1
+          - name: vir10_3c_32g
+            memory: 32768
+            aiCore: 10
+            aiCPU: 3
   {{ end }}
 {{- end }}
--- a/charts/hami/templates/scheduler/device-configmap.yaml
+++ b/charts/hami/templates/scheduler/device-configmap.yaml
@@ -443,6 +443,7 @@ data:
         aiCPU: 7
         runtimeClassName: "{{ .Values.devices.ascend.runtimeClassName }}"
         overwriteEnv: {{ .Values.scheduler.overwriteEnv | default "false" }}
+        superPod: true
         templates:
           - name: vir05_1c_16g
             memory: 16384
--- a/pkg/device/ascend/device.go
+++ b/pkg/device/ascend/device.go
@@ -131,7 +131,7 @@ func (dev *Devices) MutateAdmission(ctr *corev1.Container, p *corev1.Pod) (bool,
 	}
 
 	reqNum := count.Value()
-	if dev.config.CommonWord == Ascend910CType {
+	if dev.config.CommonWord == Ascend910CType && dev.config.SuperPod {
 		if reqNum == 1 {
 			// Since the minimum allocation unit is one physical module (2 NPUs), round up the limits and requests to 2.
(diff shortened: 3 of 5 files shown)


examples/ascend/job-910C.yaml
```diff
@@ -0,0 +1,16 @@
...
+metadata:
+  name: ascend910c-job
+spec:
+  containers:
+    - name: ubuntu-container
+      image: ascendhub.huawei.com/public-ascendhub/ascend-mindspore:23.0.RC3-centos7
+      command: ["bash", "-c", "sleep 86400"]
+      resources:
+        limits:
+          huawei.com/Ascend910C: 1 # requesting 1 NPU
```
mesutoezdil, line 12: @ouyangluwei163 before lgtm, 1 question: issue #2004 says "a single ascend 910C npu can be sliced into smaller compute instances", but mutateadmission unconditionally bumps ascend910C: 1 to 2. 
for a vnpu request like this 1, generateresourcerequests reads the bumped limit (2), fit allocates 2 physical npus each with vir05_1c_16g, and the container gets 2 vnpu instances consuming 32768 mb total, not 1 (?)

should the bump be skipped when memory is specified, so a single-npu vnpu slice works as described in the issue?

ouyangluwei163: Thanks @mesutoezdil — you're right, and thanks for the precise trace.

  Context: a 910C node exposes 16 NPUs that the driver binds in pairs (1 module = 2 NPUs). The 1→2 rounding
  was added in #1610 to match that pairing for SuperPod, so "min unit = 2 NPUs" only holds while the card is
  bound. Ascend HDK 26.0.RC1 adds npu-smi set -t multi-die-policy -d, which unbinds the pair — and that unbind
  is the prerequisite for vNPU, since a die can only be sliced after it's independent.

  Fix: I added an explicit superPod flag on the 910C config rather than inferring from "memory is specified"
  (that ties a hardware mode to request shape and misses cases like a whole-die request in split mode):

  - superPod: true → keep the module-pair logic (round 1→2, reject odd).
  - superPod: false (default) → card unbound; bump/odd-check skipped, so single-NPU vNPU slices work as in
  #2004.

  Shipped config keeps superPod: true (zero impact on existing SuperPod users); vNPU users unbind the card and
  set superPod: false. Updated the example with both prerequisites and added tests for both modes.

  Does this work for you?

mesutoezdil: > Thanks @mesutoezdil — you're right, and thanks for the precise trace.
> 
> Context: a 910C node exposes 16 NPUs that the driver binds in pairs (1 module = 2 NPUs). The 1→2 rounding was added in #1610 to match that pairing for SuperPod, so "min unit = 2 NPUs" only holds while the card is bound. Ascend HDK 26.0.RC1 adds npu-smi set -t multi-die-policy -d, which unbinds the pair — and that unbind is the prerequisite for vNPU, since a die can only be sliced after it's independent.
> 
> Fix: I added an explicit superPod flag on the 910C config rather than inferring from "memory is specified" (that ties a hardware mode to request shape and misses cases like a whole-die request in split mode):
> 
> * superPod: true → keep the module-pair logic (round 1→2, reject odd).
> * superPod: false (default) → card unbound; bump/odd-check skipped, so single-NPU vNPU slices work as in
>   [feat(ascend): support 910C vNPU templates vir05_1c_16g and vir10_3c_32g (Ascend HDK 26.0.RC1) #2004](https://github.com/Project-HAMi/HAMi/issues/2004).
> 
> Shipped config keeps superPod: true (zero impact on existing SuperPod users); vNPU users unbind the card and set superPod: false. Updated the example with both prerequisites and added tests for both modes.
> 
> Does this work for you?

makes sense.

and reminder: pls adhere to the new rules regarding responses (and codes, prs, descriptions).

https://github.com/Project-HAMi/HAMi/issues/1998#issuecomment-4888877868

archlitchi (approved):
/lgtm

Issue #1889, opened 2026-05-20
HAMi RoadMap v2.10


archlitchi:
This is the issue for HAMi RoapMap for v2.10, you can simply reply this issue to submit your ideas

Next Version: v2.10.0
Estimated release: Aug/Sep-2026
Tasks

**General**
- [x] Remove DRA components from HAMi main chart, DRA has it's own chart. @archlitchi 

**DRA**
- [x] HAMi(CRD) for KAI scheduler practice @mesutoezdil 
- [x] Ascend DRA @Shouren 
- [ ] Enflame DRA

**Scheduling Features**
- [x] Init Container Support (@maishivamhoo123 @Shouren )
- [ ] Pod Preemption (Note: Need to fix an issue in upstream kube-scheduler @maishivamhoo123 @Shouren )
- [x] Support podgroup @lin121291 @archlitchi (https://github.com/Project-HAMi/HAMi/issues/1832)
- [x] Flexible Mig @FouoF (https://github.com/Project-HAMi/HAMi/pull/2378)
- [x] New schedule policy-(mutex) @mesutoezdil 
- [x] Scheduler policy combination(https://github.com/Project-HAMi/HAMi/issues/1806, policies: mutex, numa, topology, binpack, spread) @mesutoezdil 
- [x] Handshake annotations optimize @archlitchi 
- [ ] Numa align(CPU and GPU) @saiyam1814 
- [ ] Support autoscaling @fishman @spencercjh 

**Device Support**
- [ ] Cambricon 580 series
- [x] Biren series @DSFans2014 
- [x] AMD Mi300x series @kenji-mido @FouoF 
- [x] Ascend vnpu monitor @maverick123123 
- [x] vNPU Compat mode @ouyangluwei163 
- [x] Heterogeneous Ascend Mode(vNPU and HAMi-core) Support @ouyangluwei163 
- [x] mock-device-plugin support new NPU Template @Wangmin362 
- [x] HAMi-core mode for vnpu doesn't need to align device memory to template @Wangmin362 

**HAMi-core**
- [ ] Env to disable vGPUmonitor utilization watcher @Wangmin362 
- [x] vNPU HAMi-core monitoring @maverick123123 
- [x] Change Compile image to ubi8 to support all GLIBC versions @spencercjh 

Volcano
- [x] Support HAMi-vnpu mode to volcano
- [x] Refine handshake for volcano(https://github.com/volcano-sh/volcano/issues/5361)

Kai scheduler
[...]

PR #2011, opened 2026-07-03
feat: new mutex GPU scheduling policy

### LINKED ISSUE #2009: feat: new mutex GPU scheduling policy
Implement a new \`mutex\` GPU scheduling policy.

When a pod sets \`hami.io/gpu-scheduler-policy: mutex\`, the scheduler only allocates GPUs with no existing users (\`Used == 0\`). If a GPU is already in use by another pod, it is skipped.

Changes needed:
- Add \`GPUSchedulerPolicyMutex\` constant to \`pkg/util/types.go\`
- Add mutex case to \`Less()\` in \`pkg/scheduler/policy/gpu_policy.go\` (idle GPUs sorted to tail for Fit() to pick first)
- Add \`isMutex\` filter to all device backend \`Fit()\` functions

Part of v2.10 roadmap: #1889

mesutoezdil:
Add a `mutex` GPU scheduling policy. A pod that sets `hami.io/gpu-scheduler-policy: mutex` is only placed on GPUs that have no other users (`Used == 0`).

`Less()` sorts busy GPUs first and idle GPUs to the tail, and `Fit()` skips any device with `Used > 0` under the mutex policy, across all device backends.

Based on #2012 (Score-primary sort). This PR is rebased on top of #2012 and only adds the mutex code, so it does not repeat the sort fix.

Changes

- `pkg/util/types.go`: add `GPUSchedulerPolicyMutex`.
- `pkg/scheduler/policy/gpu_policy.go`: mutex branch in `Less()`.
- device backends: `Fit()` rejects used devices under the mutex policy.
- tests for the mutex sort.

Fixes #2009
Part of #1889

## Summary by CodeRabbit

- **New Features**
  - Added a **`mutex`** GPU scheduling policy intended to allocate only idle devices.
  - Introduced **NUMA bind**-aware device ordering to preserve locality during selection.
- **Bug Fixes**
  - Updated allocation behavior across supported device types to treat **already-used** candidates as **exclusive allocation conflicts** under **mutex**, reporting `ExclusiveDeviceAllocateConflict`.
- **Tests**
  - Expanded unit tests to cover **mutex** rejection (including “used device” scenarios) and updated **NUMA bind**/policy sorting expectations.


mesutoezdil:
@coderabbitai review

mesutoezdil:
@coderabbitai full review

archlitchi (approved):
/lgtm

pkg/device/ascend/device.go
```diff
@@ -499,6 +500,11 @@ func (npu *Devices) Fit(devices []*device.DeviceUsage, request device.ContainerD
 			klog.V(5).InfoS(common.CardTimeSlicingExhausted, "pod", klog.KObj(pod), "device", dev.ID, "count", dev.Count, "used", dev.Used)
 			continue
 		}
+		if isMutex && dev.Used > 0 {
```
DSFans2014, line 503: hi @mesutoezdil  I'm a bit confused about the `mutex` policy. 
Judging from the code, the mutex scheduling policy is implemented as `restricted`, not `best-effort`.  If a card is already allocated via the `mutex` policy, can it still be shared with other Pods?
If so, why is `restricted` used here? And if it cannot be shared with other Pods, why use `mutex` instead of just requesting a full card?

mesutoezdil: > hi @mesutoezdil I'm a bit confused about the `mutex` policy. Judging from the code, the mutex scheduling policy is implemented as `restricted`, not `best-effort`. If a card is already allocated via the `mutex` policy, can it still be shared with other Pods? If so, why is `restricted` used here? And if it cannot be shared with other Pods, why use `mutex` instead of just requesting a full card?

hi @DSFans2014 as implemented, it's a placement-time preference for the requesting pod, not a device-level lock, it matches #2009's literal spec ("only allocates GPUs with no existing users, skipped if already in use"), but that spec never covered protecting the device from later non-mutex pods.
So today: a mutex pod won't join a busy card, but a later non-mutex pod can still join a card a mutex pod already claimed.
If true mutual exclusion is the goal, that needs a persistent reservation marker on the device (checked by every pod's Fit(), not just mutex ones). If this is the ‘new/ideal’ architecture, let’s open a new issue. WDYT?

DSFans2014: I got, thank you
Moreover, it would be better to also add some explanation and usage about `mutex` in the documentation.
- https://github.com/Project-HAMi/website/blob/aa282781239fc86c4f6a59a95dca36c475bfca7d/docs/userguide/configure.md?plain=1#L83
- https://github.com/Project-HAMi/website/blob/master/docs/developers/scheduling.md

mesutoezdil: @DSFans2014 sure, opened Project-HAMi/website#625 covering both pages

Issue #2017, opened 2026-07-06
docker/vgpu-init.sh breaks when the vGPU runtime directory is customized


haitwang-cloud:
**What happened**:

When the HAMi vGPU runtime directory is customized away from the default `/usr/local/vgpu` (for example, changed to `/var/lib/hami/vgpu` via Helm values), `docker/vgpu-init.sh` still writes an `ld.so.preload` whose content points to the old hardcoded path `/usr/local/vgpu/libvgpu.so`. The dynamic linker fails to preload `libvgpu.so` at the actual mounted location, and vGPU memory/core isolation is silently not enforced.

Root cause: `docker/vgpu-init.sh` treats `ld.so.preload` as a static packaged file and copies it verbatim from `/k8s-vgpu/lib/nvidia/ld.so.preload`, whose content is hardcoded to `/usr/local/vgpu/libvgpu.so`. No matter what `$DEST_DIR` is passed to the script, the resulting preload content never changes.

**What you expected to happen**:

`ld.so.preload` should always match the actual runtime destination directory, i.e.:

```
preload content = ${DEST_DIR}/libvgpu.so
```

So if the user passes `/var/lib/hami/vgpu/`, the preload file should contain `/var/lib/hami/vgpu/libvgpu.so`, not the hardcoded default.

**How to reproduce it (as minimally and precisely as possible)**:

1. Deploy HAMi with a non-default vGPU runtime directory, e.g. `/var/lib/hami/vgpu`.
2. Let `vgpu-init.sh /var/lib/hami/vgpu/` run on the node (as it normally does during init).
3. Inspect the generated preload file on the node:

   ```console
   $ cat /var/lib/hami/vgpu/ld.so.preload
   /usr/local/vgpu/libvgpu.so
   ```

   But `libvgpu.so` is actually mounted at `/var/lib/hami/vgpu/libvgpu.so`, so the preload silently fails and vGPU isolation is not enforced.

**Anything else we need to know?**:

A separate Helm value for the preload path is **not** the right fix — runtime dir and preload path must always move together, and a second config source would just drift. The script itself must derive the preload content from `$DEST_DIR`.

Proposed fix in `docker/vgpu-init.sh`:

1. Skip copying the source `ld.so.preload`.
2. Normalize `$DEST_DIR` (strip trailing slash).
3. Always rewrite `$DEST_DIR/ld.so.preload` with `${DEST_DIR}/libvgpu.so`.

```sh
[...]

PR #2018, opened 2026-07-06
fix(vgpu-init): correct source and destination path handling

haitwang-cloud:
**What type of PR is this?**

/kind bug

When the HAMi vGPU runtime directory is customized away from the default `/usr/local/vgpu` (for example, changed to `/var/lib/hami/vgpu` via Helm values), `docker/vgpu-init.sh` still writes an `ld.so.preload` whose content points to the old hardcoded path `/usr/local/vgpu/libvgpu.so`. The dynamic linker fails to preload `libvgpu.so` at the actual mounted location, and vGPU memory/core isolation is silently not enforced.

Root cause: `docker/vgpu-init.sh` treats `ld.so.preload` as a static packaged file and copies it verbatim from `/k8s-vgpu/lib/nvidia/ld.so.preload`, whose content is hardcoded to `/usr/local/vgpu/libvgpu.so`. No matter what `$DEST_DIR` is passed to the script, the resulting preload content never changes.
Fixes # https://github.com/Project-HAMi/HAMi/issues/2017

NO

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved vGPU image setup to normalize destination paths (including trailing slashes/empty values) and ensure consistent file relative-path handling.
  * Regenerate the `ld.so.preload` target at install time to point to the correct runtime library location.
  * Prevent packaging of the generated preload pointer file to avoid stale or incorrect startup behavior.
* **Helm Chart Updates**
  * Removed `ld.so.preload` from the device-plugin ConfigMap output.
  * Updated the NVIDIA device-plugin DaemonSet mounts by dropping the `deviceconfig` `ld.so.preload` subPath and conditionally mounting the host driver root to `/driver-root` (read-only).

### DIFF
--- a/charts/hami/templates/device-plugin/configmap.yaml
+++ b/charts/hami/templates/device-plugin/configmap.yaml
@@ -10,6 +10,4 @@ metadata:
 data:
   config.json: |
 {{- .Values.devicePlugin.nodeConfiguration.config | nindent 4 }}
-  ld.so.preload: |
-    {{ .Values.devicePlugin.libPath }}/libvgpu.so
 {{- end }}
--- a/charts/hami/templates/device-plugin/daemonsetnvidia.yaml
+++ b/charts/hami/templates/device-plugin/daemonsetnvidia.yaml
@@ -147,10 +147,6 @@ spec:
               subPath: device-config.yaml
             - name: cdi-root
               mountPath: /var/run/cdi
-            - name: deviceconfig
-              mountPath: /k8s-vgpu/lib/nvidia/ld.so.preload
-              subPath: ld.so.preload
-              readOnly: true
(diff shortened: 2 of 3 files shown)


mesutoezdil:
looks good, just 1 thing

fixes # #2017 won't auto-close the issue. should be Fixes #2017

(see on the right side in ui: Development: Successfully merging this pull request may close these issues.
None yet)

Shouren:
@haitwang-cloud As the content of `ld.so.preload` is finally generated by the script,  i think we should cleanup the content of `ld.so.preload` in `charts/hami/templates/device-plugin/configmap.yaml` to avoid ineffective config changes.

haitwang-cloud:
@mesutoezdil @Shouren tks for the review & valuable comments, I have addressed your comment. PTAL again

archlitchi (approved):
/lgtm

PR #2021, opened 2026-07-07
deprecation: remove labels marked as to be deprecated

FouoF:
**What type of PR is this?**

/kind deprecation

Fixes #

## Summary by CodeRabbit

* **Documentation**
  * Updated the Prometheus help text for container device memory metrics to be shorter and clearer.
* **Bug Fixes**
  * Adjusted container device memory metrics so the non-legacy metric now uses a reduced label set (removing several detailed labels), while the legacy metric continues to include the prior detailed labeling.

### DIFF
--- a/cmd/vGPUmonitor/metrics.go
+++ b/cmd/vGPUmonitor/metrics.go
@@ -110,7 +110,7 @@ var (
 	)
 	ctrDeviceMemorydesc = prometheus.NewDesc(
 		"hami_container_device_memory_bytes",
-		`Container device memory usage breakdown in bytes`,
+		`Container device memory usage in bytes`,
 		[]string{"namespace", "pod", "container", "vdevice_index", "device_uuid"}, nil,
 	)
 	ctrDeviceUtilizationdesc = prometheus.NewDesc(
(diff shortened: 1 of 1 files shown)


cmd/vGPUmonitor/metrics.go
```diff
@@ -110,8 +110,8 @@ var (
 	)
 	ctrDeviceMemorydesc = prometheus.NewDesc(
 		"hami_container_device_memory_bytes",
-		`Container device memory usage breakdown in bytes (The label "context_size", "module_size", "buffer_size" and "offset" will be deprecated in v2.10.0, use hami_vgpu_memory_context_bytes, hami_vgpu_memory_module_bytes and hami_vgpu_memory_buffer_bytes instead)`,
-		[]string{"namespace", "pod", "container", "vdevice_index", "device_uuid", "context_size", "module_size", "buffer_size", "offset"}, nil,
+		`Container device memory usage breakdown in bytes`,
+		[]string{"namespace", "pod", "container", "vdevice_index", "device_uuid"}, nil,
 	)
```
mesutoezdil, line 115: descriptor now expects 5 labels, but `collectContainerMetrics` at line 534 still passes `memoryLabels` (9 values)
prometh will return a cardinality error, causing early return and dropping all remaining container metrics 

call on line 534 needs to use `labels` instead of `memoryLabels`

mesutoezdil:
the help text still says "breakdown in bytes" but the breakdown labels are gone now.
should this just be "usage in bytes"?

archlitchi (approved):
/lgtm

PR #2024, opened 2026-07-07
fix: require Ascend core resource for soft slicing

fyuan1316:
## Summary

This PR adds an admission-time validation for Ascend soft slicing.

When a pod explicitly selects HAMi soft slicing with:

```yaml
metadata:
  annotations:
    huawei.com/vnpu-mode: hami-core
```

and the Ascend device config defines a `resourceCoreName`, the workload must request the corresponding core resource together with the memory resource. Otherwise the admission webhook now rejects the pod with a clear error.

This prevents memory-only soft-slicing requests from silently entering the template-matching path where the actual allocated memory can be greater than the requested value.

## Changes

- Validate `hami-core` Ascend workloads in `MutateAdmission`.
- Require the configured `resourceCoreName` when soft slicing is requested.
- Keep the validation scoped to configs that define a core resource, so legacy configs without `resourceCoreName` are not affected.
- Add a unit test for the memory-only rejection case.

## Related

- Related issue: Project-HAMi/ascend-device-plugin#101
- Related docs PR: Project-HAMi/ascend-device-plugin#102
- Discussion issue: #2027
## Test

- `go test ./pkg/device/ascend`
- `go test ./pkg/device`
- `go test ./pkg/scheduler`
- `git diff --check`

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened admission checks for vNPU core mode: containers are now rejected unless they specify a **positive** quantity for the configured core resource in either resource **limits** or **requests** (not just memory).
  * Updated the rejection message to require requesting the configured core resource together with the corresponding memory resource.
* **Tests**
  * Expanded vNPU core mode tests with new failure cases for missing core requests and core requests set to zero, including assertions on the expected error content.

### DIFF
--- a/pkg/device/ascend/device.go
+++ b/pkg/device/ascend/device.go
@@ -153,6 +153,10 @@ func (dev *Devices) MutateAdmission(ctr *corev1.Container, p *corev1.Pod) (bool,
 	isHAMiCore := (vnpuMode == VNPUModeHamiCore)
 
 	if isHAMiCore {
+		if dev.config.ResourceCoreName != "" && !dev.hasPositiveCoreResourceRequest(ctr) {
+			return false, fmt.Errorf("ascend soft slicing requires positive core resource %s to be requested together with %s", dev.config.ResourceCoreName, dev.config.ResourceMemoryName)
+		}
+
 		klog.V(3).Infof("Ascend core resource detected, injecting postStart lifecycle for container %s", ctr.Name)
 
 		if ctr.Lifecycle == nil {
@@ -198,6 +202,15 @@ func (dev *Devices) MutateAdmission(ctr *corev1.Container, p *corev1.Pod) (bool,
 		p.Spec.RuntimeClassName = &dev.config.RuntimeClassName
 	}
 	return true, nil
+}
+
+func (dev *Devices) hasPositiveCoreResourceRequest(ctr *corev1.Container) bool {
+	coreName := corev1.ResourceName(dev.config.ResourceCoreName)
+	if core, ok := ctr.Resources.Limits[coreName]; ok {
+		return core.Value() > 0
+	}
+	core, ok := ctr.Resources.Requests[coreName]
+	return ok && core.Value() > 0
 }
 
 func (dev *Devices) GetNodeDevices(n corev1.Node) ([]*device.DeviceInfo, error) {
--- a/pkg/device/ascend/device_test.go
+++ b/pkg/device/ascend/device_test.go
@@ -805,9 +805,10 @@ func Test_MutateAdmission_VNPUCoreMode(t *testing.T) {
 			ctr corev1.Container
 			pod corev1.Pod
 		}
-		wantPostStart bool
-		wantMem       int64
-		wantCore      int64
+		wantPostStart   bool
+		wantMem         int64
+		wantCore        int64
+		wantErrContains string
 	}{
 		{
 			name: "vNPU-mode hami-core: inject postStart and keep raw memory",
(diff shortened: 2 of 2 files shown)


DSFans2014 (commented):
/lgtm

archlitchi:
I don't think this is the right way to do so, because we will support compete mode for HAMi-core vNPU mode, like we do for NVIDIA, task without 'core' resource will compete on computing cores with each other

fyuan1316:
> I don't think this is the right way to do so, because we will support compete mode for HAMi-core vNPU mode, like we do for NVIDIA, task without 'core' resource will compete on computing cores with each other

@archlitchi:
Thanks for the clarification. I agree with your point.

After checking the design and implementation of other device types, especially the NVIDIA path, I agree that this PR is too strict. It would reject valid hami-core workloads that intentionally omit `*-core` and rely on compute competition.

What I still want to address is the memory-only hami-core case: when a user sets `huawei.com/vnpu-mode: hami-core` and requests `*-memory`, the requested memory should be used as the hami-core memory quota, while `*-core` remains optional.

I think there are two possible ways to implement this:

1. Extend `GenerateResourceRequests` to receive Pod context, so Ascend can check the `hami-core` annotation directly.

2. Keep the existing interface and add an optional pod-aware request generation path only for devices that need Pod context, falling back to the current `GenerateResourceRequests` for others.

In both cases:
- `hami-core + memory only` keeps `Coresreq=0` and supports compute-compete mode
- `hami-core + memory + core` uses an explicit core quota
- non-`hami-core` workloads keep the existing template-based behavior

Do you think either of these directions is acceptable, or would you suggest a better approach?

archlitchi:
> > I don't think this is the right way to do so, because we will support compete mode for HAMi-core vNPU mode, like we do for NVIDIA, task without 'core' resource will compete on computing cores with each other
> 
> @archlitchi: Thanks for the clarification. I agree with your point.
> 
> After checking the design and implementation of other device types, especially the NVIDIA path, I agree that this PR is too strict. It would reject valid hami-core workloads that intentionally omit `*-core` and rely on compute competition.
> 
> What I still want to address is the memory-only hami-core case: when a user sets `huawei.com/vnpu-mode: hami-core` and requests `*-memory`, the requested memory should be used as the hami-core memory quota, while `*-core` remains optional.
> 
> I think there are two possible ways to implement this:
> 
> 1. Extend `GenerateResourceRequests` to receive Pod context, so Ascend can check the `hami-core` annotation directly.
> 2. Keep the existing interface and add an optional pod-aware request generation path only for devices that need Pod context, falling back to the current `GenerateResourceRequests` for others.
> 
> In both cases:
> 
> * `hami-core + memory only` keeps `Coresreq=0` and supports compute-compete mode
> * `hami-core + memory + core` uses an explicit core quota
> * non-`hami-core` workloads keep the existing template-based behavior
> 
> Do you think either of these directions is acceptable, or would you suggest a better approach?

yeah, i think we can use Option 1, passing the pod context is reasonable for this method

pkg/device/ascend/device.go
```diff
@@ -153,6 +153,10 @@ func (dev *Devices) MutateAdmission(ctr *corev1.Container, p *corev1.Pod) (bool,
 	isHAMiCore := (vnpuMode == VNPUModeHamiCore)
 
 	if isHAMiCore {
+		if dev.config.ResourceCoreName != "" && !dev.hasPositiveCoreResourceRequest(ctr) {
+			return false, fmt.Errorf("ascend soft slicing requires positive core resource %s to be requested together with %s", dev.config.ResourceCoreName, dev.config.ResourceMemoryName)
```
mesutoezdil, line 157: pr #2035 removes the entire `if isHAMiCore` block. if #2035 merges after this, this check will be silently dropped. both prs touch the same region and need to be coordinated before either merges?

mesutoezdil:
@fyuan1316 are you planning the option 1 rework, just reminder: prs and issues inactive for 14 days get closed and this one has about 2 days left

PR #2029, opened 2026-07-08
fix(ascend): stop whole-card/memory-less hami-core requests from bypassing the node mode gate

### LINKED ISSUE #2028: [Ascend] whole-card / memory-less hami-core (soft-split) requests bypass the node soft/hard mode gate (#1812)
### What happened

#1812 added node-level soft/hard-split exclusivity for Ascend: a node is marked soft (`hami-vnpu-core`) or hard, and `Fit` rejects a mismatched pod with `ModeNotFit`. But the mode check is nested inside the memory-range condition in `pkg/device/ascend/device.go`:

```go
if request.Memreq > 0 && request.Memreq < totalMemPerCard && request.Nums > 0 {
    if !nodeSupportHamiCore && isHAMiCore { reason[common.ModeNotFit]++; return false, ... }      // soft pod on hard node
    else if nodeSupportHamiCore && !isHAMiCore { reason[common.ModeNotFit]++; return false, ... } // legacy pod on soft node
}
```

Because of the `request.Memreq < totalMemPerCard` bound, a **whole-card memory request (`Memreq == card capacity`)** or a **memory-less request (`Memreq == 0`)** never enters the block, so the mode check is skipped. A pod that explicitly declares `huawei.com/vnpu-mode: hami-core` (soft split) can therefore be scheduled onto a hard-split node, defeating the #1812 exclusivity — and a hami-core pod on a node without hami-core assets has no soft-split limiter to enforce its limits.

### Evidence (real hardware: 8× Ascend 910B4, hard node)

I ran an A/B on a single-node 910B4 cluster, submitting hami-core (soft) pods to a hard node. Same scheduler build except for the fix; `v2.9.0` and current `master` have this gate byte-for-byte identical (`device.go` L449/L462).

| case | request (all `vnpu-mode: hami-core`) | v2.9.0 (stock) | result |
|---|---|---|---|
| partial | mem 16384 | `ModeNotFit`, Pending | ✅ gate works |
| partial | mem 32767 (1 MB under capacity) | `ModeNotFit`, Pending | ✅ gate works |
| **whole-card** | **mem 32768 (= capacity)** | `FilteringSucceed` + `BindingSucceed`, **bound to the hard node** | ❌ **bypassed** |
| **memory-less** | **no memory field** | bound to the hard node | ❌ **bypassed** |

The boundary matches the code condition exactly.

### Root cause

`pkg/device/ascend/device.go` (~L462): the `ModeNotFit` check only runs for `0 < Memreq < totalMemPerCard`. `Memreq >= totalMemPerCard` (whole card) and `Memreq == 0` (memory-less) skip it.

### Proposed fix

Move the **soft-pod arm** out of the memory-range condition, so a pod requesting hami-core is checked against the node mode regardless of how much memory it asks for. The reverse arm (a legacy pod on a hami-core node) is intentionally kept inside the memory-range condition, so that plain whole-card jobs are not rejected when a node — or the global `hamiVnpuCore` default — marks the node hami-core (which would be a regression). See the linked PR.

### Honest scope note

The bypass itself is confirmed on real hardware (A/B above). The downstream runtime harm (a whole-card soft-split pod actually failing after landing on a hard-only node) is **inferred** — reproducing it needs a second node with no soft-split assets; on my single node the device-plugin always ships soft-split assets, so I can't construct a pure hard-split node. Reported as "bypass confirmed; runtime harm needs a 2-node repro".

---

*Disclosure: the analysis is based on my own testing on real Ascend 910B4 hardware; this issue was drafted with AI assistance (Claude Code) and reviewed by me.*

Wangmin362:
## What this PR does

Fixes #2028.

The node soft/hard-split exclusivity added in #1812 rejects a mismatched pod with `ModeNotFit`, but the check sits inside `request.Memreq > 0 && request.Memreq < totalMemPerCard && request.Nums > 0`. A whole-card memory request (`Memreq == card capacity`) or a memory-less request (`Memreq == 0`) skips the block, so a hami-core (soft-split) pod can be scheduled onto a hard-split node.

This PR lifts the **soft-pod arm** out of the memory-range condition: a pod that declares `vnpu-mode: hami-core` must land on a hami-core node regardless of how much memory it requests.

```go
// applies to whole-card and memory-less requests too
if isHAMiCore && !nodeSupportHamiCore {
    reason[common.ModeNotFit]++
    return false, nil, common.GenReason(reason, len(devices))
}

// reverse arm kept inside the memory-range condition on purpose
if request.Memreq > 0 && request.Memreq < totalMemPerCard && request.Nums > 0 {
    if nodeSupportHamiCore && !isHAMiCore {
        reason[common.ModeNotFit]++
        return false, nil, common.GenReason(reason, len(devices))
    }
}
```

### Why only the soft-pod arm (not a symmetric change)

The reverse arm (a legacy/whole-card pod on a hami-core node) is deliberately left inside the memory-range condition. Making it symmetric would newly reject **plain whole-card jobs** on any hami-core node, and because `nodeSupportHamiCore` defaults to the global `hamiVnpuCore` value, a cluster with `hamiVnpuCore: true` would then reject every whole-card legacy request. A hami-core pod on a hard node is a correctness problem (no soft-split limiter to enforce its limits); a legacy whole-card pod on a hami-core node is not — so only the soft-pod direction is unconditional.

### Unit tests
`pkg/device/ascend` `TestDevices_Fit`: **115/115 pass**, `go build/vet/gofmt` clean. Added 4 cases:
- whole-card (`Memreq == capacity`) hami-core on hard node → `ModeNotFit`
- memory-less (`Memreq == 0`) hami-core on hard node → `ModeNotFit`
- whole-card hami-core on hami-core node → fits (not over-blocked)
- whole-card **legacy** on hami-core node → fits (regression guard for the asymmetry above)

### Real hardware (8× Ascend 910B4, hard node) — before/after A/B
Same scheduler build, toggling only this change (`v2.9.0` and `master` have this gate byte-identical):

| case (all `vnpu-mode: hami-core`) | before | after |
|---|---|---|
| partial mem 16384 | Pending `ModeNotFit` | Pending `ModeNotFit` |
| partial mem 32767 | Pending `ModeNotFit` | Pending `ModeNotFit` |
| **whole-card mem 32768** | **bound to hard node (bypass)** | **Pending `ModeNotFit`** |
| **memory-less** | **bound to hard node (bypass)** | **Pending `ModeNotFit`** |
| legacy whole-card (no vnpu-mode) | Running | Running (not over-blocked) |

Before: `FilteringSucceed: 1 nodes fit(node-001)` + `BindingSucceed` for the whole-card hami-core pod.
After: `NodeUnfitPod ... reason="1/8 ModeNotFit"` for the same pod; the legacy whole-card pod still schedules.

## Scope note
The bypass is confirmed on real hardware. Demonstrating the *runtime* failure of a whole-card soft pod on a hard-only node needs a second node without soft-split assets, which I can't construct on a single node (the device-plugin always ships soft-split assets). See the linked issue.

---

*This PR was written with AI assistance (Claude Code): the analysis is from my own testing on real Ascend 910B4 hardware, and the code and tests were drafted with AI and reviewed/validated by me.*

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved core-mode (hami-core) scheduling by rejecting incompatible nodes earlier when the pod requests core-mode but the node doesn’t support it.
  * Tightened compatibility behavior for whole-card and memory-less requests to prevent incorrect placement decisions.
  * Preserved existing legacy workload behavior on nodes reserved for hami-core.
* **Tests**
  * Expanded core-mode gating tests with additional whole-card and memory-less scenarios, including negative cases and regression/positive coverage for reserved-node compatibility.

### DIFF
--- a/pkg/device/ascend/device.go
+++ b/pkg/device/ascend/device.go
@@ -460,21 +460,12 @@ func (npu *Devices) Fit(devices []*device.DeviceUsage, request device.ContainerD
 		totalMemPerCard = devices[0].Totalmem
 	}
 
-	// A hami-core (soft-split) pod must land on a hami-core node regardless of
-	// how much memory it requests. Whole-card (Memreq == card capacity) and
-	// memory-less (Memreq == 0) requests skip the memory-range check below, so
-	// keeping this arm outside that condition closes the bypass of the node
-	// soft/hard exclusivity introduced in #1812.
 	if isHAMiCore && !nodeSupportHamiCore {
 		reason[common.ModeNotFit]++
 		klog.V(4).InfoS("Node filtered: pod requests hami-core but node does not support it", "pod", klog.KObj(pod))
 		return false, nil, common.GenReason(reason, len(devices))
 	}
 
-	// A node reserved for hami-core rejects legacy vNPU (template-based) pods
-	// that carve out part of a card. Whole-card legacy requests are intentionally
-	// left untouched so plain jobs are not stranded when a node, or the global
-	// hamiVnpuCore default, marks the node as hami-core.
 	if request.Memreq > 0 && request.Memreq < totalMemPerCard && request.Nums > 0 {
 		if nodeSupportHamiCore && !isHAMiCore {
 			reason[common.ModeNotFit]++
--- a/pkg/device/ascend/device.go
+++ b/pkg/device/ascend/device.go
@@ -460,12 +460,23 @@ func (npu *Devices) Fit(devices []*device.DeviceUsage, request device.ContainerD
 		totalMemPerCard = devices[0].Totalmem
 	}
 
+	// A hami-core (soft-split) pod must land on a hami-core node regardless of
+	// how much memory it requests. Whole-card (Memreq == card capacity) and
+	// memory-less (Memreq == 0) requests skip the memory-range check below, so
+	// keeping this arm outside that condition closes the bypass of the node
+	// soft/hard exclusivity introduced in #1812.
+	if isHAMiCore && !nodeSupportHamiCore {
+		reason[common.ModeNotFit]++
+		klog.V(4).InfoS("Node filtered: pod requests hami-core but node does not support it", "pod", klog.KObj(pod))
+		return false, nil, common.GenReason(reason, len(devices))
+	}
+
+	// A node reserved for hami-core rejects legacy vNPU (template-based) pods
+	// that carve out part of a card. Whole-card legacy requests are intentionally
+	// left untouched so plain jobs are not stranded when a node, or the global
+	// hamiVnpuCore default, marks the node as hami-core.
 	if request.Memreq > 0 && request.Memreq < totalMemPerCard && request.Nums > 0 {
-		if !nodeSupportHamiCore && isHAMiCore {
-			reason[common.ModeNotFit]++
-			klog.V(4).InfoS("Node filtered: Node does not support hami-core mode", "node", nodeInfo.Node.Name, "pod", pod.Name)
-			return false, nil, common.GenReason(reason, len(devices))
-		} else if nodeSupportHamiCore && !isHAMiCore {
+		if nodeSupportHamiCore && !isHAMiCore {
 			reason[common.ModeNotFit]++
 			klog.V(4).InfoS("Node filtered: Reserved for hami-core but pod is legacy vNPU", "node", nodeInfo.Node.Name, "pod", pod.Name)
 			return false, nil, common.GenReason(reason, len(devices))
--- a/pkg/device/ascend/device.go
+++ b/pkg/device/ascend/device.go
@@ -478,7 +478,7 @@ func (npu *Devices) Fit(devices []*device.DeviceUsage, request device.ContainerD
 	if request.Memreq > 0 && request.Memreq < totalMemPerCard && request.Nums > 0 {
 		if nodeSupportHamiCore && !isHAMiCore {
 			reason[common.ModeNotFit]++
-			klog.V(4).InfoS("Node filtered: Reserved for hami-core but pod is legacy vNPU", "node", nodeInfo.Node.Name, "pod", pod.Name)
+			klog.V(4).InfoS("Node filtered: node reserved for hami-core but pod is legacy vNPU", "pod", klog.KObj(pod))
 			return false, nil, common.GenReason(reason, len(devices))
 		}
 	}
--- a/pkg/device/ascend/device_test.go
+++ b/pkg/device/ascend/device_test.go
@@ -1814,9 +1814,6 @@ func TestDevices_Fit(t *testing.T) {
 			nodeAnnotation: map[string]string{VNPUNodeSelectorAnnotation: "true"},
 		},
 		{
-			// F5: whole-card memory request (Memreq == card capacity) must not
-			// bypass the node mode gate; a hami-core pod still requires a
-			// hami-core node.
 			name: "fit fail: whole-card hami-core pod on legacy node (ModeNotFit)",
 			devices: []*device.DeviceUsage{{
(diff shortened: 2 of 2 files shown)


pkg/device/ascend/device.go
```diff
@@ -460,12 +460,23 @@ func (npu *Devices) Fit(devices []*device.DeviceUsage, request device.ContainerD
...
 	if request.Memreq > 0 && request.Memreq < totalMemPerCard && request.Nums > 0 {
-		if !nodeSupportHamiCore && isHAMiCore {
-			reason[common.ModeNotFit]++
-			klog.V(4).InfoS("Node filtered: Node does not support hami-core mode", "node", nodeInfo.Node.Name, "pod", pod.Name)
-			return false, nil, common.GenReason(reason, len(devices))
-		} else if nodeSupportHamiCore && !isHAMiCore {
+		if nodeSupportHamiCore && !isHAMiCore {
 			reason[common.ModeNotFit]++
 			klog.V(4).InfoS("Node filtered: Reserved for hami-core but pod is legacy vNPU", "node", nodeInfo.Node.Name, "pod", pod.Name)
 			return false, nil, common.GenReason(reason, len(devices))
```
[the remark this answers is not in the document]
Wangmin362: Good catch. Both nodeInfo.Node and pod are guarded as possibly-nil earlier in Fit, so dereferencing .Name here could panic. Switched it to klog.KObj(pod) so it matches the hami-core arm right above. Fixed in d5dc085.

DSFans2014 (commented):
/lgtm

ouyangluwei163 (commented):
Please remove the AI-generated code comments and follow the new specification #1998

pkg/device/ascend/device.go
```diff
@@ -460,14 +460,25 @@ func (npu *Devices) Fit(devices []*device.DeviceUsage, request device.ContainerD
 		totalMemPerCard = devices[0].Totalmem
 	}
 
+	// A hami-core (soft-split) pod must land on a hami-core node regardless of
+	// how much memory it requests. Whole-card (Memreq == card capacity) and
+	// memory-less (Memreq == 0) requests skip the memory-range check below, so
+	// keeping this arm outside that condition closes the bypass of the node
+	// soft/hard exclusivity introduced in #1812.
```
archlitchi: please remove these comments

pkg/device/ascend/device.go
```diff
@@ -460,14 +460,25 @@ func (npu *Devices) Fit(devices []*device.DeviceUsage, request device.ContainerD
...
+	if isHAMiCore && !nodeSupportHamiCore {
+		reason[common.ModeNotFit]++
+		klog.V(4).InfoS("Node filtered: pod requests hami-core but node does not support it", "pod", klog.KObj(pod))
+		return false, nil, common.GenReason(reason, len(devices))
+	}
+
+	// A node reserved for hami-core rejects legacy vNPU (template-based) pods
+	// that carve out part of a card. Whole-card legacy requests are intentionally
+	// left untouched so plain jobs are not stranded when a node, or the global
+	// hamiVnpuCore default, marks the node as hami-core.
```
archlitchi: and here

Wangmin362: done

pkg/device/ascend/device.go
Wangmin362: done

Wangmin362:
> Please remove the AI-generated code comments and follow the new specification #1998

done, 

> Please remove the AI-generated code comments and follow the new specification #1998

done, I have removed the redundant comments.

ouyangluwei163 (approved):
/lgtm

archlitchi (approved):
/lgtm

PR #2035, opened 2026-07-09
Feat: Allow mode-agnostic Ascend pods to schedule across both vNPU-template and HAMi-core nodes

### LINKED ISSUE #2034: Feat: Allow mode-agnostic Ascend pods to schedule across both vNPU-template and HAMi-core nodes
HAMi already supports vNPU template mode and HAMi-core mode coexisting in one cluster. But a pod
  must commit to a mode up front via the hami.io/vnpu-mode annotation: an unannotated pod is
  treated as template-only, and HAMi-core nodes reject any pod not marked hami-core. So a workload
  is hard-bound to one mode at submission and can't use whichever node has free capacity — a
  flexibility gap.

  Proposed Solution

  Make an annotation-less pod mode-agnostic, letting its effective mode follow the node it lands
  on (resolved by the device plugin at allocation), while pinned pods keep today's behavior:

  - Scheduler (Fit): only reject hami-core-requested → node-unsupported; allow unannotated pods on
  both node types.
  - PatchAnnotations: emit a superset RuntimeInfo (template name and raw memory/core) so the
  device plugin picks the right shape per node.
  - Webhook: inject one self-detecting PostStart hook that starts the limiter only when its binary
  is present, and is a no-op otherwise.

ouyangluwei163:
**What type of PR is this?**
/kind feature

Fixes #2034 

## Summary by CodeRabbit

- **Bug Fixes**
  - Device annotation generation now includes memory/core more reliably, including omitting the core field when the reported core count is zero.
  - Scheduling now rejects pods requesting `hami-core` immediately when the node can’t support it.
  - Removed legacy vNPU filtering that could incorrectly prevent compatible pods from fitting.
- **Tests**
  - Updated unit tests to validate the new annotation payload rules and the revised `hami-core` fitting/admission behavior.

### DIFF
--- a/pkg/device/ascend/device.go
+++ b/pkg/device/ascend/device.go
@@ -233,6 +233,8 @@ func (dev *Devices) PatchAnnotations(pod *corev1.Pod, annoInput *map[string]stri
 				} else {
 					_, temp := dev.trimMemory(int64(val.Usedmem))
 					info.Temp = temp
+					info.Memory = int64(val.Usedmem)
+					info.Core = val.Usedcores
 				}
 
 				rtInfo = append(rtInfo, info)
@@ -443,23 +445,10 @@ func (npu *Devices) Fit(devices []*device.DeviceUsage, request device.ContainerD
 		}
 	}
 
-	var totalMemPerCard int32 = 0
-	if len(devices) > 0 {
-		totalMemPerCard = devices[0].Totalmem
-	}
-
 	if isHAMiCore && !nodeSupportHamiCore {
 		reason[common.ModeNotFit]++
 		klog.V(4).InfoS("Node filtered: pod requests hami-core but node does not support it", "pod", klog.KObj(pod))
 		return false, nil, common.GenReason(reason, len(devices))
-	}
-
-	if request.Memreq > 0 && request.Memreq < totalMemPerCard && request.Nums > 0 {
-		if nodeSupportHamiCore && !isHAMiCore {
-			reason[common.ModeNotFit]++
-			klog.V(4).InfoS("Node filtered: node reserved for hami-core but pod is legacy vNPU", "pod", klog.KObj(pod))
-			return false, nil, common.GenReason(reason, len(devices))
-		}
 	}
 	klog.V(4).InfoS("Fit: vnpu-mode annotation", "pod", pod.Name, "vnpuMode", vnpuMode)
 
--- a/pkg/device/ascend/device_test.go
+++ b/pkg/device/ascend/device_test.go
@@ -318,7 +318,7 @@ func Test_PatchAnnotations(t *testing.T) {
 				device.InRequestDevices[dev.config.CommonWord]: "device-0,Ascend,8738,1:;",
 				device.SupportDevices[dev.config.CommonWord]:   "device-0,Ascend,8738,1:;",
 				"predicate-time":        strconv.FormatInt(time.Now().Unix(), 10),
-				"huawei.com/Ascend910A": "[{\"UUID\":\"device-0\",\"temp\":\"vir08\"}]",
+				"huawei.com/Ascend910A": "[{\"UUID\":\"device-0\",\"temp\":\"vir08\",\"memory\":8738,\"core\":1}]",
 			},
 		},
 		{
@@ -399,6 +399,62 @@ func Test_PatchAnnotations_VNPUCoreMode(t *testing.T) {
 			assert.Assert(t, !strings.Contains(val, "\"temp\""))
 		})
 	}
+}
+
+func Test_PatchAnnotations_ZeroCore(t *testing.T) {
+	dev := Devices{
+		config: VNPUConfig{
+			CommonWord:     "Ascend910A",
+			MemoryCapacity: int64(32768),
+			Templates: []Template{
+				{
+					Name:   "vir02",
+					Memory: int64(2184),
+					AICore: int32(2),
+				}, {
+					Name:   "vir04",
+					Memory: int64(4369),
+					AICore: int32(4),
+				}, {
+					Name:   "vir08",
+					Memory: int64(8738),
+					AICore: int32(8),
+				}, {
+					Name:   "vir16",
+					Memory: int64(17476),
+					AICore: int32(16),
+				},
+			},
+		},
+	}
+
+	pod := &corev1.Pod{
+		ObjectMeta: metav1.ObjectMeta{
+			Annotations: map[string]string{},
+		},
+	}
+	pd := device.PodDevices{
+		"Ascend910A": device.PodSingleDevice{
+			[]device.ContainerDevice{
+				{
+					Idx:       0,
+					UUID:      "device-zero-core",
+					Type:      "Ascend",
+					Usedcores: 0,
+					Usedmem:   8738,
+				},
+			},
+		},
+	}
+
+	annoInput := make(map[string]string)
+	result := dev.PatchAnnotations(pod, &annoInput, pd)
+
+	val, ok := result["huawei.com/Ascend910A"]
+	assert.Assert(t, ok)
+	assert.Assert(t, strings.Contains(val, "\"memory\":8738"))
+	assert.Assert(t, strings.Contains(val, "\"temp\":\"vir08\""))
+	assert.Assert(t, !strings.Contains(val, "\"core\""))
 }
 
 func Test_checkType(t *testing.T) {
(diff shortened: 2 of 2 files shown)


archlitchi:
please resolve ai-comments and conflicts

pkg/device/ascend/device.go
```diff
@@ -148,28 +148,23 @@ func (dev *Devices) MutateAdmission(ctr *corev1.Container, p *corev1.Pod) (bool,
...
-						"-c",
-						"export RUST_LOG=info\n/hami-vnpu-core/limiter > /tmp/limiter_manager.log 2>&1 &",
-					},
+	klog.V(4).Infof("Injecting self-detecting hami-vnpu-core limiter PostStart hook for container %s", ctr.Name)
+	if ctr.Lifecycle == nil {
+		ctr.Lifecycle = &corev1.Lifecycle{}
+	}
+	if ctr.Lifecycle.PostStart == nil {
+		ctr.Lifecycle.PostStart = &corev1.LifecycleHandler{
+			Exec: &corev1.ExecAction{
```
mesutoezdil: this log fires before the PostStart nil check. if PostStart is already set, injection is skipped but the message still says 'injecting'. move the log inside `if ctr.Lifecycle.PostStart == nil`.

pkg/device/ascend/device.go
```diff
@@ -465,10 +462,6 @@ func (npu *Devices) Fit(devices []*device.DeviceUsage, request device.ContainerD
 			reason[common.ModeNotFit]++
 			klog.V(4).InfoS("Node filtered: Node does not support hami-core mode", "node", nodeInfo.Node.Name, "pod", pod.Name)
 			return false, nil, common.GenReason(reason, len(devices))
-		} else if nodeSupportHamiCore && !isHAMiCore {
-			reason[common.ModeNotFit]++
-			klog.V(4).InfoS("Node filtered: Reserved for hami-core but pod is legacy vNPU", "node", nodeInfo.Node.Name, "pod", pod.Name)
-			return false, nil, common.GenReason(reason, len(devices))
 		}
 	}
```
mesutoezdil, line 452: removing this check allows legacy vNPU pods to schedule onto hami-core nodes. per CONTRIBUTING.md gate 2, changes affecting device allocation must be validated on real hardware. which device type and driver version was tested?

mesutoezdil:
the branch is out of sync with current master. 

also: line 185 in the PR branch has `"vNPU nor supported for multiple devices"` -- should be `"not"`. this also confirms the branch needs a rebase.

the 'what this PR does / why we need it' section in the PR description is empty.

pkg/device/ascend/device.go
ouyangluwei163: I have tested it again in a real-device environment. chip: 910B4 driver: 25.5.1

pkg/device/ascend/device.go
```diff
@@ -245,6 +240,8 @@ func (dev *Devices) PatchAnnotations(pod *corev1.Pod, annoInput *map[string]stri
 				} else {
 					_, temp := dev.trimMemory(int64(val.Usedmem))
 					info.Temp = temp
+					info.Memory = int64(val.Usedmem)
+					info.Core = val.Usedcores
```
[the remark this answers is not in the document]
ouyangluwei163: Intended, not a bug — an absent/0 core means "no compute cap": the limiter lets the pod compete for cores (memory is still hard-capped), so it neither fails on 0 cores nor runs unlimited, which we verified on a shared 910B4.

ouyangluwei163:
Two pods scheduled onto the **same** Ascend 910B4, both running the same continuous matmul load:

  ```yaml
  Pod A:
  huawei.com/Ascend910B4: 1
  huawei.com/Ascend910B4-memory: 16384
  # no -core requested

  Pod B :
  huawei.com/Ascend910B4: 1
  huawei.com/Ascend910B4-memory: 16384
  huawei.com/Ascend910B4-core: 30
  ```

With both pods saturating the shared card (npu-smi AICore 100%), measured throughput is ~1:1 (~49 vs ~49
TFLOPS) instead of the expected ~1:30
image

pkg/device/ascend/device.go
```diff
@@ -443,24 +445,11 @@ func (npu *Devices) Fit(devices []*device.DeviceUsage, request device.ContainerD
...
-	}
-
 	if isHAMiCore && !nodeSupportHamiCore {
 		reason[common.ModeNotFit]++
 		klog.V(4).InfoS("Node filtered: pod requests hami-core but node does not support it", "pod", klog.KObj(pod))
 		return false, nil, common.GenReason(reason, len(devices))
 	}
-
-	if request.Memreq > 0 && request.Memreq < totalMemPerCard && request.Nums > 0 {
-		if nodeSupportHamiCore && !isHAMiCore {
```
DSFans2014, line 458: @ouyangluwei163 Can we now support both vnpu-core and template on the same card? In other words, after allocating part of the card's resources using a template, can we still use soft partitioning on the remaining portion?

ouyangluwei163: I haven’t tested this scenario yet, but using different modes on different NPUs within the same node should be possible.

DSFans2014: I think it would be better to run some tests. If the issue does exist, we will need to enforce a restriction that prevents them from sharing the same card.

ouyangluwei163: Thanks for raising this. With ascend-device-plugin PR [#106](https://github.com/Project-HAMi/ascend-device-plugin/pull/106), this can't happen because mode is uniform per
  node:

  - template node: explicit hami-core pods are rejected by Fit (isHAMiCore && !nodeSupportHamiCore →
  ModeNotFit); others run as template.
  - hami-core node: explicit hami-core pods run as hami-core, and annotation-less pods follow the node (PR
  [#106](https://github.com/Project-HAMi/ascend-device-plugin/pull/106)) instead of falling back to template.

  So every container on a node uses that node's mode — template and hami-core containers can never share the
  same card

DSFans2014: I got. There is currently an issue that may confuse users: if the `hami-core` annotation is not added, the template may be used to trim the memory.
https://github.com/Project-HAMi/HAMi/blob/53da8247b1c2868b9b70de8fbf5462542950375b/pkg/device/ascend/device.go#L167-L174

ouyangluwei163: You're right — MutateAdmission runs before scheduling and doesn't know the target node, so the memory trim can only follow the annotation, not the node. Any suggestion on how you'd handle this?

DSFans2014: The template mechanism is somewhat similar to NVIDIA's MIG. Currently, NVIDIA MIG doesn't modify the request values at the webhook stage. I'm not clear on why Ascend does.  Can we remove this modification logic and trim the memory in the scheduling stage?
cc @archlitchi

archlitchi: We don't trim NVIDIA here because we use a unified resource name(nvidia.com/gpu) for all NVIDIA types, so we don't know which type of GPU it will use, but for Ascend, the device type is specified in the resourceName(huawei.com/Ascend910B4), so it would be better to trim the memory here

DSFans2014: The current implementation has the following points that may cause confusion for users
- `hami-core` may be invoked without being explicitly declared
- memory may be trimmed by template when using  `hami-core`

> But a pod must commit to a mode up front via the hami.io/vnpu-mode annotation: an unannotated pod is
treated as template-only, and HAMi-core nodes reject any pod not marked hami-core. So a workload
is hard-bound to one mode at submission and can't use whichever node has free capacity 

In #2034, the core problem is that the cards on hami-core-labeled nodes are not being fully utilized.  The root cause is that the hami-core configuration is too coarse-grained. If we support card-level hami-core, we can improve card utilization and flexibility. 

What do you think? @ouyangluwei163 @archlitchi

ouyangluwei163: Great idea — I've considered this before too. Let's track it in a separate issue.

For now, this PR focuses on the #1889 roadmap item "Heterogeneous Ascend Mode (vNPU and HAMi-core) Support".

My understanding of the roadmap intent:
- No -core requested: best-effort scheduling to maximize scheduling success — similar to how NVIDIA is handled.
- -core requested: the user already understands the HAMi-core mechanism, so no extra annotation is needed — lowering the configuration barrier.

DSFans2014: If this is a new mode and it indicates that a best-effort scheduling solution is provided, then it makes sense. Now, there are three mode as below:
- template mode
- hami-core mode
- heterogeneous mode

But how does the user specify to use heterogeneous mode?

ouyangluwei163: Here are the three modes — the request itself selects the mode, with **template as the default**:

  ### Template (default)
  No `-core`, no annotation → memory aligned to template, template nodes only.

  ```yaml
  spec:
    containers:
    - name: app
      resources:
        limits:
          huawei.com/Ascend910B4: 1
          huawei.com/Ascend910B4-memory: 16384
  ```

  ### HAMi-core
  `-core` → soft partition, raw memory, hami-core nodes only. No annotation needed.

  ```yaml
  spec:
    containers:
    - name: app
      resources:
        limits:
          huawei.com/Ascend910B4: 1
          huawei.com/Ascend910B4-memory: 16384
          huawei.com/Ascend910B4-core: 30
  ```

  ### Auto
  `vnpu-mode: auto` → template-aligned memory, best-effort scheduling on both node types.

  ```yaml
  metadata:
    annotations:
      huawei.com/vnpu-mode: auto
  spec:
    containers:
    - name: app
      resources:
        limits:
          huawei.com/Ascend910B4: 1
          huawei.com/Ascend910B4-memory: 16384
  ```

ouyangluwei163: If this looks good to everyone, I'll adjust the code accordingly.

archlitchi: i think a pod without annotation can be explained as 'auto' mode, just like nvidia

archlitchi: but i don't think it's good to trim the Memory for pod with 'auto' mode here, we should leave it a to-do  in v2.11

archlitchi (approved):
/lgtm

PR #2038, opened 2026-07-09
remove dra from charts

archlitchi:
/kind cleanup

The right way to use HAMi-DRA is by following the document here: 
https://project-hami.io/docs/installation/how-to-use-hami-dra

It's not elegant to mix that part with existing HAMi chart

## Summary by CodeRabbit

* **New Features**
  * Simplified default chart configuration by removing DRA-related settings.
* **Bug Fixes**
  * Updated Helm rendering logic so device-plugin and scheduler manifests are no longer skipped based on DRA settings, and are included based on their own enablement flags.
  * Admission webhook–related scheduler resources (including cert-manager and RBAC pieces) now follow the intended admissionWebhook setting only.
* **Chores**
  * Removed the external DRA chart dependency to reduce installation/rendering surprises.


archlitchi:
CC @Shouren

wawa0210:
----    
cc @S descendants

charts/hami/templates/scheduler/clusterrole.yaml
Shouren: @archlitchi Shall we remove this apiGroups which is only required by HAMi-DRA?

charts/hami/Chart.yaml
```diff
@@ -23,12 +23,6 @@ version: 2.9.0
 # It is recommended to use it with quotes.
```
mesutoezdil, line 23: values.yaml still has the dra block and the hami-dra block with the comment saying the scheduler extender will not be installed when dra is enabled. 
and after this pr merges, that config becomes dead and the comment is wrong? was values.yaml cleanup intentionally left out?

charts/hami/templates/scheduler/clusterrole.yaml
archlitchi: good catch

charts/hami/Chart.yaml
archlitchi: no, values.yaml has been cleaned up here: https://github.com/Project-HAMi/HAMi/pull/2038/changes#diff-05dc481173cc137563eb327a4d82b6e56d109f35858a666ad50d2a4953a27dd6

Shouren (approved):
/lgtm

PR #2042, opened 2026-07-09
fix: resolve libvgpu git metadata path for linked worktree Docker builds

### LINKED ISSUE #2041: [Bug] docker build fails from linked git worktree because Dockerfiles assume `.git/modules/libvgpu` exists

## What happened

Building HAMi from a linked Git worktree fails very early in the Docker build because the Dockerfiles assume the submodule git metadata always exists at `.git/modules/libvgpu` inside the repository root.

At least these Dockerfiles have that assumption:
- `docker/Dockerfile`
- `docker/Dockerfile.hamicore`

They do this:
- `COPY .git/modules/libvgpu /libvgpu-git`
- then rewrite `/libvgpu/.git` to point to `/libvgpu-git`

This works in a normal checkout where the repository root has a real `.git/` directory.
But in a linked worktree:
- the repository root `.git` is a file, not a directory
- the submodule gitdir is not available at `.git/modules/libvgpu`
- the Docker build context therefore does not contain that path

As a result, the build fails before it even reaches the actual `libvgpu` build logic.

A concrete failure looks like this:

```text
[nvbuild 4/8] COPY .git/modules/libvgpu /libvgpu-git
ERROR: failed to calculate checksum of ref ...: "/.git/modules/libvgpu": not found

ERROR: failed to build: failed to solve: failed to compute cache key: failed to calculate checksum of ref ...: "/.git/modules/libvgpu": not found
```

HAMi Docker builds should work the same way from:

a normal Git checkout
a linked Git worktree

The build should not depend on a checkout-local .git/modules/libvgpu layout.

Create a linked worktree from the HAMi repository:

`git worktree add .worktrees/repro-worktree -b repro/docker-worktree-gitdir`

Initialize the libvgpu submodule inside that worktree:

```bash
cd .worktrees/repro-worktree
git submodule update --init --recursive
```

Build the nvbuild stage from `docker/Dockerfile.hamicore`:

`docker build --target nvbuild -f docker/Dockerfile.hamicore .`

Observe the failure at:

`COPY .git/modules/libvgpu /libvgpu-git`

I reproduced the same issue from a linked worktree with current HAMi source.

This looks like a development/build compatibility bug rather than a runtime bug.

The root problem is not “Docker needs Git in general”.
The root problem is that the current Dockerfiles hardcode one specific Git/submodule metadata layout that is only valid for a normal checkout.

In a linked worktree, the repository root uses:

`.git` as a file

And the libvgpu submodule gitdir resolves under a worktree-specific path similar to:

`.git/worktrees/<worktree>/modules/libvgpu`

So the current `COPY .git/modules/libvgpu ...` logic is not portable across supported Git working tree layouts.

## A more portable direction may be

stop copying submodule Git internals into the image.

Pass the needed libvgpu commit/branch metadata explicitly as build args or environment variables.
Let libvgpu consume those values preferentially, rather than requiring Git metadata in the container build context.

This report is mainly to check whether maintainers also consider linked worktree builds a supported / desirable developer workflow.

pingxin403:
### What does this PR do?

Fixes #2041 — Docker build fails when the repository is a linked git worktree because `.git/modules/libvgpu` does not exist (in worktrees, `.git` is a file, not a directory).

### Root Cause

The Dockerfile uses `COPY .git/modules/libvgpu /libvgpu-git` to preserve libvgpu's git metadata for version detection inside the container. In a linked worktree, the repository root `.git` is a text file pointing to the real gitdir, so `.git/modules/libvgpu` is not a valid path.

### Fix

In `hack/build.sh`'s `docker_build()` function: detect worktrees via `[ -f .git ]`, use `git rev-parse --git-common-dir` to find the real modules path, and temporarily create `.git/modules/libvgpu` for the Docker build context. The `.git` file is restored on exit via a trap.

### Testing

- ✅ Normal checkout: `[ -f .git ]` is false, code path skipped — no change in behavior
- ✅ Linked worktree: verified path resolution logic with `git rev-parse --git-common-dir`
- ✅ Cleanup: trap restores `.git` file on both success and failure

### AI Assistance Disclosure

This PR was developed with the assistance of Claude Code (Anthropic) for code generation and debugging. The author reviewed, tested, and takes full responsibility for all changes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved build reliability when the project is checked out using linked Git worktrees, preventing Docker builds from failing due to missing or unexpected repository metadata.
  * Ensures required Git module metadata is correctly available during the Docker build, while keeping the existing build and image-tagging flow unchanged.

### DIFF
--- a/hack/build.sh
+++ b/hack/build.sh
@@ -46,14 +46,12 @@ function docker_build() {
       if [ -n "$_git_common" ] && [ -d "$_git_common/modules/libvgpu" ]; then
         _restore_git=$(mktemp)
         cp .git "$_restore_git"
+        # shellcheck disable=SC2064
+        trap "rm -rf .git && cp '$_restore_git' .git && rm -f '$_restore_git'" EXIT
         rm -f .git
         mkdir -p .git/modules
         cp -r "$_git_common/modules/libvgpu" .git/modules/
       fi
-    fi
-    if [ -n "$_restore_git" ]; then
-      # shellcheck disable=SC2064
-      trap "rm -rf .git && cp '$_restore_git' .git && rm -f '$_restore_git'" EXIT
     fi
 
     docker build --build-arg VERSION="${VERSION}" --build-arg GOLANG_IMAGE=${GOLANG_IMAGE} --build-arg NVIDIA_IMAGE=${NVIDIA_IMAGE} --build-arg DEST_DIR=${DEST_DIR} -t "${IMAGE}:${VERSION}" -f docker/Dockerfile .
--- a/hack/build.sh
+++ b/hack/build.sh
@@ -34,6 +34,28 @@ function go_build() {
 }
 
 function docker_build() {
+    # Linked git worktrees have .git as a file, not a directory.
+    # Docker build context then lacks .git/modules/libvgvu, which
+    # the Dockerfile COPYs to preserve libvgpu git metadata for
+    # 'git describe' inside the container build.
+    # Temporarily create it; restore on exit.
+    local _restore_git=""
+    if [ -f .git ]; then
+      local _git_common
+      _git_common=$(git rev-parse --git-common-dir 2>/dev/null)
+      if [ -n "$_git_common" ] && [ -d "$_git_common/modules/libvgpu" ]; then
+        _restore_git=$(mktemp)
+        cp .git "$_restore_git"
+        rm -f .git
+        mkdir -p .git/modules
+        cp -r "$_git_common/modules/libvgpu" .git/modules/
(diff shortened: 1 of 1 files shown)


spencercjh:
`RUN bash ./build.sh` is called after `COPY .git/modules/libvgpu /libvgpu-git`.

I think your fix won't work. Please check and test yourself or your agents, then create the PR.

pingxin403:
Close for now — the fix needs to be in docker_build() context preparation (pre-build), not in build.sh (runtime). Will revisit with correct approach.

PR #2043, opened 2026-07-09
fix(device-plugin): avoid nil pointer panic in CheckHealth when checkHealth returns nil

Wangmin362:
**What type of PR is this?**

/kind bug

The NVIDIA device plugin crashes with a nil pointer panic in the health-check goroutine. `nvmlResourceManager.CheckHealth` calls `err.Error()` on the error returned by `checkHealth`, but `checkHealth` legitimately returns a nil error in several cases: health checks disabled via `DP_DISABLE_HEALTHCHECKS=all`, NVML init fails while `fail-on-init-error` is off, and the stop channel is closed on plugin shutdown/restart. In all three, `CheckHealth` does `nil.Error()` and panics. The goroutine is started from `Start()` with no recover, so the panic kills the whole device-plugin process.

Symptoms: setting `DP_DISABLE_HEALTHCHECKS=all` puts the plugin into CrashLoopBackOff; on SIGHUP / kubelet-socket recreation / the 30s restart, the intended in-process restart panics and the container restarts instead. This PR guards the nil error before dereferencing it.

NONE

Tested with a regression unit test that sets `DP_DISABLE_HEALTHCHECKS=all` and asserts `CheckHealth` returns without panicking — it fails on master with a nil pointer dereference and passes with the fix.

Also verified end to end on real hardware — a node with 2x NVIDIA GeForce GTX 1080 Ti, driver 550.144.03, Kubernetes v1.29. I built the plugin from master (before) and with the fix (after) and ran each in an isolated namespace under a separate resource name so it wouldn't disturb the production plugin:
- `DP_DISABLE_HEALTHCHECKS=all`: before -> CrashLoopBackOff, panic at `nvml_manager.go` in `CheckHealth`; after -> Running.
- SIGHUP to the running plugin: before -> `Received SIGHUP, restarting.` immediately followed by the panic and a container restart; after -> re-registers with kubelet, no panic, no restart.

This PR was written with AI assistance (analysis and drafting); the diagnosis, the fix, and the on-hardware testing were done and reviewed by me.

Yes — the NVIDIA device plugin no longer crashes with a nil pointer panic when health checks are disabled (`DP_DISABLE_HEALTHCHECKS`) or when the plugin restarts.

### DIFF
--- a/pkg/device-plugin/nvidiadevice/nvinternal/rm/health_checkhealth_test.go
+++ b/pkg/device-plugin/nvidiadevice/nvinternal/rm/health_checkhealth_test.go
@@ -1,0 +1,55 @@
+/*
+Copyright 2026 The HAMi Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package rm
+
+import (
+	"fmt"
+	"testing"
+	"time"
+)
+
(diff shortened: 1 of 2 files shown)


archlitchi (approved):
/lgtm

PR #2045, opened 2026-07-09
fix(scheduler): treat an empty use/nouse gpuuuid annotation as no constraint

Wangmin362:
**What type of PR is this?**

/kind bug

`CheckUUID` splits the `use-gpuuuid` / `nouse-gpuuuid` annotation on `,` and keeps a device only if its id equals one of the parts. When the annotation is present but its value is empty, `strings.Split("", ",")` returns `[""]` and no real device id equals `""`, so an empty `nvidia.com/use-gpuuuid` (or the per-vendor equivalent, since all vendors share `CheckUUID`) makes every device fail the check and the node becomes unschedulable with `CardUuidMismatch`. This is easy to hit when the annotation value is templated and renders empty; an empty allow-list should mean "no constraint", not "match nothing".

This PR skips the check when the value is empty or whitespace. The `use` branch is the actual bug; the `nouse` branch gets the same guard only for symmetry (an empty exclude-list already excludes nothing, so its behaviour is unchanged).

NONE

Extended the existing `TestCheckUUID` with empty / whitespace-only `use-gpuuuid` and empty `nouse-gpuuuid` cases; the `use` cases fail on master (device filtered out) and pass with the fix.

Also verified end to end on real hardware — NVIDIA GeForce GTX 1080 Ti, driver 550.144.03, Kubernetes v1.29 — running the scheduler (master before / fixed after) as a separate scheduler with its own `schedulerName` in an isolated namespace:
- `nvidia.com/use-gpuuuid: ""`: before -> Pod stays Pending with `CardUuidMismatch`; after -> Pod is scheduled.
- No annotation (control): scheduled in both, confirming the empty annotation is the only difference.
- Regression check on the fixed build: a non-matching UUID still yields `CardUuidMismatch`, a matching UUID schedules — real UUID filtering is preserved.

This PR was written with AI assistance (analysis and drafting); the diagnosis, the fix, and the on-hardware testing were done and reviewed by me.

Yes — a pod with an empty `use-gpuuuid`/`nouse-gpuuuid` annotation is no longer wrongly rejected with `CardUuidMismatch`; an empty value is now treated as no constraint.

## Summary by CodeRabbit

* **Bug Fixes**
  * Empty or whitespace-only annotation values no longer unintentionally filter out devices.
  * UUID and device-type include/exclude logic now treats blank constraints as “no restriction” rather than enforcing or excluding based on empty strings.
* **Tests**
  * Expanded unit tests for UUID and device-type filtering edge cases, including new cases for empty and whitespace-only annotation values across supported device types.

### DIFF
--- a/pkg/device/nvidia/device.go
+++ b/pkg/device/nvidia/device.go
@@ -20,7 +20,6 @@ import (
 	"errors"
 	"flag"
 	"fmt"
-	"slices"
 	"strconv"
 	"strings"
 	"sync"
@@ -485,25 +484,7 @@ func resourcePresent(ctr *corev1.Container, name corev1.ResourceName) bool {
 }
 
 func checkGPUtype(annos map[string]string, cardtype string) bool {
-	cardtype = strings.ToUpper(cardtype)
-	// Empty value means "no constraint"; otherwise strings.Contains("") matches every type.
-	if inuse, ok := annos[GPUInUse]; ok && strings.TrimSpace(inuse) != "" {
-		useTypes := strings.Split(inuse, ",")
-		if !slices.ContainsFunc(useTypes, func(useType string) bool {
-			return strings.Contains(cardtype, strings.ToUpper(useType))
-		}) {
-			return false
-		}
-	}
-	if unuse, ok := annos[GPUNoUse]; ok && strings.TrimSpace(unuse) != "" {
-		unuseTypes := strings.Split(unuse, ",")
-		if slices.ContainsFunc(unuseTypes, func(unuseType string) bool {
-			return strings.Contains(cardtype, strings.ToUpper(unuseType))
-		}) {
-			return false
-		}
-	}
-	return true
+	return device.CheckType(annos, cardtype, GPUInUse, GPUNoUse)
 }
 
 func assertNuma(annos map[string]string) bool {
--- a/pkg/device/nvidia/device.go
+++ b/pkg/device/nvidia/device.go
@@ -486,7 +486,8 @@ func resourcePresent(ctr *corev1.Container, name corev1.ResourceName) bool {
 
 func checkGPUtype(annos map[string]string, cardtype string) bool {
 	cardtype = strings.ToUpper(cardtype)
-	if inuse, ok := annos[GPUInUse]; ok {
+	// Empty value means "no constraint"; otherwise strings.Contains("") matches every type.
+	if inuse, ok := annos[GPUInUse]; ok && strings.TrimSpace(inuse) != "" {
 		useTypes := strings.Split(inuse, ",")
 		if !slices.ContainsFunc(useTypes, func(useType string) bool {
 			return strings.Contains(cardtype, strings.ToUpper(useType))
@@ -494,7 +495,7 @@ func checkGPUtype(annos map[string]string, cardtype string) bool {
 			return false
 		}
 	}
-	if unuse, ok := annos[GPUNoUse]; ok {
+	if unuse, ok := annos[GPUNoUse]; ok && strings.TrimSpace(unuse) != "" {
 		unuseTypes := strings.Split(unuse, ",")
 		if slices.ContainsFunc(unuseTypes, func(unuseType string) bool {
 			return strings.Contains(cardtype, strings.ToUpper(unuseType))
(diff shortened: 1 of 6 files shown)


mesutoezdil:
checkGPUtype in pkg/device/nvidia/device.go has the same bug on the nouse-gputype branch, strings.Contains(cardtype, "") is always true so an empty nvidia.com/nouse-gputype annotation excludes every gpu type

mesutoezdil:
checkDCUtype in pkg/device/hygon/device.go has the same bug on the nouse-dcutype branch, strings.Contains(cardtype, "") is always true so an empty hygon.com/nouse-dcutype annotation excludes every dcu type

Wangmin362:
> checkDCUtype in pkg/device/hygon/device.go has the same bug on the nouse-dcutype branch, strings.Contains(cardtype, "") is always true so an empty hygon.com/nouse-dcutype annotation excludes every dcu type

Good catch — you're right. Fixed both nvidia (checkGPUtype) and hygon (checkDCUtype); the other vendors don't have this issue. Verified the nvidia path on a real GTX 1080 Ti (empty nouse-gputype now schedules; real type filtering preserved). Thanks!

pkg/device/nvidia/device.go
```diff
@@ -486,15 +486,16 @@ func resourcePresent(ctr *corev1.Container, name corev1.ResourceName) bool {
 
 func checkGPUtype(annos map[string]string, cardtype string) bool {
 	cardtype = strings.ToUpper(cardtype)
-	if inuse, ok := annos[GPUInUse]; ok {
+	// Empty value means "no constraint"; otherwise strings.Contains("") matches every type.
```
DSFans2014: I think it's better to extract a common function, similar to `CheckUUID`
https://github.com/Project-HAMi/HAMi/blob/d09c663651291d7d51b3854d5bf60ffcfb9654a8/pkg/device/devices.go#L676

Wangmin362: Good idea, done in da9254c — extracted a shared CheckType in devices.go (mirrors CheckUUID), both nvidia and hygon call it now. Tested on A100 + Hygon DCU.

DSFans2014 (commented):
/lgtm

mesutoezdil:
> > checkDCUtype in pkg/device/hygon/device.go has the same bug on the nouse-dcutype branch, strings.Contains(cardtype, "") is always true so an empty hygon.com/nouse-dcutype annotation excludes every dcu type
> 
> Good catch — you're right. Fixed both nvidia (checkGPUtype) and hygon (checkDCUtype); the other vendors don't have this issue. Verified the nvidia path on a real GTX 1080 Ti (empty nouse-gputype now schedules; real type filtering preserved). Thanks!

reminder for future: per the contribution guidelines (gate 6), replies in pr review threads must be written directly by the author, not generated by ai tools.

Wangmin362:
> > > checkDCUtype in pkg/device/hygon/device.go has the same bug on the nouse-dcutype branch, strings.Contains(cardtype, "") is always true so an empty hygon.com/nouse-dcutype annotation excludes every dcu type
> > 
> > 
> > Good catch — you're right. Fixed both nvidia (checkGPUtype) and hygon (checkDCUtype); the other vendors don't have this issue. Verified the nvidia path on a real GTX 1080 Ti (empty nouse-gputype now schedules; real type filtering preserved). Thanks!
> 
> reminder for future: per the contribution guidelines (gate 6), replies in pr review threads must be written directly by the author, not generated by ai tools.


It wasn't generated entirely by AI. I wrote it in Chinese first, then had the AI translate it into English and polish it up a bit. My English isn't very good.

mesutoezdil:
> > > > checkDCUtype in pkg/device/hygon/device.go has the same bug on the nouse-dcutype branch, strings.Contains(cardtype, "") is always true so an empty hygon.com/nouse-dcutype annotation excludes every dcu type
> > > 
> > > 
> > > Good catch — you're right. Fixed both nvidia (checkGPUtype) and hygon (checkDCUtype); the other vendors don't have this issue. Verified the nvidia path on a real GTX 1080 Ti (empty nouse-gputype now schedules; real type filtering preserved). Thanks!
> > 
> > 
> > reminder for future: per the contribution guidelines (gate 6), replies in pr review threads must be written directly by the author, not generated by ai tools.
> 
> It wasn't generated entirely by AI. I wrote it in Chinese first, then had the AI translate it into English and polish it up a bit. My English isn't very good.

no problem mate, it was just a reminder :) thanks for your works, you do all amazing, i like your prs :) always direct lgtm, all clear :) have a great day!

archlitchi (approved):
/lgtm

PR #2052, opened 2026-07-10
Remove 'deleted' annotations when registering

archlitchi:
/kind cleanup

we don't need to maintain a  'deleted' status on node annotations, we can simply remove that part

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated health-check handling to remove special-case behavior for “Deleted” handshake states across supported device types.
  * Improved node cleanup by fully removing handshake annotation keys rather than writing timestamped “Deleted_*” markers.
  * Adjusted health and reset expectations to keep reporting consistent for expired and unknown handshake scenarios.
* **Tests**
  * Revised health and cleanup test cases to match the updated handshake-state behavior and annotation-removal behavior.


archlitchi:
CC @Shouren

wawa0210:
----    
cc @S descendants

pkg/util/util_test.go
Shouren: @archlitchi I think we should also change the name of function like `TestRemoveAnnotations` or else

pkg/util/util.go
```diff
@@ -226,14 +225,36 @@ func InitKlogFlags() *flag.FlagSet {
...
 	n, err := GetNode(nn)
 	if err != nil {
 		klog.Errorln("get node failed", err.Error())
 		return err
 	}
-	return PatchNodeAnnotations(n, tmppat)
+	return RemoveNodeAnnotation(n, devType)
+}
+
+func RemoveNodeAnnotation(node *corev1.Node, annotationKey string) error {
```
DSFans2014: it's better to pass a list? If we want to delete multiple keys at once, we'd need to call this function repeatedly and send multiple requests.

archlitchi: adopted

pkg/util/util_test.go
archlitchi: adopted

DSFans2014 (commented):
/lgtm

ouyangluwei163 (commented):
/lgtm

PR #2054, opened 2026-07-10
fix: skip owner check for legacy lock format

DSFans2014:
**What type of PR is this?**

/kind bug

When using HAMi vNPU with Volcano and enabling node lock, the node lock value stored in Volcano is a timestamp. However, the `ReleaseNodeLock` function called in `ascend-device-plugin` expects the lock value to contain the pod's namespace and name. Since the timestamp format does not match the lock is never released.

Fixes #

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed node lock release handling for legacy timestamp-only annotation values, allowing locks to be released when the stored format is older than the current one.
* **Tests**
  * Expanded node lock release test coverage with new cases for legacy-format locks and additional post-release verification that the lock annotation is removed.

### DIFF
--- a/pkg/util/nodelock/nodelock_test.go
+++ b/pkg/util/nodelock/nodelock_test.go
@@ -241,9 +241,10 @@ func TestReleaseNodeLock(t *testing.T) {
 		timeout  bool
 	}
 	tests := []struct {
-		name    string
-		args    args
-		wantErr bool
+		name          string
+		args          args
+		wantErr       bool
+		checkNodeLock bool
 	}{
 		{
 			name: "node not found",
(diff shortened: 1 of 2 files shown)


pkg/util/nodelock/nodelock_test.go
```diff
@@ -325,6 +325,27 @@ func TestReleaseNodeLock(t *testing.T) {
...
+				},
+				pod: &corev1.Pod{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      "hami",
+						Namespace: "hami-ns",
+					},
+				},
+			},
+			wantErr: false,
+		},
```
mesutoezdil, line 351: this test passes with both old and new code since both return nil, the test should also assert that the node annotation is removed after the call to confirm the lock was actually released.

mesutoezdil:
first commit message uses feat: but this is a bug fix, it should be fix: to match the pr title and conventional commits.

archlitchi (approved):
/lgtm

PR #2055, opened 2026-07-10
fix: return deep copies from ListPodsInfo to prevent data race

mesutoezdil:
**What type of PR is this?**

/kind bug

`ListPodsInfo` returned raw `*PodInfo` pointers from the internal map. The read lock is released before callers in `getNodesUsage` read `p.NodeID`, `p.Devices`, and store the pointer in device state. A concurrent `AddPod` or `UpdatePod` can write through the same pointer at the same time, causing a data race.

`PodInfo.DeepCopy()` already existed. This PR uses it so callers get an independent snapshot. One line change.

Same class of race fixed in #1967.

NONE

This change is scoped to the scheduler. Unit test added: `TestListPodsInfoReturnsDeepCopy` mutates the returned slice and checks the internal map is unchanged. Passes with `-race`.

No

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved pod information listing so returned data is isolated from internal records.
  * Changes made to listed pod details no longer unintentionally alter cached pod information.

### DIFF
--- a/pkg/device/pod_test.go
+++ b/pkg/device/pod_test.go
@@ -543,6 +543,22 @@ func TestContainerDeviceDeepCopy(t *testing.T) {
 	assert.False(t, exists, "original CustomInfo should not have key2")
 }
 
+func TestListPodsInfoReturnsDeepCopy(t *testing.T) {
+	pm := NewPodManager()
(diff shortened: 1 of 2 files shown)


DSFans2014 (commented):
/lgtm

archlitchi (approved):
/lgtm

PR #2064, opened 2026-07-14
docs: design for init-container GPU resource accounting

maishivamhoo123:
Currently HAMi sums init and app container requests as if they run together, causing false quota passes, scheduler rejections, and inflated usage that never shrinks. This PR corrects all three layers – admission, scheduling, and usage recording – to use the correct formula effective = max(sum(app), max(init)). Usage is stored as the peak value and automatically reduced to app‑only once init containers terminate. Pod annotations remain untouched.
Related to  #1667 
#1773 

## Summary by CodeRabbit

* **Documentation**
  * Added a design document describing improved GPU resource accounting for workloads that include init containers.
  * Clarifies consistent behavior across admission checks, scheduling fit/scoring, and usage tracking throughout init and app phases.
  * Explains how resource accounting is adjusted after successful init completion and how usage is corrected on updates and deletions.
  * Includes safeguards to prevent repeated or incorrect post-init adjustments.

### DIFF
--- a/docs/develop/initContainer-design.md
+++ b/docs/develop/initContainer-design.md
@@ -2,12 +2,12 @@
 
 ## Problem Summary
 
-When a pod has both init containers and app containers requesting gpu resources , HAMi allocate the resources simultaneously/parallelly. But kubernetes runs the init container sequentially to completion before any app container starts,so init and app containers never execute at the same time.
+When a pod has both init containers and app containers requesting GPU resources, HAMi allocates the resources simultaneously/parallelly. But Kubernetes runs the init container sequentially to completion before any app container starts, so init and app containers never execute at the same time.
   
 
 ## The Problem in HAMi Today
 
-`device.Resourcereqs()` this is build for one request entry per container, **init
+`device.Resourcereqs()` this is built for one request entry per container, **init
 containers first, then app containers**, in this order.
 
 **1. Admission — `fitResourceQuota` (`webhook.go`)**
@@ -73,10 +73,10 @@ effective = max( sum(app container requests), max(single init container request)
 
 ## Proposal
 
-Applying this formula consistently, for all the resource dimension GPU count,
-memory, cores and per device UUID(a multi-GPU pod where init and app
-containers land on different physical devices must not have those
-devices usage merged together):
+Applying this formula consistently for all resource dimensions (GPU count,
+memory, cores, and per-device UUID; a multi-GPU pod where init and app
+containers land on different physical devices must not have usage on
+those devices merged):
 
 - **Admission quota check** computes this effective value and checks it
   against namespace quota.
@@ -146,6 +146,13 @@ different physical devices: each device gets its own effective value.
 2. `appReq` per resource = **sum** across app containers.
 3. `effectiveReq = max(appReq, initReq)`; deny if it exceeds quota.
 
+**GPU count:** today's quota check enforces **memory and cores only** —
+GPU count is not an independent quota dimension. `effectiveReq` is still
+computed for count (steps 1–2) so the scheduler and usage recorder have a
+consistent value to work with, but `FitQuota` does not compare it against
+a namespace quota. Threading count through `FitQuota` the same way memory
+and cores are is a possible future extension, not part of this change.
+
 **Memory factor:** applied **once**, to the already-computed effective
 value — not per container.
 
@@ -198,13 +205,23 @@ itself derived from container statuses by kubelet, so relying on only one
 risks acting on a momentarily stale read of the other.
 
 ```
-new_usage[uuid] = sum of app-container requests on uuid only
-delta[uuid]     = new_usage[uuid] - old_usage[uuid]  
+new_usage[uuid] = sum of app-container usage values on uuid only
+                  (the same per-container fields — e.g. Usedmem — that
+                  AddUsage/getNodesUsage already collapse; not raw
+                  requests)
+delta[uuid]     = new_usage[uuid] - old_usage[uuid]
 apply delta[uuid] to QuotaManager and PodManager
 ```
+
+Requests stay the basis for admission and scheduling decisions above;
+once a pod is running, the recorded/shrunk value — the number quota and
+node-capacity accounting compare against — is always derived from the
+same usage fields `AddUsage`/`getNodesUsage` already track, so it can
+never drift from what `CollapseInitContainerUsage` produced when the pod
+started.
 
 Only ever runs **after** `pod.Status` confirms completion.
 
 **Idempotency:** `PodManager` stores a boolean `initContainersShrunk` flag
 per pod. The shrink applies only while `false`; once applied it flips to
-`true` and is never re-applie.
+`true` and is never re-applied.
--- a/docs/develop/initContainer-design.md
+++ b/docs/develop/initContainer-design.md
@@ -199,17 +199,39 @@ add and remove always operate on the same number, so there's no drift.
 
 ![The journey of one pod](./imgs/pod-lifecycle-journey.svg)
 
-**Condition:** `pod.Status.Phase == Running` **and** every init container
-has `Status.Terminated{ExitCode: 0}`. Checked together — `Phase` is
-itself derived from container statuses by kubelet, so relying on only one
-risks acting on a momentarily stale read of the other.
+**Condition:** every init container has `Status.Terminated` set (finished,
+regardless of exit code), checked on every reconcile against current
+object state — not gated on observing any particular `Phase` value.
+`Phase == Running` is not required, since it was found to be unreliable
+(does not consistently appear before `Succeeded`).
+
+Two actions (shrink) and one no‑op (hold) once every init container has terminated:
+
+- **Pod reaches a terminal phase (`Succeeded` or `Failed`):** every
+  container in the pod — init and app — has finished. Usage shrinks to
+  zero regardless of exit codes, since nothing in the pod is using GPU
+  anymore. Unlike `Running`, `Succeeded`/`Failed` are true terminal
+  states with no further transitions, so `Phase` is safe to use here.
+- **All init containers exited `ExitCode: 0`, pod not yet terminal:**
+  app containers have started or are starting → shrink to
+  app-containers-only usage.
+- **Any init container exited non-zero, pod not yet terminal:** pod may still restart (depending on restartPolicy). It still holds its allocated GPU devices, so usage is not shrunk at this point. The pod will either restart and eventually succeed, or be terminated permanently (phase → Failed), which will then trigger a shrink.
 
 ```
-new_usage[uuid] = sum of app-container usage values on uuid only
-                  (the same per-container fields — e.g. Usedmem — that
-                  AddUsage/getNodesUsage already collapse; not raw
-                  requests)
-delta[uuid]     = new_usage[uuid] - old_usage[uuid]
+for each uuid:
+if pod.Status.Phase in (Succeeded, Failed):
+    new_usage[uuid] = 0   
+else if all init containers terminated with ExitCode == 0:
+    new_usage[uuid] = sum of app-container usage values on uuid only
+                      (the same per-container fields — e.g. Usedmem —
+                      that AddUsage/getNodesUsage already collapse;
+                      not raw requests)
+else :
+   // init containers still running, or some exited non-zero, pod not terminal.
+  // No shrink: keep current effective usage unchanged.
+        continue   
+
+delta[uuid] = new_usage[uuid] - old_usage[uuid]
 apply delta[uuid] to QuotaManager and PodManager
 ```
 
@@ -222,6 +244,15 @@ started.
 
 Only ever runs **after** `pod.Status` confirms completion.
 
-**Idempotency:** `PodManager` stores a boolean `initContainersShrunk` flag
-per pod. The shrink applies only while `false`; once applied it flips to
-`true` and is never re-applied.
+**Idempotency:** `PodManager` stores a `PodShrinkState` per pod —
+`NotShrunk`, `AppOnly`, or `Zero` — replacing a single boolean, since the
+three outcomes above mean a pod can legitimately shrink twice (once to
+`AppOnly`, later to `Zero`), not just once.
+
+Allowed transitions:
+- `NotShrunk → AppOnly`: when all init containers exit `0` and the pod is
+  not yet terminal.
+- `NotShrunk → Zero` or `AppOnly → Zero`: when the pod reaches a terminal
+  phase(Succeeded or Failed).
+
+  (Note: the case “any init container exits non‑zero” is removed; it will naturally end up in a terminal phase later if the pod does not restart, so no separate branch is needed.)
--- a/docs/develop/initContainer-design.md
+++ b/docs/develop/initContainer-design.md
@@ -3,6 +3,8 @@
 ## Problem Summary
 
 When a pod has both init containers and app containers requesting GPU resources, HAMi allocates the resources simultaneously/parallelly. But Kubernetes runs the init container sequentially to completion before any app container starts, so init and app containers never execute at the same time.
+
+Note : This design covers init and app containers only. Sidecar containers are out of scope here and will be handled in a separate PR.
   
 
 ## The Problem in HAMi Today
@@ -70,6 +72,7 @@ So the correct formula for a pod's GPU footprint at any instant is:
 ```
 effective = max( sum(app container requests), max(single init container request) )
 ```
+**Assumption:**  The resources requested by the init container will always be the same as one of the app containers.
 
 ## Proposal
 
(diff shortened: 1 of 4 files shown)


maishivamhoo123:
@DSFans2014  @archlitchi  and @Shouren can i added the docs for #1773 can you please review this ?
Thank you!

docs/develop/initContainer-design.md
```diff
@@ -0,0 +1,227 @@
...
+
+```
+effective_mem = max(app_mem_sum, init_mem_peak)
+if memoryFactor > 1:
+    quota_check_mem = effective_mem * memoryFactor
+```
+
+### Scheduler Fit & Scoring
+
+1. **Init pass:** fit each init container independently against a fresh
```
DSFans2014, line 170: If the reqNum value for all initContainers is 1, will all initContainers be allocated on the same cardID?

docs/develop/initContainer-design.md
```diff
@@ -0,0 +1,227 @@
...
+
+**On deletion**, `TakeAndDeletePod` returns exactly the currently-stored
+(possibly already-shrunk) value, and `RmUsage` subtracts exactly that —
+add and remove always operate on the same number, so there's no drift.
+
+### Shrink on Init Container Completion
+
+![The journey of one pod](./imgs/pod-lifecycle-journey.svg)
+
+**Condition:** `pod.Status.Phase == Running` **and** every init container
```
DSFans2014: Is there any issue if the main container exits immediately after being started?

docs/develop/initContainer-design.md
maishivamhoo123: @DSFans2014  thank you for review yes when all init containers have reqNum=1, each independent fresh-copy fit will select the same physical card whose score will be highest on a clean node state.because init containers are sequential, they can genuinely reuse the same device. The per-UUID max formula handles both outcomes correctly: if all land on GPU0, effective_mem[GPU0] = max(IC1, IC2, and so on) .
if they land on different devices, each gets its own independent peak. Neither case over-counts. 
am i correct ?

docs/develop/initContainer-design.md
maishivamhoo123: No i think there is no any problem because if the app container exits immediately and the pod goes straight to Succeeded/failed state before our shrink logic runs that's fine. because When the pod gets deleted, we always remove exactly whatever number is stored at that point. So even if the shrink never fired and we still have the full max(app, init) value stored, deletion will remove exactly that. End result is still zero. Nothing leaks. The initContainersShrunk flag just makes sure the shrink can't accidentally fire after deletion has already cleaned everything up.
am i correct ?

DSFans2014: > Condition: pod.Status.Phase == Running and every init container has Status.Terminated{ExitCode: 0}

My concern is whether the `Running` state will appear in this case. I'm not sure whether Kubernetes merges transient intermediate states and sends a single final state.

maishivamhoo123: @DSFans2014 thank you for review The Running phase will always appear, even if the app container exits immediately after init containers finish. Kubernetes does not merge transient phases.
as we can see in this function 
```
// updateStatusInternal updates the internal status cache, and queues an update to the api server if
// necessary.
// This method IS NOT THREAD SAFE and must be called from a locked function.
func (m *manager) updateStatusInternal(logger klog.Logger, pod *v1.Pod, status v1.PodStatus, forceUpdate, podIsFinished bool) (bool, *podStatusNotification) {
	var oldStatus v1.PodStatus
	cachedStatus, isCached := m.podStatuses[pod.UID]
	if isCached {
```

https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/status/status_manager.go#L849 

In updateStatusInternal, the new status is cached, its version is incremented, and a message is pushed onto podStatusChannel to trigger a sync. There is no any logic that skips intermediate phases. So whenever the kubelet starts the first app container, it computes a Running phase and calls SetPodStatus, which immediately triggers a patch to the API. If the container then terminates, a second status (Succeeded) is computed and sent in the same way. Both updates are visible to any informer or controller.
and i verified it locally also :- 
Screenshot 2026-07-16 101704

The sleep 1 just makes the phase long enough to observe manually.
does it make sense ?

docs/develop/initContainer-design.md
```diff
@@ -0,0 +1,227 @@
+# HAMi Support for Init Container GPU Resource Accounting — Design
+
+## Problem Summary
+
+When a pod has both init containers and app containers requesting GPU resources, HAMi allocates the resources simultaneously/parallelly. But Kubernetes runs the init container sequentially to completion before any app container starts, so init and app containers never execute at the same time.
```
mesutoezdil, line 5: this thread got marked resolved but sidecar containers are still not mentioned anywhere in the doc, same gap i found in pr 1773

docs/develop/initContainer-design.md
```diff
@@ -0,0 +1,227 @@
...
+Requests stay the basis for admission and scheduling decisions above;
+once a pod is running, the recorded/shrunk value — the number quota and
+node-capacity accounting compare against — is always derived from the
+same usage fields `AddUsage`/`getNodesUsage` already track, so it can
+never drift from what `CollapseInitContainerUsage` produced when the pod
+started.
+
+Only ever runs **after** `pod.Status` confirms completion.
+
+**Idempotency:** `PodManager` stores a boolean `initContainersShrunk` flag
```
mesutoezdil: this thread got marked resolved too but the doc still only describes one boolean flag, no crash safe delta reconciliation across the two managers

mesutoezdil:
issue 1667 was about wrong device assignment which pr 1650 already fixed, this doc only changes accounting and does not touch device assignment, might be better to change fixes 1667 to part of 1667

docs/develop/initContainer-design.md
DSFans2014: We can test it as follows. First, start the monitoring script
```
NS="phase-race"

rv=$(kubectl get pods -n "$NS" -o jsonpath='{.metadata.resourceVersion}' 2>/dev/null)

kubectl get --raw "/api/v1/namespaces/$NS/pods?watch=1&resourceVersion=$rv&allowWatchBookmarks=true" \
| jq -r --unbuffered 'select(.type != "BOOKMARK" and .object != null) | [.type, .object.metadata.name, .object.status.phase] | @tsv' \
| while IFS=$'\t' read -r type name phase; do
    timestamp=$(date '+%H:%M:%S.%3N') 
    echo "[$timestamp] $type | $name | $phase"
done

```
then start test pod as below:
```
NS=phase-race
ITER=5

kubectl create ns "$NS" --dry-run=client -o yaml | kubectl apply -f -

for i in $(seq 1 "$ITER"); do
P="race-$i"
cat <<EOF | kubectl apply -f -
apiVersion: v1
kind: Pod
metadata:
  name: $P
  namespace: $NS
  labels:
    race: phase
spec:
  restartPolicy: Never
  initContainers:
  - name: init-1
    image: busybox:latest
    command: ["sh","-c","sleep 0.2"]
  - name: init-2
    image: busybox:latest
    command: ["sh","-c","sleep 0.3"]
  containers:
  - name: main
    image: busybox:latest
    command: ["sh","-c","echo start; sleep 0.01; exit 0"]
EOF
kubectl -n "$NS" wait --for=jsonpath='{.status.phase}'=Succeeded pod/"$P" --timeout=120s >/dev/null || true
done

```
we may see the output as below:
ScreenShot_2026-07-16_155124_394

If there are any issues with the testing process, please let me know

docs/develop/initContainer-design.md
maishivamhoo123: @mesutoezdil This is totally intentional i will create a different  pr for the sidecar container as it have a lot of difference compared to the init container so i am thinking to raise a different issue and pr for sidecar container.
is it ok ?

docs/develop/initContainer-design.md
maishivamhoo123: @mesutoezdil the boolean flag is only there to avoid running the same shrink twice while the controller is still running. It’s not the crash-safety mechanism.For crashes, the real safety net is:-  every time the controller checks the pod, it will recalculate the correct usage directly from the pod annotations, spec, and status. So if we crash mid-shrink, on restart the next check will recompute the right numbers and apply only the difference (delta) to both managers. The flag is just a temporary inmemory guard that will disappears on restart, so the recalculation fixes everything automatically.
does it make sence?

docs/develop/initContainer-design.md
maishivamhoo123: @DSFans2014 thank you for pointing this out , I checked it properly with the help of the above conditions . I ran the test again and watched both Phase and initContainerStatuses together while creating pods live.
For race-1: both init containers show Terminated{ExitCode:0} while the pod is still Pending. Then it jumps straight to Succeeded. Running never shows up at all.
But for race-2 in the same run, Running does show up before Succeeded.

Screenshot 2026-07-16 150410

Screenshot 2026-07-16 145354


So it's random — same pod spec, but one pod skips Running and the next one doesn't. It depends on timing, not something we can control. That means we can't rely on Phase == Running in the condition, since it might just never happen for a given pod.
so i am thinking to remove pod.Status.Phase == Running  so the new condition will now become : all init containers show Terminated{ExitCode:0}. That field is stable once it's set and doesn't depend on this timing issue.

am i correct? can i move forward with this?

DSFans2014: Yeah, looks like Running status doesn’t always show up from this result. We should change the condition

maishivamhoo123:
@DSFans2014  i resolved all your comments can you please look in to this once and let me know is everything is perfect or not ?
Thank you!

docs/develop/initContainer-design.md
```diff
@@ -0,0 +1,258 @@
...
+Think of a pod like a kitchen: one chef (the **init container**) preps
+ingredients and leaves; only then does a second chef (the **app
+container**) start cooking. They are never both in the kitchen at once.
+
+![Init and app containers never overlap](./imgs/timeline-init-vs-app.svg)
+
+So the correct formula for a pod's GPU footprint at any instant is:
+
+```
+effective = max( sum(app container requests), max(single init container request) )
```
DSFans2014, line 73: I'm not sure whether a single machine can take different devices.  Should we also consider the scenario where a request contains multiple devices? @archlitchi

archlitchi: yes and no, a single request shouldn't contain multiple kinds of devices, but a single node can have multiple kinds of devices

DSFans2014: > a single request shouldn't contain multiple kinds of devices

Can different containers (including init containers ) in the same Pod request different devices? @archlitchi

Shouren: @DSFans2014 @archlitchi Can we simplify the scenario for init containers by assuming, for now, that the resources requested by the init container will always be the same as one of the app containers? If other needs come up later, we can implement them then.

docs/develop/initContainer-design.md
```diff
@@ -0,0 +1,258 @@
...
+
+- **Pod reaches a terminal phase (`Succeeded` or `Failed`):** every
+  container in the pod — init and app — has finished. Usage shrinks to
+  zero regardless of exit codes, since nothing in the pod is using GPU
+  anymore. Unlike `Running`, `Succeeded`/`Failed` are true terminal
+  states with no further transitions, so `Phase` is safe to use here.
+- **All init containers exited `ExitCode: 0`, pod not yet terminal:**
+  app containers have started or are starting → shrink to
+  app-containers-only usage.
+- **Any init container exited non-zero, pod not yet terminal:** pod may still restart (depending on restartPolicy). It still holds its allocated GPU devices, so usage is not shrunk at this point. The pod will either restart and eventually succeed, or be terminated permanently (phase → Failed), which will then trigger a shrink.
```
DSFans2014, line 221:
```
spec:
  restartPolicy: OnFailure
  initContainers:
  - name: init-1
    image: busybox
    command: ["sh","-c","sleep 0.2"]
    resources:
      limits:
        nvidia.com/gpu: 1
        nvidia.com/gpumem: 1024
  - name: init-2
    image: busybox
    command: ["sh","-c","sleep 0.3; exit 1"]
    resources:
      limits:
        nvidia.com/gpu: 1
        nvidia.com/gpumem: 512
  containers:
  ...
```
Considering the above scenario—where init-1 has completed successfully and init-2 is stuck in a restart loop due to failures (may be waiting for some condition to be met before it can succeed) —should the extra resources claimed by init-1 be released?

docs/develop/initContainer-design.md
archlitchi: > @DSFans2014 @archlitchi Can we simplify the scenario for init containers by assuming, for now, that the resources requested by the init container will always be the same as one of the app containers? If other needs come up later, we can implement them then.

yeah, we can assume this for now

docs/develop/initContainer-design.md
maishivamhoo123: @DSFans2014  currently we are using the all-or-nothing approach: if any init container fails, we don't shrink usage at all. But you're right we are not releasing the extra resource use by init-1. I think we can improve this with the watermark approach track the highest index of an init container that finished with exit 0 (since they run in order). After that, we only count the requests of the remaining init containers. In your example, once init-1 succeeds, the effective peak drops from 1024Mi to 512Mi (only init-2 matters). It's a bit more state to track (an index instead of a simple enum), but it's a utilization improvement, not a correctness fix — we'd never over-admit either way.

@archlitchi  @Shouren  @DSFans2014  What are your thoughts? Should we include this now, or keep it as a future improvement?
Thank you!

DSFans2014: I don't think this a critical problem. We can record a TODO item and address it in a future iteration.

docs/develop/initContainer-design.md
```diff
@@ -0,0 +1,258 @@
...
+once a pod is running, the recorded/shrunk value — the number quota and
+node-capacity accounting compare against — is always derived from the
+same usage fields `AddUsage`/`getNodesUsage` already track, so it can
+never drift from what `CollapseInitContainerUsage` produced when the pod
+started.
+
+Only ever runs **after** `pod.Status` confirms completion.
+
+**Idempotency:** `PodManager` stores a `PodShrinkState` per pod —
+`NotShrunk`, `AppOnly`, or `Zero` — replacing a single boolean, since the
```
DSFans2014: What are these statuses used for?  Also, the names are a bit confusing

maishivamhoo123: @DSFans2014 PodShrinkState guards against two specific problems. First, idempotency: the system checks the same pod over and over, and once it shrinks the usage, it must recognize the shrink is already done on subsequent checks so it doesn't subtract the same memory twice. Second, one-directional transitions: usage should only ever go down from full to app-only, or full straight to zero but never back up.
A simple hasShrunk boolean cannot handle this because it only tracks whether any shrink happened. Once the first shrink (full  to app-only) occurs, the boolean flips to true. Later, when the pod dies and a second shrink (app-only to  zero) is needed, the system looks at the true boolean, assumes everything is already handled, and skips the second shrink leaving the pod's memory permanently stuck at the app-only level even after the pod is gone.
What each value means 
NotShrunk : still the original max(app_sum, init_peak) from admission.
AppOnly : init containers all exited 0 recorded usage = app containers only.
Zero : pod hit a terminal phase recorded usage = 0.
i guess now it is clear ? there is a scope of changing name from Zero to Released  other than this is looking fine i am open for any suggestions if you have any suggestions please let me know?
Thank you !

DSFans2014: > First, idempotency: the system checks the same pod over and over, and once it shrinks the usage, it must recognize the shrink is already done on subsequent checks so it doesn't subtract the same memory twice. 

If we address the issue mentioned in https://github.com/Project-HAMi/HAMi/pull/2064#discussion_r3600241171 in the future, how should the state transition?

maishivamhoo123: @DSFans2014  in that condition Instead of using a fixed three state flag, we can track a single number called the watermark ,  basically a counter that keeps track of how many init containers have successfully finished in the order.
this  NotShrunk is just watermark 0, and AppOnly is just the final watermark number (when all inits have succeeded). The big win is the middle ground , lets take  your example only when init‑1 (1024Mi) succeeds, the watermark moves from 0 to 1, and we immediately release that 1024Mi the effective usage drops to only what the remaining init containers (like init‑2) are asking for. If init‑2 keeps failing and retrying, the watermark stays at 1 and no further shrink happens until either init‑2 finally succeeds (moving the watermark to the top, which then includes app containers) or the whole pod hits a terminal phase and jumps straight to Zero/Released, releasing everything. This watermark only ever moves forward, one step at a time, never backward, and since it's recomputed from the pod's actual status on every check, repeated reconciliations without any status change naturally produce zero delta — so our idempotency stays intact. so we can safely keep it as a future TODO without reworking anything in this PR.
is it a good approach or do you have any further suggestions?

DSFans2014: From my point of view, we only need one bool variable  to track whether the init container's resources have been reclaimed. Moreover, the problems in https://github.com/Project-HAMi/HAMi/pull/2064#discussion_r3600241171 can be resolved by the watermark number described above.
```
// We use initContainerResourceReleased to indicate whether the init container's resources have been released
// with an initial value of false

// If all init containers have completed and initContainerResourceReleased remains false
// we reclaim the init container resources
if all initContainer terminated and !initContainerResourceReleased:
    // release the resource of initContainer
    initContainerResourceReleased = true

// If the pod terminates, we free up the app containers resources.
if pod.Status.Phase in (Succeeded, Failed):
   // release the resource of app containers
```

If there are any issues with the above solution, feel free to point them out.

maishivamhoo123: i also think this is the correct approach i will include this and raise a PR .
 thank you!

maishivamhoo123:
@DSFans2014 I've applied all the suggested changes , Could you please take another look? Also, let me know if the design is ready for me to start coding, or if anything else i need to change.
  Thanks you!

DSFans2014 (commented):
/lgtm

archlitchi (approved):
/lgtm

PR #2066, opened 2026-07-14
feat(scheduler): retry NodeLock in Bind for PodGroup members

### LINKED ISSUE #1832: feat: support Kubernetes Coscheduling Plugin for PodGroup

Support Kubernetes Coscheduling Plugin for PodGroup 

**What type of PR is this?**

/kind feature

**Background**

HAMi currently schedules pods individually. When running distributed training jobs (e.g., PyTorch DDP, MPI), all workers must start simultaneously — if some pods acquire GPU resources but others cannot, the job stalls and holds resources indefinitely. This is the gang scheduling problem.

HAMi should not implement gang scheduling semantics itself. Instead, it should integrate correctly with the Kubernetes [Coscheduling Plugin](https://github.com/kubernetes-sigs/scheduler-plugins/tree/master/pkg/coscheduling), which already handles group-level permit/deny logic. HAMi only needs to fix two behaviors that break in a Coscheduling environment.

**Goals**

- Fix `LockNode` to retry on lock contention instead of failing immediately, allowing concurrent PodGroup bind operations to serialize gracefully
- Fix `onAddPod` cache handling to evict pods that return to Pending with stale annotations, preventing phantom GPU occupancy
- Fix `Bind` failure path to clean up `podManager` cache, preventing ghost entries on failed binds
- No HAMi API changes — users only need to install scheduler-plugins and annotate pods

**Non-Goals**

- HAMi-native PodGroup CRD
- Gang scheduling logic inside HAMi (min-member waiting, group-level permit/deny)
- Volcano integration

**Problem Details**

**1. Concurrent Bind failures**

When Coscheduling releases a PodGroup, kube-scheduler binds all member pods near-simultaneously. Multiple `/bind` calls targeting the same node race on `LockNode`. Without retry, losing pods fail their Bind immediately and force the entire PodGroup to restart scheduling.

Before (current behavior):
```
Pod-A /bind → LockNode(node-1) → success
Pod-B /bind → LockNode(node-1) → error: node has been locked → PodGroup restarts
Pod-C /bind → LockNode(node-1) → error: node has been locked → PodGroup restarts
```

After (proposed behavior):
```
Pod-A /bind → LockNode(node-1) → success, completes, releases lock
Pod-B /bind → LockNode(node-1) → lock contention, retry after 1s → success
Pod-C /bind → LockNode(node-1) → lock contention, retry after 1s → success
```

**2. Phantom GPU occupancy**

When Coscheduling denies a PodGroup, pods return to Pending with stale HAMi annotations still attached. HAMi's `podManager` re-adds these pods to its cache on the next `onAddPod` event, creating ghost GPU reservations that make nodes appear fuller than they are.

```
PodGroup denied → Pod returns to Pending (hami.io/vgpu-node annotation still set)
               → onAddPod fires → podManager.AddPod() → ghost GPU reservation
               → next scheduling round sees fewer available GPUs than reality
```

Fix: in `onAddPod`, detect pods that are Pending with stale annotations (`bind-phase != success`) and evict them from cache instead of re-adding.

A related leak exists in the `Bind` failure path: when `Bind` fails, `ReleaseNodeLocks` releases the node lock but does not call `podManager.DelPod()`, leaving a ghost cache entry. This should also be fixed.

**Usage**

Install [scheduler-plugins](https://github.com/kubernetes-sigs/scheduler-plugins) alongside HAMi. No HAMi API changes — users only need standard PodGroup annotations:

```yaml
apiVersion: scheduling.sigs.k8s.io/v1alpha1
kind: PodGroup
metadata:
  name: my-training-job
spec:
  minMember: 4
  minResources:
    nvidia.com/vgpu: "4"
```

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: worker-0
  annotations:
    scheduling.sigs.k8s.io/pod-group: my-training-job
spec:
  schedulerName: hami-scheduler
  ...
```

**New Flag**

```
--node-lock-retry-timeout duration   timeout for retrying node lock acquisition when locked by valid pod (default 30s)
```

Fixes #

Yes — HAMi works correctly alongside scheduler-plugins Coscheduling. New flag: `--node-lock-retry-timeout` (default: 30s).

lin121291:
**What type of PR is this?**

/kind feature

When multiple members of the same PodGroup are bound concurrently to
the same node, they contend on the `hami.io/mutex.lock` annotation.
The losing pod fails immediately and waits for kube-scheduler backoff
(~2s). This PR adds a retry loop in `Bind` so PodGroup pods wait for
the lock to clear (~22ms on real hardware with a device plugin) instead.

- Add `ErrNodeLockContention` sentinel in `nodelock` package
- Add `IsPodGroupMember()` based on `scheduling.x-k8s.io/pod-group` label
- Add `--node-lock-retry-timeout` flag (default 28s, align with extender
  `httpTimeout` in KubeSchedulerConfiguration)
- Non-PodGroup pod behavior is unchanged

Fixes #1832

The retry timeout (28s) should be less than the extender `httpTimeout`
configured in KubeSchedulerConfiguration (HAMi chart default: 30s).

Yes. New CLI flag `--node-lock-retry-timeout` (default 28s). No breaking changes.

## Summary by CodeRabbit

* **New Features**
  * Added `--node-lock-retry-timeout` to configure how long the scheduler retries node locks for grouped (PodGroup) pods (28s default; `0` disables).
* **Bug Fixes**
  * Grouped pods now retry node-lock acquisition on contention until the configured timeout, and release locks correctly on failure.
  * Non-grouped pods and non-contention failures no longer trigger retry behavior.
* **Tests**
  * Added unit tests covering contention, retry timing, and error handling for grouped vs non-grouped pods.

### DIFF
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -753,7 +753,9 @@ func (s *Scheduler) lockAllDevices(node *corev1.Node, pod *corev1.Pod) error {
 
 func (s *Scheduler) releaseAllDevices(node *corev1.Node, pod *corev1.Pod) {
 	for _, val := range device.GetDevices() {
-		val.ReleaseNodeLock(node, pod)
+		if err := val.ReleaseNodeLock(node, pod); err != nil {
+			klog.ErrorS(err, "Failed to release node lock", "node", node.Name, "pod", klog.KObj(pod))
+		}
 	}
 }
 
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -768,6 +768,7 @@ func (s *Scheduler) acquireNodeLocks(node *corev1.Node, pod *corev1.Pod) error {
 		if err == nil {
 			return nil
 		}
+		s.releaseAllDevices(node, pod)
 		if !nodelockutil.IsNodeLockContention(err) {
 			return err
 		}
@@ -775,8 +776,11 @@ func (s *Scheduler) acquireNodeLocks(node *corev1.Node, pod *corev1.Pod) error {
 			return fmt.Errorf("timed out after %v waiting for node %s to be unlocked: %w",
 				config.NodeLockRetryTimeout, node.Name, nodelockutil.ErrNodeLockContention)
 		}
-		s.releaseAllDevices(node, pod)
-		time.Sleep(100 * time.Millisecond)
+		select {
+		case <-s.stopCh:
+			return fmt.Errorf("scheduler shutting down while waiting for node lock: %w", nodelockutil.ErrNodeLockContention)
+		case <-time.After(100 * time.Millisecond):
+		}
 	}
 }
 
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -742,6 +742,44 @@ func (s *Scheduler) cleanupStalePodAllocation(pod *corev1.Pod) {
 	}
 }
 
+func (s *Scheduler) lockAllDevices(node *corev1.Node, pod *corev1.Pod) error {
+	for _, val := range device.GetDevices() {
+		if err := val.LockNode(node, pod); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (s *Scheduler) releaseAllDevices(node *corev1.Node, pod *corev1.Pod) {
+	for _, val := range device.GetDevices() {
+		val.ReleaseNodeLock(node, pod)
+	}
+}
+
+func (s *Scheduler) acquireNodeLocks(node *corev1.Node, pod *corev1.Pod) error {
+	if !util.IsPodGroupMember(pod) || config.NodeLockRetryTimeout <= 0 {
+		return s.lockAllDevices(node, pod)
+	}
+
+	deadline := time.Now().Add(config.NodeLockRetryTimeout)
+	for {
+		err := s.lockAllDevices(node, pod)
+		if err == nil {
+			return nil
+		}
+		if !nodelockutil.IsNodeLockContention(err) {
+			return err
+		}
+		if time.Now().After(deadline) {
+			return fmt.Errorf("timed out after %v waiting for node %s to be unlocked: %w",
+				config.NodeLockRetryTimeout, node.Name, nodelockutil.ErrNodeLockContention)
+		}
+		s.releaseAllDevices(node, pod)
+		time.Sleep(100 * time.Millisecond)
+	}
+}
+
 func (s *Scheduler) Bind(args extenderv1.ExtenderBindingArgs) (*extenderv1.ExtenderBindingResult, error) {
 	klog.InfoS("Attempting to bind pod to node", "pod", args.PodName, "namespace", args.PodNamespace, "node", args.Node)
 	var res *extenderv1.ExtenderBindingResult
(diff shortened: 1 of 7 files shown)


pkg/scheduler/scheduler_test.go
```diff
@@ -1844,3 +1844,132 @@ func Test_Bind_DelPodOnGetNodeFailure(t *testing.T) {
...
+var errContention = fmt.Errorf("contended: %w", nodelockutil.ErrNodeLockContention)
+
+func setupBindLockRetryTest(t *testing.T, retryTimeout time.Duration, pod *corev1.Pod, mock *bindLockMockDevice) (*Scheduler, extenderv1.ExtenderBindingArgs, func()) {
+	t.Helper()
+
+	oldRetry := config.NodeLockRetryTimeout
+	config.NodeLockRetryTimeout = retryTimeout
+	oldDevicesMap := device.DevicesMap
+	device.DevicesMap = map[string]device.Devices{"bind-lock-mock": mock}
+	cleanup := func() {
```
mesutoezdil, line 1880: nit: cleanup never closes s.stopCh so the informer goroutines started below leak past each test

pkg/scheduler/scheduler.go
```diff
@@ -742,6 +742,48 @@ func (s *Scheduler) cleanupStalePodAllocation(pod *corev1.Pod) {
...
+		if err := val.LockNode(node, pod); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (s *Scheduler) releaseAllDevices(node *corev1.Node, pod *corev1.Pod) {
+	for _, val := range device.GetDevices() {
+		val.ReleaseNodeLock(node, pod)
```
mesutoezdil: a swallowed release error leaves this pod holding its own stale lock so the retry loop then contends with itself instead of the real winner until timeout, worth at least logging the error

mesutoezdil:
and as i see it s  only implements the lock retry goal from #1832, the onAddPod stale pending cache eviction and the podManager cleanup on the Bind failure path are still unaddressed, 
might be worth changing fixes to part of #1832 so the issue does not auto close?

mesutoezdil:
still asking to change fixes #1832 to part of #1832 in the description since this pr only covers the lock retry goal, the other two goals in that issue are untouched and it would auto close prematurely

lin121291:
Thanks @mesutoezdil for the review.

On the scope of `Fixes #1832`: the fixes proposed in the original issue were written before real-hardware testing. The discussion on the issue narrowed things down to just the retry piece:

- Kind test ([gist](https://gist.github.com/lin121291/525be433aafc414744582ba2c5da83d7)) showed 5-min pending → this motivated the stale-annotation and Bind-cleanup fixes.
- Real T4 test ([gist](https://gist.github.com/lin121291/7a36b2c915553b589be7ed5d89453fc8)) showed the 5-min was a kind artifact (no device plugin → `Allocate()` never runs). On real hardware `Allocate → Release` is ~22 ms and a 5-pod gang converges in 13 s via normal kube-scheduler backoff.
- [Consensus reached with @Shouren](https://github.com/Project-HAMi/HAMi/issues/1832#issuecomment-4766971350): only the retry gap is actionable.

The other two fixes also turn out to be already covered by existing self-healing paths in the scheduler (`PodManager` is UID-keyed and idempotent, `Filter` clears stale entries at the start of every round, `LockNode` has dangling-owner detection).

I'll update issue #1832 to reflect this narrower scope with the gist evidence linked, then this PR can stay as `Fixes #1832`.

pkg/scheduler/scheduler_test.go
```diff
@@ -1844,3 +1844,133 @@ func Test_Bind_DelPodOnGetNodeFailure(t *testing.T) {
...
+
+	oldRetry := config.NodeLockRetryTimeout
+	config.NodeLockRetryTimeout = retryTimeout
+	oldDevicesMap := device.DevicesMap
+	device.DevicesMap = map[string]device.Devices{"bind-lock-mock": mock}
+
+	s := NewScheduler()
+	cleanup := func() {
+		config.NodeLockRetryTimeout = oldRetry
+		device.DevicesMap = oldDevicesMap
```
mesutoezdil, line 1882: nit: client.KubeClient is set but never restored in cleanup so later tests can inherit this fake client

mesutoezdil (approved):
/lgtm

archlitchi (approved):
/lgtm

PR #2067, opened 2026-07-14
design(amd): update amd design docs

FouoF:
**What type of PR is this?**

/kind design

This pr update the design docs in following points:
major changes:
1. using HSA_CU_MASK instead of ROC_GLOBAL_CU_MASK
HSA_CU_MASK supports per GPU CU mask which is essential for multi-gpu tasks. As tasks request different num of vGPU will create a cross using for GPUs. For example, task1 request 25% one gpu, it get gpu 0 CU 0-76. Then task2 requests 25% 
 but 2 GPUs, it can not use gpu 1 CU 0-76 though they are ideal for gpu 0 is allocated.
2. gpucore represents CU percentage instead of CU count
This design is intended to align with other device apis and it also guarantee the consistency in metrics. Besides, CU total count may not exposed to the user.
3. using HIP_DEVICE_MEMORY_LIMIT for all device HBM limit
HAMi api only allow same gpumem allocation for all request devices. So one env for all devices can satisfy it.
4. device plugin calculates concrete CUs instead of scheduler
Device related logic better place in device plugin.

minor changes:
1. using cu range instead of cu mask to describe allocation for better human readable
2. add hami.io/amd-devices-to-allocate for better compatibility of exists progress 
3. add know limit Not support mixed type in one node

Fixes #

## Summary by CodeRabbit

* **Documentation**
  * Updated AMD Instinct vGPU guidance to use fractional GPU memory in MiB (`amd.com/gpumem`, reflected as `HIP_DEVICE_MEMORY_LIMIT=<MiB>m`) and compute-unit partitioning as an inclusive 1–100% (`amd.com/gpucores`).
  * Reworked CU selection to replace the prior CU-mask hex-bitmask flow with container startup using `ROCR_VISIBLE_DEVICES` ordering and `HSA_CU_MASK` expressed as `GPU_list:CU_list` for non-overlapping CU ranges.
  * Refreshed WGP behavior/limitations, including adjacent CU-pair alignment for WGP-capable devices and that zero-interference isn’t guaranteed.

### DIFF
--- a/docs/develop/amd-vgpu.md
+++ b/docs/develop/amd-vgpu.md
@@ -106,7 +106,7 @@ HSA_CU_MASK=0:0-75;1:0-75
 Each `CU_list` uses HSA's CU ID-list grammar, for example `0-3,8,10-12`.
 
 Exclusivity of CU ranges across pods on a device is enforced under the AMD
-node lock (`AMDDevices.LockNode` and `ReleaseNodeLock`).
+node lock (`AMDDevices.LockNode` and `ReleaseNodeLock` which are unimplemented now).
 
 ## 5. Resource model and core_limit -> CU mask
 
--- a/docs/develop/amd-vgpu.md
+++ b/docs/develop/amd-vgpu.md
@@ -7,7 +7,7 @@ AMD ROCm GPUs, so multiple pods can share one Instinct GPU.
 
 ## 2. Goals
 
-- Fractional allocation by memory (`amd.com/gpumem`, MB) and compute (`amd.com/gpucores`, CU count).
+- Fractional allocation by memory (`amd.com/gpumem`, MiB) and compute (`amd.com/gpucores`, percentage).
 - Exclusive, non-overlapping CU partitioning across pods.
 - Following HAMi's existing architecture design to achieve the functionality.
 
@@ -20,7 +20,7 @@ Switching to LD_AUDIT (`la_symbind64`), which intercepts only cross-library bind
 resolved it. The existing NVIDIA LD_PRELOAD path is unchanged.
 
 **Advantages of CU masking, compared to hardware partitioning (CPX/NPS).**
-Masking (`ROC_GLOBAL_CU_MASK`) assigns an arbitrary and fine-grained per-pod CU partitioning at container start ([AMD Docs](https://rocm.docs.amd.com/en/latest/how-to/setting-cus.html)). 
+Masking (`HSA_CU_MASK`) assigns an arbitrary and fine-grained per-pod CU partitioning at container start ([AMD Docs](https://rocm.docs.amd.com/en/latest/how-to/setting-cus.html)).
 On the other hand, hardware partitioning (CPX/NPS) slices only at fixed XCD granularity and is set per physical GPU.
 
 ## 4. Protocol (scheduler <-> device-plugin)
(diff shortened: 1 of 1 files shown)


docs/develop/amd-vgpu.md
```diff
@@ -47,44 +47,33 @@ Registered under `hami.io/node-amd-register`, in JSON format — an array of `De
...
 
 ### 4.2 Pod allocation (scheduler -> pod annotations -> Allocate)
 
-The allocation result is written under the **AMD-specific** key (each vendor has its own):
+The scheduler allocation result is written under the **AMD-specific** keys:
 
 ```text
-hami.io/amd-devices-allocated: <UUID>,AMDGPU,<memMB>,<cuCount>:;
+hami.io/amd-devices-to-allocate: <UUID>,amd,<memMiB>,<cuCount>:;
+hami.io/amd-devices-allocated:   <UUID>,amd,<memMiB>,<cuCount>:;
```
mesutoezdil: question: is lowercase amd here intentional, since the existing hami.io/amd-devices-allocated encoding in pkg/device/amd/device.go already uses the AMDDevice constant which is AMDGPU not amd?

docs/develop/amd-vgpu.md
```diff
@@ -94,39 +83,37 @@ Example Pod request:
...
+**Zero-interference is not guaranteed.** Even with non-overlapping masks,
+residual interference remains (as reported in #1707).
 
 ## 6. Known limitations
 
 - **No `amd-smi` / `rocm-smi` virtualization.** 
   These read sysfs/drm, not HIP, so
   LD_AUDIT cannot intercept them; in-container tools may report physical resources.
+- **Not support mixed type in one node.**
+  Currently, device plugin get gpu type form the label `amd.com/gpu.product-name`, This label can not describe more than one gpu types.
```
mesutoezdil: nit: get should be gets, form should be from, and the comma before This should be a period

mesutoezdil:
pls also address the coderabbit points on wgp cu pairing, the gpu index ambiguity in hsa_cu_mask, and the percentage rounding rule, those look like real gaps not just style

docs/develop/amd-vgpu.md
FouoF: This string may be used as prefix to construct annotation key. To keep the annotation key full lowercase, I change it to lowercase. But I found it is not actually used after checking the code, so keep it `AMDGPU` is ok.

docs/develop/amd-vgpu.md
```diff
@@ -94,39 +116,55 @@ Example Pod request:
...
+result is clamped to `[1, devcore]`. For example, `25` on a 304-CU MI300X
+yields `76` CUs; `33` yields `100` CUs (`floor(100.32)`); `67` yields
+`203` CUs (`floor(203.68)`).
 
-The count -> bitmap conversion is a **scheduler's** responsibility, because non-overlap requires knowledge of the device's current
-allocation state. 
+The scheduler records this `cuCount` in `hami.io/amd-devices-allocated`. The
+device-plugin selects a non-overlapping CU range of that size and injects it
+as `HSA_CU_MASK`. On WGP-capable devices, the chosen range must additionally
+align to adjacent CU pairs within each WGP.
```
mesutoezdil, line 149: what happens when floor gives an odd cuCount on a wgp device is still undefined, worth spelling out whether it rounds, rejects, or adjusts before building the mask

docs/develop/amd-vgpu.md
```diff
@@ -47,45 +68,46 @@ Registered under `hami.io/node-amd-register`, in JSON format — an array of `De
...
 ```text
-amd.com/cu-mask-assigned: "false" -> "true"  (set by device-plugin)
+# two GPUs: UUID-A (index 0) gets CUs 0-75, UUID-B (index 1) gets CUs 0-75
+HSA_CU_MASK=0:0-75;1:0-75
 ```
 
+Each `CU_list` uses HSA's CU ID-list grammar, for example `0-3,8,10-12`.
+
+Exclusivity of CU ranges across pods on a device is enforced under the AMD
+node lock (`AMDDevices.LockNode` and `ReleaseNodeLock`).
```
mesutoezdil: this reads as already enforced but LockNode and ReleaseNodeLock in pkg/device/amd/device.go still just return nil, worth wording it as intended behavior for now

docs/develop/amd-vgpu.md
FouoF: How to deal with WGP can be determined when implement. I think point out the limit in the doc is enough.

mesutoezdil: fair enough

archlitchi (approved):
/lgtm

PR #2068, opened 2026-07-14
fix(scheduler): avoid the racy read of the node cache in register()

Wangmin362:
**What type of PR is this?**
/kind bug

`register()` logged a node's device set by reading it back from the shared `s.nodes` map immediately after `addNode`:

```go
s.addNode(val.Name, nodeInfo)
if s.nodes[val.Name] != nil && len(nodeInfo.Devices) > 0 {   // unlocked read of s.nodes
    ... "totalDevices", s.nodes[val.Name].Devices ...        // unlocked read again
}
```

That read takes no lock, while the informer delete callback `onDelNode` -> `rmNode` deletes from the same map under the manager lock. The two run on different goroutines with no shared lock, so they form a data race — `concurrent map read and map write`, which is fatal and crashes the scheduler.

Rather than guarding the read, this drops it: the log now prints the `nodeInfo` we just built. It carries the same per-vendor device set for the current node, needs no lock, and removes the shared access entirely. The only observable change is that the log no longer shows the cross-vendor accumulated `totalDevices` field (the line still carries `deviceVendor` and `nodeInfo`).

NONE

Verification (control-plane concurrency, hardware-independent):

- Added `Test_register_NodeCacheConcurrency`. It drives the two real production paths that touch `s.nodes` — `register()` and `onDelNode` -> `rmNode` — concurrently and lets the race detector judge.
  - On the previous code (unlocked read), `go test -race` reports a `DATA RACE` between `rmNode` deleting the map and `register()` reading it, and the test fails.
  - With this change, `go test -race ./pkg/scheduler/` passes.

Test design notes:
- To make `register()` reach the cache access every round, the test models a node whose GPU flaps between healthy and unhealthy — the case the node health check exists for. Only the health flag in the register annotation changes; the rest of the hardware is fixed (a card's memory does not change). The flip is what makes `CheckHealth` report `needUpdate`.
- Reusing one node name keeps that map key hot, so the read and the delete overlap reliably instead of depending on timing luck.
- The lister is seeded directly through its indexer, so the informer is never started and no background goroutine leaks past the test.
- `onDelNode` does extra per-call work, so several delete goroutines are used to delete often enough to overlap `register()`'s brief cache access.
- Work is bounded by an iteration count rather than a `time.Sleep` window, so the run is deterministic.

The HAMi scheduler no longer risks a fatal concurrent-map-access crash when a node is deleted while the scheduler is registering node devices.

**AI Assistant disclosure**:

This PR was written primarily by Claude Code (code change and test). The author reviewed and understands the change; the analysis and commit message are the author's own.

### DIFF
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -422,11 +422,12 @@ func (s *Scheduler) register(labelSelector labels.Selector, printedLog map[strin
 				nodeInfo.Devices[deviceinfo.DeviceVendor] = append(nodeInfo.Devices[deviceinfo.DeviceVendor], *deviceinfo)
 			}
 			s.addNode(val.Name, nodeInfo)
-			if s.nodes[val.Name] != nil && len(nodeInfo.Devices) > 0 {
+			// Log the locally built nodeInfo; reading it back from s.nodes raced with onDelNode->rmNode.
+			if len(nodeInfo.Devices) > 0 {
 				if printedLog[val.Name] {
-					klog.V(5).InfoS("Node device updated", "nodeName", val.Name, "deviceVendor", devhandsk, "nodeInfo", nodeInfo, "totalDevices", s.nodes[val.Name].Devices)
+					klog.V(5).InfoS("Node device updated", "nodeName", val.Name, "deviceVendor", devhandsk, "nodeInfo", nodeInfo)
 				} else {
-					klog.InfoS("Node device added", "nodeName", val.Name, "deviceVendor", devhandsk, "nodeInfo", nodeInfo, "totalDevices", s.nodes[val.Name].Devices)
+					klog.InfoS("Node device added", "nodeName", val.Name, "deviceVendor", devhandsk, "nodeInfo", nodeInfo)
 					printedLog[val.Name] = true
 				}
 			}
--- a/pkg/scheduler/register_race_test.go
+++ b/pkg/scheduler/register_race_test.go
@@ -1,0 +1,125 @@
+/*
+Copyright 2026 The HAMi Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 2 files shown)


pkg/scheduler/scheduler.go
```diff
@@ -422,14 +422,16 @@ func (s *Scheduler) register(labelSelector labels.Selector, printedLog map[strin
...
 			s.addNode(val.Name, nodeInfo)
-			if s.nodes[val.Name] != nil && len(nodeInfo.Devices) > 0 {
+			s.mutex.RLock()
+			if cachedNode, exists := s.nodes[val.Name]; exists && cachedNode != nil && len(nodeInfo.Devices) > 0 {
 				if printedLog[val.Name] {
-					klog.V(5).InfoS("Node device updated", "nodeName", val.Name, "deviceVendor", devhandsk, "nodeInfo", nodeInfo, "totalDevices", s.nodes[val.Name].Devices)
+					klog.V(5).InfoS("Node device updated", "nodeName", val.Name, "deviceVendor", devhandsk, "nodeInfo", nodeInfo, "totalDevices", cachedNode.Devices)
 				} else {
-					klog.InfoS("Node device added", "nodeName", val.Name, "deviceVendor", devhandsk, "nodeInfo", nodeInfo, "totalDevices", s.nodes[val.Name].Devices)
+					klog.InfoS("Node device added", "nodeName", val.Name, "deviceVendor", devhandsk, "nodeInfo", nodeInfo, "totalDevices", cachedNode.Devices)
```
mesutoezdil: nit: klog calls run while holding the rlock here, and in register_race_test.go stopch is never closed so the informer goroutines leak past the test, both worth a quick fix but not blocking

pkg/scheduler/scheduler.go
```diff
@@ -422,14 +422,16 @@ func (s *Scheduler) register(labelSelector labels.Selector, printedLog map[strin
 				nodeInfo.Devices[deviceinfo.DeviceVendor] = append(nodeInfo.Devices[deviceinfo.DeviceVendor], *deviceinfo)
 			}
 			s.addNode(val.Name, nodeInfo)
-			if s.nodes[val.Name] != nil && len(nodeInfo.Devices) > 0 {
+			s.mutex.RLock()
```
DSFans2014: > register() reads s.nodes directly without taking the nodeManager lock, while onDelNode deletes from the same map under the lock

I think it's better to call a locked member function like `GetNode` rather than doing explicit locking

archlitchi (approved):
/lgtm

PR #2075, opened 2026-07-16
fix(scheduler): do not exit on a transient error updating the leader label

Wangmin362:
**What type of PR is this?**
/kind bug

`updateSchedulerLabel` calls `klog.Fatalf` (i.e. `os.Exit`) when listing the scheduler pods or patching the leader label fails. Both hit the live API server, so a transient error (apiserver restart, throttling, a 409) kills the scheduler. The follower branch already logs the same patch error with `klog.ErrorS` and keeps going — this just makes the leader branch and the list do the same.

NONE

Tested on a kind cluster with the real scheduler image: dropping `patch pods` from RBAC makes the leader patch fail — the current code goes to CrashLoopBackOff, this change stays Running and logs the error each cycle. Unit test added; `go test ./pkg/scheduler/`, `go vet` and `gofmt` are clean.

The scheduler no longer exits on a transient API error while updating its role label; it logs and retries on the next cycle.

**AI Assistant disclosure**:

This PR was written with the help of Claude Code. I reviewed and understand the change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved resilience of scheduler label updates: failures during pod discovery or label patching are now handled gracefully without stopping the scheduler process.
  * Error details are recorded in logs while the scheduler continues operating.

* **Tests**
  * Added unit tests to confirm the scheduler does not exit when label patching fails.
  * Added unit tests to confirm the scheduler does not exit when pod listing fails.

### DIFF
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -448,10 +448,11 @@ func (s *Scheduler) updateSchedulerLabel() {
 	schedulerSelector := labels.Set(map[string]string{util.HAMiComponentLabel: util.HAMiComponentScheduler}).AsSelector()
 	schedulerPods, err := s.podLister.Pods(os.Getenv("POD_NAMESPACE")).List(schedulerSelector)
 	if err != nil {
-		klog.Fatalf("Failed to list hami scheduler pods from lister: namespace %s selector %s",
-			os.Getenv("POD_NAMESPACE"),
-			schedulerSelector.String(),
+		klog.ErrorS(err, "Failed to list hami scheduler pods from lister",
+			"namespace", os.Getenv("POD_NAMESPACE"),
+			"selector", schedulerSelector.String(),
 		)
+		return
 	}
 
 	for idx := range schedulerPods {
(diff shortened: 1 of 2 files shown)


pkg/scheduler/scheduler.go
```diff
@@ -448,10 +448,11 @@ func (s *Scheduler) updateSchedulerLabel() {
 	schedulerSelector := labels.Set(map[string]string{util.HAMiComponentLabel: util.HAMiComponentScheduler}).AsSelector()
 	schedulerPods, err := s.podLister.Pods(os.Getenv("POD_NAMESPACE")).List(schedulerSelector)
 	if err != nil {
-		klog.Fatalf("Failed to list hami scheduler pods from lister: namespace %s selector %s",
-			os.Getenv("POD_NAMESPACE"),
-			schedulerSelector.String(),
+		klog.ErrorS(err, "Failed to list hami scheduler pods from lister",
+			"namespace", os.Getenv("POD_NAMESPACE"),
+			"selector", schedulerSelector.String(),
 		)
```
mesutoezdil, line 454: this list error path is why codecov is failing, no test hits it yet

Wangmin362: ok， thank you ,i hava add this test path

archlitchi (approved):
/lgtm

PR #2088, opened 2026-07-18
fix(scheduler): bound-check MIG template/instance index parsed from UUID annotation

mesutoezdil:
/kind bug

getNodesUsage reads a MIG template index and instance number from a pod's UUID annotation and used both without checking them. 
A stale or mismatched index can panic the scheduler's usage-scrape goroutine instead of just being skipped.

This PR adds bounds and consistency checks: an out-of-range index is rejected, and an index that disagrees with the device's already-recorded MIG template is rejected too.

Two new tests: one for an out-of-range index, one for two pods that disagree on the template index. 
Confirmed both panic on the old code and pass after the fix. go test with race, go vet, and golangci-lint all pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of invalid, stale, or mismatched MIG UUID and index data to prevent unsafe access.
  * The scheduler now safely validates MIG template and instance indexes; on inconsistencies it logs the issue and skips marking MIG usage to avoid incorrect scheduling.
* **Tests**
  * Added unit tests covering stale template indices, unparsable MIG UUIDs, out-of-range MIG instances, and mismatched MIG index scenarios.

### DIFF
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -646,11 +646,26 @@ func (s *Scheduler) getNodesUsage(nodes *[]string, task *corev1.Pod) (*map[strin
 									d.Device.Health = false
 									continue
 								}
-								tmpIdx, Instance, _ := device.ExtractMigTemplatesFromUUID(udevice.UUID)
+								tmpIdx, instanceIdx, err := device.ExtractMigTemplatesFromUUID(udevice.UUID)
+								if err != nil {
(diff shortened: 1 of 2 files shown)


DSFans2014 (commented):
/lgtm

ouyangluwei163 (commented):
/lgtm

archlitchi (approved):
/lgtm

PR #2091, opened 2026-07-20
feat(helm): add switch for admission webhook namespaceSelector

### LINKED ISSUE #2039: [Feature] add an escape hatch to stop managing admission webhook namespaceSelector on AKS-like platforms

Add an explicit chart option to disable rendering / managing `scheduler.admissionWebhook.namespaceSelector` for environments where the webhook `namespaceSelector` field is mutated and owned by the platform.

For example:

    scheduler:
      admissionWebhook:
        manageNamespaceSelector: true

Expected behavior:

- `true` (default): keep current behavior and render `namespaceSelector` exactly as today
- `false`: do not render `namespaceSelector`, so the field can be fully managed by the platform / external controller

**What type of PR is this?**

/kind feature

**why we need it**:

On AKS, the control plane may mutate `MutatingWebhookConfiguration` / `ValidatingWebhookConfiguration` and take ownership of `namespaceSelector` via `admissionsenforcer`.

Reference:

- Azure/AKS#4002

In HAMi, the scheduler webhook currently always renders `namespaceSelector`.
That works on normal Kubernetes distributions, but on AKS it can create a server-side apply ownership conflict:

- Helm / chart wants to manage `.webhooks[name="vgpu.hami.io"].namespaceSelector`
- AKS `admissionsenforcer` also updates and owns the same field
- subsequent `helm upgrade` / `helmfile sync` may fail with an SSA conflict on that field

Important point: the problem is not that `namespaceSelector` must not exist on AKS.
The real problem is that the chart currently cannot opt out of managing that field when the platform manages it.

A concrete example from our HAMi deployment on AKS:

- the live webhook still works
- AKS injects extra namespace exclusions such as:
  - `control-plane NotIn [true]`
  - `kubernetes.azure.com/managedby NotIn [aks]`
- but Helm upgrade fails because `namespaceSelector` ownership has been taken by `admissionsenforcer`

We verified that if HAMi stops rendering `namespaceSelector`, the webhook can still exist and the release can be upgraded successfully on AKS, because Helm no longer fights for ownership of that field.

Why an explicit switch is useful:

- keeps the current default behavior unchanged for normal clusters
- preserves existing namespace filtering semantics by default
- gives AKS users and other managed-platform users an escape hatch
- avoids pushing a provider-specific workaround into the default chart behavior

This seems safer and easier to accept than removing `namespaceSelector` by default.

This request is mainly about field ownership compatibility, not about changing HAMi webhook semantics by default.

Suggested implementation direction:

- introduce a new boolean like `scheduler.admissionWebhook.manageNamespaceSelector`
- default it to `true`
- gate the entire `namespaceSelector:` block on that value

That allows:

- existing users: no behavior change
- AKS users: explicitly set it to `false` and let the platform own the field

Yes.
It introduces an optional chart value for managed Kubernetes environments where webhook `namespaceSelector` is mutated / owned externally.
**Default behavior remains unchanged.**

spencercjh:
**Most of this Pull Request was generated or revised with the assistance of AI tools(kimi cli with K3 max). I have reviewed the resulting content and take full responsibility for its accuracy, security, licensing compliance, and inclusion in this project.**

**What type of PR is this?**

/kind feature

Adds an explicit chart option `scheduler.admissionWebhook.manageNamespaceSelector` (default `true`) to control whether the chart renders and manages the webhook `namespaceSelector` field:

- `true` (default): keep current behavior and render `namespaceSelector` exactly as today
- `false`: do not render `namespaceSelector`, so the field can be fully managed by the platform / external controller

On managed platforms such as AKS, the control plane mutates `MutatingWebhookConfiguration` and takes ownership of `namespaceSelector` via `admissionsenforcer` (Azure/AKS#4002). Since the chart currently always renders that field, Helm fights the platform for field ownership and subsequent `helm upgrade` / `helmfile sync` may fail with a server-side apply conflict. Setting this option to `false` lets those environments opt out.

Fixes #2039

This change is about field ownership compatibility only; it does not change HAMi webhook semantics by default.

- Default rendering verified byte-identical to master (`helm template` diff against master is empty).
- With `scheduler.admissionWebhook.manageNamespaceSelector=false`, the entire `namespaceSelector:` block is omitted while `objectSelector` and all other webhook fields render unchanged.
- `helm lint` and the `trivy config` chart check (CI `lint_chart` equivalent) both pass.

Yes. It introduces an optional chart value for managed Kubernetes environments where the webhook `namespaceSelector` is mutated / owned externally. Default behavior remains unchanged.

## Summary by CodeRabbit

- **New Features**
  - Added a Helm configuration option to control whether the scheduler admission webhook manages its namespace selector.
  - The option defaults to enabled; setting it to disabled will omit the namespace selector from the webhook configuration.

- **Documentation**
  - Updated the chart values documentation to describe the new `scheduler.admissionWebhook.manageNamespaceSelector` setting and its default value.

### DIFF
--- a/charts/hami/README.md
+++ b/charts/hami/README.md
@@ -111,6 +111,7 @@ This document provides detailed descriptions of all configurable values paramete
 | `scheduler.admissionWebhook.customURL.host` | Custom URL host | `127.0.0.1` |
 | `scheduler.admissionWebhook.customURL.port` | Custom URL port | `31998` |
 | `scheduler.admissionWebhook.customURL.path` | Custom URL path | `/webhook` |
+| `scheduler.admissionWebhook.manageNamespaceSelector` | Whether the chart renders and manages the webhook namespaceSelector field | `true` |
 | `scheduler.admissionWebhook.reinvocationPolicy` | Reinvocation policy | `Never` |
 | `scheduler.admissionWebhook.failurePolicy` | Failure policy | `Ignore` |
 
--- a/charts/hami/templates/scheduler/webhook.yaml
+++ b/charts/hami/templates/scheduler/webhook.yaml
@@ -23,6 +23,7 @@ webhooks:
     failurePolicy: {{ .Values.scheduler.admissionWebhook.failurePolicy }}
     matchPolicy: Equivalent
     name: vgpu.hami.io
+    {{- if .Values.scheduler.admissionWebhook.manageNamespaceSelector }}
     namespaceSelector:
       {{- if .Values.scheduler.admissionWebhook.namespaceSelector.matchLabels }}
       matchLabels:
@@ -42,6 +43,7 @@ webhooks:
       {{- if .Values.scheduler.admissionWebhook.namespaceSelector.matchExpressions }}
       {{- toYaml .Values.scheduler.admissionWebhook.namespaceSelector.matchExpressions | nindent 6 }}
       {{- end }}
+    {{- end }}
     objectSelector:
       {{- if .Values.scheduler.admissionWebhook.objectSelector.matchLabels }}
       matchLabels:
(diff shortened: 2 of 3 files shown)


wawa0210:
----    
/assign mesutoezdil

archlitchi (approved):
/lgtm

PR #2098, opened 2026-07-20
chore: remove dead script, boilerplate file, and unused ParseConfig stubs

mesutoezdil:
**What type of PR is this?**

/kind cleanup

Removes code and files that are confirmed dead:

- `hack/update-generated-api.sh`: runs protoc against `pkg/api/*.proto`, a directory that no longer exists in the repo, so the script can only fail.
- `hack/boilerplate/boilerplate.go.txt`: referenced by nothing (no script, Makefile target, or workflow uses it).
- Empty `ParseConfig` stubs in `pkg/device/amd` and `pkg/device/awsneuron`: unlike every other vendor backend, these were never wired into `InitDevicesWithConfig` in `pkg/scheduler/config/config.go`, and their bodies are empty.

All four confirmed unreachable with `golang.org/x/tools/cmd/deadcode` (whole-program analysis rooted at the three binaries plus tests) and a repo-wide reference scan.

Fixes #

Deliberately left out: the unreachable helpers in `pkg/device-plugin/nvidiadevice/nvinternal` (kept to minimize drift from the upstream NVIDIA device plugin fork) and `P2PLinkType.String` (Stringer methods can be reached via interfaces/reflection that static analysis misses).


DSFans2014 (commented):
/lgtm

archlitchi (approved):
/lgtm

PR #2100, opened 2026-07-20
fix: harden HTTP servers and migconfig file permissions

mesutoezdil:
**What type of PR is this?**

/kind bug

Fixes gosec findings: sets `ReadHeaderTimeout` (15s) on the scheduler HTTP/HTTPS servers, the scheduler metrics server, and the vGPUmonitor metrics server (G112/G114, Slowloris); writes `/tmp/migconfig.yaml` with `0600` instead of `os.ModePerm` (G306). Also fixes a "webiste" typo in a workflow name.

Fixes #

The fixed `/tmp/migconfig.yaml` path itself is unchanged; nvidia-mig-parted reads the same path and it lives inside the container.

## Summary by CodeRabbit

* **Security**
  * Tightened permissions on the generated MIG configuration file so only the owning process can read it.
  * Improved logging when saving the generated MIG configuration fails.
* **Reliability**
  * Increased scheduler and GPU monitoring HTTP resilience by adding request header and read timeouts.
  * Updated metrics endpoints to run on dedicated routing instead of the default HTTP mux.
* **Tests**
  * Added unit coverage for MIG configuration file writing (content, permissions, and failure behavior).
* **Chores**
  * Corrected the release workflow display name.

### DIFF
--- a/cmd/scheduler/main.go
+++ b/cmd/scheduler/main.go
@@ -160,6 +160,7 @@ func start() error {
 			Addr:              config.HTTPBind,
 			Handler:           router,
 			ReadHeaderTimeout: 15 * time.Second,
+			ReadTimeout:       60 * time.Second,
 		}
 		if err := server.ListenAndServe(); err != nil {
 			return fmt.Errorf("listen and Serve error, %v", err)
@@ -188,6 +189,7 @@ func start() error {
 			Handler:           handler,
 			TLSConfig:         tlsCfg,
 			ReadHeaderTimeout: 15 * time.Second,
+			ReadTimeout:       60 * time.Second,
 		}
 		klog.InfoS("Starting HTTPS server", "address", addr)
 		if err := server.ListenAndServeTLS("", ""); err != nil {
--- a/cmd/scheduler/main.go
+++ b/cmd/scheduler/main.go
@@ -156,7 +156,12 @@ func start() error {
 	}
 
 	if len(tlsCertFile) == 0 || len(tlsKeyFile) == 0 {
-		if err := http.ListenAndServe(config.HTTPBind, router); err != nil {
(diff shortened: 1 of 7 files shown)


cmd/scheduler/main.go
```diff
@@ -156,7 +156,12 @@ func start() error {
 	}
 
 	if len(tlsCertFile) == 0 || len(tlsKeyFile) == 0 {
-		if err := http.ListenAndServe(config.HTTPBind, router); err != nil {
+		server := &http.Server{
+			Addr:              config.HTTPBind,
+			Handler:           router,
+			ReadHeaderTimeout: 15 * time.Second,
```
DSFans2014, line 162: Why not set `ReadTimeout` and `WriteTimeout` together?

mesutoezdil: i skipped WriteTimeout on purpose.. in Go it also counts the handler's own run time, so a slow /bind or /filter under load would get cut off mid-response. 
The callers (kube-scheduler and apiserver) already have their own timeouts, so it adds risk without real benefit. controller-runtime does the same, only ReadHeaderTimeout.     
but added ReadTimeout: 60s to all four servers
thanks for feedback

archlitchi:
by changing the permission to 0600, will it cause migconfig.yaml unreadable by mig-parted process?

mesutoezdil:
It won't break: nvidia-mig-parted is exec'd by the device plugin itself (util.go), so it runs as the same user (root, since the container is privileged). 0600 is owner read/write, so the reader is unaffected. It actually matters here: /tmp is a hostPath mount of the host's /tmp, so the old os.ModePerm left the MIG config world-writable on the host before root applies it.

archlitchi (approved):
/lgtm

PR #2105, opened 2026-07-21
fix(scheduler): scope per-type device list in fitInDevices

### LINKED ISSUE #2104: [Bug] fitInDevices leaks one device type's devices into another type's allocation for multi-type containers
## What happened

In `pkg/scheduler/score.go`, `fitInDevices` declares its `devs` accumulator once, *above* the `for _, k := range requests` loop, and never resets it between device types:

```go
func fitInDevices(...) (bool, string) {
	devs := device.ContainerDevices{}   // declared once
	...
	for _, k := range requests {        // one iteration per requested device type
		...
		devs = append(devs, tmpDevs[k.Type]...)                  // never reset: grows across types
		(*devinput)[k.Type] = append((*devinput)[k.Type], devs)  // stores the growing slice
	}
}
```

When a single container requests two device types, the entry stored for the second-processed type contains that type's devices **plus** every earlier type's devices. `*devinput` becomes the winning node's device allocation; during bind, `PatchAnnotations` → `EncodePodSingleDevice` serializes it verbatim into the pod's per-type allocation annotation. `EncodePodSingleDevice` does not filter by type, so a wrong-type device UUID ends up in a type's annotation and is read back by the device plugin during `Allocate`.

Each device type's allocation entry should contain only that type's devices.

## How to reproduce it

Create a Pod with a single container that requests two heterogeneous device types (for example an NVIDIA GPU resource together with a second vendor's resource). The device type processed second in the request map has the other type's device present in its allocation annotation.

The common single-device-type case is unaffected: the loop runs once and `devs` starts empty either way.

## Root cause

Loop-scoped accumulator declared at function scope. Moving `devs := device.ContainerDevices{}` to the first statement inside the `for _, k := range requests` loop resets it per device type.

This is scheduler-extender scope and is fully covered by a unit test in `pkg/scheduler` (no GPU hardware needed to validate). I have a fix and a regression test ready and will open a PR referencing this issue.

---

imantaba:
## What this fixes

Fixes #2104.

`fitInDevices` in `pkg/scheduler/score.go` declared its `devs` accumulator once, above the `for _, k := range requests` loop, and never reset it between device types. When a single container requests two device types, the second-processed type's entry in the returned `PodDevices` accumulated the first type's devices as well. That list is later serialized verbatim by `PatchAnnotations` → `EncodePodSingleDevice` into the pod's per-type allocation annotation (which does not filter by type), so a wrong-type device UUID is written into a type's annotation and read back by the device plugin during `Allocate`.

## The change

Declare `devs := device.ContainerDevices{}` as the first statement inside the per-type loop so each device type records only its own devices.

- Behavior-preserving for single-device-type requests (the common case): the loop runs once and `devs` starts empty either way.
- Correct for multi-type requests: each type's entry contains only its own devices.
- `devs` is not referenced anywhere after the loop, so narrowing its scope is safe.

## Testing

Added `Test_fitInDevices_MultiTypePartition`, which drives `fitInDevices` with two device types in one container and asserts each type's entry in the output `PodDevices` contains only its own device. The assertions check both type entries, so the test fails on the previous code regardless of Go's random map iteration order and passes after the fix.

- `go test ./pkg/scheduler/... -race -count=1` — pass
- `make verify` — pass (license headers, import aliases, golangci-lint, goimports)

This change is scoped to the scheduler extender and is validated by unit tests; no GPU hardware is required.

---

## AI assistance disclosure

This contribution used AI assistance (Claude Code): the bug was surfaced during an AI-assisted code review, and the one-line fix plus the regression test were drafted with Claude Code under my review. I have read and verified the change, understand the root cause and the fix, and take full responsibility for its correctness and for responding to review feedback.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed device allocation handling for requests that require multiple device types, ensuring results are correctly partitioned per type.
  * Prevented cross-type allocation leakage between device types within a single scheduling decision.
* **Tests**
  * Added a unit test covering multi-type device requests to verify per-type isolation of allocated device UUIDs.

### DIFF
--- a/pkg/scheduler/score.go
+++ b/pkg/scheduler/score.go
@@ -51,7 +51,6 @@ func getNodeResources(list NodeUsage, t string) []*device.DeviceUsage {
 
 func fitInDevices(node *NodeUsage, requests device.ContainerDeviceRequests, pod *corev1.Pod, nodeInfo *device.NodeInfo, devinput *device.PodDevices) (bool, string) {
 	//devmap := make(map[string]device.ContainerDevices)
-	devs := device.ContainerDevices{}
 	total, totalCore, totalMem := int32(0), int32(0), int32(0)
 	free, freeCore, freeMem := int32(0), int32(0), int32(0)
 	sums := 0
@@ -93,11 +92,10 @@ func fitInDevices(node *NodeUsage, requests device.ContainerDeviceRequests, pod 
 					klog.V(5).Infoln("After AddResourceUsage:", node.Devices.DeviceLists[nidx].Device)
 				}
 			}
-			devs = append(devs, tmpDevs[k.Type]...)
 		} else {
 			return false, reason
 		}
-		(*devinput)[k.Type] = append((*devinput)[k.Type], devs)
+		(*devinput)[k.Type] = append((*devinput)[k.Type], tmpDevs[k.Type])
 	}
 	return true, ""
 }
--- a/pkg/scheduler/score_test.go
+++ b/pkg/scheduler/score_test.go
@@ -3571,3 +3571,91 @@ func Test_Nvidia_GPU_Topology(t *testing.T) {
 		})
 	}
 }
+
+// fitMockDevice is a minimal device.Devices implementation whose Fit always
+// succeeds and returns a single device of its own type. Used to exercise
+// fitInDevices with more than one device type in a single container.
+type fitMockDevice struct {
+	typeName string
+	uuid     string
+}
+
+func (m *fitMockDevice) CommonWord() string { return m.typeName }
+func (m *fitMockDevice) MutateAdmission(_ *corev1.Container, _ *corev1.Pod) (bool, error) {
+	return false, nil
+}
+func (m *fitMockDevice) CheckHealth(_ string, _ *corev1.Node) (bool, bool) { return true, true }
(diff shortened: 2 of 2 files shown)


pkg/scheduler/score.go
```diff
@@ -61,6 +60,9 @@ func fitInDevices(node *NodeUsage, requests device.ContainerDeviceRequests, pod
 	}
 	//This loop is for requests for different devices
 	for _, k := range requests {
+		// devs is scoped per device type so one type's allocation does not
+		// leak into another type's entry in devinput.
+		devs := device.ContainerDevices{}
```
mesutoezdil: since `devs` now only ever holds a copy of `tmpDevs[k.Type]`, you can delete it and append `tmpDevs[k.Type]` to `(*devinput)[k.Type]` directly

mesutoezdil:
thx for pr, dco is red, pls solve it

mesutoezdil:
@coderabbitai full review

pkg/scheduler/score.go
imantaba: Done. `devs` variable deleted and `tmpDevs[k.Type]` appended.

imantaba:
> thx for pr, dco is red, pls solve it

dco is green now

archlitchi (approved):
/lgtm

FouoF (approved):

PR #1278, opened 2025-08-18
feat(webhook): Add additional check for initContainers

mayooot:
**What type of PR is this?**

Add one of the following kinds:

/kind cleanup
/kind feature

**What this PR does / why we need it**:
Add additional check for initContainers in webhook

PR #2116, opened 2026-07-23
feat: init-container GPU resource accounting

mayooot:
**What type of PR is this?**

Implements GPU resource accounting for init containers, following the design in
  [`docs/develop/initContainer-design.md`](https://github.com/Project-HAMi/HAMi/blob/master/docs/develop/initContainer-design.md) (#2064).

  Init containers run sequentially to completion before app containers start, so their
  GPU usage should never be summed with app usage. The effective per-device footprint is:

  effective = max( sum(app requests), max(init request) )

  This is applied consistently across the three accounting paths:

  - **Admission** (`webhook.go`): quota check uses the effective value; `MutateAdmission`
    now also runs over init containers.
  - **Scheduling** (`score.go`): each init container is fit independently against a fresh
    node copy with an isolated allocation context; app containers accumulate on the real
    node. Init allocations no longer pollute app quota/filter checks.
  - **Usage recording** (`scheduler.go`, `pods.go`): stores the collapsed view, and shrinks
    to app-only once init containers finish (one-time, with symmetric quota delta).

  ## Context

  Follow-up to my earlier attempt in #1278, which only touched the webhook and was
  incomplete. This PR implements the full model per the merged design doc.

## Summary by CodeRabbit

- **New Features**
  - Improved GPU resource accounting for pods with init containers using init-aware effective usage semantics.
  - Added automatic transition from init-inclusive usage to app-only usage once init containers complete successfully.
  - GPU admission, scheduling, and scoring/fit now consistently account for both init and app containers.

- **Bug Fixes**
  - Prevented init-container device allocations from polluting app container allocation/footprints.
  - Fixed quota usage tracking to remain consistent across pod updates, shrink transitions, and deletion.

- **Tests**
  - Added coverage for init-container accounting, quota evaluation, and allocation separation.

### DIFF
--- a/pkg/scheduler/score.go
+++ b/pkg/scheduler/score.go
@@ -144,10 +144,27 @@ func (s *Scheduler) calcScoreWithOptions(nodes *map[string]*NodeUsage, resourceR
 				}
 			}
 
+			// Init containers run sequentially to completion before any app
+			// container starts, so an init container never uses GPU at the same
+			// instant as an app container (or another init container). The correct
+			// per-device footprint is max(sum(app requests), max(init request)),
+			// which fits the node iff the app containers fit cumulatively AND each
+			// init container fits independently. resourceReqs is ordered init-first
+			// (see device.Resourcereqs), so entries with index < numInit are init
+			// containers. App containers accumulate on the real node copy as before;
+			// each init container is checked against a fresh copy of the node's
+			// entry state so its usage never piles onto the app containers'.
+			numInit := len(task.Spec.InitContainers)
+			var pristine *NodeUsage
+			if numInit > 0 {
+				pristine = node.DeepCopy()
+			}
+
 			// Assume the node is a fit by default. This handles pods with no device
 			// requests, which should be schedulable on any node.
 			ctrfit := true
 			deviceType := ""
+			appDevices := make(device.PodDevices)
 			//This loop is for different container request
 			for ctrid, n := range resourceReqs {
 				sums := 0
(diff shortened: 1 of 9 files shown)


pkg/scheduler/score.go
```diff
@@ -160,8 +177,26 @@ func (s *Scheduler) calcScoreWithOptions(nodes *map[string]*NodeUsage, resourceR
 					score.Devices[deviceType] = append(score.Devices[deviceType], device.ContainerDevices{})
 					continue
 				}
+				// Init containers fit independently against a fresh copy of the
+				// node's entry state; app containers accumulate on the real node.
+				fitNode := node
+				fitDevices := &appDevices
+				if ctrid < numInit {
```
mesutoezdil, line 184: each init container fits against its own fresh copy of the node so it could end up on a diff gpu than the app containers or other init containers in the same pod, is that intended or should they be pinned to the same device

pkg/scheduler/webhook.go
```diff
@@ -69,22 +69,39 @@ func (h *webhook) Handle(_ context.Context, req admission.Request) admission.Res
...
-				klog.Errorf("validating pod failed:%s", err.Error())
-				return admission.Errored(http.StatusInternalServerError, err)
+				return false, err
 			}
-			hasResource = hasResource || found
+			found = found || f
+		}
+		return found, nil
+	}
+	for idx := range pod.Spec.InitContainers {
```
mesutoezdil, line 90: mutateadmission now gets called on init containers for every vendor backend, did we confirm all the non nvidia impls are ok being invoked on an init ctr since before this diff they never saw one

pkg/scheduler/score_test.go
```diff
@@ -2658,6 +2658,82 @@ func Test_calcScore(t *testing.T) {
 	}
 }
 
+func TestCalcScoreInitContainerDoesNotPolluteAppAllocatedContext(t *testing.T) {
```
mesutoezdil, line 2661: only testing 1 init + 1 app on the same uuid here, might be worth adding a case w 2 init containers that land on diff uuids to lock in the behavior mentioned in score.go

DSFans2014 (commented):
hi @mayooot  Thank you very much for your contribution.  However, this overlaps with #1773 which @maishivamhoo123  is currently working on. And the documentation https://github.com/Project-HAMi/HAMi/pull/2064 was also designed by him

PR #2134, opened 2026-07-26
fix(mthreads): guard against zero device count in GenerateResourceRequests

### LINKED ISSUE #2133: Scheduler panics with integer divide by zero when a container requests mthreads.com/vgpu: 0
## What happened

`GenerateResourceRequests` in `pkg/device/mthreads/device.go` divides the requested memory and cores by the device count when building the `ContainerDeviceRequest`:

```go
if n, ok := v.AsInt64(); ok {   // n is the mthreads.com/vgpu count
    ...
    return device.ContainerDeviceRequest{
        Nums:             int32(n),
        Type:             MthreadsGPUDevice,
        Memreq:           int32(memnum) / int32(n),   // n can be 0
        MemPercentagereq: int32(mempnum),
        Coresreq:         corenum / int32(n),          // n can be 0
    }
}
```

`n` comes from the resource quantity via `AsInt64()`. A quantity of `"0"` returns `(0, true)`, so it passes the `ok` check and reaches the division, causing a `runtime error: integer divide by zero` panic.

This request is built from `device.Resourcereqs()`, which the scheduler's `Filter` handler calls synchronously for every registered device backend on every pod being scheduled (all backends are always loaded, so this fires even on clusters with no Moore Threads hardware). The route uses `julienschmidt/httprouter` with no `PanicHandler`, so the panic takes down the `Filter` handler for that pod on every scheduling attempt, leaving the pod stuck in `Pending` and spamming panic stack traces to the scheduler log.

A container that requests `0` of the count resource should be treated as "no request" (as already happens when the count resource is absent), not crash the scheduler.

## How to reproduce it

Create a Pod whose container sets the Moore Threads count resource to `0` alongside a memory or core request, e.g.:

```yaml
resources:
  limits:
    mthreads.com/vgpu: "0"
    mthreads.com/sgpu-memory: "1000"
```

The scheduler's `Filter` handler panics with `integer divide by zero` when scoring this pod.

## Root cause

Missing zero/negative guard on the device count before dividing `Memreq`/`Coresreq` by it. No other vendor backend divides these fields by the count (checked the whole `pkg/device` tree), so this is a mthreads-specific omission.

Fix is a small guard that returns an empty `ContainerDeviceRequest` when the count is `<= 0`, matching the existing fall-through for an absent count resource. Fully covered by a unit test in `pkg/device/mthreads` (no hardware needed). I have the fix and regression test ready and will open a PR referencing this issue.

---

imantaba:
## What this fixes

Fixes #2133.

`GenerateResourceRequests` in `pkg/device/mthreads/device.go` divides the requested memory and cores by the device count (`Nums`) when building the `ContainerDeviceRequest`. The count is read from the resource quantity via `AsInt64()`, and a value of `"0"` returns `(0, true)`, so it passes the existing `ok` check and reaches the division — producing a `runtime error: integer divide by zero` panic.

This request is built from `device.Resourcereqs()`, which the scheduler's `Filter` handler calls synchronously for every registered device backend on every pod. The route uses `httprouter` with no `PanicHandler`, so the panic takes down the `Filter` handler for that pod on every scheduling attempt, leaving the pod stuck in `Pending`.

## The change

Return an empty `ContainerDeviceRequest` when the device count is `<= 0`, before the division. This matches the existing fall-through behavior when the count resource is absent (both mean "no request"). No other vendor backend divides these fields by the count, so the change is scoped to mthreads.

## Testing

Added a `Test_GenerateResourceRequests` table case (`count set to zero with memory and cores`) that sets `mthreads.com/vgpu: "0"` alongside a memory and core request. On the previous code this case panics with `integer divide by zero`; after the fix it returns an empty request.

- `go test ./pkg/device/mthreads/ -race -count=1` — pass
- `make verify` — pass (license headers, import aliases, golangci-lint, goimports, staticcheck)

This change is scoped to the mthreads device backend and is validated by unit tests; no GPU hardware is required.

---

## AI assistance disclosure

This contribution used AI assistance (Claude Code): the bug was surfaced during an AI-assisted code review, and the one-line guard plus the regression test were drafted with Claude Code under my review. I have read and verified the change, understand the root cause and the fix, and take full responsibility for its correctness and for responding to review feedback.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved `mthreads` device admission validation by rejecting non-positive (`<= 0`) device count values with an error.
  * If the configured `mthreads` count limit is unset, admission now falls back to the container’s Requests.
  * Resource request generation now returns no requests when the resolved count is non-positive or exceeds 32-bit integer bounds.
* **Tests**
  * Added/expanded test cases for `0`, negative values, and overflow scenarios, including explicit error assertions.

### DIFF
--- a/pkg/device/mthreads/device.go
+++ b/pkg/device/mthreads/device.go
@@ -91,8 +91,6 @@ func ParseConfig(fs *flag.FlagSet) {
 func (dev *MthreadsDevices) MutateAdmission(ctr *corev1.Container, p *corev1.Pod) (bool, error) {
 	count, ok := ctr.Resources.Limits[corev1.ResourceName(MthreadsResourceCount)]
 	if !ok {
-		// Mirror GenerateResourceRequests, which also honours Requests; otherwise a
-		// request-only count would skip this validation and be consumed unchecked.
 		count, ok = ctr.Resources.Requests[corev1.ResourceName(MthreadsResourceCount)]
 	}
 	if ok {
@@ -207,10 +205,6 @@ func (dev *MthreadsDevices) GenerateResourceRequests(ctr *corev1.Container) devi
 				"container", ctr.Name,
 				"deviceCount", n)
 			if n <= 0 || n > math.MaxInt32 {
-				// A non-positive count is invalid. A count above math.MaxInt32 is
-				// rejected too: the Memreq/Coresreq below divide by int32(n), and a
-				// value such as 1<<32 truncates to 0 (or wraps negative), which would
-				// panic the scheduler's Filter handler with a divide-by-zero.
 				return device.ContainerDeviceRequest{}
 			}
 			memnum := 0
--- a/pkg/device/mthreads/device.go
+++ b/pkg/device/mthreads/device.go
@@ -90,6 +90,9 @@ func ParseConfig(fs *flag.FlagSet) {
 func (dev *MthreadsDevices) MutateAdmission(ctr *corev1.Container, p *corev1.Pod) (bool, error) {
 	count, ok := ctr.Resources.Limits[corev1.ResourceName(MthreadsResourceCount)]
 	if ok {
+		if count.Value() <= 0 {
+			return false, fmt.Errorf("%s must be greater than 0", MthreadsResourceCount)
+		}
 		if count.Value() > 1 {
 			ctr.Resources.Limits[corev1.ResourceName(MthreadsResourceCores)] = *resource.NewQuantity(count.Value()*int64(coresPerMthreadsGPU), resource.DecimalSI)
 			ctr.Resources.Limits[corev1.ResourceName(MthreadsResourceMemory)] = *resource.NewQuantity(count.Value()*int64(memoryPerMthreadsGPU), resource.DecimalSI)
@@ -197,6 +200,11 @@ func (dev *MthreadsDevices) GenerateResourceRequests(ctr *corev1.Container) devi
 			klog.InfoS("Detected mthreads device request",
 				"container", ctr.Name,
 				"deviceCount", n)
+			if n <= 0 {
+				// A zero or negative device count is not a valid request; return an
+				// empty request to avoid the divide-by-zero in the Memreq/Coresreq below.
+				return device.ContainerDeviceRequest{}
+			}
 			memnum := 0
 			mem, ok := ctr.Resources.Limits[mthreadsResourceMem]
 			if !ok {
--- a/pkg/device/mthreads/device.go
+++ b/pkg/device/mthreads/device.go
@@ -20,6 +20,7 @@ import (
 	"errors"
 	"flag"
 	"fmt"
+	"math"
 	"slices"
 	"strconv"
 	"strings"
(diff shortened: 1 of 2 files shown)


FouoF:
The `MutateAdmission` has already checked it, so it dose not need to check again.

mesutoezdil:
> The `MutateAdmission` has already checked it, so it dose not need to check again.

but mutateAdmission only checks the memory value, never count <= 0, so vgpu: "0" passes the webhook and still hits the division in GenerateResourceRequests. isnt it?

mesutoezdil:
i think the main gap s fix stops the panic but MutateAdmission still doesn't reject count <= 0, so the pod is silently admitted with no gpu instead of a clear error, unlike enflame's  pattern. 
as yu see, also missing a test for negative count.

imantaba:
Thanks both. @FouoF   `MutateAdmission` doesn't actually cover this case: it only branches on `count > 1` and then validates the memory value, so a request with `mthreads.com/vgpu: "0"` (or negative) plus a legal memory value passes the webhook untouched and reaches GenerateResourceRequests, where `memnum / n` and `corenum / n` divide by zero and panic the scheduler's Filter handler.

@mesutoezdil  I pushed an update.
`MutateAdmission` now rejects `count <= 0` with a clear error, matching the enflame backend's `count.Value() <= 0` pattern, so the user gets a helpful message instead of a pod silently admitted with no GPU.
Kept the guard in GenerateResourceRequests (empty request for `count <= 0`) as defense-in-depth for any path that skips the webhook.
Added tests: `MutateAdmission` rejects` count == 0`, and `GenerateResourceRequests` returns an empty request (no panic) for count `0` and `-1`.

imantaba:
Could a maintainer re-run the e2e job?

pkg/device/mthreads/device.go
```diff
@@ -197,6 +200,11 @@ func (dev *MthreadsDevices) GenerateResourceRequests(ctr *corev1.Container) devi
 			klog.InfoS("Detected mthreads device request",
 				"container", ctr.Name,
 				"deviceCount", n)
+			if n <= 0 {
```
mesutoezdil: the guard is on n but the div below uses int32(n), so a count like 4294967296 wraps to 0 and still panics, can u check int32(n) <= 0 instead?

imantaba: Good catch, thanks @mesutoezdil.
Fixed.

pkg/device/mthreads/device.go
```diff
@@ -90,6 +90,9 @@ func ParseConfig(fs *flag.FlagSet) {
 func (dev *MthreadsDevices) MutateAdmission(ctr *corev1.Container, p *corev1.Pod) (bool, error) {
 	count, ok := ctr.Resources.Limits[corev1.ResourceName(MthreadsResourceCount)]
 	if ok {
+		if count.Value() <= 0 {
+			return false, fmt.Errorf("%s must be greater than 0", MthreadsResourceCount)
+		}
```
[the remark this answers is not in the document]
imantaba: fixed

pkg/device/mthreads/device_test.go
```diff
@@ -160,7 +160,34 @@ func Test_MutateAdmission(t *testing.T) {
...
+				p: &corev1.Pod{
+					ObjectMeta: metav1.ObjectMeta{
+						Annotations: map[string]string{
+							"mthreads.com/request-gpu-num": "test123",
+						},
+					},
+				},
+			},
+			want:    false,
+			wantErr: true,
```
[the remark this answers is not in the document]
imantaba: Added

imantaba:
Can someone please fix unhealthy pod on deploy hami helm step in e2e-test job

pkg/device/mthreads/device.go
```diff
@@ -89,7 +90,15 @@ func ParseConfig(fs *flag.FlagSet) {
 
 func (dev *MthreadsDevices) MutateAdmission(ctr *corev1.Container, p *corev1.Pod) (bool, error) {
 	count, ok := ctr.Resources.Limits[corev1.ResourceName(MthreadsResourceCount)]
+	if !ok {
+		// Mirror GenerateResourceRequests, which also honours Requests; otherwise a
```
archlitchi: please remove comments here and all other unnecessary comments

imantaba: Done, removed both comments.
The Requests-fallback note in `MutateAdmission` and the overflow note in `GenerateResourceRequests`. The code carries the intent on its own (`count.Value() <= 0` guard and the math.MaxInt32 check), so they weren't needed.

imantaba: Pushed

archlitchi (approved):
/lgtm

PR #2139, opened 2026-07-27
fix webhook to deny privileged containers

### LINKED ISSUE #2138: [Bug] Mutating webhook allows pods with privileged containers when other containers request GPU resources
## What happened

When a pod contains both a **privileged container** and a **GPU-requesting container**, the mutating admission webhook logs that it is denying admission for the privileged container, but still **allows** the pod and applies scheduler/GPU patches.

Any pod with a privileged container should be rejected with `admission.Denied(...)`.

## How to reproduce

1. Configure the HAMi webhook with a GPU resource name (e.g. `hami.io/gpu`).
2. Submit a pod with two containers:
   - Container A: `securityContext.privileged: true`
   - Container B: requests `hami.io/gpu: "1"`
3. Observe the webhook response: `Allowed: true` with patches applied.

Minimal pod spec:

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: privileged-gpu-mixed
  namespace: default
spec:
  containers:
  - name: privileged-sidecar
    image: busybox
    securityContext:
      privileged: true
  - name: gpu-workload
    image: busybox
    resources:
      limits:
        hami.io/gpu: "1"
```

## Root cause

In `pkg/scheduler/webhook.go`, the privileged-container check uses `continue` instead of returning `admission.Denied`:

```go
if ctr.SecurityContext.Privileged != nil && *ctr.SecurityContext.Privileged {
    klog.Warningf(..., "Denying admission as container %s is privileged", ...)
    continue  // should return admission.Denied(...)
}
```

The loop continues to the next container, GPU resources are mutated, and the pod is admitted.

## Impact

- Security/policy bypass for mixed privileged + GPU pods
- Webhook logs say "Denying admission" but the request is allowed

## Environment

- HAMi master @ c343242
- Verified locally with a unit test against `pkg/scheduler/webhook.go`

Jay2006sawant:
**What type of PR is this?**

/kind bug

The mutating admission webhook in `pkg/scheduler/webhook.go` logged "Denying admission" for privileged containers but used `continue` instead of returning `admission.Denied`. When a pod had both a privileged container and a GPU-requesting container, the webhook still allowed the pod and applied scheduler/GPU patches.

This PR returns `admission.Denied` immediately when any container is privileged, and adds unit tests for privileged-only and mixed privileged+GPU pods.

Fixes #2138 

- Change is limited to `webhook.go` and `webhook_test.go`
- No GPU hardware required to verify
- `make test` and `make verify` pass locally

Yes. Pods containing any privileged container are now rejected by the HAMi mutating webhook. Previously, a pod with a privileged sidecar plus a GPU workload could be admitted.

## Summary by CodeRabbit

* **Bug Fixes**
  * Admission requests that include privileged containers are now denied when the workload also triggers GPU mutation, with a denial message identifying the privileged container.
  * Privileged containers are handled consistently across both init and regular containers (including privileged sidecars).

* **Tests**
  * Added coverage to verify denial for privileged-only, privileged + GPU workload, and scheduler-name mismatch scenarios, and to confirm no patches are returned on denial.

### DIFF
--- a/pkg/scheduler/webhook.go
+++ b/pkg/scheduler/webhook.go
@@ -62,10 +62,6 @@ func (h *webhook) Handle(_ context.Context, req admission.Request) admission.Res
 		klog.Warningf(template+" - Denying admission as pod has no containers", pod.Namespace, pod.Name, pod.UID)
 		return admission.Denied("pod has no containers")
 	}
-	if name, privileged := privilegedContainerName(pod); privileged {
-		klog.Warningf(template+" - Denying admission as container %s is privileged", pod.Namespace, pod.Name, pod.UID, name)
-		return admission.Denied(fmt.Sprintf("container %s is privileged", name))
-	}
 	if pod.Spec.SchedulerName != "" &&
 		(pod.Spec.SchedulerName != corev1.DefaultSchedulerName || !config.ForceOverwriteDefaultScheduler) &&
 		(len(config.SchedulerName) == 0 || pod.Spec.SchedulerName != config.SchedulerName) {
@@ -73,6 +69,7 @@ func (h *webhook) Handle(_ context.Context, req admission.Request) admission.Res
 		return admission.Allowed("pod already has different scheduler assigned")
 	}
 	klog.V(5).Infof(template, pod.Namespace, pod.Name, pod.UID)
+	privilegedName, hasPrivileged := privilegedContainerName(pod)
 	hasResource := false
 	for idx := range pod.Spec.Containers {
 		c := &pod.Spec.Containers[idx]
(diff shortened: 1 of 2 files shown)


pkg/scheduler/webhook.go
```diff
@@ -61,6 +62,10 @@ func (h *webhook) Handle(_ context.Context, req admission.Request) admission.Res
 		klog.Warningf(template+" - Denying admission as pod has no containers", pod.Namespace, pod.Name, pod.UID)
 		return admission.Denied("pod has no containers")
 	}
+	if name, privileged := privilegedContainerName(pod); privileged {
```
mesutoezdil: wait this check runs b4 the schedulername gate so it denies literally any privileged container even if the pod isnt even using hami sched, kinda turns this into a general policy check and it still fails open anyway since failurepolicy is ignore in the chart

Jay2006sawant: Thanks @mesutoezdil ! good point. Updated in d9bb61e: the scheduler gate runs first again, and privileged containers are denied only when the pod also requests HAMi GPU resources. Privileged-only pods and pods on other schedulers are no longer blocked.

archlitchi (approved):
/lgtm

PR #2141, opened 2026-07-27
fix(scheduler): deny privileged containers in webhook

### LINKED ISSUE #2138: [Bug] Mutating webhook allows pods with privileged containers when other containers request GPU resources

aryansri05:
## What changed

- Return a denied admission response as soon as the webhook sees a privileged container.
- Add a regression test for a pod that combines a privileged sidecar with a GPU-requesting workload container.

## Why

The webhook previously logged that it was denying privileged containers but then continued processing the pod. If another container requested GPU resources, the webhook could still allow and mutate the pod.

Fixes #2138.

## Validation

- `go test ./pkg/scheduler`

## Summary by CodeRabbit

* **Bug Fixes**
  * Pod admission is now denied when any container is configured as privileged.
  * Prevents privileged containers from bypassing admission checks while other containers continue processing.

* **Tests**
  * Added coverage verifying that pods containing privileged containers are rejected.

### DIFF
--- a/pkg/scheduler/webhook.go
+++ b/pkg/scheduler/webhook.go
@@ -74,7 +74,7 @@ func (h *webhook) Handle(_ context.Context, req admission.Request) admission.Res
 		if ctr.SecurityContext != nil {
 			if ctr.SecurityContext.Privileged != nil && *ctr.SecurityContext.Privileged {
 				klog.Warningf(template+" - Denying admission as container %s is privileged", pod.Namespace, pod.Name, pod.UID, c.Name)
-				continue
+				return admission.Denied("privileged containers are not allowed")
 			}
 		}
 		for _, val := range device.GetDevices() {
--- a/pkg/scheduler/webhook_test.go
+++ b/pkg/scheduler/webhook_test.go
@@ -545,6 +545,91 @@ func TestEmptyContainersDenied(t *testing.T) {
 	}
 }
 
+func TestPrivilegedContainerDeniedAdmission(t *testing.T) {
+	prevSchedulerName := config.SchedulerName
+	prevForceOverwrite := config.ForceOverwriteDefaultScheduler
+	t.Cleanup(func() {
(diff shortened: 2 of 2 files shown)


archlitchi:
Fixed in https://github.com/Project-HAMi/HAMi/pull/2139

Issue #2126, opened 2026-07-24
[LFX Mentorship] Improve HAMi GPU observability metrics and dashboards


rootsongjc:
## Summary

Improve the usefulness, consistency, and documentation of HAMi GPU observability for operators running shared GPU workloads.

This issue is the umbrella upstream tracking issue for a proposed CNCF LFX Mentorship program. Existing reports cover individual metrics, labels, exporters, and dashboards, but there is no single open issue that defines a term-sized observability outcome.

## Scope

- Inventory the current HAMi metrics, exporters, labels, dashboards, and documentation.
- Identify and prioritize gaps with mentors using real operator workflows.
- Improve a focused set of metrics and labels needed to understand allocation, utilization, memory, and failures.
- Deliver and validate a Grafana dashboard for the selected metrics.
- Document metric semantics, units, labels, cardinality considerations, and example PromQL.
- Produce an OpenTelemetry trace design and a bounded proof of concept for one mentor-approved control-plane path; full production tracing across every HAMi component is not required for this term.
- Add practical alerting or SLO guidance based on the resulting telemetry.

## Expected deliverables

- An observability gap analysis and mentor-approved implementation plan.
- Maintainer-reviewed improvements to selected metrics/exporters and their tests.
- A versioned Grafana dashboard with setup and validation instructions.
- Metrics reference documentation and example queries.
- An OpenTelemetry design plus a small, demonstrable tracing prototype for one selected workflow.
- Operator guidance for alerts or SLOs grounded in the delivered signals.

## Acceptance criteria

- Metric names, types, units, labels, and cardinality are documented and reviewed.
- Dashboard panels use documented queries and are validated against a reproducible HAMi deployment.
- Changes preserve compatibility where practical; breaking telemetry changes are explicitly documented.
- The tracing prototype has a clearly bounded path and does not block completion of the metrics/dashboard deliverables.
- Relevant code, tests, dashboards, and documentation pass project CI and maintainer review.

## Out of scope

- Full production-grade tracing across webhook, scheduler extension, device plugin, monitor, and every runtime path in one mentorship term.
- Solving every vendor-specific telemetry incompatibility.
- Replacing Prometheus, DCGM Exporter, Grafana, or OpenTelemetry.

## Related issues

- #1778
- #2015
- #1721
- #1775
- #410
- #498
- #24

Aryanbhargava18:
[...]

PR #2142, opened 2026-07-27
fix(metrics): resolve high-cardinality leak in memory_allocated_bytes

Aryanbhargava18:
## What happened
The `hami_gpu_memory_allocated_bytes` metric (and its legacy counterpart `GPUDeviceMemoryAllocated`) defines a `device_cores` label but incorrectly injects `fmt.Sprint(devs.Device.Usedcores)` as its value instead of `Totalcore`. 

Because `Usedcores` is a dynamic state variable (changing every time a pod is scheduled or unscheduled), this causes a severe **Prometheus High Cardinality Leak**. Every time the used cores change, Prometheus is forced to create a completely new time series, abandoning the old one. On a cluster with high pod churn, this combinatorial explosion of stale time series will eventually cause Prometheus OOMs.

Prometheus best practices dictate that labels must represent stable identity attributes, not dynamic state. The `device_cores` label should represent the static core capacity of the device (`Totalcore`), while the dynamic state should remain as metric values (e.g., via `hami_gpu_core_allocated_ratio`). 

## How to reproduce
1. Deploy HAMi and scrape `:31993/metrics`.
2. Schedule a pod that requests GPU cores.
3. Observe `hami_gpu_memory_allocated_bytes`.
4. Delete the pod.
5. Observe that a **new** time series is created for `hami_gpu_memory_allocated_bytes` with a different `device_cores` label, instead of updating the existing series. 

## Root cause
In `cmd/scheduler/metrics.go`:
```go
ch <- prometheus.MustNewConstMetric(
    nodevGPUMemoryAllocatedDesc,
    prometheus.GaugeValue,
    float64(devs.Device.Usedmem)*float64(1024)*float64(1024),
    nodeID, devs.Device.ID, fmt.Sprint(devs.Device.Index), 
    fmt.Sprint(devs.Device.Usedcores), // <-- BUG: Injecting dynamic state into label
    devs.Device.Type,
)
```

## Fix
Changed `devs.Device.Usedcores` to `devs.Device.Totalcore` when populating the `device_cores` label, stabilizing the time series identity.

**Note:** This aligns directly with the gap analysis currently open in #2126 regarding metric cardinality risks.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated GPU-related Prometheus metrics to align label definitions and semantics across both current and legacy formats.
  * For GPU memory allocation, the `device_cores` label now reflects total device cores (while the underlying memory allocation value remains based on used memory).
  * For GPU node overview, the label set was simplified by removing the container-sharing dimension, and the `device_cores` / `device_memory_limit` labels now report total cores and memory limits.

### DIFF
--- a/cmd/scheduler/metrics.go
+++ b/cmd/scheduler/metrics.go
@@ -65,7 +65,7 @@ func (cc ClusterManagerCollector) Collect(ch chan<- prometheus.Metric) {
 	nodevGPUMemoryAllocatedDesc := prometheus.NewDesc(
 		"hami_gpu_memory_allocated_bytes",
 		"Device memory allocated for a certain GPU",
-		[]string{"node", "device_uuid", "device_index", "device_cores", "device_type"}, nil,
+		[]string{"node", "device_uuid", "device_index", "device_type"}, nil,
 	)
 	nodevGPUSharedNumDesc := prometheus.NewDesc(
 		"hami_gpu_shared_count",
@@ -80,7 +80,7 @@ func (cc ClusterManagerCollector) Collect(ch chan<- prometheus.Metric) {
 	nodeGPUOverview := prometheus.NewDesc(
 		"hami_node_gpu_overview",
 		"GPU overview on a certain node",
-		[]string{"node", "device_uuid", "device_index", "device_cores", "shared_containers", "device_memory_limit", "device_type"}, nil,
+		[]string{"node", "device_uuid", "device_index", "device_type"}, nil,
 	)
 	nodeGPUMemoryPercentage := prometheus.NewDesc(
 		"hami_node_gpu_memory_allocated_ratio",
@@ -121,7 +121,7 @@ func (cc ClusterManagerCollector) Collect(ch chan<- prometheus.Metric) {
 		legacyMemoryAllocatedDesc = prometheus.NewDesc(
 			"GPUDeviceMemoryAllocated",
 			"Device memory allocated for a certain GPU",
-			[]string{"nodeid", "deviceuuid", "deviceidx", "devicecores", "devicetype"}, nil,
+			[]string{"nodeid", "deviceuuid", "deviceidx", "devicetype"}, nil,
 		)
 		legacySharedNumDesc = prometheus.NewDesc(
 			"GPUDeviceSharedNum",
@@ -136,7 +136,7 @@ func (cc ClusterManagerCollector) Collect(ch chan<- prometheus.Metric) {
 		legacyOverview = prometheus.NewDesc(
 			"nodeGPUOverview",
 			"GPU overview on a certain node",
-			[]string{"nodeid", "deviceuuid", "deviceidx", "devicecores", "sharedcontainers", "devicememorylimit", "devicetype"}, nil,
+			[]string{"nodeid", "deviceuuid", "deviceidx", "devicetype"}, nil,
 		)
 		legacyMemoryPercentage = prometheus.NewDesc(
 			"nodeGPUMemoryPercentage",
@@ -208,7 +208,7 @@ func (cc ClusterManagerCollector) Collect(ch chan<- prometheus.Metric) {
 				nodevGPUMemoryAllocatedDesc,
 				prometheus.GaugeValue,
 				float64(devs.Device.Usedmem)*float64(1024)*float64(1024),
-				nodeID, devs.Device.ID, fmt.Sprint(devs.Device.Index), fmt.Sprint(devs.Device.Totalcore), devs.Device.Type,
+				nodeID, devs.Device.ID, fmt.Sprint(devs.Device.Index), devs.Device.Type,
 			)
 			ch <- prometheus.MustNewConstMetric(
 				nodevGPUSharedNumDesc,
@@ -226,7 +226,7 @@ func (cc ClusterManagerCollector) Collect(ch chan<- prometheus.Metric) {
 				nodeGPUOverview,
 				prometheus.GaugeValue,
 				float64(devs.Device.Usedmem)*float64(1024)*float64(1024),
-				nodeID, devs.Device.ID, fmt.Sprint(devs.Device.Index), fmt.Sprint(devs.Device.Totalcore), fmt.Sprint(devs.Device.Used), fmt.Sprint(devs.Device.Totalmem), devs.Device.Type,
+				nodeID, devs.Device.ID, fmt.Sprint(devs.Device.Index), devs.Device.Type,
 			)
 			ch <- prometheus.MustNewConstMetric(
 				nodeGPUMemoryPercentage,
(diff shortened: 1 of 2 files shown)


cmd/scheduler/metrics.go
```diff
@@ -208,7 +208,7 @@ func (cc ClusterManagerCollector) Collect(ch chan<- prometheus.Metric) {
 				nodevGPUMemoryAllocatedDesc,
 				prometheus.GaugeValue,
 				float64(devs.Device.Usedmem)*float64(1024)*float64(1024),
-				nodeID, devs.Device.ID, fmt.Sprint(devs.Device.Index), fmt.Sprint(devs.Device.Usedcores), devs.Device.Type,
+				nodeID, devs.Device.ID, fmt.Sprint(devs.Device.Index), fmt.Sprint(devs.Device.Totalcore), devs.Device.Type,
```
mesutoezdil, line 211: same exact bug is still left at line 229 (nodegpuoverview) and 273 (legacyoverview), device_cores label there is also filled with usedcores instead of totalcore so those 2 metrics keep leaking cardinality too

FouoF:
Better remove these changing labels. Actually other metrics can fully replace them.

Aryanbhargava18:
Makes sense. I've completely removed the `device_cores`, `shared_containers`, and `device_memory_limit` labels from those metrics, since the other gauges already cover them. Pushed the update.

cmd/scheduler/metrics.go
```diff
@@ -80,7 +80,7 @@ func (cc ClusterManagerCollector) Collect(ch chan<- prometheus.Metric) {
 	nodeGPUOverview := prometheus.NewDesc(
 		"hami_node_gpu_overview",
 		"GPU overview on a certain node",
-		[]string{"node", "device_uuid", "device_index", "device_cores", "shared_containers", "device_memory_limit", "device_type"}, nil,
+		[]string{"node", "device_uuid", "device_index", "device_type"}, nil,
```
mesutoezdil: w/ those 3 labels gone this metric now has the same label set and the same value as hami_gpu_memory_allocated_bytes so its a pure dup, maybe keep the static ones (totalcore/totalmem) or deprecate the metric?

cmd/scheduler/metrics.go
```diff
@@ -65,7 +65,7 @@ func (cc ClusterManagerCollector) Collect(ch chan<- prometheus.Metric) {
 	nodevGPUMemoryAllocatedDesc := prometheus.NewDesc(
 		"hami_gpu_memory_allocated_bytes",
 		"Device memory allocated for a certain GPU",
-		[]string{"node", "device_uuid", "device_index", "device_cores", "device_type"}, nil,
+		[]string{"node", "device_uuid", "device_index", "device_type"}, nil,
```
mesutoezdil: dropping labels here breaks existing dashboards so it needs a release note, and skill/hami-vgpu-metrics-summary/SKILL.md:248 still documents the overview labels, can u update it?

Aryanbhargava18:
@mesutoezdil Great catch on the dashboard breakage and the overview deduplication. 

I just pushed a fix that handles both:
1. Restored `device_cores` to `hami_gpu_memory_allocated_bytes` (and legacy equivalents) so existing dashboards won't break. However, to permanently fix the high cardinality leak, the label is now populated with the static `Totalcore` value instead of the dynamic `Usedcores` value.
2. Restored `device_cores` and `device_memory_limit` to `hami_node_gpu_overview` (and legacy), populated with `Totalcore` and `Totalmem` respectively. 
3. Permanently dropped the `shared_containers` label from the overview metric, since it is inherently dynamic and impossible to stabilize without leaking cardinality. 
4. Updated `skill/hami-vgpu-metrics-summary/SKILL.md` to reflect the static overview labels. 

This should completely solve the cardinality leak while keeping the static dimensions intact for PromQL aggregations.

mesutoezdil:
> @mesutoezdil Great catch on the dashboard breakage and the overview deduplication.
> 
> I just pushed a fix that handles both:
> 
> 1. Restored `device_cores` to `hami_gpu_memory_allocated_bytes` (and legacy equivalents) so existing dashboards won't break. However, to permanently fix the high cardinality leak, the label is now populated with the static `Totalcore` value instead of the dynamic `Usedcores` value.
> 2. Restored `device_cores` and `device_memory_limit` to `hami_node_gpu_overview` (and legacy), populated with `Totalcore` and `Totalmem` respectively.
> 3. Permanently dropped the `shared_containers` label from the overview metric, since it is inherently dynamic and impossible to stabilize without leaking cardinality.
> 4. Updated `skill/hami-vgpu-metrics-summary/SKILL.md` to reflect the static overview labels.
> 
> This should completely solve the cardinality leak while keeping the static dimensions intact for PromQL aggregations.

Reminder: Answers must be written by **human being**. You can view the relevant rule here.

"4. Review replies. The reply you post must be written by you and must address the specific point raised. Verbatim or canned AI replies, or replies that do not engage the comment, lead to the **PR being closed**."

Aryanbhargava18:
Apologies for the formatting, I didn't mean to sound like a bot. I was just using a list to track the labels you mentioned so I didn't miss anything.

TL;DR on the push:
- I put `Totalcore` and `Totalmem` back into the descriptors to keep the dashboards working.
- I dropped `shared_containers` entirely because the container count constantly fluctuates and thrashes the time series, which is what caused the OOMs in the first place.

Let me know if the PromQL queries in the grafana dashboards still complain with this label set.

mesutoezdil:
> Apologies for the formatting, I didn't mean to sound like a bot. I was just using a list to track the labels you mentioned so I didn't miss anything.
> 
> TL;DR on the push:
> 
> * I put `Totalcore` and `Totalmem` back into the descriptors to keep the dashboards working.
> * I dropped `shared_containers` entirely because the container count constantly fluctuates and thrashes the time series, which is what caused the OOMs in the first place.
> 
> Let me know if the PromQL queries in the grafana dashboards still complain with this label set.

thanks

btw: rules exist not bcs you are a bot, but to ensure you convey the relevant feedback to the reviewer/maintainer in the most concise and pertinent manner
thereby avoiding unnecessary clutter and overly long responses..

Aryanbhargava18:
> > Apologies for the formatting, I didn't mean to sound like a bot. I was just using a list to track the labels you mentioned so I didn't miss anything.
> > TL;DR on the push:
> > 
> > * I put `Totalcore` and `Totalmem` back into the descriptors to keep the dashboards working.
> > * I dropped `shared_containers` entirely because the container count constantly fluctuates and thrashes the time series, which is what caused the OOMs in the first place.
> > 
> > Let me know if the PromQL queries in the grafana dashboards still complain with this label set.
> 
> thanks
> 
> btw: rules exist not bcs you are a bot, but to ensure you convey the relevant feedback to the reviewer/maintainer in the most concise and pertinent manner thereby avoiding unnecessary clutter and overly long responses..

yeah that's fair, appreciate the clarification. Let me know if the updated static labels work for the dashboard queries on your end.

archlitchi (approved):
/lgtm

archlitchi (approved):
/lgtm

PR #2143, opened 2026-07-27
feat(monitor): implement hami_container_oom_kills_total metric to track workload stability

Aryanbhargava18:
## What happened
Currently, HAMi operators have no way of knowing if a GPU workload crashed due to Out-Of-Memory (OOM) without manually tailing Kubernetes events or running a separate observability stack. This observability gap makes it very difficult to tune `hami_vgpu_memory_limit_bytes` effectively, as operators cannot easily correlate allocation limits with workload stability.

The `vGPUMonitor` should natively export OOM metrics to Prometheus so that operators can immediately alert on GPU workloads that are crashing due to memory starvation.

## How to reproduce
1. Run a PyTorch/TensorFlow container with a restrictive `hami.io/vgpu-memory` limit.
2. Force the model to allocate more memory than the limit, causing an OOM kill.
3. Observe that `:31993/metrics` exports no indication that the container died from OOM.

## Fix
This PR implements **Phase 1** of the Observability gap analysis detailed in #2126.

It adds a new metric: `hami_container_oom_kills_total` (and legacy `Container_oom_kills_total`).
During the `collectPodAndContainerInfo` scrape loop, it inspects the Kubernetes `Pod.Status.ContainerStatuses` to check if the `LastTerminationState.Reason` is `OOMKilled`. If it is, the gauge is set to `1`. 

This allows operators to easily write PromQL alerts like:
`sum by (namespace, pod, container) (hami_container_oom_kills_total) > 0`

By natively integrating this into `vGPUMonitor`, we avoid forcing operators to deploy `kube-state-metrics` just to debug HAMi memory isolation boundaries.

## Summary by CodeRabbit

* **New Features**
  * Added Prometheus metrics for per-container OOM-kill status.
  * Provides OOM-kill visibility with namespace, pod, and container labels.
  * When legacy metrics are enabled, exports both the new metric and the legacy OOM-kill variant.

### DIFF
--- a/cmd/vGPUmonitor/metrics.go
+++ b/cmd/vGPUmonitor/metrics.go
@@ -140,6 +140,11 @@ var (
 		[]string{"namespace", "pod", "container", "vdevice_index", "device_uuid"}, nil,
 	)
 
+	ctrOOMKillsDesc = prometheus.NewDesc(
+		"hami_container_oom_kill_status",
+		"0 or 1 gauge indicating if the container's last termination was an OOM kill",
+		[]string{"namespace", "pod", "container"}, nil,
+	)
 	ctrDeviceMemoryBufferDesc = prometheus.NewDesc(
 		"hami_vgpu_memory_buffer_bytes",
 		"Container device memory buffer size in bytes",
@@ -157,6 +162,7 @@ var (
 	legacyCtrDeviceUtilizationdesc *prometheus.Desc
 	legacyCtrDeviceLastKernelDesc  *prometheus.Desc
 	legacyCtrDeviceMigInfo         *prometheus.Desc
+	legacyCtrOOMKillsDesc          *prometheus.Desc
 )
 
 func initLegacyDescriptors() {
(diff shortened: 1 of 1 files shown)


cmd/vGPUmonitor/metrics.go
```diff
@@ -559,6 +573,24 @@ func (cc ClusterManagerCollector) collectContainerMetrics(ch chan<- prometheus.M
...
+		if status.Name == ctr.Name {
+			if status.LastTerminationState.Terminated != nil && status.LastTerminationState.Terminated.Reason == "OOMKilled" {
+				oomKills = 1
+			}
+			break
+		}
+	}
+
+	labelsOOM := []string{pod.Namespace, pod.Name, ctr.Name}
+	if err := sendMetric(ch, ctrOOMKillsDesc, prometheus.GaugeValue, oomKills, labelsOOM...); err != nil {
```
mesutoezdil, line 588: naming it _total but sending it as a 0/1 gauge based only on lastterminationstate is kinda misleading, its not an actual cumulative kill count so rate()/increase() in promql on this would give garbage, maybe use restartcount or a real counter instead

FouoF:
@Aryanbhargava18 Thanks for the contribution. 

One key concern: Kubernetes OOMKilled reflects host/cgroup memory pressure (system RAM), not GPU / vGPU memory exhaustion. Exceeding hami.io/vgpu-memory(maybe hami.io/gpumem by default) typically surfaces as a CUDA OOM inside the process, and does not set the container termination reason to OOMKilled. If you want to implement it, you need to do some work on https://github.com/Project-HAMi/HAMi-core.

Aryanbhargava18:
Makes total sense. I was conflating host OOM with CUDA OOM. Since Kubernetes doesn't catch the CUDA memory crashes as OOMKilled, this PR won't work. I'll close it and look into doing this in HAMi-core instead. Thanks

PR #2145, opened 2026-07-27
fix: bound int32 conversion in enflame device parsing

mesutoezdil:
/kind bug

CodeQL flagged 4 spots in pkg/device/enflame/device.go where strconv.Atoi results were cast to int32 with no range check. A large value could silently wrap into a wrong or negative number. Added a clamp helper and used it everywhere the conversion happens.


dongjiang1989 (approved):
thanks

archlitchi (approved):
/lgtm

Issue #2144, opened 2026-07-27
Deferred code-scanning findings: release workflow permissions, mig-parted CVE, CodeQL PR trigger


mesutoezdil:
Three findings from GitHub code scanning were reviewed and deliberately left as-is for now. Tracking here so they aren't lost.

1. permissions: write-all in auto-release.yaml
Flagged by Scorecard (TokenPermissionsID). Narrowing it to per-job least-privilege touches the release pipeline's GITHUB_TOKEN scope and can only be verified against a real tag push. Present since the workflow was first added (aaafe144, 2024-09-23), no incidents reported.

2. CVE-2026-39824 in nvidia-mig-parted
Trivy flags golang.org/x/sys v0.31.0, a transitive dependency of github.com/NVIDIA/mig-parted@v0.12.2, installed in docker/Dockerfile, Dockerfile.hamicore, Dockerfile.hamimaster, and Dockerfile.withlib. Not present in our own go.mod (already v0.47.0). The CVE is a Windows-only NTUnicodeString overflow; HAMi only runs on Linux, so real-world impact is effectively none. Fixing it means bumping mig-parted to v0.14.4 (two minor versions), which could change MIG partitioning behavior and needs testing before rollout.

3. CodeQL not running on pull_request (SAST score)
.github/workflows/codeql-analysis.yml only triggers on push to master/dev and a weekly cron, not on pull_request. Scorecard's SAST check penalizes this. Adding a pull_request trigger would let GITHUB_TOKEN run in the context of external fork PRs, a known attack surface, so this needs a considered approach (e.g. pull_request_target with restricted checkout, or scoping to internal PRs only) rather than a blind trigger addition.

4. HAMICORE_IMAGE not pinnable in Dockerfile.hamimaster
This build-arg is produced by a separate CI job at build time (docker/build-push-action digest output from call-release-image-hamicore.yaml), not a static value, so it cannot be given a pinned default in the Dockerfile. Real fix needs wiring the digest through job outputs across auto-release.yaml, call-release-image-hamicore.yaml, and call-release-image.yaml.

5. BranchProtectionID
Not a code fix, needs a repo admin to configure branch protection rules under Settings > Branches.

[...]

PR #2147, opened 2026-07-27
ci: pin actions, images and install scripts by hash

mesutoezdil:
**What type of PR is this?**

/kind cleanup

OpenSSF Scorecard flagged 89 unpinned dependencies across workflows, Dockerfiles and install scripts. This pins them:
- all GitHub Actions to commit SHA, with the version kept as a trailing comment so it stays readable
- base Docker images (golang, nvidia/cuda, vllm, python) by digest
- the helm and golangci-lint install scripts now download to a file, verify sha256, then run, instead of piping curl into a shell
- pip install in the benchmark client Dockerfile pinned to an exact version

One image ref (HAMICORE_IMAGE build-arg in Dockerfile.hamimaster) is produced at build time by another job and can't be statically pinned, tracked in #2144.

Fixes #

All workflow YAML validated with a parser, both shell scripts pass bash -n, repo still builds clean.

No

## Summary by CodeRabbit

* **Security**
  * Improved supply-chain security by pinning GitHub Actions and automation tooling to immutable commit SHAs.
  * Added checksum verification for Helm installation and pinned, checksum-verified `golangci-lint` provisioning.
  * Pinned Docker base images and build/runtime images by digest.
* **Reliability**
  * CI, release, scanning, and chart/lint workflows now use deterministic action/tool versions for more consistent runs.
* **Maintenance**
  * Updated benchmark and image dependencies to fixed versions (including `requests` and `nvidia-mig-parted`).

### DIFF
--- a/.github/workflows/auto-label-pr.yaml
+++ b/.github/workflows/auto-label-pr.yaml
@@ -12,7 +12,7 @@ jobs:
    labeling:
     runs-on: ubuntu-latest
     steps:
-    - uses: github/issue-labeler@v3.4
+    - uses: github/issue-labeler@c1b0f9f52a63158c4adc09425e858e87b32e9685 # v3.4
       with:
         configuration-path: .github/labeler.yml
         enable-versioned-regex: 0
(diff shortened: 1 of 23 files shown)


Shouren (approved):
/lgtm

archlitchi (approved):
/lgtm

PR #2153, opened 2026-07-28
feat(metrics): add native node_name and pod labels to vGPUmonitor metrics

### LINKED ISSUE #2152: [Feature/Metrics] Instrument vGPUmonitor metrics with native node_name and pod labels
Instrument `vGPUmonitor` Prometheus metrics in `cmd/vGPUmonitor/metrics.go` to natively emit `node_name` and consistent Pod/Container identification labels (`node_name`, `namespace`, `pod`, `container`) directly from the binary.

**Why is this needed**:
Currently, host-level metrics exported by `vGPUmonitor` (such as `hami_host_gpu_memory_used_bytes` and `hami_host_gpu_utilization_ratio`) completely omit the `node_name` label. Meanwhile, container-level metrics (such as `hami_vgpu_memory_used_bytes` and `hami_container_device_memory_bytes`) include `namespace`, `pod`, and `container`, but also omit `node_name`.

In Kubernetes clusters with multiple GPU nodes, `vGPUmonitor` runs as a DaemonSet. When metrics are scraped in custom Prometheus deployments without ServiceMonitors (or without explicit Prometheus relabeling rules), host and container metrics lack node identification. This makes multi-node/pod metric joins, cross-node aggregations, and Grafana node/pod filtering difficult or inaccurate.

Natively instrumenting the metric descriptors inside `vGPUmonitor/metrics.go` to include `node_name` alongside container/pod metadata ensures all HAMi metrics are self-contained and consistent across all scraping configurations.

**Additional context**:
N.A.

swastikCommits:
**What type of PR is this?**

/kind feature

Currently, `vGPUmonitor` metrics (like `hami_host_gpu_memory_used_bytes` and `hami_vgpu_memory_used_bytes`) don't include the `node_name` label natively in their metric descriptors. Instead, they rely on external Prometheus `relabel_config` rules in Helm charts to add the node name. 
In custom Prometheus setups or standalone scrapers without ServiceMonitors, this causes metrics from different GPU nodes to lack node identification, making Grafana filtering and cross-node metric joins difficult.

In this PR we update `cmd/vGPUmonitor/metrics.go` so that `node_name` is natively attached to all host and container metric descriptors directly at the exporter level.

Fixes #2152 

Added unit test coverage in `cmd/vGPUmonitor/metrics_test.go` to check that all standard and legacy descriptors include the `node_name` / `nodename` label.

```release-note
Adds 'node_name' label key to standard vGPUmonitor Prometheus metrics (and 'nodename' to legacy metrics) sourced from Downward API NODE_NAME. Existing Grafana dashboards and Prometheus alerts querying vGPUmonitor series will now include node-level dimension labels.
```

**AI assistance disclosure**:
I used AI assistance (Gemini CLI) to help consult codebase patterns, dig deep into the issue and structure unit tests.
All code logic and changes were manually inspected and validated.

## Summary by CodeRabbit

* **New Features**
  * Added node identity labels as the first metric label for host GPU, container/vGPU, and MIG monitoring metrics (`node_name`; legacy `nodename`).

* **Bug Fixes**
  * Aligned label ordering between current and legacy metric descriptors and the emitted metric values.

* **Tests**
  * Added unit tests covering descriptor contents, node-name resolution from the node-name environment value, and correct metric delivery for both new and legacy emissions.

### DIFF
--- a/cmd/vGPUmonitor/metrics.go
+++ b/cmd/vGPUmonitor/metrics.go
@@ -88,62 +88,62 @@ var (
 	hostGPUdesc = prometheus.NewDesc(
 		"hami_host_gpu_memory_used_bytes",
 		"GPU device memory usage in bytes",
-		[]string{"device_index", "device_uuid", "device_type"}, nil,
+		[]string{"node_name", "device_index", "device_uuid", "device_type"}, nil,
 	)
 
 	hostGPUUtilizationdesc = prometheus.NewDesc(
 		"hami_host_gpu_utilization_ratio",
 		"GPU core utilization ratio (0-100)",
-		[]string{"device_index", "device_uuid", "device_type"}, nil,
+		[]string{"node_name", "device_index", "device_uuid", "device_type"}, nil,
 	)
 
 	ctrvGPUdesc = prometheus.NewDesc(
 		"hami_vgpu_memory_used_bytes",
 		"vGPU device memory usage in bytes",
-		[]string{"namespace", "pod", "container", "vdevice_index", "device_uuid"}, nil,
+		[]string{"node_name", "namespace", "pod", "container", "vdevice_index", "device_uuid"}, nil,
 	)
 
 	ctrvGPUlimitdesc = prometheus.NewDesc(
 		"hami_vgpu_memory_limit_bytes",
 		"vGPU device memory limit in bytes",
-		[]string{"namespace", "pod", "container", "vdevice_index", "device_uuid"}, nil,
+		[]string{"node_name", "namespace", "pod", "container", "vdevice_index", "device_uuid"}, nil,
 	)
 	ctrDeviceMemorydesc = prometheus.NewDesc(
 		"hami_container_device_memory_bytes",
 		`Container device memory usage in bytes`,
-		[]string{"namespace", "pod", "container", "vdevice_index", "device_uuid"}, nil,
+		[]string{"node_name", "namespace", "pod", "container", "vdevice_index", "device_uuid"}, nil,
 	)
 	ctrDeviceUtilizationdesc = prometheus.NewDesc(
 		"hami_container_device_utilization_ratio",
 		"Container device SM utilization ratio",
-		[]string{"namespace", "pod", "container", "vdevice_index", "device_uuid"}, nil,
+		[]string{"node_name", "namespace", "pod", "container", "vdevice_index", "device_uuid"}, nil,
 	)
 	ctrDeviceLastKernelDesc = prometheus.NewDesc(
 		"hami_container_last_kernel_elapsed_seconds",
 		"Seconds since last kernel execution in container",
-		[]string{"namespace", "pod", "container", "vdevice_index", "device_uuid"}, nil,
+		[]string{"node_name", "namespace", "pod", "container", "vdevice_index", "device_uuid"}, nil,
 	)
 	ctrDeviceMigInfo = prometheus.NewDesc(
 		"hami_mig_device_info",
 		"MIG device information for container",
-		[]string{"namespace", "pod", "container", "vdevice_index", "device_uuid", "instance_id"}, nil,
+		[]string{"node_name", "namespace", "pod", "container", "vdevice_index", "device_uuid", "instance_id"}, nil,
 	)
 	ctrDeviceMemoryContextDesc = prometheus.NewDesc(
 		"hami_vgpu_memory_context_bytes",
 		"Container device memory context size in bytes",
-		[]string{"namespace", "pod", "container", "vdevice_index", "device_uuid"}, nil,
+		[]string{"node_name", "namespace", "pod", "container", "vdevice_index", "device_uuid"}, nil,
 	)
 
 	ctrDeviceMemoryModuleDesc = prometheus.NewDesc(
 		"hami_vgpu_memory_module_bytes",
 		"Container device memory module size in bytes",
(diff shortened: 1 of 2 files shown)


cmd/vGPUmonitor/metrics.go
```diff
@@ -360,6 +360,14 @@ func (cc ClusterManagerCollector) collectGPUDeviceMetrics(ch chan<- prometheus.M
 	return nil
 }
 
+func getNodeName() string {
+	nodeName := os.Getenv(util.NodeNameEnvName)
+	if nodeName == "" {
+		nodeName, _ = os.Hostname()
```
mesutoezdil: os.Hostname() gives the pod name not the node name unless the ds runs w/ hostNetwork (its false by default in our chart), so this fallback silently mislabels every metric and makes the pod selector below match nothing, better to keep the hard error?

cmd/vGPUmonitor/metrics.go
```diff
@@ -508,7 +518,7 @@ func (cc ClusterManagerCollector) collectContainerMetrics(ch chan<- prometheus.M
 		smUtil := c.Info.DeviceSmUtil(i)
 		lastKernelTime := c.Info.LastKernelTime()
 
-		labels := []string{pod.Namespace, pod.Name, ctr.Name, fmt.Sprint(i), uuid}
+		labels := []string{getNodeName(), pod.Namespace, pod.Name, ctr.Name, fmt.Sprint(i), uuid}
```
mesutoezdil: getNodeName() runs per device inside this loop while the two host paths hoist it, and collectPodAndContainerInfo already has nodeName in scope, can u just pass it down?

cmd/vGPUmonitor/metrics.go
swastikCommits: You're right, `os.Hostname()` inside a container defaults to the Pod name when `hostNetwork` is false, which would produce misleading `node_name` labels. I will remove the `os.Hostname()` fallback and enforce a hard error when `NODE_NAME` is missing.

mesutoezdil:
thanks, pls refrain from force pushing once you have started getting reviews, 
it makes it much harder to tell what has changed from review to review

swastikCommits:
Sure I will keep that in mind. Thanks for the review :)

swastikCommits:
hi @mesutoezdil @archlitchi @wawa0210 does this require any further changes?

mesutoezdil (commented):
the body says user-facing change N.A. but this adds node_name as the first label to 11 exported metrics and nodename to 8 legacy ones, so anything keyed on those series changes. pls fill in the block w/ a release-note.

no ai disclosure either, see https://github.com/Project-HAMi/HAMi/blob/master/CONTRIBUTING.md#ai-assistance-notice

cmd/vGPUmonitor/metrics_test.go
```diff
@@ -0,0 +1,98 @@
...
+				t.Errorf("standard descriptor %s does not contain node_name label", descStr)
+			}
+		} else {
+			if !strings.Contains(descStr, "nodename") {
+				t.Errorf("legacy descriptor %s does not contain nodename label", descStr)
+			}
+		}
+	}
+
+	expectedCount := 17 // 9 standard + 8 legacy descriptors
```
mesutoezdil: Describe() only lists 9 of the 11 standard descs, ctrDeviceLastKernelDesc and ctrDeviceMigInfo arent in it, so what is a test called AllMetricDescriptors actually asserting for those 2?

cmd/vGPUmonitor/metrics.go
```diff
@@ -360,7 +360,16 @@ func (cc ClusterManagerCollector) collectGPUDeviceMetrics(ch chan<- prometheus.M
 	return nil
 }
 
+func getNodeName() string {
+	return os.Getenv(util.NodeNameEnvName)
+}
+
 func (cc ClusterManagerCollector) collectGPUMemoryMetrics(ch chan<- prometheus.Metric, hdev nvml.Device, index int) error {
+	nodeName := getNodeName()
```
mesutoezdil: still one getenv per device per scrape, collectGPUDeviceMetrics runs in the device loop and both helpers call getNodeName again, can u hoist it and pass it in like u did for the ctr path?

cmd/vGPUmonitor/metrics_test.go
swastikCommits: Thanks for the catch! `ctrDeviceLastKernelDesc` and `ctrDeviceMigInfo` were omitted from Describe(). 
I've added both descriptors to Describe() in `cmd/vGPUmonitor/metrics.go` and updated `metrics_test.go` to assert all 19 descriptors (11 standard + 8 legacy).

swastikCommits:
Hi @mesutoezdil! I've updated the PR body with the release-note and AI disclosure, hoisted `getNodeName()` to avoid redundant `os.Getenv` calls, and added the missing descriptors to `Describe()` (asserting all 19 in `metrics_test.go`).

All unit tests pass, its ready for a re-review :)

FouoF:
Thanks for the metrics work. This PR overlaps the same public label-contract decision being discussed in #2343 and #2398, but with a larger conflicting diff and incomplete coverage. Keeping all three open would create duplicate review work and competing metric schemas. We are closing this implementation so maintainers can select one contract in #2126 before reviewing code. Please document any requirement that the selected path does not cover in that issue.

PR #2155, opened 2026-07-28
fix leader election hostname prefix matching

### LINKED ISSUE #2154: [Bug] Leader election isHolderOf false positive when hostname is a prefix of another node's hostname
## What happened

## How to reproduce

```go
```

## Root cause

```go
```

```go
```

## Impact

## Environment

- HAMi master @ 7db31bf
- No GPU hardware required

Jay2006sawant:
**What type of PR is this?**

/kind bug

`isHolderOf` in `pkg/util/leaderelection/leaderelection.go` used `strings.HasPrefix(holder, hostname)` to match lease holder identities. Kubernetes holder IDs use the format `hostname_<uuid>`, but the check did not require the `_` separator.

This caused false positives when one node's hostname is a prefix of another's (e.g. `dev` matching `dev-server_<uuid>`). A node could incorrectly believe it holds the lease, trigger leader callbacks, and run leader-only scheduler work on a non-leader replica.

This PR matches holder identities with `hostname + "_"` and adds regression tests for prefix collision and valid local holder cases.

Fixes #2154 

- One-line logic change in `isHolderOf`
- Tests added in existing ginkgo suite (`27` specs pass)
- No GPU hardware required
- `make verify` passes locally

Yes, when scheduler leader election is enabled. Nodes whose hostnames are prefixes of other node hostnames will no longer falsely report as leader. This improves HA correctness in clusters with names like `dev` / `dev-server` or `worker` / `worker2`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved leader detection so a host is recognized only when the lease holder identity exactly matches the expected hostname and valid UUID suffix, preventing false matches and malformed identities.

* **Tests**
  * Expanded leader-detection coverage for valid identities, hostname-prefix collisions, missing suffixes, malformed UUIDs, and nil lease fields.

### DIFF
--- a/pkg/util/leaderelection/leaderelection_test.go
+++ b/pkg/util/leaderelection/leaderelection_test.go
@@ -451,7 +451,7 @@ var _ = ginkgo.Describe("Nil checks for lease fields", func() {
 					Namespace: namespace,
 				},
 				Spec: coordinationv1.LeaseSpec{
-					HolderIdentity:       func() *string { s := hostname; return &s }(),
+					HolderIdentity:       func() *string { s := generateHolderIdentity(hostname); return &s }(),
 					LeaseDurationSeconds: nil, // nil duration
 				},
 			}
@@ -487,6 +487,16 @@ var _ = ginkgo.Describe("isHolderOf hostname prefix boundary", func() {
 			},
 		}
 		g.Expect(lm.isHolderOf(lease)).Should(g.BeTrue())
+	})
+
+	ginkgo.It("should not match a bare hostname without uuid suffix", func() {
+		bareHolder := "dev"
+		lease := &coordinationv1.Lease{
+			Spec: coordinationv1.LeaseSpec{
+				HolderIdentity: &bareHolder,
+			},
+		}
+		g.Expect(lm.isHolderOf(lease)).Should(g.BeFalse())
 	})
 })
 
--- a/pkg/util/leaderelection/leaderelection_test.go
+++ b/pkg/util/leaderelection/leaderelection_test.go
@@ -462,6 +462,34 @@ var _ = ginkgo.Describe("Nil checks for lease fields", func() {
 	})
 })
 
+var _ = ginkgo.Describe("isHolderOf hostname prefix boundary", func() {
+	var lm *leaderManager
+
+	ginkgo.BeforeEach(func() {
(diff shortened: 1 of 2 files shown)


pkg/util/leaderelection/leaderelection_test.go
```diff
@@ -462,6 +462,34 @@ var _ = ginkgo.Describe("Nil checks for lease fields", func() {
 	})
 })
 
+var _ = ginkgo.Describe("isHolderOf hostname prefix boundary", func() {
```
mesutoezdil, line 465: the existing fixture at line 454 still builds the holder as the bare hostname so it no longer models "we are the holder", can u switch it to generateHolderIdentity(hostname)?

pkg/util/leaderelection/leaderelection_test.go
```diff
@@ -462,6 +462,34 @@ var _ = ginkgo.Describe("Nil checks for lease fields", func() {
...
+	})
+
+	ginkgo.It("should match the local hostname holder identity", func() {
+		localHolder := generateHolderIdentity("dev")
+		lease := &coordinationv1.Lease{
+			Spec: coordinationv1.LeaseSpec{
+				HolderIdentity: &localHolder,
+			},
+		}
+		g.Expect(lm.isHolderOf(lease)).Should(g.BeTrue())
```
mesutoezdil, line 489: the one input this line actually flips is a bare "dev" holder w/ no suffix (was true, now false), can u pin that case too?

Jay2006sawant: Hello @mesutoezdil ! addressed in 3c6cef3:

Line 454 now uses generateHolderIdentity(hostname).
Added a regression test for bare "dev" holder (no suffix) - returns false.

mesutoezdil (commented):
fyi #2178 was a closed dup of this w/ a different fix, `SplitN(holder, "_", 2)[0] == hostname`. worth noting yours is the safer one: theirs breaks for a hostname that itself contains `_`, since the split would cut it short and the pod would never see itself as leader.

no ai disclosure in the description, see https://github.com/Project-HAMi/HAMi/blob/master/CONTRIBUTING.md#ai-assistance-notice

Jay2006sawant:
@mesutoezdil Thanks for the review . addressed in `9542fa2`.PTAL!

PR #2156, opened 2026-07-28
fix(nvidia): treat gpumem-percentage of 0 as unset

Wangmin362:
**What type of PR is this?**

/kind bug

`nvidia.com/gpumem-percentage: 0` is accepted today, so the scheduler computes `memreq = Totalmem * 0 / 100 = 0`: the pod fits any card however full it is, is booked with 0 memory, and the device plugin injects `CUDA_DEVICE_MEMORY_LIMIT_0=0m`, which hami-core reads as "no limit". The container can then use the whole card while the scheduler still thinks it is free.

`nvidia.com/gpumem: 0` is fine because it leaves the percentage unset and falls through to the defaultMemory / whole-card branch below. This makes a non-positive percentage take the same path.

NONE

Clamping to 100 instead would also stop the 0-memory booking, but it skips that defaultMemory branch, and unlike the 101 sentinel, 100 is not excluded in `ComputeScore`, so scoring would count a whole card.

On a V100-32GB node, a pod with `gpumem-percentage: 0` got `0m` and allocated 8 GiB while still booked as 0; a second pod asking for `gpumem: 30000` on the same card was then admitted and died with `cuMemoryAllocate failed res=2`. I also ran a before/after A/B on an A100-80GB node with two shadow schedulers differing only in this change: an empty card is now booked as 81920 instead of 0, with 80000/81920 already booked the same request goes Pending with `CardInsufficientMemory`, and with `defaultMemory: 512` it is booked as 512 and the container gets `512m` instead of `0m`.

The tests added for a zero percentage fail on master and pass here. This PR was written primarily by Claude Code, and I reviewed and tested it.

A `nvidia.com/gpumem-percentage` of 0 is now treated as unset: the container gets the whole card, or `defaultMemory` when one is configured, instead of running with no GPU memory limit.

## Summary by CodeRabbit

* **Bug Fixes**
  * NVIDIA GPU memory requests with zero or negative percentages now use the default full-card allocation.
  * Percentage values above 100 remain capped at 100, while positive values are honored.
  * Explicit GPU memory requests continue to take precedence over percentage settings.
  * Resource accounting now correctly reserves full GPU memory for zero-percentage requests, preventing allocation when insufficient memory remains.

### DIFF
--- a/pkg/device/nvidia/device.go
+++ b/pkg/device/nvidia/device.go
@@ -561,11 +561,17 @@ func (dev *NvidiaGPUDevices) GenerateResourceRequests(ctr *corev1.Container) dev
 			if ok {
 				mempnums, ok := mem.AsInt64()
 				if ok {
-					if mempnums < 0 || mempnums > 100 {
+					if mempnums > 100 {
 						klog.ErrorS(nil, "memory percentage request out of range, clamping to 100", "container", ctr.Name, "requested", mempnums)
 						mempnums = 100
 					}
-					mempnum = int32(mempnums)
+					if mempnums > 0 {
+						mempnum = int32(mempnums)
+					} else {
+						// 0 would inject CUDA_DEVICE_MEMORY_LIMIT=0m, which hami-core reads as "no limit", so keep the "unset" sentinel and let the default below apply, like nvidia.com/gpumem: 0.
+						klog.ErrorS(nil, "memory percentage request is not positive, ignoring it", "container", ctr.Name, "requested", mempnums)
+						mempnum = 101
+					}
 				}
 			}
 			if mempnum == 101 && memnum == 0 {
--- a/pkg/device/nvidia/device_test.go
+++ b/pkg/device/nvidia/device_test.go
@@ -1940,6 +1940,79 @@ func TestGenerateResourceRequests(t *testing.T) {
 				Coresreq:         0,
 			},
 		},
+		{
+			name: "gpu count + memory percentage of 0 — treated as unset",
+			ctr: &corev1.Container{
+				Resources: corev1.ResourceRequirements{
+					Limits: corev1.ResourceList{
+						"nvidia.com/gpu":               *resource.NewQuantity(1, resource.BinarySI),
+						"nvidia.com/gpumem-percentage": *resource.NewQuantity(0, resource.DecimalSI),
+					},
+				},
+			},
+			want: device.ContainerDeviceRequest{
+				Nums:             1,
+				Type:             NvidiaGPUDevice,
+				Memreq:           0,
+				MemPercentagereq: 100,
+				Coresreq:         0,
+			},
+		},
+		{
+			name: "memory percentage of 0 in Requests — treated as unset",
+			ctr: &corev1.Container{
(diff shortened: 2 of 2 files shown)


pkg/device/nvidia/device.go
```diff
@@ -400,8 +400,9 @@ func (dev *NvidiaGPUDevices) MutateAdmission(ctr *corev1.Container, p *corev1.Po
 
 func (dev *NvidiaGPUDevices) validateMemoryPercentage(ctr *corev1.Container) error {
 	if pct, ok := resourceValue(ctr, corev1.ResourceName(dev.config.ResourceMemoryPercentageName)); ok {
-		if pct < 0 || pct > 100 {
-			return fmt.Errorf("invalid %s value %d in container %s: must be an integer between 0 and 100", dev.config.ResourceMemoryPercentageName, pct, ctr.Name)
+		// 0 reserves nothing and makes the plugin inject CUDA_DEVICE_MEMORY_LIMIT=0m, which means "no limit".
+		if pct <= 0 || pct > 100 {
```
mesutoezdil: the chart ships failurePolicy: Ignore by default so this error doesnt actually reject the pod, it just gets admitted unmutated and fails later at the kubelet, can u soften the release note to say that?

pkg/device/nvidia/device_test.go
```diff
@@ -1948,6 +2082,12 @@ func TestGenerateResourceRequests_DefaultMemory(t *testing.T) {
 	result := dev.GenerateResourceRequests(ctr)
 	assert.Equal(t, result.Memreq, int32(512))
 	assert.Equal(t, result.MemPercentagereq, int32(101))
+
+	// an out-of-range percentage is clamped to the whole card, it does not fall back to defaultMem
```
mesutoezdil: w/ defaultMem set a percentage of 0 now jumps from defaultMem to the whole card, is that the intent or shd an invalid pct fall back to defaultMem?

ouyangluwei163:
Could we handle `nvidia.com/gpumem-percentage: 0` the same way `nvidia.com/gpumem: 0` is already handled, entirely inside `Generate Resource Requests` — instead of rejecting it at admission?
https://github.com/Project-HAMi/HAMi/blob/master/pkg/device/nvidia/device.go#L557-L577

pkg/device/nvidia/device.go
Wangmin362: I have dropped the admission change, thank you

pkg/device/nvidia/device_test.go
Wangmin362: i have fixed it, thank you

Wangmin362:
> Could we handle `nvidia.com/gpumem-percentage: 0` the same way `nvidia.com/gpumem: 0` is already handled, entirely inside `Generate Resource Requests` — instead of rejecting it at admission? https://github.com/Project-HAMi/HAMi/blob/master/pkg/device/nvidia/device.go#L557-L577

done, handled entirely in GenerateResourceRequests now. thank you

archlitchi (approved):
/lgtm

PR #2159, opened 2026-07-28
fix(vGPUmonitor): scan all devices in CheckBlocking

### LINKED ISSUE #2158: vGPUmonitor CheckBlocking only inspects the first device, missing contention on multi-GPU containers
## What happened

`CheckBlocking` in `cmd/vGPUmonitor/feedback.go` is supposed to report whether any GPU a container uses is being contended by a higher-priority task. It loops over the container's devices, but returns `false` as soon as it finds the **first** device present in the utilization switch map:

```go
func CheckBlocking(utSwitchOn map[string]UtilizationPerDevice, p int, c *nvidia.ContainerUsage) bool {
	for i := range c.Info.DeviceMax() {
		uuid := c.Info.DeviceUUID(i)
		_, ok := utSwitchOn[uuid]
		if ok {
			for i := range min(p, len(utSwitchOn[uuid])) {
				if utSwitchOn[uuid][i] > 0 {
					return true
				}
			}
			return false   // <-- exits after checking only the first device
		}
	}
	return false
}
```

For a container that uses more than one GPU, that `return false` ends the whole function after the first matched device. If an earlier device has no contention but a later device does, `CheckBlocking` wrongly reports "not blocking".

The sibling function `CheckPriority`, just below it in the same file, has the identical structure **without** the early `return false` — it scans every device and only returns early on a positive match. `CheckBlocking` was clearly meant to do the same.

`CheckBlocking` should report blocking if **any** device the container uses is contended, regardless of device order.

## Impact

`Observe()` uses `CheckBlocking` to decide whether to pause (`SetRecentKernel(-1)`) a lower-priority container's kernel execution when a higher-priority job wants a GPU. With multi-GPU containers, this bug can let a low-priority container keep running unthrottled on a GPU that a higher-priority job is contending for — but only when a different device of that same container is scanned first and shows no contention. This weakens the priority-preemption / oversubscription mechanism.

## How to reproduce it

A container uses two GPUs, `gpu-0` and `gpu-1`, at priority `p = 1`. `gpu-0` has no contention, `gpu-1` has a higher-priority contender:

```
utSwitchOn = {"gpu-0": [0,0], "gpu-1": [1,0]}
CheckBlocking(utSwitchOn, 1, container) => false   // wrong, should be true
CheckPriority(utSwitchOn, 1, container) => true    // correct, on identical data
```

## Root cause

Stray early `return false` inside the per-device loop. Removing it lets the loop scan all of the container's devices, with the trailing `return false` handling the "no device contended" case — matching `CheckPriority`.

Fully unit-testable with no hardware (`nvidia.UsageInfo` is an interface; `cmd/vGPUmonitor/feedback_test.go` already mocks it). I have a one-line fix and a regression test ready and will open a PR referencing this issue.

---

imantaba:
## What this fixes

Fixes #2158.

`CheckBlocking` in `cmd/vGPUmonitor/feedback.go` returned `false` as soon as it found the first device the container uses in the utilization switch map, so for a container using more than one GPU it never inspected the remaining devices. If an earlier device had no contention but a later one did, `CheckBlocking` reported "not blocking" and `Observe()` let the lower-priority container keep running unthrottled on a GPU a higher-priority job was contending for, weakening priority preemption.

The sibling `CheckPriority`, just below it, has the identical loop structure **without** the early `return false` — it scans every device and only returns early on a positive match. `CheckBlocking` was meant to do the same.

## The change

Drop the stray early `return false` inside the per-device loop. The loop now scans all of the container's devices, and the trailing `return false` handles the "no device contended" case — mirroring `CheckPriority`.

## Testing

Added `TestCheckBlocking_MultiDeviceContention`: a two-device container where only the second device (`gpu-1`) has contention. On the previous code `CheckBlocking` returns `false`; after the fix it returns `true`, matching `CheckPriority` on the same input.

- `go test ./cmd/vGPUmonitor/ -race -count=1` — pass
- `make verify` — pass (license headers, import aliases, golangci-lint, goimports, staticcheck)

Unit-testable with no hardware — `nvidia.UsageInfo` is an interface and the existing `feedback_test.go` already mocks it.

---

## AI assistance disclosure

This contribution used AI assistance (Claude Code): the bug was surfaced during an AI-assisted code review, and the one-line fix plus the regression test were drafted with Claude Code under my review. I have read and verified the change, understand the root cause and the fix, and take full responsibility for its correctness and for responding to review feedback.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved GPU contention detection for workloads that use multiple GPU devices.
  * Blocking decisions now evaluate all associated devices before concluding no blocking is needed, ensuring priority results remain consistent.

* **Tests**
  * Expanded coverage with a new table-driven test for multi-device contention scenarios, including handling when utilization data is missing for some or all device UUIDs.

### DIFF
--- a/cmd/vGPUmonitor/feedback_test.go
+++ b/cmd/vGPUmonitor/feedback_test.go
@@ -86,21 +86,74 @@ func (s *multiStubInfo) SetRecentKernel(int32)              {}
 func (s *multiStubInfo) GetUtilizationSwitch() int32        { return 0 }
 func (s *multiStubInfo) SetUtilizationSwitch(int32)         {}
 
-// TestCheckBlocking_MultiDeviceContention verifies that CheckBlocking inspects
-// every device the container uses, not just the first one that appears in the
-// switch map. Here the first device (gpu-0) has no contention while a second
-// device (gpu-1) does; CheckBlocking must still report blocking.
-func TestCheckBlocking_MultiDeviceContention(t *testing.T) {
-	c := &nvidia.ContainerUsage{Info: &multiStubInfo{priority: 1, uuids: []string{"gpu-0", "gpu-1"}}}
-	sw := map[string]UtilizationPerDevice{
-		"gpu-0": {0, 0},
-		"gpu-1": {1, 0},
+// TestCheckBlocking_MultiDevice verifies that CheckBlocking inspects every device
+// the container uses, not just the first one that appears in the switch map. The
+// cases cover several device counts, all-clear, contention isolated to the last
+// device, and UUIDs missing from the switch map. In every case contention (when
+// present) sits at an index below the priority, so CheckPriority agrees and is
+// asserted for parity.
+func TestCheckBlocking_MultiDevice(t *testing.T) {
+	tests := []struct {
+		name     string
+		priority int
+		uuids    []string
+		sw       map[string]UtilizationPerDevice
+		want     bool
+	}{
+		{
+			name:     "two devices, contention on the second",
+			priority: 1,
+			uuids:    []string{"gpu-0", "gpu-1"},
+			sw:       map[string]UtilizationPerDevice{"gpu-0": {0, 0}, "gpu-1": {1, 0}},
+			want:     true,
+		},
+		{
+			name:     "three devices, all clear",
+			priority: 1,
+			uuids:    []string{"gpu-0", "gpu-1", "gpu-2"},
+			sw:       map[string]UtilizationPerDevice{"gpu-0": {0, 0}, "gpu-1": {0, 0}, "gpu-2": {0, 0}},
+			want:     false,
+		},
+		{
+			name:     "four devices, contention only on the last",
+			priority: 1,
+			uuids:    []string{"gpu-0", "gpu-1", "gpu-2", "gpu-3"},
+			sw:       map[string]UtilizationPerDevice{"gpu-0": {0, 0}, "gpu-1": {0, 0}, "gpu-2": {0, 0}, "gpu-3": {1, 0}},
+			want:     true,
+		},
+		{
+			name:     "some device UUIDs missing from switch map, present one contended",
+			priority: 1,
+			uuids:    []string{"gpu-0", "gpu-1", "gpu-2"},
+			sw:       map[string]UtilizationPerDevice{"gpu-1": {1, 0}},
+			want:     true,
+		},
+		{
+			name:     "none of the container UUIDs are in the switch map",
+			priority: 1,
+			uuids:    []string{"gpu-0", "gpu-1"},
+			sw:       map[string]UtilizationPerDevice{"gpu-9": {1, 0}},
+			want:     false,
+		},
+		{
+			name:     "empty switch map",
+			priority: 1,
+			uuids:    []string{"gpu-0", "gpu-1"},
+			sw:       map[string]UtilizationPerDevice{},
+			want:     false,
+		},
 	}
-	if !CheckBlocking(sw, 1, c) {
-		t.Error("CheckBlocking: expected true (gpu-1 has contention), got false")
-	}
-	// Sanity: the sibling CheckPriority already scans all devices and agrees.
-	if !CheckPriority(sw, 1, c) {
-		t.Error("CheckPriority: expected true (gpu-1 has contention), got false")
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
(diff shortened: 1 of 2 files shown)


cmd/vGPUmonitor/feedback_test.go
blackdragoon26, line 1: Hi, I was reviewing this and just had a suggestion, could we add more cases, i.e.,  table-driven tests for 3–4 devices as well? Maybe cover all-clear, contention only on the last device, and UUIDs missing from `utSwitchOn`,as in this we only cover one case with 2 GPUs.

imantaba: Good idea, thanks. I converted the multi-device test into a table-driven `TestCheckBlocking_MultiDevice` covering  hree- and four-device containers, an all-clear set, contention isolated to the last device, and container UUIDs missing from `utSwitchOn` (both a present-but-contended device and none-present). CheckPriority is asserted alongside CheckBlocking for parity on the same inputs.

cmd/vGPUmonitor/feedback_test.go
```diff
@@ -58,3 +58,102 @@ func TestCheckFunctionsHighPriority(t *testing.T) {
...
 }
+
+// multiStubInfo mocks a container that uses several devices, so the check
+// functions can be exercised over more than one UUID.
+type multiStubInfo struct {
+	priority int
+	uuids    []string
+}
+
+func (s *multiStubInfo) DeviceMax() int                     { return len(s.uuids) }
```
mesutoezdil: the real Spec returns a constant 16 here, not the device count, and unlike Observe just below neither check guards w/ IsValidUUID, so what does the loop actually do on a 2 gpu node now that the early return is gone?

cmd/vGPUmonitor/feedback_test.go
```diff
@@ -58,3 +58,102 @@ func TestCheckFunctionsHighPriority(t *testing.T) {
 		t.Error("CheckBlocking: expected false")
 	}
 }
+
+// multiStubInfo mocks a container that uses several devices, so the check
+// functions can be exercised over more than one UUID.
+type multiStubInfo struct {
```
mesutoezdil: this is a near copy of stubInfo above, can u just give stubInfo a uuids slice instead of a 2nd stub?

mesutoezdil (commented):
minor: the description doesnt follow the pr template, theres no "what type of pr is this?" block or `/kind` line, the label came from the labeler. ai disclosure is there, thanks for that.

cmd/vGPUmonitor/feedback_test.go
imantaba: You're right: `Spec.DeviceMax()` returns the constant `maxDevices = 16`, not the live count, so the loop always walks 16 fixed slots — on a 2-GPU node slots 0–1 hold real UUIDs and 2–15 are zeroed (`uuid[0] == 0`).
The checks stay correct without an `IsValidUUID` guard because `utSwitchOn` is only ever keyed by valid UUIDs - `Observe` guards with `IsValidUUID` before it inserts/increments, so a zeroed slot's UUID can't be a key. `CheckBlocking/CheckPriority` only read the map, so those trailing slots drop out on the `_, ok := utSwitchOn[uuid]` miss; the map lookup filters them the same way `Observe` does explicitly. Removing the early `return false` just lets the loop reach the later real devices instead of bailing after the first match.
You're right that the stub hides this by returning `len(uuids)` for `DeviceMax`. I'll rework it to mirror production - a fixed `DeviceMax` with `DeviceUUID/IsValidUUID` valid only for the populated prefix - so the test actually exercises the 16-slot walk.

cmd/vGPUmonitor/feedback_test.go
imantaba: Agreed

cmd/vGPUmonitor/feedback_test.go
mesutoezdil: Reminder: Answers must be written by **human being**. You can view the relevant rule here.
"4. Review replies. The reply you post must be written by you and must address the specific point raised. Verbatim or canned AI replies, or replies that do not engage the comment, lead to the **PR being closed**."

imantaba: Fair. But i do understand the change `DeviceMax` id fixed at 16, the empty slot never get into`utSwitchOn` so they just miss the lookup and the stray `return false` was stopping before the second real device.

archlitchi (approved):
/lgtm

PR #2160, opened 2026-07-28
fix(nvidia): handle zero-score topology combinations

Eshiv-Pandey:
**What type of PR is this?**

/kind bug

`computeBestCombination()` initializes `bestScore` to `0` and only records a device combination when `totalScore > bestScore`.

When topology-aware scheduling evaluates valid NVIDIA GPU combinations and every pair score is `0`, no combination is selected. The scheduler can return an empty allocation even though the request can be satisfied.

This PR initializes `bestScore` to `-1` so a valid zero-score combination can still be selected.

A regression test covers a 2-GPU topology-aware request where all available GPU pair scores are `0`.

None filed.

Before fix:

```text
=== RUN   TestFit_TopologyBestCombinationZeroScores
device_test.go:2817: assertion failed: 0 (int) != 2 (int)
--- FAIL: TestFit_TopologyBestCombinationZeroScores
```

After fix:

```text
=== RUN   TestFit_TopologyBestCombinationZeroScores
--- PASS: TestFit_TopologyBestCombinationZeroScores
PASS
ok   github.com/Project-HAMi/HAMi/pkg/device/nvidia
```

Broader validation:

```text
ok   github.com/Project-HAMi/HAMi/pkg/device/nvidia
ok   github.com/Project-HAMi/HAMi/pkg/scheduler
ok   github.com/Project-HAMi/HAMi/pkg/scheduler/config
ok   github.com/Project-HAMi/HAMi/pkg/scheduler/policy
ok   github.com/Project-HAMi/HAMi/pkg/scheduler/routes
```

`make verify` was attempted in a Linux Go 1.26.5 Docker environment, but Docker Desktop returned `unexpected EOF` before the command completed.

No breaking change. Topology-aware NVIDIA scheduling now selects a valid GPU combination when all valid pair scores are zero.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved NVIDIA topology-based scheduling to handle edge scoring cases more reliably, including when pair scores are zero or negative.
  * Corrected device combination selection for both single-GPU and multi-GPU requests so allocations match expected “best/worst” outcomes.
* **Tests**
  * Added unit coverage verifying two-GPU requests succeed when all topology pair scores are zero.
  * Added unit coverage verifying single- and two-GPU requests succeed with negative pair scores, with expected device UUIDs selected.

### DIFF
--- a/pkg/device/nvidia/device.go
+++ b/pkg/device/nvidia/device.go
@@ -937,8 +937,9 @@ func getDevicePairScoreMap(nodeInfo *device.NodeInfo) map[string]*device.DeviceP
 }
 
 func computeWorstSingleCard(nodeInfo *device.NodeInfo, request device.ContainerDeviceRequest, tmpDevs map[string]device.ContainerDevices) device.ContainerDevices {
-	worstScore := -1
+	worstScore := 0
 	worstDevices := device.ContainerDevices{}
+	found := false
 	deviceScoreMap := getDevicePairScoreMap(nodeInfo)
 	// Iterate through all devices to find the one with the lowest score
 	devices := tmpDevs[request.Type]
@@ -952,7 +953,8 @@ func computeWorstSingleCard(nodeInfo *device.NodeInfo, request device.ContainerD
 			}
 			totalScore += scoreMapDev1.Scores[dev2.UUID]
 		}
-		if totalScore < worstScore || worstScore == -1 {
+		if !found || totalScore < worstScore {
+			found = true
 			worstScore = totalScore
 			worstDevices = device.ContainerDevices{dev1}
 		}
@@ -961,8 +963,9 @@ func computeWorstSingleCard(nodeInfo *device.NodeInfo, request device.ContainerD
 }
 
 func computeBestCombination(nodeInfo *device.NodeInfo, combinations []device.ContainerDevices) device.ContainerDevices {
-	bestScore := 0
+	bestScore := -1
 	bestCombination := device.ContainerDevices{}
+	found := false
 	deviceScoreMap := getDevicePairScoreMap(nodeInfo)
 	// Iterate through all combinations to find the one with the highest score
 	for _, partition := range combinations {
@@ -977,7 +980,8 @@ func computeBestCombination(nodeInfo *device.NodeInfo, combinations []device.Con
 			}
 		}
 
-		if totalScore > bestScore {
+		if !found || totalScore > bestScore {
+			found = true
 			bestScore = totalScore
 			bestCombination = partition
 		}
--- a/pkg/device/nvidia/device_test.go
+++ b/pkg/device/nvidia/device_test.go
@@ -2783,3 +2783,86 @@ func TestFit_TopologyBestCombination(t *testing.T) {
 	assert.Assert(t, uuids["dev-0"])
 	assert.Assert(t, uuids["dev-2"])
 }
+
+func TestFit_TopologyBestCombinationZeroScores(t *testing.T) {
+	config := NvidiaConfig{
+		ResourceCountName:            "nvidia.com/gpu",
+		ResourceMemoryName:           "nvidia.com/gpumem",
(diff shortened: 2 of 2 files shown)


pkg/device/nvidia/device.go
```diff
@@ -961,7 +961,7 @@ func computeWorstSingleCard(nodeInfo *device.NodeInfo, request device.ContainerD
 }
 
 func computeBestCombination(nodeInfo *device.NodeInfo, combinations []device.ContainerDevices) device.ContainerDevices {
-	bestScore := 0
+	bestScore := -1
```
mesutoezdil, line 966: computeWorstSingleCard right above uses -1 as a real sentinel w/ `|| worstScore == -1`, here its just a magic init, so a `found bool` or seeding w/ combinations[0] would close this for good instead of moving the floor down by one.

pkg/device/nvidia/device_test.go
```diff
@@ -2783,3 +2783,36 @@ func TestFit_TopologyBestCombination(t *testing.T) {
...
+	nv := InitNvidiaDevice(config)
+	devices := []*device.DeviceUsage{
+		{ID: "dev-0", Index: 0, Used: 0, Count: 10, Totalmem: 8192, Totalcore: 100, Type: NvidiaGPUDevice, Health: true},
+		{ID: "dev-1", Index: 1, Used: 0, Count: 10, Totalmem: 8192, Totalcore: 100, Type: NvidiaGPUDevice, Health: true},
+		{ID: "dev-2", Index: 2, Used: 0, Count: 10, Totalmem: 8192, Totalcore: 100, Type: NvidiaGPUDevice, Health: true},
+	}
+	nodeInfo := &device.NodeInfo{
+		Devices: map[string][]device.DeviceInfo{
+			NvidiaGPUDevice: {
+				{ID: "dev-0", DevicePairScore: device.DevicePairScore{Scores: map[string]int{"dev-1": 0, "dev-2": 0}}},
```
mesutoezdil, line 2803: these scores come straight out of DecodePairScores on a node annotation, so what stops one from being negative, and would this test still pass if u set them to -1?

mesutoezdil (commented):
heads up that ci hasnt run any of it, the unit test and compile jobs are still awaiting maintainer approval so the checks list only has the bots.
also fyi #2090 is open and reworks Fit in this same file, worth keeping an eye on the order.

pkg/device/nvidia/device.go
Eshiv-Pandey: @mesutoezdil Ohh nice suggestion! I've updated both `computeBestCombination` and `computeWorstSingleCard` to use the explicit `found bool` pattern instead of relying on magic number sentinels. Jotting down the changes real quick --->

-> `computeWorstSingleCard`: Changed from `worstScore := -1` with `|| worstScore == -1` check to `found bool` pattern
->`computeBestCombination`: Now uses `found bool` instead of just changing the initialization to `-1`

Both functions now follow the same idiomatic pattern: `if !found || condition { found = true; ... }`
have also added `TestFit_TopologyNegativeScores` covering both paths. 
Pls have a look if i did that correctly :)

pkg/device/nvidia/device_test.go
Eshiv-Pandey: @mesutoezdil Yeah right ... scores can be negative from `DecodePairScores`. 

As i said above have added `TestFit_TopologyNegativeScores` to validate exactly that scenario with negative pair scores (-5, -3, -1)
- **SingleGPU subtest**: Verifies `computeWorstSingleCard` correctly selects a device when all scores are negative
- **MultiGPU subtest**: Verifies `computeBestCombination` picks the combination with the highest (least negative) total score

and both subtests pass, confirming the same. The scheduler now selects the "best worst" option even when everything has negative affinity, and yes keeping an eye on the PR you mentioned sir!

archlitchi (approved):
/lgtm

PR #2166, opened 2026-07-28
perf: remove unused all-device scan in metrics collector

### LINKED ISSUE #2165: perf: Remove unused all-device lookup loop in scheduler metrics collector
### Description
During Prometheus metrics collection in `cmd/scheduler/metrics.go`, an nested loop iterates over all cluster nodes and node devices to compute `totaldev` (device total memory) for every scheduled container device. 
The computed `totaldev` value is logged at verbosity 4 (`klog.V(4)`), but is never used in any emitted Prometheus metric or downstream calculation.
### Impact
For clusters with $A$ allocated container devices and $D$ total registered devices, this creates an $O(A \times D)$ redundant scan on every Prometheus scrape, incurring unnecessary CPU and latency overhead.
### Code Reference
- [`cmd/scheduler/metrics.go:358-376`](https://github.com/Project-HAMi/HAMi/blob/master/cmd/scheduler/metrics.go#L358-L376):
```go
var totaldev int32
found := false
for _, ni := range *nu {
    for _, nodedev := range ni.Devices.DeviceLists {
        if strings.Compare(nodedev.Device.ID, ctrdevval.UUID) == 0 {
            totaldev = nodedev.Device.Totalmem
            found = true
            break
        }
    }
    if found {
        break
    }
}

### PR DESCRIPTION
**Pull request by Gaurav-205**
### Summary
Fixes #2165.

During Prometheus metrics collection in `cmd/scheduler/metrics.go`, an nested loop was iterating over all cluster nodes and node devices to calculate `totaldev` for every scheduled container device. This computed `totaldev` value was only logged at `klog.V(4)` and never used in any emitted metric or calculation.

This PR removes the redundant $O(A \times D)$ device scan loop and unused `strings` package import, eliminating unnecessary CPU and memory allocation overhead during Prometheus scrapes.

### Changes
- Removed unneeded nested node/device search loop in `cmd/scheduler/metrics.go`.
- Removed unused `strings` import.

## Summary

- **Refactor**
  - Streamlined scheduler metrics collection by removing legacy device memory lookup and its associated logging.
  - Improved scheduled pod snapshot isolation: returned pod/device data is now deep-copied so callers can’t mutate cached state, with custom metadata intentionally omitted from the returned copies.

- **Tests**
  - Updated pod and deep-copy tests to match the new snapshot and metadata omission behavior.

- **Chores**
  - Removed unused imports and simplified internal collection/copying logic.

### DIFF
--- a/cmd/scheduler/metrics.go
+++ b/cmd/scheduler/metrics.go
@@ -20,7 +20,6 @@ import (
 	"fmt"
 	"log"
 	"net/http"
-	"strings"
 	"time"
 
 	"github.com/prometheus/client_golang/prometheus"
@@ -355,25 +354,6 @@ func (cc ClusterManagerCollector) Collect(ch chan<- prometheus.Metric) {
 							float64(ctrdevval.Usedcores),
 							val.Namespace, val.NodeID, val.Name, fmt.Sprint(ctridx), ctrdevval.UUID)
 					}
-					var totaldev int32
-					found := false
-					for _, ni := range *nu {
-						for _, nodedev := range ni.Devices.DeviceLists {
-							if strings.Compare(nodedev.Device.ID, ctrdevval.UUID) == 0 {
-								totaldev = nodedev.Device.Totalmem
-								found = true
-								break
-							}
-						}
-						if found {
-							break
-						}
-					}
-					klog.V(4).InfoS("Total memory for device",
-						"deviceUUID", ctrdevval.UUID,
-						"totalMemory", totaldev,
-						"nodeID", val.NodeID,
-					)
 				}
 			}
 		}

### CODE REVIEW

**Comment by Gaurav-205** [comment]
@wawa0210 @lengrongfu hey, can you please review this pr!!

File: cmd/scheduler/metrics.go [hunk 1]
```diff
@@ -355,25 +354,6 @@ func (cc ClusterManagerCollector) Collect(ch chan<- prometheus.Metric) {
...
-								break
-							}
-						}
-						if found {
-							break
-						}
-					}
-					klog.V(4).InfoS("Total memory for device",
-						"deviceUUID", ctrdevval.UUID,
-						"totalMemory", totaldev,
```
Eshiv-Pandey, line 374: **Great find!** This looks reasonable tbh. The removed scan only feeds a verbose log line and does not affect any emitted metric. One question tho... was this `totalMemory` log used for any debugging flow, or is the node memory limit metric enough for that case?

Gaurav-205: Thanks! I checked the full metrics collector flow. totaldev was only used by that klog.V(4) message; it does not affect emitted metrics, scheduling, or allocation.
For debugging device capacity, the collector already exposes the same per-device total-memory information through hami_gpu_memory_limit_bytes and hami_node_gpu_overview (device_memory_limit). So the verbose log is redundant, and removing the scan does not reduce Prometheus visibility.
If maintainers prefer retaining a V(4) log for troubleshooting, I can add one using the already available per-device data, without restoring the all-node scan.

mesutoezdil (commented):
also the description doesnt follow the pr template, no "what type of pr is this?" block or `/kind` line, and no ai disclosure. if any ai tool was used it has to be disclosed, see https://github.com/Project-HAMi/HAMi/blob/master/CONTRIBUTING.md#ai-assistance-notice

fyi #2170 is open on this same file, worth watching the order.

cmd/scheduler/metrics.go
```diff
@@ -355,25 +354,6 @@ func (cc ClusterManagerCollector) Collect(ch chan<- prometheus.Metric) {
...
-						"nodeID", val.NodeID,
-					)
 				}
 			}
 		}
```
mesutoezdil, line 359: the loop u deleted broke on the first uuid hit while ranging `*nu`, which is a map, and the pods node was already sitting in `val.NodeID`, so what did that V(4) line print if the same uuid ever turned up under 2 nodes?

PR #2170, opened 2026-07-28
fix(metrics): update scheduler metric descriptors label key from node to node_name

### LINKED ISSUE #2161: BUG: scheduler metrics use label node instead of node_name, breaking cross-component PromQL joins
**What happened**:
In `cmd/vGPUmonitor/metrics.go`, host and container GPU metrics use the standard label key `node_name`. However, in `cmd/scheduler/metrics.go`, all scheduler GPU metrics use `node` as the label key name.
Because the label key names differ between the `vGPUmonitor` exporter and the `scheduler` exporter, operators cannot perform PromQL joins across components in Prometheus or Grafana.

For example, the following query silently returns no results:
```
hami_gpu_memory_allocated_bytes * on(node_name) hami_host_gpu_utilization_ratio
```

All HAMi telemetry exporters (both `vGPUmonitor` and `scheduler`) should use consistent label names (`node_name`) so cross component Prometheus queries and Grafana dashboards work out of the box.

1. Deploy HAMi with metrics enabled.
2. Scrape metrics from both `vGPUmonitor` (port 9400) and `scheduler` (port 9401) in Prometheus.
3. Observe that `vGPUmonitor` metrics export label `node_name="<node>"`, while scheduler metrics export label `node="<node>"`
4. Attempt a Prometheus metric join on `node_name`, which fails to match.

Fixing this requires updating the metric descriptors in `cmd/scheduler/metrics.go` to rename "node" to "node_name", aligning it with vGPUmonitor (Issue #2152 )

**Environment**:
- HAMi version: latest
- Others:

swastikCommits:
**What type of PR is this?**
/kind bug

In `cmd/vGPUmonitor/metrics.go`, host and container GPU metrics export standard Prometheus labels using key `node_name`. However, `cmd/scheduler/metrics.go` exported its metric descriptors with label key `"node"`. This label inconsistency prevents the operators from performing PromQL metric joins across components in Prometheus or Grafana (e.g., joining `vGPUmonitor` metrics with `scheduler` metrics on node name).

In this PR we update `"node"` to `"node_name"` across all standard scheduler metric descriptors to align them with `vGPUmonitor` and added unit test coverage. We also added a nil-guard for `sher` inside `ClusterManagerCollector.Collect()` to safely allow unit tests and uninitialized collector calls to execute without panicking.

Fixes #2161 

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated GPU and container allocation metrics to consistently identify nodes by name.
  * Prevented invalid memory-ratio metrics from being emitted when total memory is zero or negative.
  * Preserved core-limit metrics in these conditions.

* **Tests**
  * Added coverage for standard and legacy metric descriptors.
  * Added validation that expected scheduler metrics are emitted correctly.

### DIFF
--- a/cmd/scheduler/metrics.go
+++ b/cmd/scheduler/metrics.go
@@ -77,42 +77,46 @@ func (cc ClusterManagerCollector) Collect(ch chan<- prometheus.Metric) {
 	nodevGPUMemoryLimitDesc := prometheus.NewDesc(
 		"hami_gpu_memory_limit_bytes",
 		"Device memory limit for a certain GPU",
-		[]string{"node", "device_uuid", "device_index", "device_type"}, nil,
+		[]string{"node_name", "device_uuid", "device_index", "device_type"}, nil,
 	)
 	nodevGPUCoreLimitDesc := prometheus.NewDesc(
 		"hami_gpu_core_limit_ratio",
 		"Device core limit for a certain GPU",
-		[]string{"node", "device_uuid", "device_index", "device_type"}, nil,
+		[]string{"node_name", "device_uuid", "device_index", "device_type"}, nil,
 	)
 	nodevGPUMemoryAllocatedDesc := prometheus.NewDesc(
 		"hami_gpu_memory_allocated_bytes",
 		"Device memory allocated for a certain GPU",
-		[]string{"node", "device_uuid", "device_index", "device_cores", "device_type"}, nil,
+		[]string{"node_name", "device_uuid", "device_index", "device_cores", "device_type"}, nil,
 	)
 	nodevGPUSharedNumDesc := prometheus.NewDesc(
 		"hami_gpu_shared_count",
 		"Number of containers sharing this GPU",
-		[]string{"node", "device_uuid", "device_index", "device_type"}, nil,
+		[]string{"node_name", "device_uuid", "device_index", "device_type"}, nil,
 	)
 	nodeGPUCoreAllocatedDesc := prometheus.NewDesc(
 		"hami_gpu_core_allocated_ratio",
 		"Device core allocated for a certain GPU",
-		[]string{"node", "device_uuid", "device_index", "device_type"}, nil,
+		[]string{"node_name", "device_uuid", "device_index", "device_type"}, nil,
 	)
 	nodeGPUOverview := prometheus.NewDesc(
 		"hami_node_gpu_overview",
 		"GPU overview on a certain node",
+<<<<<<< HEAD
 		[]string{"node", "device_uuid", "device_index", "device_cores", "device_memory_limit", "device_type"}, nil,
+=======
+		[]string{"node_name", "device_uuid", "device_index", "device_cores", "shared_containers", "device_memory_limit", "device_type"}, nil,
+>>>>>>> 793c056 (fix(metrics): update scheduler metric descriptors label key from node to node_name)
 	)
 	nodeGPUMemoryPercentage := prometheus.NewDesc(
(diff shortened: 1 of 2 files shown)


swastikCommits:
Hi @mesutoezdil @archlitchi @wawa0210 @ouyangluwei163 can you please review this PR whenever you have a moment
Thanks.

cmd/scheduler/metrics_test.go
```diff
@@ -0,0 +1,83 @@
...
+		descStr := desc.String()
+		// Ensure standard GPU descriptors (excluding namespace-scoped quota metrics) contain node_name
+		if strings.Contains(descStr, "fqName: \"hami_") && !strings.Contains(descStr, "hami_resource_quota_used") {
+			if !strings.Contains(descStr, "node_name") {
+				t.Errorf("standard descriptor %s does not contain node_name label", descStr)
+			}
+		}
+	}
+
+	if foundDescriptors == 0 {
```
mesutoezdil: both tests fail here w/ `expected at least 1 descriptor`, so the node_name loop above never even runs.

cmd/scheduler/metrics.go
```diff
@@ -165,6 +165,10 @@ func (cc ClusterManagerCollector) Collect(ch chan<- prometheus.Metric) {
 		)
 	}
 
+	if sher == nil {
```
mesutoezdil: thats why: `sher` is nil in tests, nothing hits ch before this, and Describe is DescribeByCollect, so 0 descs come out. pls inject the scheduler into ClusterManagerCollector instead of an early return.

cmd/scheduler/metrics_test.go
```diff
@@ -0,0 +1,83 @@
...
+	collector.Describe(ch)
+	close(ch)
+
+	foundDescriptors := 0
+	for desc := range ch {
+		foundDescriptors++
+		descStr := desc.String()
+		// Ensure standard GPU descriptors (excluding namespace-scoped quota metrics) contain node_name
+		if strings.Contains(descStr, "fqName: \"hami_") && !strings.Contains(descStr, "hami_resource_quota_used") {
+			if !strings.Contains(descStr, "node_name") {
```
mesutoezdil: this only checks node_name is there, not that `node` is gone, so a half done rename still passes.

cmd/scheduler/metrics.go
```diff
@@ -55,42 +55,42 @@ func (cc ClusterManagerCollector) Collect(ch chan<- prometheus.Metric) {
 	nodevGPUMemoryLimitDesc := prometheus.NewDesc(
 		"hami_gpu_memory_limit_bytes",
 		"Device memory limit for a certain GPU",
-		[]string{"node", "device_uuid", "device_index", "device_type"}, nil,
+		[]string{"node_name", "device_uuid", "device_index", "device_type"}, nil,
```
mesutoezdil, line 80: on master vGPUmonitor sets no node label at all, that comes from your other pr #2153. so what does the join in #2161 return if this one lands alone?

mesutoezdil (commented):
ran the 2 new tests locally, both fail, notes inline

this also renames a public prom label, so it breaks any dashboard or alert keyed on `node`. pls add the "does this pr introduce a user-facing change?" block w/ a release-note for the rename.

no ai disclosure either, see https://github.com/Project-HAMi/HAMi/blob/master/CONTRIBUTING.md#ai-assistance-notice

swastikCommits:
Closing this PR as the branch history got cluttered during intermediate merge updates.

PR #2171, opened 2026-07-28
Fix dockerfile git worktree

### LINKED ISSUE #2041: [Bug] docker build fails from linked git worktree because Dockerfiles assume `.git/modules/libvgpu` exists

shellyco-code:
**What type of PR is this?**
/kind bug

This PR resolves an issue where the Docker builds (`docker/Dockerfile` and `docker/Dockerfile.hamicore`) would fail early when building from a linked Git worktree. The build context assumed that `.git/modules/libvgpu` would always exist as a directory, which is false in worktree environments where `.git` is a file.

By removing the hardcoded `COPY .git/modules/libvgpu` steps, we make the Dockerfiles portable across all supported Git working tree layouts. The internal `libvgpu/build.sh` script already handles missing git metadata gracefully by defaulting `CI_COMMIT_SHA` to `"unknown"`.

Fixes #2041

This ensures compatibility with developers who use `git worktree` for maintaining multiple branches simultaneously without needing multiple full clones.

```release-note
NONE


## Summary by CodeRabbit

* **Bug Fixes**
  * Docker build commands now work reliably when run from Git worktrees.
  * Repository metadata is restored automatically after successful or failed builds, preventing changes to the local checkout.
  * Improved handling of the `libvgpu` submodule during container builds.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,6 +21,13 @@ GO_BUILD_LDFLAGS=\
 all: build
 
 docker:
+	@if [ -f .git ]; then \
+		echo "Detected git worktree, preparing .git/modules/libvgpu"; \
+		REAL_GITDIR=$$(git rev-parse --git-common-dir); \
+		mv .git .git.bak; \
+		mkdir -p .git/modules; \
+		cp -r $$REAL_GITDIR/modules/libvgpu .git/modules/libvgpu; \
+	fi
 	docker build \
 	--platform ${TARGET_PLATFORMS} \
 	--build-arg GOLANG_IMAGE=${GOLANG_IMAGE} \
@@ -28,9 +35,17 @@ docker:
 	--build-arg DEST_DIR=${DEST_DIR} \
 	--build-arg VERSION=${VERSION} \
 	--build-arg GOPROXY=https://goproxy.cn,direct \
-	. -f=docker/Dockerfile -t ${IMG_NAME}:${IMG_TAG}
+	. -f=docker/Dockerfile -t ${IMG_NAME}:${IMG_TAG} || (if [ -f .git.bak ]; then rm -rf .git; mv .git.bak .git; fi; exit 1)
+	@if [ -f .git.bak ]; then rm -rf .git; mv .git.bak .git; fi
 
 dockerwithlib:
+	@if [ -f .git ]; then \
+		echo "Detected git worktree, preparing .git/modules/libvgpu"; \
+		REAL_GITDIR=$$(git rev-parse --git-common-dir); \
+		mv .git .git.bak; \
+		mkdir -p .git/modules; \
+		cp -r $$REAL_GITDIR/modules/libvgpu .git/modules/libvgpu; \
+	fi
 	docker build \
 	--no-cache \
 	--platform ${TARGET_PLATFORMS} \
@@ -39,7 +54,8 @@ dockerwithlib:
 	--build-arg DEST_DIR=${DEST_DIR} \
 	--build-arg VERSION=${VERSION} \
 	--build-arg GOPROXY=https://goproxy.cn,direct \
-	. -f=docker/Dockerfile.withlib -t ${IMG_NAME}:${IMG_TAG}
+	. -f=docker/Dockerfile.withlib -t ${IMG_NAME}:${IMG_TAG} || (if [ -f .git.bak ]; then rm -rf .git; mv .git.bak .git; fi; exit 1)
+	@if [ -f .git.bak ]; then rm -rf .git; mv .git.bak .git; fi
 
 tidy:
 	$(GO) mod tidy
--- a/hack/build.sh
+++ b/hack/build.sh
@@ -34,9 +34,24 @@ function go_build() {
 }
 
 function docker_build() {
+    local git_restored=false
+    if [ -f .git ]; then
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Review by mesutoezdil — COMMENTED** [review]
this branch carries #2151 too, c4847d0 and 3cd40b7 are both in here and only aec9056 is the docker fix. worth rebasing onto master so this pr is one commit.

related: 3cd40b7 is titled "address PR feedback for health check config", which is the commit u said u pushed to #2151 on the 28th. it landed here instead, #2151 is still on c4847d0 and still fails lint on the unused os import.

File: docker/Dockerfile [hunk 1]
```diff
@@ -13,8 +13,6 @@ RUN go install github.com/NVIDIA/mig-parted/cmd/nvidia-mig-parted@v0.12.3
 FROM $NVIDIA_IMAGE AS nvbuild
 RUN dnf install -y cmake git && dnf clean all
 COPY ./libvgpu /libvgpu
-COPY .git/modules/libvgpu /libvgpu-git
-RUN rm -rf /libvgpu/.git && echo "gitdir: /libvgpu-git" > /libvgpu/.git
 WORKDIR /libvgpu
 RUN rm -rf /libvgpu/build
 RUN bash ./build.sh
```
mesutoezdil, line 20: dropping the copy means git describe fails in every build, not just worktrees, so CI_COMMIT_SHA falls back to unknown for released images too. #2042 tried to keep it by resolving the real gitdir, what does libvgpu report as its version after this?

pkg/device-plugin/nvidiadevice/nvinternal/plugin/factory.go
```diff
@@ -128,9 +132,9 @@ func (o *options) getResourceManagers() ([]rm.ResourceManager, error) {
 			_ = o.nvmllib.Shutdown()
 		}()
 
-		return rm.NewNVMLResourceManagers(o.infolib, o.nvmllib, o.devicelib, o.config.Config)
+		return rm.NewNVMLResourceManagers(o.infolib, o.nvmllib, o.devicelib, o.config.Config, disableHealthChecks, enableHealthChecks)
```
DSFans2014: Please remove changes that are unrelated to the pr title

shellyco-code:
@mesutoezdil ,@DSFans2014  PTAL

FouoF:
Thanks for working on Git worktree builds. The current approach temporarily renames the checkout's .git entry to .git.bak. That can leave the repository metadata displaced when the copy fails or the process is interrupted, and it is unsafe when builds run concurrently. The existing major review concern is therefore a correctness issue, not a minor cleanup request. We are closing this implementation. A replacement must leave the working tree metadata untouched and include a real worktree-build regression test.

PR #2172, opened 2026-07-28
enforce resource quota for all device backends

### LINKED ISSUE #2157: [Bug] ResourceQuota enforcement only applies to NVIDIA devices, not other accelerator backends
## What happened

## How to reproduce

   - Used: 9000
   - Limit: 10000

## Verified behavior

```go
```

```go
```

## Root cause

```go
// Only supports NVIDIA
    continue
}
```

## Impact

## Proposed fix

## Environment

- HAMi master
- Verified with unit tests

Jay2006sawant:
**What type of PR is this?**

/kind bug

HAMi's `QuotaManager` tracks namespace memory and core usage in a device agnostic way, but quota enforcement was wired only for NVIDIA. The admission webhook skipped all non-NVIDIA devices, and only NVIDIA's `Fit()` path called `FitQuota()` during scheduling.

This PR adds shared quota helpers in `pkg/device/quota.go` and enforces namespace ResourceQuota for all registered device backends at admission time. It also adds pre-allocation quota checks in the scheduler `Fit()` path for Cambricon, Ascend, and Hygon, matching the existing NVIDIA behavior.

Fixes #2157

- Shared helpers: `PodQuotaRequests`, `FitPodQuota`, `FitAllocationQuota`
- Webhook `fitResourceQuota()` now iterates all devices in `DevicesMap`
- NVIDIA `fitQuota()` refactored to use `FitAllocationQuota`
- Added quota checks in `Fit()` for Cambricon, Ascend, and Hygon
- New tests cover non-NVIDIA webhook denial and quota request calculation
- `make test` and `make verify` pass locally
- No GPU hardware required

Yes. In multi-vendor clusters, pods requesting Cambricon, Ascend, Hygon, or other registered backends can now be denied at admission or scheduler Filter time when namespace ResourceQuota limits for vendor memory or core resources would be exceeded. Previously, only NVIDIA quota enforcement was active.

**AI assistance disclosure:**
I used AI assistance to understand the codebase and explore the quota enforcement flow. All code changes, tests, and verification were written and run by me. I reviewed and understand the full diff.

## Summary by CodeRabbit

* **New Features**
  * Enhanced namespace quota validation to account for pod memory and compute requests across supported device types.
  * Applied device-specific memory scaling when calculating quota consumption.
* **Bug Fixes**
  * Prevented allocations that exceed namespace quotas, including pending and already allocated resources.
  * Extended quota enforcement to non-NVIDIA device types.
* **Tests**
  * Added coverage for quota calculations, memory scaling, allocation rejection, and scheduler admission enforcement.

### DIFF
--- a/pkg/device/hygon/device.go
+++ b/pkg/device/hygon/device.go
@@ -84,6 +84,10 @@ func (dev *DCUDevices) CommonWord() string {
 	return HygonDCUCommonWord
 }
 
+func (dev *DCUDevices) MemoryFactor() int32 {
+	return MemoryFactor
+}
+
 func ParseConfig(fs *flag.FlagSet) {
 	fs.StringVar(&HygonResourceCount, "dcu-name", "hygon.com/dcunum", "dcu resource count")
 	fs.StringVar(&HygonResourceMemory, "dcu-memory", "hygon.com/dcumem", "dcu memory resource")
@@ -300,6 +304,11 @@ func (dcu *DCUDevices) Fit(devices []*device.DeviceUsage, request device.Contain
 		if k.MemPercentagereq != 101 && k.Memreq == 0 {
 			//This incurs an issue
 			memreq = dev.Totalmem * k.MemPercentagereq / 100
+		}
+		if !device.FitAllocationQuota(pod.Namespace, k.Type, MemoryFactor, int64(memreq), int64(k.Coresreq), tmpDevs, allocated) {
+			reason[common.ResourceQuotaNotFit]++
+			klog.V(3).InfoS(common.ResourceQuotaNotFit, "pod", pod.Name, "memreq", memreq, "coresreq", k.Coresreq)
+			continue
 		}
 		if dev.Totalmem-dev.Usedmem < memreq {
 			reason[common.CardInsufficientMemory]++
--- a/pkg/scheduler/webhook.go
+++ b/pkg/scheduler/webhook.go
@@ -30,7 +30,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 
 	"github.com/Project-HAMi/HAMi/pkg/device"
-	"github.com/Project-HAMi/HAMi/pkg/device/nvidia"
 	"github.com/Project-HAMi/HAMi/pkg/scheduler/config"
 )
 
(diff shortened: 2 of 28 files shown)


pkg/device/hygon/device.go
```diff
@@ -301,6 +301,11 @@ func (dcu *DCUDevices) Fit(devices []*device.DeviceUsage, request device.Contain
 			//This incurs an issue
 			memreq = dev.Totalmem * k.MemPercentagereq / 100
 		}
+		if !device.FitAllocationQuota(pod.Namespace, k.Type, 1, int64(memreq), int64(k.Coresreq), tmpDevs, allocated) {
```
mesutoezdil: hygon already scales memnums by its own MemoryFactor in GenerateResourceRequests, so the hardcoded 1 here leaves the quota limit unscaled. what happens on a cluster w/ memoryFactor 10? ascend has the same issue.

pkg/scheduler/webhook.go
```diff
@@ -129,47 +129,12 @@ func isPrivilegedContainer(ctr *corev1.Container) bool {
 }
 
 func fitResourceQuota(pod *corev1.Pod) bool {
-	for deviceName, dev := range device.GetDevices() {
-		// Only supports NVIDIA
-		if deviceName != nvidia.NvidiaGPUDevice {
-			continue
+	for deviceName := range device.GetDevices() {
+		memoryFactor := int32(1)
+		if deviceName == nvidia.NvidiaGPUDevice {
```
mesutoezdil: this swaps one hardcoded nvidia branch for another, a MemoryFactor() on the Devices interface would cover every backend.

pkg/scheduler/webhook_test.go
```diff
@@ -425,6 +425,78 @@ func TestFitResourceQuota(t *testing.T) {
...
+		deviceName: &quotaWebhookDevice{
+			word: deviceName,
+			names: device.ResourceNames{
+				ResourceCountName:  "huawei.com/Ascend910B",
+				ResourceMemoryName: memName,
+				ResourceCoreName:   "huawei.com/Ascend910B-core",
+			},
+		},
+	}
+	qm := device.NewQuotaManager()
```
mesutoezdil, line 480: NewQuotaManager is a sync.Once singleton, so this writes to the global w/ no cleanup, unlike DevicesMap above.

mesutoezdil (commented):
no ai disclosure, see https://github.com/Project-HAMi/HAMi/blob/master/CONTRIBUTING.md#ai-assistance-notice. fyi #2090 touches nvidia/device.go too.

DSFans2014 (commented):
> This PR adds shared quota helpers in pkg/device/quota.go and enforces namespace ResourceQuota for all registered device backends at admission time. It also adds pre-allocation quota checks in the scheduler Fit() path for Cambricon, Ascend, and Hygon, matching the existing NVIDIA behavior.

Why check all devices at admission but only Cambricon, Ascend, Hygon and NVIDIA in fit?

Jay2006sawant:
@DSFans2014 Good catch. Admission now enforces quota for all registered devices (the #2157 bug). Fit() quota was NVIDIA-only before; I added it for Cambricon/Ascend/Hygon since they share the same vGPU allocation path. Other backends are covered at admission but not yet in Fit().happy to extend that in a follow-up if you'd like.

mesutoezdil:
> @DSFans2014 Good catch. Admission now enforces quota for all registered devices (the #2157 bug). Fit() quota was NVIDIA-only before; I added it for Cambricon/Ascend/Hygon since they share the same vGPU allocation path. Other backends are covered at admission but not yet in Fit().happy to extend that in a follow-up if you'd like.

PR #2173, opened 2026-07-28
fix(awsneuron): handle init-container annotation indexes

Eshiv-Pandey:
**What type of PR is this?**

/kind bug

AWS Neuron `PatchAnnotations()` reads container resources with `pod.Spec.Containers[ctridx]`.

HAMi stores allocated devices using a combined container index: init containers first, then regular containers. When an AWS Neuron allocation belongs to an init container, `ctridx` can point at the wrong regular container. In that case, a whole-device init-container request can be patched as neuron-core IDs instead of the assigned Neuron device index.

This PR resolves the combined index against init containers first, then regular containers.

A regression test covers an init container requesting `aws.amazon.com/neuron` while a regular container is also present.

None filed.

Before fix:

```text
=== RUN   Test_PatchAnnotations/init_container_neuron_device
device_test.go:329: assertion failed: 0,1 (result[AWSNeuronAssignedIndex] string) != 0 (test.want[AWSNeuronAssignedIndex] string)
--- FAIL: Test_PatchAnnotations/init_container_neuron_device
```

After fix:

```text
=== RUN   Test_PatchAnnotations/init_container_neuron_device
--- PASS: Test_PatchAnnotations/init_container_neuron_device
```

Validation:

```text
ok   github.com/Project-HAMi/HAMi/pkg/device/awsneuron
ok   github.com/Project-HAMi/HAMi/pkg/device
```

No breaking change. AWS Neuron annotations now use the correct container resources for init-container allocations.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved AWS Neuron resource handling for workloads with init containers.
  * Prevented annotation failures when container indices are invalid or outside the available container list.

* **Tests**
  * Added coverage for AWS Neuron resources requested by init containers.
  * Expanded validation of returned resource assignments and types.

### DIFF
--- a/pkg/device/awsneuron/device.go
+++ b/pkg/device/awsneuron/device.go
@@ -135,8 +135,12 @@ func (dev *AWSNeuronDevices) PatchAnnotations(pod *corev1.Pod, annoinput *map[st
 		value := ""
 		for ctridx, dp := range devlist {
 			if len(dp) > 0 {
+				ctr := getContainerByIndex(pod, ctridx)
+				if ctr == nil {
+					continue
+				}
 				for _, val := range dp {
-					devValue, ok := pod.Spec.Containers[ctridx].Resources.Limits[corev1.ResourceName(dev.resourceCountName)]
+					devValue, ok := ctr.Resources.Limits[corev1.ResourceName(dev.resourceCountName)]
 					if ok {
 						c, _ := devValue.AsInt64()
 						if c > 0 {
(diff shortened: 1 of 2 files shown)


pkg/device/awsneuron/device.go
```diff
@@ -135,8 +135,12 @@ func (dev *AWSNeuronDevices) PatchAnnotations(pod *corev1.Pod, annoinput *map[st
 		value := ""
 		for ctridx, dp := range devlist {
 			if len(dp) > 0 {
+				ctr := getContainerByIndex(pod, ctridx)
+				if ctr == nil {
```
mesutoezdil, line 139: devlist comes from Resourcereqs which always sizes to init+regular, so when can this actually be nil, and is silently skipping the right answer if it ever is?

mesutoezdil (commented):
fyi #1773 reworks init container accounting across pods.go, quota.go and score.go, worth watching the order.

pkg/device/awsneuron/device.go
Eshiv-Pandey: @mesutoezdil 

I did traced through the code and found tht the **nil** you referred to is actually unreachable.
I originally added the check as a defensive guard since it's just a couple of integer comparisons and would avoid a panic if the invariant were ever broken.
for eg - the PR you mention may induce the panic? 
Do lemme know wht you think of the guards.

mesutoezdil:
cc @archlitchi @DSFans2014 @fouof @shouren for a second opinion on the defensive nil-check thread above.

pkg/device/awsneuron/device.go
mesutoezdil: #1773 already merged now. resourcereqs and decodepoddevices both still size to init+regular always. so this nil check stays unreachable, but safe to keep as a guard.

FouoF:
@mesutoezdil  I think it is ok to keep it as the guard.

FouoF (approved):

PR #2174, opened 2026-07-28
test(nvidia): raise calculate_score.go coverage from 18% to 94%

adity1raut:
**What type of PR is this?**
/kind feature

Adds unit tests for `pkg/device/nvidia/calculate_score.go`, raising its
statement coverage from ~18% to ~94%.

- `Filter`, `calculateGPUScore`, and every branch of
  `calculateGPUPairScore` (all P2PLink/NVLink types, nil handling, and
  the asymmetric-link panic) are now directly tested.
- `newDevice` and `deviceListBuilder.build()` (including its NVML
  Init/GetDevices/GetP2PLink/GetNVLink error paths) are exercised using
  fake `nvml.Interface` and `nvlib/device.Interface` implementations
  built on the moq-generated mocks already vendored under
  `go-nvml/pkg/nvml/mock`, so no real GPU/driver is required.
- `CalculateGPUScore`'s success path still isn't covered: `NewDevices()`
  calls `nvml.New()` directly, so exercising it would need either a real
  NVML driver or a source change to inject the interface — out of scope
  for this test-only PR.

Fixes #

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive coverage for NVIDIA device discovery, construction, enumeration, and PCI identifier handling.
  * Added validation for device filtering, GPU pair scoring, topology links, P2P connections, and NVLink levels.
  * Added checks for expected error handling across initialization, linking, and unknown-device scenarios.

### DIFF
--- a/pkg/device/nvidia/calculate_score_test.go
+++ b/pkg/device/nvidia/calculate_score_test.go
@@ -218,6 +218,7 @@ func Test_DeviceList_Filter(t *testing.T) {
 			got, err := list.Filter(test.uuids)
 			if test.wantErr {
 				assert.ErrorContains(t, err, "no device with uuid")
+				assert.Assert(t, got == nil)
 				return
 			}
 			assert.NilError(t, err)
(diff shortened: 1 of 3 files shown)


pkg/device/nvidia/calculate_score_test.go
```diff
@@ -111,22 +111,164 @@ func Test_calculateGPUScore(t *testing.T) {
...
-						continue
-					}
-					score[gpuJ.UUID] = calculateGPUPairScore(gpuI, gpuJ)
-				}
-				scoreList[i] = DeviceScore{
-					UUID:  gpuI.UUID,
-					Score: score,
-				}
-			}
+			scoreList := calculateGPUScore(test.args)
```
mesutoezdil: the old version reimplemented the loop and never called calculateGPUScore at all.

pkg/device/nvidia/calculate_score_test.go
```diff
@@ -111,22 +111,164 @@ func Test_calculateGPUScore(t *testing.T) {
...
+			Links:       map[int][]P2PLink{0: {{Type: linkType}}},
+		}
+
+		want := (level + 1) * 100
+		got := calculateGPUPairScore(gpu0, gpu1)
+		assert.Equal(t, got, want, "nvlink level %d (%s)", level+1, linkType)
+	}
+}
+
+func Test_calculateGPUPairScore_panicsOnAsymmetricLinks(t *testing.T) {
```
mesutoezdil: this pins the panic at calculate_score.go:215 as expected behaviour, but the only caller is the device plugin at register.go:215, so asymmetric nvlink data takes the plugin down. is the panic the contract we want to lock in?

mesutoezdil (commented):
2 notes inline.

mesutoezdil:
Production behavior also changed (panic to error), worth reflecting in the title/description.

adity1raut:
> Production behavior also changed (panic to error), worth reflecting in the title/description.

You're right, I missed that. I'll update the title and description to reflect the behavior change as well. Thanks!

mesutoezdil:
asymmetric pair errors out here, so RegisterInAnnotation never patches the node, blocking all device registration on it. #2306 scores it 0 instead and keeps registering.

/hold

archlitchi:

FouoF:
Thanks for the NVIDIA scoring tests. This test-only branch has remained on hold, and the production issue it was intended to support has since been addressed by the merged #2306. Keeping the old branch would preserve tests against a superseded implementation and require another large review without a current gap. We are closing this PR. If #2306 still lacks an important regression case, please propose only that case from the latest master and explain the failure it protects.

PR #2175, opened 2026-07-28
ci: scope reusable workflow permissions, pin build.sh images, stop persisting checkout credentials, bump mig-parted

mesutoezdil:
/kind cleanup
Closes four gaps in the build/CI path: five read-only workflows no longer persist the job token via `actions/checkout`, the six reusable workflows called by `auto-release.yaml` now declare the scope their caller already grants instead of `write-all`, `hack/build.sh` reads its base images from the digest-pinned values in `version.mk` instead of keeping a second unpinned copy, and `nvidia-mig-parted` moves from v0.12.2 to v0.14.4 to drop the `golang.org/x/sys` CVE it pulled in. The benchmark client image also drops `requests` and `curl`, which removes the unpinned `pip install` and covers both call sites with `urllib.request`.
The reusable-workflow scopes were copied from the existing per-job `permissions` blocks in `auto-release.yaml`, its only caller, so the effective token is unchanged. HAMi only shells out to mig-parted `export` and `apply -f`, and neither subcommand nor the config schema changed across 0.13 and 0.14.

### DIFF
--- a/.github/workflows/auto-release.yaml
+++ b/.github/workflows/auto-release.yaml
@@ -91,6 +91,7 @@ jobs:
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
         with:
           ref: ${{ env.RUN_TAG }}
+          persist-credentials: false
         # if branch exists, the action will no fail, and it output created=false
       - name: release Y branch
         uses: peterjgrainger/action-create-branch@4b81ce657e255acd677cc6c55c9c763654be3aef # v4.0.0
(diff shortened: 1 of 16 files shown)


mesutoezdil:
it clears 3 open code scanning alerts on master: 1164 (write-all, this pr fixes all 6 workflows that still have it), 1159 (unpinned pip and curl in Dockerfile.client) and 1045 (mig-parted v0.12.2 ships x/sys v0.31.0, v0.14.4 ships v0.45.0) @archlitchi

archlitchi (approved):
/lgtm

PR #2178, opened 2026-07-28
fix: leader election false positive when hostname is prefix

### LINKED ISSUE #2154: [Bug] Leader election isHolderOf false positive when hostname is a prefix of another node's hostname

Aayush-207:
**What type of PR is this?**

/kind bug

Fixes false positives in leader election hostname matching.
Previously, isHolderOf used strings.HasPrefix to check if the lease holder identity matched the local hostname. This caused nodes with shorter hostnames (e.g. dev) to incorrectly match longer hostnames (e.g. dev-server_<uuid>).
The fix uses strings.SplitN(holder, "_", 2) and compares only the first segment to the hostname, ensuring exact matches only.

Fixes #2154

Added unit tests using Ginkgo v2 DescribeTable to cover edge cases (dev vs dev-server, worker vs worker2, empty hostname, no underscore cases).

Verified locally with go test ./pkg/util/leaderelection/... -v.

Leader election now correctly validates lease holder identity by requiring exact hostname match before the "_" separator. This prevents false positives when hostnames are prefixes of one another, improving scheduler reliability in clusters with similar node names.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved leader election identity matching by comparing only the first hostname segment, avoiding incorrect matches when one hostname is a prefix of another.
  * Added safeguards so leader-holder determination returns false when lease or holder identity information is missing or incomplete.
* **Tests**
  * Added coverage for hostname-segment matching behavior, including edge cases for empty values and missing underscore delimiters.

### DIFF
--- a/pkg/util/leaderelection/leaderelection.go
+++ b/pkg/util/leaderelection/leaderelection.go
@@ -163,10 +163,13 @@ func (m *leaderManager) onDelete(obj any) {
 
 func (m *leaderManager) isHolderOf(lease *coordinationv1.Lease) bool {
 	// kube-scheduler lease id take format of `hostname + "_" + string(uuid.NewUUID())`
+	// We split on "_" and compare only the first segment to avoid false positives
+	// when one hostname is a prefix of another (e.g. "dev" vs "dev-server").
 	if lease == nil || lease.Spec.HolderIdentity == nil {
 		return false
 	}
-	return strings.HasPrefix(*lease.Spec.HolderIdentity, m.hostname)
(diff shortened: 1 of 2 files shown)


archlitchi:
Found a earlier PR on https://github.com/Project-HAMi/HAMi/pull/2155

PR #2179, opened 2026-07-28
fix: use atomic load/store for shared-region limit fields

om7057:
**What type of PR is this?**

/kind bug

**What this PR does / why we need it**:

`DeviceMemoryLimit`, `SetDeviceMemoryLimit`, `SetDeviceSmLimit`, `GetRecentKernel`/`SetRecentKernel`, `GetUtilizationSwitch`/`SetUtilizationSwitch`, and `LastKernelTime` (v1) in `pkg/monitor/nvidia/{v0,v1}` read and wrote `limit`/`smLimit`/`recentKernel`/`utilizationSwitch`/`lastKernelTime` in `sharedRegionT` with plain Go assignments. These are the same shared-memory words HAMi-core's `shared_region_t` marks `_Atomic` on the C side, which the companion PR above makes atomic. All of them now use `sync/atomic` on the Go side too. `proc_num` is the one field still left out: it's read through `activeProcs()`/`DeviceNum()` rather than a direct field access this PR owns, and converting it isn't as mechanical as the fields above since multiple callers slice `procs` off of it; out of scope here.

`recentKernel`, `utilizationSwitch`, and `lastKernelTime` were originally going to be covered by #2180. @shivv23 agreed to fold those into this PR instead of keeping two overlapping PRs open, and is credited as co-author on the commit that added them.

I checked and there is currently no caller of `SetDeviceMemoryLimit`/`SetDeviceSmLimit` in this repo outside tests, so switching those two setters is not fixing an observed bug. `DeviceMemoryLimit` (the getter) does have a live caller at `cmd/vGPUmonitor/metrics.go:504`, so making that read atomic has effect today, not just for a hypothetical future caller.

**Which issue(s) this PR fixes**:
Closes part of #2127. Companion PR on the C side: Project-HAMi/HAMi-core#238.

**Special notes for your reviewer**:

AI assistance disclosure: I used Claude Code in this PR to help reproduce and verify the ABI/struct-layout details raised in review (the v0/v1 `sharedRegionT` offset comparisons) and to cover edge cases like the legacy v0 cache format, and to draft the review replies and this description text, all of which I've read, verified, and stand behind.

Changes:
- `DeviceMemoryLimit`, `SetDeviceMemoryLimit`, `SetDeviceSmLimit` now use `sync/atomic.LoadUint64`/`StoreUint64` on the same shared-memory words, in both `v0/spec.go` and `v1/spec.go`.
[...]

PR #2180, opened 2026-07-28
fix(monitor): use atomic load/store for shared memory access

### LINKED ISSUE #2127: Should we use atomic load/store when accessing shared memory?
I just checked https://github.com/Project-HAMi/HAMi/pull/624, and it used ordinary Go assignments to set limits in shared memory.

Even if we do not worry about data tearing, Atomic operations can help invalidate caches and make the code more robust.

shivv23:

The vGPUmonitor writes GPU memory limits, SM limits, utilization switch, and recent kernel values to shared memory using ordinary Go assignments. When multiple container processes on the same node read/write these values concurrently, this causes torn reads and stale cache values on multi-core systems.

This PR replaces all direct shared-memory assignments with \sync/atomic\ load/store operations, ensuring memory visibility guarantees across goroutines and preventing data races on the mmap'd shared region.

Fixes #2127

No. This is an internal correctness fix with no API or behavioral change.

## How has this been tested?

- [x] \go test -race ./pkg/monitor/nvidia/v0/...\ passes
- [x] \go test -race ./pkg/monitor/nvidia/v1/...\ passes
- [x] \go vet ./pkg/monitor/nvidia/...\ passes
- [x] Added \TestSpec_ConcurrentAtomicAccess\ with 50 goroutines x 200 iterations to verify race-free concurrent access

## AI Assistance Disclosure

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reliability when NVIDIA monitoring values are read or updated concurrently.
  - Prevented inconsistent device memory, SM limits, recent kernel timing, and utilization settings during simultaneous operations.
  - Enhanced stability for multithreaded access to shared GPU monitoring data.
- **Tests**
  - Added concurrency coverage that exercises parallel setters/getters to verify consistent behavior under load.

### DIFF
--- a/pkg/monitor/nvidia/v0/spec.go
+++ b/pkg/monitor/nvidia/v0/spec.go
@@ -16,7 +16,10 @@ limitations under the License.
 
 package v0
 
-import "unsafe"
+import (
+	"sync/atomic"
+	"unsafe"
+)
 
 const maxDevices = 16
 
@@ -130,10 +133,9 @@ func (s Spec) DeviceSmUtil(idx int) uint64 {
 }
 
 func (s Spec) SetDeviceSmLimit(l uint64) {
-	idx := uint64(0)
-	for idx < s.sr.num {
-		s.sr.smLimit[idx] = l
-		idx += 1
+	n := s.sr.num
+	for idx := uint64(0); idx < n; idx++ {
+		atomic.StoreUint64(&s.sr.smLimit[idx], l)
 	}
 }
 
@@ -146,14 +148,13 @@ func (s Spec) DeviceUUID(idx int) string {
 }
 
 func (s Spec) DeviceMemoryLimit(idx int) uint64 {
-	return s.sr.limit[idx]
+	return atomic.LoadUint64(&s.sr.limit[idx])
 }
 
 func (s Spec) SetDeviceMemoryLimit(l uint64) {
-	idx := uint64(0)
-	for idx < s.sr.num {
-		s.sr.limit[idx] = l
-		idx += 1
+	n := s.sr.num
+	for idx := uint64(0); idx < n; idx++ {
+		atomic.StoreUint64(&s.sr.limit[idx], l)
 	}
 }
 
@@ -176,17 +177,17 @@ func (s Spec) GetPriority() int {
 }
 
 func (s Spec) GetRecentKernel() int32 {
-	return s.sr.recentKernel
+	return atomic.LoadInt32(&s.sr.recentKernel)
 }
 
 func (s Spec) SetRecentKernel(v int32) {
-	s.sr.recentKernel = v
+	atomic.StoreInt32(&s.sr.recentKernel, v)
 }
 
 func (s Spec) GetUtilizationSwitch() int32 {
-	return s.sr.utilizationSwitch
+	return atomic.LoadInt32(&s.sr.utilizationSwitch)
 }
 
 func (s Spec) SetUtilizationSwitch(v int32) {
-	s.sr.utilizationSwitch = v
+	atomic.StoreInt32(&s.sr.utilizationSwitch, v)
 }
--- a/pkg/monitor/nvidia/v0/spec_test.go
+++ b/pkg/monitor/nvidia/v0/spec_test.go
@@ -18,7 +18,9 @@ package v0
 
 import (
 	"reflect"
+	"sync"
 	"testing"
+	"unsafe"
 )
 
 type specTest struct {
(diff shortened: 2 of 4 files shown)


pkg/monitor/nvidia/v0/spec.go
```diff
@@ -146,13 +149,13 @@ func (s Spec) DeviceUUID(idx int) string {
 }
 
 func (s Spec) DeviceMemoryLimit(idx int) uint64 {
-	return s.sr.limit[idx]
+	return atomic.LoadUint64((*uint64)(unsafe.Pointer(&s.sr.limit[idx])))
```
mesutoezdil: the cast is a no op, `&s.sr.limit[idx]` is already a `*uint64`, so this just pulls in unsafe for nothing and turns off the type check, pls drop it here and in the other 7 spots.

mesutoezdil (commented):
heads up this overlaps #2179, which is older and already under review w/ a hold. yours covers more fields, theirs is limited to limit/smLimit. 

three notes inline. also no ai disclosure in the description, see https://github.com/Project-HAMi/HAMi/blob/master/CONTRIBUTING.md#ai-assistance-notice

pkg/monitor/nvidia/v1/spec.go
```diff
@@ -183,17 +186,17 @@ func (s Spec) GetPriority() int {
 }
 
 func (s Spec) GetRecentKernel() int32 {
-	return s.sr.recentKernel
+	return atomic.LoadInt32((*int32)(unsafe.Pointer(&s.sr.recentKernel)))
```
mesutoezdil: c has `_Atomic int recent_kernel` and `_Atomic int utilization_switch` so these 2 line up, but it also has `_Atomic uint64_t last_kernel_time` and LastKernelTime() at line 167 still reads it plainly, on a path vGPUmonitor hits every scrape. why is that one left out?

pkg/monitor/nvidia/v0/spec_test.go
```diff
@@ -567,3 +569,36 @@ func TestSpec_LastKernelTime(t *testing.T) {
...
+
+				_ = s.DeviceMemoryLimit(0)
+				_ = s.DeviceMemoryLimit(1)
+				_ = s.GetRecentKernel()
+				_ = s.GetUtilizationSwitch()
+			}
+		}(i)
+	}
+
+	wg.Wait()
```
mesutoezdil, line 603: v1 got the identical change but no test, can u add the same one there?

pkg/monitor/nvidia/v0/spec.go
shivv23: @mesutoezdil Fixed!

pkg/monitor/nvidia/v1/spec.go
shivv23: @mesutoezdil Fixed!

pkg/monitor/nvidia/v0/spec_test.go
shivv23: @mesutoezdil Fixed!

shivv23:
@mesutoezdil Hi! I have made all required changes to the PR. Do have a look,please.

om7057:
The loop bound `s.sr.num` in `SetDeviceSmLimit`/`SetDeviceMemoryLimit` is still re-read from the shared region on every iteration (all four spots, both v0 and v1), even though the store inside the loop is now atomic. Worth hoisting into a local before the loop, e.g. `n := s.sr.num` then `for idx := uint64(0); idx < n; idx++`.

om7057:
@shivv23  Any interest in combining into one instead of two? #2179 is the older one and already has a round of review from mesutoezdil, plus a companion C-side PR (HAMi-core#238) that's what actually makes the atomic fix matter cross-process. Your PR covers recentKernel/utilizationSwitch/lastKernelTime that mine doesn't, so happy to bring those into #2179 with you credited as co-author on that commit. No pressure if you'd rather keep #2180 going on its own; just thought I'd ask before we both burn review cycles on overlapping work.

shivv23:
@om7057 Yes sure, you can go ahead.

om7057:
> @om7057 Yes sure, you can go ahead.

@shivv23 I included `recentKernel`, `utilizationSwitch`, and `lastKernelTime` in #2179, along with a concurrency test that covers all four shared fields (`recentKernel`, `utilizationSwitch`, `lastKernelTime`, and the existing `limit`/`smLimit` changes).

I also added you as a co-author on the commit since your work was incorporated: https://github.com/Project-HAMi/HAMi/commit/f0fe1a0b69271e0229b1cc968b38db6f04ce9354

I think it makes sense to close this PR in favor of #2179 to keep everything in one place.

shivv23:
@om7057 Thanks and this PR is now closed.

PR #2190, opened 2026-07-29
fix(enflame): bound profile values before int32 conversion in Fit

mesutoezdil:
/kind bug
`Fit` in `pkg/device/enflame/device.go` converts three profile fields to `int32` without an upper bound. All three trace back to `strconv.Atoi` in `parseProfile`, so a profile name such as `4294967299g.4294967299gb` wraps to 3 slices and 3072 MiB and the profile is accepted. The bogus `drsSlice` also reaches the container annotation. This is CodeQL alert `go/incorrect-integer-conversion` at `pkg/device/enflame/device.go:394`. #2145 added `clampToInt32` and applied it elsewhere in the file but left these three sites.

### DIFF
--- a/pkg/device/enflame/device.go
+++ b/pkg/device/enflame/device.go
@@ -391,11 +391,6 @@ func (enf *EnflameDevices) Fit(devices []*device.DeviceUsage, request device.Con
 		reason[common.ModeNotFit]++
 		return false, tmpDevs, common.GenReason(reason, len(devices))
 	}
-	// Reject rather than clamp: these come from strconv.Atoi on a profile name,
-	// so an out of range value means the profile is malformed, and clamping
-	// would let it through with a plausible looking size. MemoryGB is bounded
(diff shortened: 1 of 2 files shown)


pkg/device/enflame/device.go
```diff
@@ -391,20 +391,28 @@ func (enf *EnflameDevices) Fit(devices []*device.DeviceUsage, request device.Con
 		reason[common.ModeNotFit]++
 		return false, tmpDevs, common.GenReason(reason, len(devices))
 	}
-	requiredSlice := int32(profile.Size)
-	if requiredSlice <= 0 {
+	// Reject rather than clamp: these come from strconv.Atoi on a profile name,
```
archlitchi: please remove these comments here and others

mesutoezdil: sure, done 👍🏽

FouoF (approved):

PR #2194, opened 2026-07-29
fix(monitor): fix use-after-unmap crash in loadCache + raise cudevshr.go coverage to 90%

adity1raut:
**What type of PR is this?**
/kind feature

- Fixes a crash bug and adds tests for `pkg/monitor/nvidia/cudevshr.go`.

**Bug fix:** `loadCache()` unmapped `usage.data` and then read from it afterward to build an error message — a use-after-unmap that crashes the whole vGPU monitor process whenever a cache file has a valid magic flag but an unknown version/size. Fixed by reading the values before unmapping.

**Tests:** Added unit tests covering `ContainerLister` and `loadCache`, raising `cudevshr.go` coverage from 0% to ~91%. Covers the constructor's error paths, `Update()`'s directory-scanning logic, and `loadCache`'s cache-file parsing (using real mmap'd temp files since the code reads raw memory via `unsafe.Pointer`).

image

Fixes #

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cache loading safety by requiring both the expected version and minimum file size before parsing modern formats.
  * Preserved detected version details in unsupported-format errors.
  * Prevented concurrent container memory cleanup while metrics are being collected.
  * Corrected cache memory layout compatibility for 64-bit process identifiers.
* **New Features**
  * Added a `MinSize()` helper to define the minimum safe cache size.
* **Tests**
  * Expanded coverage for cache parsing, container discovery, informer timeouts, pod events, and directory reconciliation.

### DIFF
--- a/pkg/monitor/nvidia/cudevshr_test.go
+++ b/pkg/monitor/nvidia/cudevshr_test.go
@@ -35,8 +35,14 @@ import (
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/cache"
 
+	v1 "github.com/Project-HAMi/HAMi/pkg/monitor/nvidia/v1"
 	"github.com/Project-HAMi/HAMi/pkg/util"
 )
+
+// v1CacheFileSize is large enough for v1.CastSpec to safely dereference
+// every field of the real v1 sharedRegionT; using a smaller fixture would
+// leave usage.Info backed by an out-of-bounds pointer.
+var v1CacheFileSize = v1.MinSize()
 
 // fakePodLister is a minimal corelisters.PodLister used to drive Update()'s
 // error path without touching a real (or fake) Kubernetes API server.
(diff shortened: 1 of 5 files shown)


pkg/monitor/nvidia/cudevshr_test.go
```diff
@@ -0,0 +1,435 @@
...
+
+func Test_ContainerLister_Accessors(t *testing.T) {
+	l := &ContainerLister{
+		containers: map[string]*ContainerUsage{"a": {}},
+	}
+	assert.Equal(t, len(l.ListContainers()), 1)
+	assert.Assert(t, l.Clientset() == nil)
+
+	l.Lock()
+	l.UnLock()
```
mesutoezdil, line 93: nothing in the repo actually calls these Lock/UnLock, and metrics.go `Collect()` reads `c.Info` on the prometheus scrape goroutine while `Update()` munmaps it, same use-after-unmap class as the one this pr fixes, in scope here?

pkg/monitor/nvidia/v1/spec.go
```diff
@@ -174,6 +174,12 @@ func CastSpec(data []byte) Spec {
 	}
 }
 
+// MinSize returns the minimum byte length data must have for CastSpec to be
+// memory-safe.
+func MinSize() int {
+	return int(unsafe.Sizeof(sharedRegionT{}))
```
mesutoezdil, line 187: `MinSize()` uses go's sizeof but the file is written by libvgpu's C `shared_region_t`, and the totals match only by padding luck (go `ownerPid uint32` vs C `size_t owner_pid`), can u add sth to catch drift?

mesutoezdil:

pkg/monitor/nvidia/cudevshr_test.go
adity1raut: I am Fixed This now

archlitchi (approved):
/lgtm

PR #2197, opened 2026-07-29
fix(scheduler): preserve concurrent node locks

ShiroKSH:
/kind bug

Prevents concurrent scheduler replicas from replacing or clearing a node lock after a Kubernetes API conflict retry.

Fixes none.

`SetNodeLock` and `ReleaseNodeLock` previously checked the annotation before entering `retry.OnError`. After a conflict, the retry fetched a newer Node and patched it without rechecking the lock. An acquisition could overwrite a concurrent holder; a release could remove it.

The retry now treats the annotation observed before patching as a compare-and-set target. Acquisition returns `ErrNodeLockContention` when a refreshed Node is locked. Release only removes the exact lock value it originally observed. Focused fake-client regression tests cover both conflict paths.

Validation:

- `make verify`
- `make test`
- `go test ./pkg/util/nodelock -run 'Test(SetNodeLockPreservesConcurrentLockAfterConflict|ReleaseNodeLockPreservesConcurrentLockAfterConflict)$' -count=1 -v -timeout=60s`

This is scheduler-only; no device-allocation or in-container isolation path changed.

No.

**AI assistance disclosure**:

AI assistance was used for repository exploration, race analysis, and drafting the focused implementation and tests. I reviewed the affected scheduler path and the validation results, and take responsibility for this contribution.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved node lock contention handling with clearer contention errors and no unnecessary retries.
  * Revalidated lock ownership before acquiring or releasing locks to prevent overwriting concurrent changes.
  * Preserved active locks during conflicting updates and improved compatibility with legacy lock records.
  * Updated release reporting so it occurs only when a lock is successfully removed.

* **Tests**
  * Expanded coverage for concurrent lock acquisition, release conflicts, restamped locks, and retry behavior.

### DIFF
--- a/pkg/util/nodelock/nodelock.go
+++ b/pkg/util/nodelock/nodelock.go
@@ -140,6 +140,7 @@ func SetNodeLock(nodeName string, lockname string, pods *corev1.Pod) error {
 	if _, ok := node.Annotations[NodeLockKey]; ok {
 		return fmt.Errorf("node %s is locked: %w", nodeName, ErrNodeLockContention)
 	}
+	lockOwner := NodeLockSep + GeneratePodNamespaceName(pods, NodeLockSep)
 	err = retry.OnError(DefaultStrategy, func(err error) bool {
 		return !IsNodeLockContention(err)
 	}, func() error {
@@ -148,7 +149,11 @@ func SetNodeLock(nodeName string, lockname string, pods *corev1.Pod) error {
 			klog.ErrorS(err, "Failed to get node when retry to patch", "node", nodeName)
 			return err
 		}
-		if _, ok := node.Annotations[NodeLockKey]; ok {
+		lockStr, ok := node.Annotations[NodeLockKey]
+		if ok && strings.Contains(lockStr, NodeLockSep) && strings.HasSuffix(lockStr, lockOwner) {
+			return nil
+		}
+		if ok {
 			return fmt.Errorf("node %s is locked: %w", nodeName, ErrNodeLockContention)
 		}
 		patchData := fmt.Sprintf(`{"metadata":{"annotations":{"%s":"%s"},"resourceVersion":"%s"}}`, NodeLockKey, GenerateNodeLockKeyByPod(pods), node.ResourceVersion)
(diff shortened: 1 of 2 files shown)


pkg/util/nodelock/nodelock.go
```diff
@@ -199,19 +202,26 @@ func ReleaseNodeLock(nodeName string, lockname string, pod *corev1.Pod, skipNode
 			klog.ErrorS(err, "Failed to get node when retry to patch", "node", nodeName)
 			return err
 		}
+		currentLock, ok := node.Annotations[NodeLockKey]
+		if !ok || currentLock != lockStr {
```
mesutoezdil: the outer check matches only the pod ns/name suffix but this one compares the full string, so a re-stamped lock of the same pod is skipped silently.

mesutoezdil (commented):
if you wouldn't mind doing merge commits instead of force pushing, that would make it a bit clearer what (if anything) has changed.

pkg/util/nodelock/nodelock.go
```diff
@@ -138,17 +138,19 @@ func SetNodeLock(nodeName string, lockname string, pods *corev1.Pod) error {
...
-		return true
+		return !IsNodeLockContention(err)
 	}, func() error {
 		node, err = client.GetClient().CoreV1().Nodes().Get(ctx, nodeName, metav1.GetOptions{})
 		if err != nil {
 			klog.ErrorS(err, "Failed to get node when retry to patch", "node", nodeName)
 			return err
 		}
+		if _, ok := node.Annotations[NodeLockKey]; ok {
+			return fmt.Errorf("node %s is locked: %w", nodeName, ErrNodeLockContention)
```
mesutoezdil, line 157: if our first patch succeeded but the resp was lost, this retry sees our own lock and fails w/ contention even tho we hold it, pls return nil when the lock suffix matches this pod

Norway-02:
Looked through the unresolved review threads:

**Re: the restamped-lock silent-skip in `ReleaseNodeLock`** (@mesutoezdil's inline comment)

The issue is in this retry block in `ReleaseNodeLock`:

```go
if skipNodeLockOwnerCheck || !strings.Contains(currentLock, NodeLockSep) {
    if currentLock != lockStr {
        return nil  // silent exit
    }
}
```

If `skipNodeLockOwnerCheck` is true, or the lock is legacy (no separator), but the scheduler has since re-stamped the same pod's lock with a new timestamp, `currentLock != lockStr` is true and the release silently aborts — leaving the node locked indefinitely.

The outer check for owner-encoded locks correctly uses suffix matching (pod ns/name), but this branch compares the **exact original lock string**, which breaks when the timestamp refreshes.

A straightforward fix would be to apply the same suffix check here:

```go
if !skipNodeLockOwnerCheck {
    if strings.Contains(currentLock, NodeLockSep) {
        if !strings.HasSuffix(currentLock, lockOwner) {
            return nil
        }
    } else if currentLock != lockStr {
        // legacy format: only abort if another holder took over
        return nil
    }
}
```

This is consistent with CodeRabbit's suggestion that was marked resolved, but `mesutoezdil`'s non-line comment suggests the issue still exists in the current diff.

**Re: line 157 (SetNodeLock)** — The current code already returns `nil` when the suffix matches this pod's lock, so that case looks correctly handled in the current version of the PR.

Happy to assist test or iterate on this if helpful.

Shouren (commented):
> Prevents concurrent scheduler replicas from replacing or clearing a node lock after a Kubernetes API conflict retry.

@ShiroKSH HAMi does not currently support active-active scheduler replicas. Its high-availability model uses leader election, so only the leader is expected to perform scheduling and binding while the other replicas remain standby. Therefore, concurrent node-lock updates from multiple scheduler replicas are not expected in a supported deployment, and the scenario this PR aims to fix does not appear to exist.

archlitchi (approved):
/lgtm

PR #2201, opened 2026-07-29
Fix device plugin recovery from Unhealthy state

### LINKED ISSUE #2200: Device Plugin: Devices cannot recover from Unhealthy state
**What happened**:
In `pkg/device-plugin/nvidiadevice/nvinternal/plugin/server.go`, there is a known `FIXME` indicating that devices have no mechanism to recover once they are marked as `Unhealthy`. When the health check channel emits an event, the device plugin unconditionally marks the device as `Unhealthy` and reports it to the kubelet. If a GPU encounters a transient error (e.g., a temporary XID error that clears, or a thermal throttling event that resolves), it will remain permanently unavailable to the Kubernetes cluster. The only way to restore the GPU's capacity is to manually restart the device plugin pod on that node.

The health monitoring system should distinguish between health events that are transient vs. permanent, and periodically re-evaluate the status of `Unhealthy` devices to transition them back to `Healthy` when they recover, so manual intervention is not required.

Identified directly from the `FIXME` in the source code.

N/A

**Environment**:
- HAMi version: Latest main branch
- Others: N/A

itsvishalyadav:
**What type of PR is this?**

/kind bug
/kind cleanup

This PR refactors the device plugin's health signaling mechanism to support bidirectional health reporting (both `Healthy` and `Unhealthy` events) as groundwork for recovering devices from an `Unhealthy` state. 

*Note: Initially, this PR attempted to implement an automatic recovery mechanism by polling `GetMemoryInfo` every 5 seconds. However, as noted in review by @mesutoezdil, `GetMemoryInfo` can still succeed on a GPU that just hit an XID fault, which would unsafely hide real hardware faults. Therefore, the automatic polling mechanism has been reverted. This PR now focuses purely on the API rename and mock fixes to lay the foundation for a safe (likely manual/endpoint-driven) recovery mechanism in the future.*

**AI Assistance Notice:**
> I consulted ChatGPT to understand the codebase and help draft the initial attempt at this feature.

Fixes #2200

N/A

No

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved GPU health monitoring so devices affected by placement failures, monitoring errors, or critical hardware events are reported with the appropriate health status.
  * Ensured health state updates are handled consistently across supported device management implementations.
  * Improved reliability when health information cannot be determined by marking affected devices appropriately.

### DIFF
--- a/pkg/device-plugin/nvidiadevice/nvinternal/rm/health.go
+++ b/pkg/device-plugin/nvidiadevice/nvinternal/rm/health.go
@@ -88,6 +88,8 @@ func (r *nvmlResourceManager) checkHealth(stop <-chan interface{}, devices Devic
 		_ = eventSet.Free()
 	}()
 
+	unhealthyDevices := make(map[string]*Device)
+
 	parentToDeviceMap := make(map[string]*Device)
 	deviceIDToGiMap := make(map[string]uint32)
 	deviceIDToCiMap := make(map[string]uint32)
@@ -125,11 +127,10 @@ func (r *nvmlResourceManager) checkHealth(stop <-chan interface{}, devices Devic
 		if ret != nvml.SUCCESS {
 			klog.Infof("Marking device %v as unhealthy: %v", d.ID, ret)
 			d.Health = kubeletdevicepluginv1beta1.Unhealthy
+			unhealthyDevices[d.ID] = d
 			health <- d
 		}
 	}
-
-	unhealthyDevices := make(map[string]*Device)
 
 	for {
 		select {
--- a/pkg/device-plugin/nvidiadevice/nvinternal/rm/health.go
+++ b/pkg/device-plugin/nvidiadevice/nvinternal/rm/health.go
@@ -40,6 +40,7 @@ import (
 
 	"github.com/NVIDIA/go-nvml/pkg/nvml"
 	"k8s.io/klog/v2"
+	kubeletdevicepluginv1beta1 "k8s.io/kubelet/pkg/apis/deviceplugin/v1beta1"
 )
 
 const (
@@ -56,8 +57,8 @@ const (
 	envEnableHealthChecks = "DP_ENABLE_HEALTHCHECKS"
 )
 
-// CheckHealth performs health checks on a set of devices, writing to the 'unhealthy' channel with any unhealthy devices
-func (r *nvmlResourceManager) checkHealth(stop <-chan interface{}, devices Devices, unhealthy chan<- *Device, disableNVML <-chan bool) error {
+// CheckHealth performs health checks on a set of devices, writing to the 'health' channel with any unhealthy devices
+func (r *nvmlResourceManager) checkHealth(stop <-chan interface{}, devices Devices, health chan<- *Device, disableNVML <-chan bool) error {
 	xids := getHealthCheckXids()
 	if xids.IsAllDisabled() {
 		return nil
@@ -123,9 +124,12 @@ func (r *nvmlResourceManager) checkHealth(stop <-chan interface{}, devices Devic
 		}
 		if ret != nvml.SUCCESS {
 			klog.Infof("Marking device %v as unhealthy: %v", d.ID, ret)
-			unhealthy <- d
+			d.Health = kubeletdevicepluginv1beta1.Unhealthy
+			health <- d
 		}
 	}
+
+	unhealthyDevices := make(map[string]*Device)
 
 	for {
 		select {
@@ -141,12 +145,31 @@ func (r *nvmlResourceManager) checkHealth(stop <-chan interface{}, devices Devic
 
 		e, ret := eventSet.Wait(5000)
 		if ret == nvml.ERROR_TIMEOUT {
+			for id, d := range unhealthyDevices {
+				uuid, _, _, err := r.getDevicePlacement(d)
+				if err != nil {
+					continue
+				}
+				gpu, ret := r.nvml.DeviceGetHandleByUUID(uuid)
+				if ret != nvml.SUCCESS {
+					continue
+				}
+				_, ret = gpu.GetMemoryInfo()
+				if ret == nvml.SUCCESS {
+					klog.Infof("Device %s has recovered", d.ID)
+					d.Health = kubeletdevicepluginv1beta1.Healthy
+					health <- d
+					delete(unhealthyDevices, id)
+				}
+			}
 			continue
 		}
 		if ret != nvml.SUCCESS {
 			klog.Infof("Error waiting for event: %v; Marking all devices as unhealthy", ret)
 			for _, d := range devices {
-				unhealthy <- d
+				d.Health = kubeletdevicepluginv1beta1.Unhealthy
+				unhealthyDevices[d.ID] = d
+				health <- d
 			}
 			continue
 		}
@@ -167,7 +190,9 @@ func (r *nvmlResourceManager) checkHealth(stop <-chan interface{}, devices Devic
 			// If we cannot reliably determine the device UUID, we mark all devices as unhealthy.
 			klog.Infof("Failed to determine uuid for event %v: %v; Marking all devices as unhealthy.", e, ret)
 			for _, d := range devices {
-				unhealthy <- d
+				d.Health = kubeletdevicepluginv1beta1.Unhealthy
+				unhealthyDevices[d.ID] = d
+				health <- d
 			}
 			continue
 		}
@@ -188,7 +213,9 @@ func (r *nvmlResourceManager) checkHealth(stop <-chan interface{}, devices Devic
 		}
 
 		klog.Infof("XidCriticalError: Xid=%d on Device=%s; marking device as unhealthy.", e.EventData, d.ID)
-		unhealthy <- d
+		d.Health = kubeletdevicepluginv1beta1.Unhealthy
+		unhealthyDevices[d.ID] = d
+		health <- d
 	}
 }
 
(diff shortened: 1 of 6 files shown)


itsvishalyadav:
Hi @dsfans2014, @FouoF and @ouyangluwei163
Could you please take a look at this PR when you have a moment? It introduces a mechanism to automatically recover devices from the Unhealthy state without requiring a full device plugin restart, addressing a known \FIXME\ in the code. I've also addressed the initial feedback from CodeRabbit. Let me know if there's anything else you'd like me to change. Thank you!

pkg/device-plugin/nvidiadevice/nvinternal/rm/health.go
```diff
@@ -56,8 +57,8 @@ const (
 	envEnableHealthChecks = "DP_ENABLE_HEALTHCHECKS"
 )
 
-// CheckHealth performs health checks on a set of devices, writing to the 'unhealthy' channel with any unhealthy devices
-func (r *nvmlResourceManager) checkHealth(stop <-chan interface{}, devices Devices, unhealthy chan<- *Device, disableNVML <-chan bool) error {
+// CheckHealth performs health checks on a set of devices, writing to the 'health' channel with any unhealthy devices
+func (r *nvmlResourceManager) checkHealth(stop <-chan interface{}, devices Devices, health chan<- *Device, disableNVML <-chan bool) error {
```
mesutoezdil, line 61: the rename stops at the sig, l.102/112/119 still send to `unhealthy` so this pkg doesn't compile.

pkg/device-plugin/nvidiadevice/nvinternal/rm/health.go
```diff
@@ -87,6 +88,8 @@ func (r *nvmlResourceManager) checkHealth(stop <-chan interface{}, devices Devic
 		_ = eventSet.Free()
 	}()
 
+	unhealthyDevices := make(map[string]*Device)
```
mesutoezdil: devices marked unhealthy in the registration loop below never land in this map, so they can never recover.

pkg/device-plugin/nvidiadevice/nvinternal/rm/health.go
```diff
@@ -123,7 +126,9 @@ func (r *nvmlResourceManager) checkHealth(stop <-chan interface{}, devices Devic
 		}
 		if ret != nvml.SUCCESS {
 			klog.Infof("Marking device %v as unhealthy: %v", d.ID, ret)
-			unhealthy <- d
+			d.Health = kubeletdevicepluginv1beta1.Unhealthy
```
mesutoezdil: we now write `d.Health` from the health goroutine while listandwatch reads the same `*Device` through `apiDevices()`, that's a data race afaict.

pkg/device-plugin/nvidiadevice/nvinternal/rm/health.go
```diff
@@ -141,12 +146,31 @@ func (r *nvmlResourceManager) checkHealth(stop <-chan interface{}, devices Devic
 
 		e, ret := eventSet.Wait(5000)
 		if ret == nvml.ERROR_TIMEOUT {
+			for id, d := range unhealthyDevices {
+				uuid, _, _, err := r.getDevicePlacement(d)
```
mesutoezdil: for a mig device `getDevicePlacement` gives back the parent gpu uuid, so the probe below says nothing abt the mig instance itself.

pkg/device-plugin/nvidiadevice/nvinternal/rm/health.go
```diff
@@ -141,12 +146,31 @@ func (r *nvmlResourceManager) checkHealth(stop <-chan interface{}, devices Devic
...
+				if err != nil {
+					continue
+				}
+				gpu, ret := r.nvml.DeviceGetHandleByUUID(uuid)
+				if ret != nvml.SUCCESS {
+					continue
+				}
+				_, ret = gpu.GetMemoryInfo()
```
mesutoezdil: `GetMemoryInfo` still succeeds on a gpu that just hit an xid, so imo this flips nearly every device back to healthy after 5s and we lose the signal for real faults.

mesutoezdil (commented):
could you add a unit test for the recovery path? nothing covers it rn.

and no ai assistance disclosure is present. if any ai tool was used, it must be disclosed per CONTRIBUTING.md: https://github.com/Project-HAMi/HAMi/blob/master/CONTRIBUTING.md#ai-assistance-notice

pkg/device-plugin/nvidiadevice/nvinternal/rm/rm_mock.go
```diff
@@ -152,16 +152,16 @@ func (mock *ResourceManagerMock) CheckHealth(stop <-chan interface{}, unhealthy
...
 // Check the length with:
 //
 //	len(mockedResourceManager.CheckHealthCalls())
 func (mock *ResourceManagerMock) CheckHealthCalls() []struct {
-	Stop      <-chan interface{}
-	Unhealthy chan<- *Device
+	Stop   <-chan interface{}
+	Health chan<- *Device
 } {
 	var calls []struct {
```
mesutoezdil, line 166: this local still says `Unhealthy` while the return type above says `Health`, so it doesn't compile.

itsvishalyadav:
Thank you for the detailed review @mesutoezdil! You raised excellent points. I have pushed a new commit that addresses your feedback:

1. **Flawed Recovery Signal & MIG Support:** You are entirely right that \GetMemoryInfo\ is not a safe signal for recovery from an XID fault since it can still succeed, meaning we would blindly restore faulty devices. Since NVML does not provide a robust Boolean flag for 'is recovered' without a workload test, I have reverted the automatic 5-second polling recovery mechanism completely to avoid hiding real hardware faults.
2. **Data Race:** By reverting the automatic recovery, we no longer mutate \d.Health\ asynchronously inside \health.go\'s goroutine. The write to \d.Health\ now happens synchronously back inside \ListAndWatch\, fixing the data race you correctly identified.
3. **Mock Compilation:** The un-updated mock field names have been fixed.
4. **AI Assistance Disclosure:** Added to the PR description per the guidelines.

This PR now serves purely as a refactoring groundwork to change the internal channel from \unhealthy\ to \health\ (supporting bidirectional signaling). Since automatic recovery via polling is unsafe with our current signals, what is the project's preferred approach to resolving the original \FIXME\? Should we provide an API endpoint or SIGHUP signal to allow cluster admins to manually clear the Unhealthy state without a full pod restart?

mesutoezdil:
> Thank you for the detailed review @mesutoezdil! You raised excellent points. I have pushed a new commit that addresses your feedback:
> 
> 1. **Flawed Recovery Signal & MIG Support:** You are entirely right that \GetMemoryInfo\ is not a safe signal for recovery from an XID fault since it can still succeed, meaning we would blindly restore faulty devices. Since NVML does not provide a robust Boolean flag for 'is recovered' without a workload test, I have reverted the automatic 5-second polling recovery mechanism completely to avoid hiding real hardware faults.
> 2. **Data Race:** By reverting the automatic recovery, we no longer mutate \d.Health\ asynchronously inside \health.go's goroutine. The write to \d.Health\ now happens synchronously back inside \ListAndWatch, fixing the data race you correctly identified.
> 3. **Mock Compilation:** The un-updated mock field names have been fixed.
> 4. **AI Assistance Disclosure:** Added to the PR description per the guidelines.
> 
> This PR now serves purely as a refactoring groundwork to change the internal channel from \unhealthy\ to \health\ (supporting bidirectional signaling). Since automatic recovery via polling is unsafe with our current signals, what is the project's preferred approach to resolving the original \FIXME? Should we provide an API endpoint or SIGHUP signal to allow cluster admins to manually clear the Unhealthy state without a full pod restart?

itsvishalyadav:
> > Thank you for the detailed review @mesutoezdil! You raised excellent points. I have pushed a new commit that addresses your feedback:
> > 
> > 1. **Flawed Recovery Signal & MIG Support:** You are entirely right that \GetMemoryInfo\ is not a safe signal for recovery from an XID fault since it can still succeed, meaning we would blindly restore faulty devices. Since NVML does not provide a robust Boolean flag for 'is recovered' without a workload test, I have reverted the automatic 5-second polling recovery mechanism completely to avoid hiding real hardware faults.
> > 2. **Data Race:** By reverting the automatic recovery, we no longer mutate \d.Health\ asynchronously inside \health.go's goroutine. The write to \d.Health\ now happens synchronously back inside \ListAndWatch, fixing the data race you correctly identified.
> > 3. **Mock Compilation:** The un-updated mock field names have been fixed.
> > 4. **AI Assistance Disclosure:** Added to the PR description per the guidelines.
> > 
> > This PR now serves purely as a refactoring groundwork to change the internal channel from \unhealthy\ to \health\ (supporting bidirectional signaling). Since automatic recovery via polling is unsafe with our current signals, what is the project's preferred approach to resolving the original \FIXME? Should we provide an API endpoint or SIGHUP signal to allow cluster admins to manually clear the Unhealthy state without a full pod restart?
> 
> Reminder: Answers must be written by human being. You can view the relevant rule here. https://github.com/Project-HAMi/HAMi/blob/master/CONTRIBUTING.md#contribution-gates "4. Review replies. The reply you post must be written by you and must address the specific point raised. Verbatim or canned AI replies, or replies that do not engage the comment, lead to the PR being closed."

My bad for the previous comment. 

I have made the requested changes in a new commit. Since the condition on GetMemoryInfo was not sufficient as a condition for the GPU recovery from XID error (since it might conceal a problem with the hardware), I have completely deleted the polling mechanism, thus solving the data race problem with ListAndWatch. Additionally, I have fixed the mock compilation errors.

The safe polling scheme will not work for the recovery of the devices, so do you think it would make sense to tackle the original FIXME? Can we proceed with the idea that I expose an API or a signal for the administrators of the cluster to mark the device as healthy without having to restart the pod?

FouoF:
@itsvishalyadav Actually these codes are forked form https://github.com/NVIDIA/k8s-device-plugin and they will be synced periodically. We'd better not change them if not necessary to avoid potential conflict.

PR #2206, opened 2026-07-30
Fix: adopt to gangScheduling feature gates on k8s v1.36+

archlitchi:
/kind bug

PR:https://github.com/Project-HAMi/HAMi/pull/2066 only support coScheduling plugin, but doesn't support GangScheduling feature gate implemented in v1.36.

How to use:

1. Enable
```
  --feature-gates=GenericWorkload=true,GangScheduling=true
```
on `kube-apiserver`, `kube-scheduler`, `kube-controller-manager`

2. Then, for `kube-apiserver`, add the parameter below: 
```
  --runtime-config=scheduling.k8s.io/v1alpha2=true
```

Example:
```
apiVersion: scheduling.k8s.io/v1alpha2
kind: PodGroup
metadata:
  name: test-fail-group
  namespace: default
spec:
  schedulingPolicy:
    gang:
      minCount: 2
---
apiVersion: v1
kind: Pod
metadata:
  name: e2e-test2-fail-pod1
  namespace: default
spec:
  schedulerName: hami-scheduler
  restartPolicy: Never
  schedulingGroup:
    podGroupName: test-fail-group
  containers:
  - name: c
    image: registry.k8s.io/pause:3.10.1
    resources:
      limits:
        hami-test.com/gpu: "1"
        hami-test.com/gpumem: "9000"
---
apiVersion: v1
kind: Pod
metadata:
  name: e2e-test2-fail-pod2
  namespace: default
spec:
  schedulerName: hami-scheduler
  restartPolicy: Never
  schedulingGroup:
    podGroupName: test-fail-group
  containers:
  - name: c
    image: registry.k8s.io/pause:3.10.1
    resources:
      limits:
        hami-test.com/gpu: "1"
        hami-test.com/gpumem: "9000"
```

Result:
cbbead1a-a0ea-4706-983c-40fef9597af6

I'll later add a user-guide to website

## Summary by CodeRabbit

* **Bug Fixes**
  * Node lock contention errors are now consistently identified as retryable when multiple workloads attempt to lock the same node.
  * Pod group membership detection now recognizes both scheduler labels and native scheduling group configuration.
* **Tests**
  * Added coverage for concurrent node-lock attempts and pod group membership scenarios, including missing and empty configuration values.


archlitchi:
CC @mesutoezdil @FouoF

pkg/util/util.go
```diff
@@ -281,10 +281,15 @@ func AllContainersCreated(pod *corev1.Pod) bool {
...
-// Coscheduling PodGroup, based on the presence of the PodGroupLabel.
 func IsPodGroupMember(pod *corev1.Pod) bool {
 	if pod == nil {
 		return false
 	}
-	return pod.Labels[PodGroupLabel] != ""
+	if pod.Labels[PodGroupLabel] != "" {
+		return true
+	}
+	if sg := pod.Spec.SchedulingGroup; sg != nil && sg.PodGroupName != nil && *sg.PodGroupName != "" {
```
mesutoezdil, line 291: 1 thing i am not sure about: filter already writes vgpu-node and the device annos, so a native gang that never reaches minCount seems to keep those bookings while the pods stay pending, is that worth a follow-up or out of scope here?

mesutoezdil, line 291: tiny note for the docs later: afaik this field only landed in 1.36 (1.35 used spec.workloadRef, which is tombstoned in the api now), so the v1.35+ wording in your notes may need a small tweak.

pkg/util/nodelock/nodelock_test.go
```diff
@@ -454,25 +454,47 @@ func TestConcurrentNodeLocks(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("Failed to create node: %v", err)
+	}
+	winner := &corev1.Pod{ObjectMeta: metav1.ObjectMeta{Name: "winner", Namespace: "test-ns"}}
+	loser := &corev1.Pod{ObjectMeta: metav1.ObjectMeta{Name: "loser", Namespace: "test-ns"}}
+
+	if err := SetNodeLock(nodeName, "", winner); err != nil {
+		t.Fatalf("winner SetNodeLock failed: %v", err)
+	}
+	err = SetNodeLock(nodeName, "", loser)
```
mesutoezdil: both calls here are sequential so this covers the already-locked branch rather than the race the name suggests, maybe just rename it or fire 2 concurrent LockNode calls?

pkg/util/util.go
FouoF: What about making the `getNodeUsage` ignore those pods pending and not the same PodGroup with the scheduled one?

archlitchi: yeah, we'll note that it only support v1.36+

pkg/util/nodelock/nodelock_test.go
archlitchi: yeah, make sense

pkg/util/util.go
archlitchi: > What about making the `getNodeUsage` ignore those pods pending and not the same PodGroup with the scheduled one?

make sense, but i think it needs carefully design, we can have a follow-up about that

FouoF (approved):

PR #2213, opened 2026-07-30
fix(device-plugin): periodically recover unhealthy devices via NVML re-check

### LINKED ISSUE #2200: Device Plugin: Devices cannot recover from Unhealthy state

shivv23:

The NVIDIA device plugin's \ListAndWatch\ has a known FIXME: devices marked \Unhealthy\ by the NVML health check (e.g., transient XID errors, thermal throttling) can never transition back to \Healthy\. Once a GPU encounters a transient error, it remains permanently unavailable to the Kubernetes cluster until the device plugin pod is manually restarted.

This PR adds a periodic recovery mechanism that re-checks unhealthy devices via NVML every 30 seconds. When a device responds successfully to \DeviceGetHandleByUUID\, it is marked \Healthy\ again and the kubelet is notified, eliminating the need for manual intervention.

Fixes #2200

No. This is an internal correctness fix with no API or behavioral change.

## How has this been tested?

- [x] \go build ./pkg/device-plugin/nvidiadevice/nvinternal/rm/...\ compiles
- [x] \go vet ./pkg/device-plugin/nvidiadevice/nvinternal/...\ (blocked on Windows NVML dependency, verified manually)

## AI Assistance Disclosure

## Summary by CodeRabbit

* **New Features**
  * Added automatic recovery attempts for unhealthy NVIDIA devices.
  * Device health status now updates when recovery succeeds.
  * Health updates are reported only when device status changes.

* **Bug Fixes**
  * Prevented devices from remaining permanently marked as unhealthy when they become recoverable.

### DIFF
--- a/pkg/device-plugin/nvidiadevice/nvinternal/plugin/server.go
+++ b/pkg/device-plugin/nvidiadevice/nvinternal/plugin/server.go
@@ -456,15 +456,35 @@ func (plugin *NvidiaDevicePlugin) GetDevicePluginOptions(context.Context, *kubel
 func (plugin *NvidiaDevicePlugin) ListAndWatch(e *kubeletdevicepluginv1beta1.Empty, s kubeletdevicepluginv1beta1.DevicePlugin_ListAndWatchServer) error {
 	s.Send(&kubeletdevicepluginv1beta1.ListAndWatchResponse{Devices: plugin.apiDevices()})
 
+	unhealthyDevices := make(map[string]bool)
+	recoveryTicker := time.NewTicker(30 * time.Second)
+	defer recoveryTicker.Stop()
+
 	for {
 		select {
 		case <-plugin.stop:
 			return nil
 		case d := <-plugin.health:
-			// FIXME: there is no way to recover from the Unhealthy state.
 			d.Health = kubeletdevicepluginv1beta1.Unhealthy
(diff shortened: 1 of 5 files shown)


mesutoezdil:
Thanks. This repeats the approach already rejected in #2201: a successful NVML call is not a safe recovery signal, and this forked code should be fixed upstream instead.

shivv23:
@mesutoezdil Oh I didn't see that. Apologies for this mistake.

PR #2216, opened 2026-07-30
fix(kunlun): validate interconnect topology parsing to prevent panics

### LINKED ISSUE #2215: kunlun: scheduler panics on malformed interconnect topology string in topo.go
**What happened**:

`pkg/device/kunlun/topo.go` parses the interconnect topology strings (`InterGroupConnection`, `GroupConnection`, `InterGroupConnection2`) by splitting on `-` and indexing the result without any length check, and every `strconv.Atoi` error is discarded:

https://github.com/Project-HAMi/HAMi/blob/master/pkg/device/kunlun/topo.go#L99-L128

```go
lw, _ := strconv.Atoi(strings.Split(pair, "-")[0])
rw, _ := strconv.Atoi(strings.Split(pair, "-")[1])  // panics if pair has no "-"
```

If any pair is malformed (e.g. `"3"` with no dash, or a trailing comma producing an empty element), `strings.Split(pair, "-")[1]` panics with index out of range. `interconnect()` runs this parsing inside the scheduler's device-fit path, so a malformed string would crash the scheduler extender at scheduling time. The same pattern appears at 7 call sites across `parseInterconnection()`, `parseInterconnection2()`, and `interconnect()`.

Additionally, in `parseInterconnection2()`, `connect := make([]int, 4)` is indexed with `connect[i] = v` for every split element, so a group with more than 4 values also panics with index out of range.

To be precise about severity: today these three strings are `const` in `device.go` and well-formed, so the panic is **latent** — it becomes reachable the moment they are edited or (more likely) made configurable, and nothing except luck guards it. Two problems are live right now though:

1. Silently-discarded `Atoi` errors mean a typo like `"0-x"` is treated as device index `0`, silently selecting a wrong device pair rather than failing.
2. `interconnect()` re-splits and re-parses the topology string **inside a triple-nested loop** over devices, on every scheduling attempt — the topology is constant and should be parsed once.

Topology parsing should be validated (skip/reject malformed pairs with a log instead of panicking), and parsed once instead of on every scheduling pass.

Call `parseInterconnection()` with `InterGroupConnection = "0-1,3"` — index out of range panic.

`topo.go` currently has no test file.

I'd like to work on this: a single validated parse helper used by all call sites, parse-once for the constant topology, plus a new `topo_test.go`. Will open a PR right away.

**Environment**:
- HAMi version: master (05e6c80)

Aamod007:
**What type of PR is this?**

/kind bug

`pkg/device/kunlun/topo.go` parsed the interconnect topology strings by indexing `strings.Split` results without length checks and discarding every `strconv.Atoi` error. A malformed pair (e.g. `"3"` with no dash, or a trailing comma) panics with index out of range inside the scheduler's fit path, and a typo like `"0-x"` was silently treated as device index `0`, selecting a wrong device pair.

This PR:
- introduces a shared `parseGroup` helper that validates element count and integer parsing, skipping malformed groups with a warning instead of panicking;
- hoists the `InterGroupConnection` parsing out of `interconnect()`'s triple-nested device loop, so the constant topology is parsed once per call instead of once per device pair;
- adds `topo_test.go` (the file previously had no tests) covering malformed inputs, the default topology constants, and 2-device pair selection.

Fixes #2215

Behavior for the current well-formed `const` topology strings is unchanged — `TestParseInterconnection`/`TestParseInterconnection2` pin the exact output for the defaults.

AI assistance was used during development; all changes were reviewed and tested by me.

```release-note
fix(kunlun): malformed interconnect topology strings are now skipped with a warning instead of panicking the scheduler
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Kunlun topology parsing with stricter validation of group arity and numeric tokens.
  * Malformed interconnection definitions are now ignored safely, reducing inconsistent behavior.
  * Enhanced interconnection selection to reuse pre-parsed pairs for more reliable, deterministic compatibility matching.

* **Tests**
  * Added unit tests covering valid and malformed topology/interconnection inputs.
  * Added scenarios to verify correct interconnection pair selection and behavior when no suitable devices are available.

### DIFF
--- a/pkg/device/kunlun/topo.go
+++ b/pkg/device/kunlun/topo.go
@@ -96,33 +96,47 @@ func calcscore(p []int, c []int) float32 {
 	return 1000
 }
 
+// parseGroup parses a "-"-separated group of device indices (e.g. "0-4" or
+// "0-1-4-5") and requires it to contain exactly want valid integers.
+// Malformed groups are skipped with a warning instead of panicking.
+func parseGroup(group string, want int) ([]int, bool) {
+	values := strings.Split(group, "-")
+	if len(values) != want {
+		klog.Warningf("skipping malformed interconnect group %q: expected %d values", group, want)
+		return nil, false
+	}
+	connect := make([]int, want)
+	for i, value := range values {
+		v, err := strconv.Atoi(value)
+		if err != nil {
+			klog.Warningf("skipping malformed interconnect group %q: %v", group, err)
+			return nil, false
+		}
+		connect[i] = v
+	}
+	return connect, true
+}
+
+func parsePairs(s string) [][]int {
+	var pairs [][]int
+	for pair := range strings.SplitSeq(s, ",") {
+		if p, ok := parseGroup(pair, 2); ok {
+			pairs = append(pairs, p)
+		}
+	}
+	return pairs
+}
+
 func parseInterconnection() [][]int {
-	var interconnection [][]int
-	pairs := strings.Split(InterGroupConnection, ",")
-	for _, pair := range pairs {
-		lw, _ := strconv.Atoi(strings.Split(pair, "-")[0])
-		rw, _ := strconv.Atoi(strings.Split(pair, "-")[1])
-		interconnection = append(interconnection, []int{lw, rw})
-	}
-	pairs = strings.Split(GroupConnection, ",")
-	for _, pair := range pairs {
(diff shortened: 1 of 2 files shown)


mesutoezdil:
these topology strings are compile-time constants defined in device.go,
and not user or config input,
so the described panic is unreachable at runtime.

pr can be closed.

PR #2218, opened 2026-07-30
Fix/multi device quota enforcement

### LINKED ISSUE #2157: [Bug] ResourceQuota enforcement only applies to NVIDIA devices, not other accelerator backends

manmathbh:
**What type of PR is this?**

/kind bug

fitResourceQuota had a hard skip for every device except NVIDIA. Pods
requesting Ascend, Cambricon, Hygon or any other backend could blow right
past namespace ResourceQuota at admission time.

Removed the guard. Non-NVIDIA devices use memoryFactor=1, NVIDIA keeps
nvidia.MemoryFactor as before.

Fixes #2157

FitQuota in pkg/device/quota.go was already device-agnostic — call it with
an Ascend device name and it returns the right answer. The gap was purely
in fitResourceQuota skipping the call for non-NVIDIA.

Test setup registers a stub Ascend device with a 1000mb quota (0 used) and
verifies deny/allow for over/under-quota requests plus a no-count-resource
edge case.

Pods requesting non-NVIDIA accelerator memory or core resources are now
checked against namespace ResourceQuota at admission time, same as NVIDIA.
Previously they were silently excluded.

## Summary by CodeRabbit

* **Bug Fixes**
  * Quota checks now account for all supported device types, not only NVIDIA GPUs.
  * Improved handling of device-specific memory quotas, including Ascend devices.
* **Tests**
  * Expanded quota validation coverage for mixed device types and memory-only requests.
  * Added scenarios for exceeded and successfully fitted Ascend quotas.

### DIFF
--- a/pkg/scheduler/webhook.go
+++ b/pkg/scheduler/webhook.go
@@ -130,11 +130,10 @@ func isPrivilegedContainer(ctr *corev1.Container) bool {
 
 func fitResourceQuota(pod *corev1.Pod) bool {
 	for deviceName, dev := range device.GetDevices() {
-		// Only supports NVIDIA
-		if deviceName != nvidia.NvidiaGPUDevice {
-			continue
+		memoryFactor := int32(1)
+		if deviceName == nvidia.NvidiaGPUDevice {
+			memoryFactor = nvidia.MemoryFactor
 		}
-		memoryFactor := nvidia.MemoryFactor
 		resourceNames := dev.GetResourceNames()
 		resourceName := corev1.ResourceName(resourceNames.ResourceCountName)
 		memResourceName := corev1.ResourceName(resourceNames.ResourceMemoryName)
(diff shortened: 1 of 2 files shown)


mesutoezdil:
the gap is real, but hardcoding memoryfactor=1 for non-nvidia devices is wrong..

quota usage is accumulated from factor-scaled values (ascend/hygon multiply by their configured memoryfactor in generateresourcerequests), 
so with factor > 1 the check compares mixed units and denies pods that are within quota. 

the factor must come from each backend's config. 
also note the issue author had already volunteered to work on #2157.

DSFans2014 (commented):
@manmathbh  thank you for your contribution. it seems duplicated with #2172

manmathbh:
Didn't see #2172 was open for this. Closing, thanks for the review.

PR #2223, opened 2026-07-31
fix(ascend): update ascend config according to latest driver

FouoF:
**What type of PR is this?**

/kind bug

Sync with https://github.com/Project-HAMi/ascend-device-plugin/pull/57, new driver has corrected the template name.

Fixes #

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated Ascend 910B4-1 device configuration template names to accurately reflect their memory capacities.
  * Preserved existing memory, core, and CPU resource settings.

### DIFF
--- a/charts/hami/templates/scheduler/device-configmap.yaml
+++ b/charts/hami/templates/scheduler/device-configmap.yaml
@@ -391,15 +391,11 @@ data:
         runtimeClassName: "{{ .Values.devices.ascend.runtimeClassName }}"
         overwriteEnv: {{ .Values.scheduler.overwriteEnv | default "false" }}
         templates:
-          # NOTE: Names of vnpu template for 910B4-1 are fixed by Ascend runtime and must not be changed.
-          # The memory is used for scheduling so the correct values must be set.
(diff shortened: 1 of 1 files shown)


mesutoezdil:
lgtm

just 1 ask: this breaks 910B4-1 vNPU creation on old drivers, maybe fill the releasenote field with smth like "910B4-1 vNPU template names updated, requires latest Ascend driver". 
Optional: keep a 1-line comment noting template names depend on driver version. WDYT?

FouoF:
Former comments indicate that the template name is not consistent with template spec. So i think it should be treated as a driver bug and the best practice is to update the driver. 

https://github.com/Project-HAMi/ascend-device-plugin/commit/531a70a863ab0c2a1e65c6bfaee4c96b688a9af2 changed it and mark as **fix**, I'm not sure whether it is behavior change according to driver update.  So I can't provide a config and driver version compatibility matrix.

DSFans2014 (commented):
/lgtm

archlitchi (approved):
/lgtm

PR #2224, opened 2026-07-31
fix(util): re-evaluate lock state in optimistic locking retry loop

Norway-02:
fix(util): re-evaluate lock state in optimistic locking retry loop

## What happened
The HAMi scheduler and device plugins rely on `pkg/util/nodelock` to implement a distributed node lock using Kubernetes node annotations (`hami.io/mutex.lock`).

Currently, `SetNodeLock` and `ReleaseNodeLock` correctly use an optimistic locking pattern via `retry.OnError`, but the condition checking occurs *outside* the retry function. Inside the `retry.OnError` block, the code blindly fetches the new `resourceVersion` and patches the node without evaluating whether the lock is still available (or still owned by the caller).

This leads to a race condition where concurrent scheduler instances blindly overwrite each other's locks upon a patch conflict.

The `retry.OnError` loop should explicitly re-evaluate the lock state after fetching the node (inside the loop). If the node is already locked by another process (for `SetNodeLock`) or the lock is no longer owned by the current pod (for `ReleaseNodeLock`), the retry loop should abort rather than blindly applying the patch.

1. Run two instances of `SetNodeLock` concurrently for the same node.
2. The first instance patches the node successfully.
3. The second instance fails with a 409 Conflict.
4. The second instance's `retry.OnError` triggers, fetches the node, and unceremoniously overwrites the first instance's lock because it fails to check `node.Annotations[NodeLockKey]` inside the retry block.

## Environment
* HAMi version: master

--
In a perfect world, AI assistance would produce equal or higher quality
work than any human. That isn't the world we live in today, and in many cases
AI-generated code can contain subtle bugs or not adhere to project-specific
best practices. I say this despite being a fan of and using them successfully
myself (with heavy supervision)!

When using AI assistance, we expect contributors to understand the code
that is produced and be able to answer critical questions about it. It
isn't a maintainer's job to review a PR so broken that it requires
significant rework to be acceptable.

Please be respectful to maintainers and disclose AI assistance.

**AI Assistance Disclosure:** This PR was identified and written with the assistance of an AI agent. The AI was used to identify the concurrency race condition within the optimistic locking retry logic, write the fix, and draft this PR description. The solution was fully reviewed, vetted, and manually verified.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved node-lock acquisition when ownership changes during retries.
  * Prevented lock releases from removing another pod’s lock.
  * Treats safely aborted releases as successful no-ops.
  * Added handling for lock contention without unnecessary retries.

### DIFF
--- a/pkg/util/nodelock/nodelock.go
+++ b/pkg/util/nodelock/nodelock.go
@@ -45,6 +45,9 @@ const (
 // valid pod. Callers may retry this error when the caller is a PodGroup member.
 var ErrNodeLockContention = errors.New("node lock contention")
 
+// ErrNodeLockAborted indicates the lock operation was safely aborted to avoid blindly overwriting state.
+var ErrNodeLockAborted = errors.New("node lock operation aborted")
+
 func IsNodeLockContention(err error) bool {
 	return errors.Is(err, ErrNodeLockContention)
 }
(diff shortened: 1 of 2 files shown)


Norway-02:
@DSFans2014 Hi! I've opened this PR to fix a critical concurrency race condition in the distributed node locking mechanism (`pkg/util/nodelock`). The `retry.OnError` loop was previously bypassing lock state validation upon a patch conflict, which could cause concurrent schedulers to blindly overwrite each other's locks. I've added the lock state check inside the retry block to ensure robust optimistic locking. Could you take a look when you have a chance? Thank you!

mesutoezdil (commented):

Norway-02:
@mesutoezdil Thanks for the review! All feedback has been addressed:
1. Ran `gofmt` to fix the trailing whitespace.
2. Added a test in `nodelock_test.go` covering the new retry paths for both Set and Release node locks.
3. Formatted the AI assistance disclosure exactly as requested in `CONTRIBUTING.md` within the PR description block.
4. Added `ErrNodeLockAborted` as a sentinel error instead of abusing `context.Canceled`.
5. Fixed `SetNodeLock` to return `nil` if the lock is already owned by this pod to handle lost responses.

PTAL when you have a moment. Thanks!

mesutoezdil:
closing per the comparison above, #2197 is the earlier and more complete base. pls check existing prs before opening a new one.

PR #2236, opened 2026-07-31
ci: move release workflow write permissions to job level

mesutoezdil:
/kind cleanup
Scorecard flags any top-level write permission in a workflow (TokenPermissionsID). The three call-release workflows declared contents/packages/id-token writes at the top level; this scopes them to the single job that needs them and keeps the top level read-only. Behavior is unchanged since each workflow has exactly one job.

Resolves code scanning alerts:
- https://github.com/Project-HAMi/HAMi/security/code-scanning/1165
- https://github.com/Project-HAMi/HAMi/security/code-scanning/1166
- https://github.com/Project-HAMi/HAMi/security/code-scanning/1167

Same pattern as #2192. The alerts will close on the next scheduled Scorecard run.

### DIFF
--- a/.github/workflows/call-release-image-hamicore.yaml
+++ b/.github/workflows/call-release-image-hamicore.yaml
@@ -21,15 +21,17 @@ on:
       tagoverride:
         required: false
         type: string
-# id-token is needed for keyless cosign signing
 permissions:
   contents: read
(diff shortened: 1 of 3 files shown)


Shouren (approved):
/lgtm

PR #2238, opened 2026-07-31
fix(device): correct AllocatedCardsInsufficientRequest count during partial allocation

### LINKED ISSUE #2237: [bug] Incorrect AllocatedCardsInsufficientRequest count in device Fit failure reason
What happened:  While looking through the Fit() implementations, I noticed that the failure reason records len(tmpDevs) instead of len(tmpDevs[k.Type]).

if len(tmpDevs) > 0 {
    reason[common.AllocatedCardsInsufficientRequest] = len(tmpDevs)
    klog.V(5).InfoS(common.AllocatedCardsInsufficientRequest, "pod", klog.KObj(pod), "request", originReq, "allocated", len(tmpDevs))
}

Since tmpDevs is a map[string]device.ContainerDevices, len(tmpDevs) returns the number of map entries rather than the number of allocated devices.For example, if a request asks for 3 devices but only 2 matching devices are available, tmpDevs[k.Type] contains 2 allocated devices, but the generated failure reason reports an allocated count of 1.
From tracing the code path, this seems to only affect the failure reason returned by Fit() and the corresponding scheduler logs/events. The scheduling decision itself does not appear to change.

What you expected to happen: The allocated count reported in AllocatedCardsInsufficientRequest should match the number of devices stored in tmpDevs[k.Type]. In the example above, it should report 2 instead of 1.

How to reproduce it (as minimally and precisely as possible):
1. Create a node with 2 devices of the same type.
2. Create a request for 3 devices.
3. Call the corresponding backend's Fit() method.
4. tmpDevs[k.Type] contains 2 allocated devices, but the returned reason reports an allocated count of 1.

Anything else we need to know? : I found the same implementation in the following files:

- pkg/device/nvidia/device.go
- pkg/device/amd/device.go
- pkg/device/ascend/device.go
- pkg/device/cambricon/device.go
- pkg/device/hygon/device.go
- pkg/device/iluvatar/device.go
- pkg/device/metax/device.go
- pkg/device/mthreads/device.go
- pkg/device/vastai/device.go
- pkg/device/biren/device.go

I also noticed that pkg/device/enflame/device.go already uses len(tmpDevs[k.Type]) for the same failure reason, and pkg/device/metax/sdevice.go similarly records the number of allocated candidate devices instead of the number of map entries.

Replacing len(tmpDevs) with len(tmpDevs[k.Type]) (and updating the corresponding log statement) appears to produce the expected allocated count.

Environment:
- HAMi version: Latest master branch (v2.9.0+)
- Go version: 1.24+

princexpoddar:
What type of PR is this?
/kind bug

In several device Fit() implementations, the logic for recording common.AllocatedCardsInsufficientRequest used len(tmpDevs) instead of len(tmpDevs[k.Type]).

Since tmpDevs is a map[string]device.ContainerDevices, len(tmpDevs) returns the number of map entries (which is 1) rather than the number of allocated devices in tmpDevs[k.Type].For example, if a container requests 3 GPUs on a node with 2 available GPUs, tmpDevs[k.Type] contains 2 allocated devices, but the reported failure reason is 1/2 AllocatedCardsInsufficientRequest instead of 2/2 AllocatedCardsInsufficientRequest.

This PR updates the allocated count to use len(tmpDevs[k.Type]) across the affected 10 device drivers and adds regression unit tests covering the partial allocation scenario.

AI Assistance : Consulted Claude for codebase inspection and verification.

Which issue(s) this PR fixes:
Fixes #2237

- Verified with regression unit tests in pkg/device/amd/device_test.go and pkg/device/nvidia/device_test.go.
- Note that pkg/device/enflame/device.go and pkg/device/metax/sdevice.go already use this pattern.

No

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved device allocation failure reporting when multiple device types are available.
  * Insufficient-allocation counts and messages now accurately reflect only the requested device type.
  * Prevented misleading allocation results when a request exceeds available devices of that type.

* **Tests**
  * Added coverage for partial allocation requests across supported device types, including accurate failure reasons and allocation counts.


mesutoezdil (commented):
one miss: pkg/device/enflame/gcu.go line 183 has the same wrong len(tmpDevs) pattern but is not in this sweep, pls fix it too.

pkg/device/amd/device_test.go
```diff
@@ -551,6 +551,40 @@ func TestDevices_Fit(t *testing.T) {
 			wantDevIDs: []string{},
 			wantReason: "1/1 ExclusiveDeviceAllocateConflict",
 		},
+		{
```
mesutoezdil, line 554: pls replicate it for the other touched drivers that already have a fit test table so the sweep is covered everywhere

princexpoddar:
@mesutoezdil Thanks for the feedback! 

I have:
1. Fixed the missing GCU driver at pkg/device/enflame/gcu.go to use len(tmpDevs[k.Type]).
2. Replicated the multi-card partial allocation regression test case (requesting 3 devices on a 2-device node) across all other touched drivers containing a Fit test table:
   - pkg/device/ascend/device_test.go
   - pkg/device/biren/device_test.go
   - pkg/device/cambricon/device_test.go
   - pkg/device/enflame/gcu_test.go
   - pkg/device/hygon/device_test.go
   - pkg/device/mthreads/device_test.go
   - pkg/device/vastai/device_test.go

I've pushed the updates to the branch and verified that the entire unit test suite passes cleanly!

princexpoddar:
Additionally, I have also updated the `NumaNotFit` logic as suggested by CodeRabbit to use `len(tmpDevs[k.Type])` instead of `len(tmpDevs)` across all 7 affected drivers (`nvidia`, `ascend`, `cambricon`, `hygon`, `iluvatar`, `metax`, `mthreads`). 

All changes have been successfully pushed and verified!

archlitchi (approved):
/lgtm

PR #2245, opened 2026-07-31
fix(device,plugin): stop MIG usage corruption and fd leak

adity1raut:
**What type of PR is this?**

/kind bug

Two independent bugs found while auditing MIG resource accounting in `pkg/device/nvidia/device.go` and `pkg/device-plugin/nvidiadevice/nvinternal/plugin/lock.go`:

1. **Silent MIG usage-accounting corruption for percentage-based memory requests.**
   `CustomFilterRule` checked MIG template/slot sizes against the raw, unresolved `request.Memreq`. For percentage-based memory requests (`MemPercentagereq` set, `Memreq == 0`), that comparison is always trivially true, so the filter could admit a device even when no MIG slot is actually big enough for the real (percentage-resolved) request.
   `AddResourceUsage`'s "fresh template" branch had no guard for the case where no template fits — unlike its sibling ("reuse existing slot") branch, which already returns an error in that case. So if no template fit, the function silently fell through and still incremented `Usedmem`/`Usedcores`, corrupting the node's device usage accounting and leaving `ContainerDevice.UUID` without a valid MIG index suffix.

   Fix: resolve the actual memory request before calling `CustomFilterRule` (instead of passing the still-zero `Memreq`), and add the same `found`-guard to `AddResourceUsage`'s fresh-template branch that the sibling branch already has.

2. **File descriptor leak in `createMigApplyLock`.**
   `os.Create(file)` returns an open `*os.File` that was discarded without calling `.Close()`. This function runs on every `ApplyMigTemplate()` call (via `DisableOtherNVMLOperation`), so every MIG reconfiguration on a node leaked one fd.

   Fix: close the file handle after creating it.

Added regression tests:
- `TestAddResourceUsage_MigResetNoFit` — verifies `AddResourceUsage` now returns an error (and doesn't touch usage counters) when no MIG template fits during a fresh template selection.
- `TestFit_MigPercentageRequestRejectsUndersizedTemplate` — verifies `Fit()` now correctly rejects a MIG device whose only template is too small for a percentage-based memory request (this previously passed incorrectly).

Fixes #

Both fixes are minimal and mirror existing patterns already in the file (the `found`-guard mirrors the sibling branch in `AddResourceUsage`; the fd close follows normal `os.Create` usage elsewhere in the codebase).

```release-note
Fix MIG device usage accounting corruption for percentage-based memory requests, and fix a file descriptor leak in the MIG apply lock file creation.
```

**AI Assistance Disclosure:** I used Geminy to help investigate the codebase for bugs, diagnose the root cause of both issues, and draft the fix and regression tests. I reviewed, understood, and verified all AI-generated changes (build, tests, lint) before submitting.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved MIG resource allocation to report errors when no suitable template is available.
  * Prevented failed allocations from changing usage records or assigning invalid MIG identifiers.
  * Corrected memory filtering for percentage-based resource requests.
  * Improved lock handling so file-closing errors are reported correctly.

* **Tests**
  * Added coverage for failed MIG allocations, percentage-based memory requests, and quota configurations.

### DIFF
--- a/pkg/device/nvidia/device.go
+++ b/pkg/device/nvidia/device.go
@@ -678,7 +678,6 @@ func (dev *NvidiaGPUDevices) migNeedsReset(n *device.DeviceUsage) bool {
 }
 
 func (dev *NvidiaGPUDevices) AddResourceUsage(pod *corev1.Pod, n *device.DeviceUsage, ctr *device.ContainerDevice) error {
-	n.Used++
 	if n.Mode == MigMode {
 		if dev.migNeedsReset(n) {
 			found := false
@@ -728,6 +727,7 @@ func (dev *NvidiaGPUDevices) AddResourceUsage(pod *corev1.Pod, n *device.DeviceU
 			}
 		}
 	}
+	n.Used++
 	n.Usedcores += ctr.Usedcores
 	n.Usedmem += ctr.Usedmem
 	return nil
--- a/pkg/device/nvidia/device.go
+++ b/pkg/device/nvidia/device.go
@@ -681,6 +681,7 @@ func (dev *NvidiaGPUDevices) AddResourceUsage(pod *corev1.Pod, n *device.DeviceU
 	n.Used++
 	if n.Mode == MigMode {
 		if dev.migNeedsReset(n) {
+			found := false
 		OuterLoop:
 			for tidx, templates := range n.MigTemplate {
 				for idx, template := range templates {
@@ -700,9 +701,13 @@ func (dev *NvidiaGPUDevices) AddResourceUsage(pod *corev1.Pod, n *device.DeviceU
 						}
 						n.MigUsage.Index = int32(tidx)
 						n.MigUsage.UsageList[usageListIdx].InUse = true
+						found = true
 						break OuterLoop
 					}
 				}
+			}
+			if !found {
+				return errors.New("mig template allocate resource fail")
 			}
 		} else {
 			found := false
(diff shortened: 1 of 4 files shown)


pkg/device/nvidia/device.go
```diff
@@ -694,10 +695,14 @@ func (dev *NvidiaGPUDevices) AddResourceUsage(pod *corev1.Pod, n *device.DeviceU
 						}
 						n.MigUsage.Index = int32(tidx)
 						n.MigUsage.UsageList[usageListIdx].InUse = true
+						found = true
 						break OuterLoop
 					}
 				}
 			}
+			if !found {
+				return errors.New("mig template allocate resource fail")
```
mesutoezdil, line 709: n.Used++ already ran above so this error path leaves it bumped, the sibling branch has the same problem, maybe fix both while u are here

mesutoezdil (commented):

adity1raut:
@archlitchi  PTAL This Pr ready to Merged

pkg/device/nvidia/device.go
```diff
@@ -833,7 +838,14 @@ func (nv *NvidiaGPUDevices) Fit(devices []*device.DeviceUsage, request device.Co
 			klog.V(5).InfoS(common.CardComputeUnitsExhausted, "pod", klog.KObj(pod), "device", dev.ID, "device index", i)
 			continue
 		}
-		if !nv.CustomFilterRule(allocated, request, tmpDevs[k.Type], dev) {
+		// CustomFilterRule must see the resolved memory request: for
```
archlitchi: remove these comments here and comments in device_test file

archlitchi (approved):
/lgtm

PR #2234, opened 2026-07-31
refactor: replace NVML util panics with error returns

shellyco-code:
**What type of PR is this?**

/kind bug

**What this PR does / why we need it**:
This PR fixes a critical node-level reliability issue where the HAMi device plugin crashes via `panic(0)` if it encounters any NVML query failures (e.g., driver glitches) in utility functions.

Instead of panicking and crashing the plugin, `GetIndexAndTypeFromUUID` and `GetMigUUIDFromIndex` now return an `error`. The caller (`GetContainerDeviceStrArray`) logs the error via `klog.Errorf` and safely skips the faulty device, allowing the plugin to continue operating with the remaining healthy devices on the node.

**Which issue(s) this PR fixes**:
Fixes #2230

**Special notes for your reviewer**:
- Replaced all `panic(0)` calls with `fmt.Errorf()` returns and handled them at call sites.
- I have added DCO sign-off (`Signed-off-by`) to the commit.

**Does this PR introduce a user-facing change?**:
```release-note
Fix: the device plugin will no longer crash entirely if it encounters NVML utility failures while fetching device information.

PR #2246, opened 2026-07-31
fix(plugin): guard mig smi output parsing against malformed lines

AdityaShome:
**What type of PR is this?**

/kind bug

`GetMigUUIDFromSmiOutput` (nvidia-smi fallback for MIG UUID lookup) indexed into `strings.Split` results with no bounds checking, so any MIG-block line not matching the exact expected shape caused an `index out of range` panic instead of being skipped turning a recoverable NVML error into a device plugin crash. Same general pattern as the scheduler-side fix in #2088 (unguarded indexing while parsing a MIG index/UUID string) a different component and code path, not otherwise related.

This PR bounds-checks both splits and skips the malformed line instead of indexing blindly.

Not raised issue as it is a small fix.

- Reproduced the panic first (crafted `nvidia-smi -L` input missing the expected `"Device"` keyword), then fixed it. Regression tests added covering the happy path and both previously-panicking malformed cases.
- Not validated on physical MIG hardware, verified via unit tests reproducing the malformed-output panic and its fix, plus `go build`/`go vet`/`go test -race`. Diff: 10+/4- (fix) + 40+ (tests).

```release-note
Fix a panic in the NVIDIA device plugin's nvidia-smi-based MIG UUID fallback lookup when the parsed output line doesn't match the expected format.
```

---

**AI assistance disclosure**: AI assistance was used to help trace the fallback path, reproduce the panic and suggest an initial fix. I manually investigated the issue, validated the root cause, refined and implemented the solution, verified the behavior, added the tests and reviewed all changes before opening this PR.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of malformed GPU management output during MIG UUID lookup.
  - Invalid device indices, incomplete lines, and entries without valid MIG UUIDs are now safely skipped.
  - Valid device entries continue to be processed as expected, reducing errors in GPU device detection.

- **Tests**
  - Added coverage for successful MIG UUID lookup, malformed device lines, non-numeric indices, and incomplete output.

### DIFF
--- a/pkg/device-plugin/nvidiadevice/nvinternal/plugin/util.go
+++ b/pkg/device-plugin/nvidiadevice/nvinternal/plugin/util.go
@@ -188,7 +188,7 @@ func GetMigUUIDFromSmiOutput(output string, uuid string, idx int) string {
 	return ""
 }
 
-func GetMigUUIDFromIndex(uuid string, idx int) string {
+func GetMigUUIDFromIndex(uuid string, idx int) (string, error) {
 	defer nvml.Shutdown()
 	if nvret := nvml.Init(); nvret != nvml.SUCCESS {
 		klog.Errorln("nvml Init err: ", nvret)
@@ -212,15 +212,18 @@ func GetMigUUIDFromIndex(uuid string, idx int) string {
 			klog.Fatalf("nvidia-smi -L failed with %s\n", err)
 		}
 		outStr := stdout.String()
-		uuid := GetMigUUIDFromSmiOutput(outStr, originuuid, idx)
-		return uuid
+		migUUID := GetMigUUIDFromSmiOutput(outStr, originuuid, idx)
+		if migUUID == "" {
+			return "", fmt.Errorf("failed to resolve MIG UUID for device %s at index %d via nvidia-smi fallback", originuuid, idx)
+		}
+		return migUUID, nil
 	}
 	res, ret := migdev.GetUUID()
 	if ret != nvml.SUCCESS {
 		klog.Error(`nvml get mig uuid error ret=`, ret)
 		panic(0)
 	}
-	return res
+	return res, nil
 }
 
 func GetMigGpuInstanceIdFromIndex(uuid string, idx int) (int, error) {
@@ -467,7 +470,7 @@ func deepCopyMigConfig(src nvidia.MigConfigSpec) nvidia.MigConfigSpec {
 	return dst
 }
 
-func (nv *NvidiaDevicePlugin) GetContainerDeviceStrArray(c device.ContainerDevices) []string {
+func (nv *NvidiaDevicePlugin) GetContainerDeviceStrArray(c device.ContainerDevices) ([]string, error) {
 	tmp := []string{}
 	needsreset := false
 	position := 0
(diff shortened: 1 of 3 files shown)


AdityaShome:
1. Non numeric device index now logs and skips the line instead of calling `klog.Fatal` and killing the process.
2. Returned value now requires a UUID field and MIG prefix, so a line without a real UUID is skipped instead of returned as is.

AdityaShome:
Last 3 commits are just adding signoff by rebase.

pkg/device-plugin/nvidiadevice/nvinternal/plugin/util.go
```diff
@@ -161,17 +161,27 @@ func GetMigUUIDFromSmiOutput(output string, uuid string, idx int) string {
...
+			colonParts := strings.Split(val, ":")
+			if len(colonParts) < 3 || !strings.Contains(colonParts[1], "UUID") {
+				continue
+			}
+			outputStr := strings.TrimSpace(colonParts[2])
 			outputStr = strings.TrimRight(outputStr, ")")
+			if !strings.HasPrefix(outputStr, "MIG-") {
+				continue
+			}
 			return outputStr
```
mesutoezdil, line 185: the caller in GetContainerDeviceStrArray still appends the result blindly even when this returns empty, maybe handle that there as a follow up

mesutoezdil (commented):
pls rename the title to the conventional format, sth like "fix(plugin): guard mig smi output parsing against malformed lines".

AdityaShome:
It now skips appending and logs an error when the MIG UUID lookup returns empty.

mesutoezdil:
sorry, my "follow up" nit was not meant as skip and continue, thats the same silent under allocation we rejected in #2234. pls make GetMigUUIDFromIndex's empty result an error that fails the allocate instead, or drop the caller change from this pr and keep the parser fix only.

AdityaShome:
Made `GetMigUUIDFromIndex` and `GetContainerDeviceStrArray` return an error on unresolved MIG UUID instead of skipping, so Allocate() now fails hard instead of under allocating, validated on real GPU hardware (RTX 3050, driver 592.82, MIG unsupported), confirming the error path triggers and fails the request as expected.

AdityaShome:
A small follow up, so it does not go stale.

mesutoezdil:
ok error propagation looks good now, getmiguuidfromindex returns err, getcontainerdevicestrarray passes it up, allocate fails hard instead of quietly under allocating. nice that u tested on real hw too.

only thing left, pr conflicts w master now. #2378 refactored mig alloc in the same files a few days ago. pls rebase and make sure ur fix still applies clean on top.

pkg/device-plugin/nvidiadevice/nvinternal/plugin/util.go
```diff
@@ -161,24 +161,34 @@ func GetMigUUIDFromSmiOutput(output string, uuid string, idx int) string {
 			continue
 		}
 		klog.Infoln("inspecting", val)
-		num := strings.Split(val, "Device")[1]
-		num = strings.Split(num, ":")[0]
+		deviceParts := strings.Split(val, "Device")
```
mesutoezdil, line 164: getmiguuidfromsmioutput and getmiguuidfromindex are gone in master now. #2378 replaced this whole nvidia-smi fallback with nvml miginstancemanager. pls check if the bug still exists in new code.

AdityaShome: Great work on the new MIG design overall, while going through the code I found two smaller issues:

1. gpuUUIDToIndex and nvmlBusyGPUs (mig_startup.go) call nvml.Init() with no matching nvml.Shutdown() unlike every other NVML entry point in the codebase. gpuUUIDToIndex runs on every Allocate() and every 5s via the reconciler, so this leaks indefinitely on nodes with active MIG allocations.
2. migmgr.go's MigInstanceManager never calls nvml.Init() itself it silently relies on nvmlBusyGPUs() having already leaked one at startup. Fixing 1 by adding a proper Shutdown() would break this.

Planning two follow-up PRs for these.

mesutoezdil:

Issue #2243, opened 2026-07-31
[Bug] Multi device pods fail to bind because backends acquire the same node lock


iemAnshuman:
## What happened?

While following the multi-device scheduling path added around #2105, I noticed that a pod requesting two device types can successfully pass filtering but then fail during Bind.

`lockAllDevices` calls `LockNode` for every registered backend. Several backends use the common `pkg/util/nodelock` helper, which always operates on the same annotation:

```text
hami.io/mutex.lock
```

This means a pod requesting two of those device types effectively does this:

```text
backend A: acquire lock for pod X -> success
backend B: acquire lock for pod X -> node lock contention
```

The second call sees the lock created by the first call, but `LockNode` does not recognize that it belongs to the same pod. Bind then fails and releases the lock. The next scheduling attempt repeats the same sequence.

I reproduced this with two mock device backends using the shared node-lock helper. No concurrency or device hardware is required.

The failure is:

```text
node node-1 has been locked within 5m0s: node lock contention
```

Calling `LockNode` twice directly for the same pod produces the same result.

## Expected behavior

A pod should not contend with a lock acquired earlier in its own Bind operation.

This seems especially relevant after #2105, which added coverage for a container requesting multiple device types. That test exercises device selection, but its mock backends do not exercise the real node-lock path.

## Is this covered by #2197 or #2224?

I don't think so.

Those PRs handle ownership changes during API conflict retries. This case is sequential and uses the same owner, so it happens without a conflict or a second scheduler process:

```text
LockNode(pod X)
LockNode(pod X)
```

## Fix consideration

Making same-pod acquisition idempotent would prevent the immediate Bind failure, but release probably needs some thought as well.

If two device plugins share one lock and complete independently, the first plugin to finish must not release the lock while the second plugin is still allocating the same pod.

So I think the intended lifecycle needs to be clarified:

* acquire the shared lock once for the whole pod and release it after all requested device types finish; or
* give each backend a genuinely separate lock that matches what its device plugin uses.

I’m happy to work on a regression test and fix once the intended ownership/release model is agreed.

## Environment

* HAMi current master
* Reproduced with Kubernetes fake clients
* No accelerator hardware required

AI assistance was used while tracing the code and drafting the initial test. I reviewed and reproduced the behavior myself.

PR #2252, opened 2026-08-01
fix(cambricon): replace hand-rolled node lock with nodelock delegation

### LINKED ISSUE #2251: Cambricon ReleaseNodeLock can return nil without removing the node lock
**What happened**:

`ReleaseNodeLock` in the Cambricon backend can return `nil` without removing `cambricon.com/dsmlu.lock`, so the node stays unschedulable for MLU until the 2 minute expiry in `LockNode`. Three ways it goes wrong, all in `pkg/device/cambricon/device.go`.

`setNodeLock` patches the annotation onto the apiserver but never puts it back on the caller's `*corev1.Node`. So when `Bind` fails and `releaseAllDevices` runs with that same object, `ReleaseNodeLock` looks for `DsmluLockTime` at line 159, doesn't find it, logs `Node lock not set` and returns nil. Lock taken, release says it worked, annotation still there. No concurrency needed for this one.

The retry loop at 164 to 174 builds `newNode` once and sends the same object six times. Same `resourceVersion` every attempt, so retrying can't clear a 409. Backoff is `rand.Intn(i+1)*10ms`, so the whole loop is over in about 100ms.

Line 175 deletes from `n.Annotations` on the caller's node. In `Bind` that node comes from `s.nodeLister.Get()` at `scheduler.go:845`, which is the shared informer object with no deep copy. `-race` catches it against a `Filter` reading the same map.

The annotation comes off, or the release returns an error.

Found it reading the Bind path, not on a real MLU cluster, so no cluster logs. Reproduces in unit tests with a fake clientset, can post those. The first one is small: call `LockNode`, then `ReleaseNodeLock` with the same node, then read the annotation back off the apiserver.

Cambricon is the only backend that hand rolls this. The others that take a node lock use `pkg/util/nodelock`, which gets the node first and merge patches with the fresh `resourceVersion`. Reading the node from the apiserver instead of trusting the one passed in would close all three.

Fix it in place, or have Cambricon call `nodelock` like the rest? The second picks up whatever owner checks land from #2197 and #2224, so I'd rather ask first.

manmathbh:
**What type of PR is this?**

/kind bug

Replaced Cambricon's hand-rolled node lock with pkg/util/nodelock. Every
other backend already delegates to nodelock — Cambricon was the only one
rolling its own.

Three bugs in the old code:

- setNodeLock patched the apiserver but did not write back to the caller's
  Node. ReleaseNodeLock read the stale in-memory object, found no lock
  annotation, logged "Node lock not set", returned nil. Lock stayed.

- Retry loop sent the same object every attempt. Same resourceVersion,
  no backoff. Could not clear 409 conflicts.

- delete(n.Annotations, DsmluLockTime) on the shared informer's Node
  object, no deep copy. Race with Filter reading the same map.

Fixes #2251

Old DsmluLockTime (cambricon.com/dsmlu.lock) replaced with the shared
hami.io/mutex.lock annotation. Any stale lock from a previous version
expires via the 5-minute nodelock timeout.

Cambricon MLU node locking uses the shared nodelock infrastructure.
Old lock annotations expire naturally (5-minute TTL).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability and consistency of Cambricon node locking and unlocking.
  * Updated lock handling to use the shared node-locking mechanism.
  * Prevented unnecessary locks for pods without MLU resource requests.
  * Removed obsolete legacy lock annotations during lock processing.

* **Tests**
  * Expanded coverage for lock creation, release, annotation handling, and non-MLU pod behavior.

### DIFF
--- a/pkg/device/cambricon/device.go
+++ b/pkg/device/cambricon/device.go
@@ -17,24 +17,18 @@ limitations under the License.
 package cambricon
 
 import (
-	"context"
-	"encoding/json"
 	"flag"
 	"fmt"
-	"math/rand"
 	"slices"
 	"strings"
-	"time"
 
 	"github.com/Project-HAMi/HAMi/pkg/device"
 	"github.com/Project-HAMi/HAMi/pkg/device/common"
 	"github.com/Project-HAMi/HAMi/pkg/util"
-	"github.com/Project-HAMi/HAMi/pkg/util/client"
+	"github.com/Project-HAMi/HAMi/pkg/util/nodelock"
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/klog/v2"
 )
 
@@ -46,14 +40,11 @@ const (
 	MluMemSplitEnable      = "CAMBRICON_SPLIT_ENABLE"
 	MLUInUse               = "cambricon.com/use-mlutype"
 	MLUNoUse               = "cambricon.com/nouse-mlutype"
-	// MLUUseUUID annotation specifies a comma-separated list of MLU UUIDs to use.
-	MLUUseUUID = "cambricon.com/use-gpuuuid"
-	// MLUNoUseUUID annotation specifies a comma-separated list of MLU UUIDs to exclude.
-	MLUNoUseUUID          = "cambricon.com/nouse-gpuuuid"
-	DsmluLockTime         = "cambricon.com/dsmlu.lock"
+	MLUUseUUID             = "cambricon.com/use-gpuuuid"
+	MLUNoUseUUID           = "cambricon.com/nouse-gpuuuid"
 	DsmluProfile          = "CAMBRICON_DSMLU_PROFILE"
 	DsmluResourceAssigned = "CAMBRICON_DSMLU_ASSIGNED"
-	retry                 = 5
+	NodeLockMLU           = "hami.io/mutex.lock"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


pkg/device/cambricon/device.go
```diff
@@ -46,14 +40,11 @@ const (
...
-	MLUUseUUID = "cambricon.com/use-gpuuuid"
-	// MLUNoUseUUID annotation specifies a comma-separated list of MLU UUIDs to exclude.
-	MLUNoUseUUID          = "cambricon.com/nouse-gpuuuid"
-	DsmluLockTime         = "cambricon.com/dsmlu.lock"
+	MLUUseUUID             = "cambricon.com/use-gpuuuid"
+	MLUNoUseUUID           = "cambricon.com/nouse-gpuuuid"
 	DsmluProfile          = "CAMBRICON_DSMLU_PROFILE"
 	DsmluResourceAssigned = "CAMBRICON_DSMLU_ASSIGNED"
-	retry                 = 5
+	NodeLockMLU           = "hami.io/mutex.lock"
```
mesutoezdil: moving mlu onto the shared lock pulls it into the #2243 multi device pod problem (mixed mlu + other vendor pods now contend on the same annotation), are u aware of that tradeoff? pls drop a note in #2243 so the central fix covers this too

mesutoezdil (commented):

manmathbh:
> no ai assistance disclosure is present. if any ai tool was used, it must be disclosed per CONTRIBUTING.md: https://github.com/Project-HAMi/HAMi/blob/master/CONTRIBUTING.md#ai-assistance-notice

Used Copilot for autocomplete while writing the change, no generated blocks.
Read the issue, read nodelock.go, matched the NVIDIA pattern by hand.

pkg/device/cambricon/device.go
manmathbh: Aware, but the old cambricon.com/dsmlu.lock was already a single
annotation on the whole node, so the contention surface didn't change.
It was just invisible because it wasn't sharing the key with anyone.
Now at least the central fix in #2243 covers it. Happy to leave a note
there.

iemAnshuman:
@manmathbh the body says an old cambricon.com/dsmlu.lock times out after 5 minutes, but that timeout belongs to hami.io/mutex.lock. nothing reads or deletes dsmlu.lock once this lands, so whatever is sitting on a node at upgrade time stays there for good. does cambricon's device plugin read that key? the cambricon.com/ prefix and the CAMBRICON_DSMLU_* envs right below it make me think it might, but I havent looked at the plugin side. would also wait for #2255 before this merges, otherwise a pod asking for mlu and nvidia takes the lock on the first call and then blocks on it on the second

manmathbh:
@iemAnshuman @mesutoezdil Thanks for the catches! I've pushed a commit addressing the feedback:

Added cleanupLegacyLock() to strip the orphaned dsmlu.lock via JSON patch so nodes don't stay locked after an upgrade.
Applied the CodeRabbit refactors (hasMLURequest() helper, removed duplicate constants, fixed test assertions).
Totally agree on waiting for #2255 to merge first so we don't deadlock mixed-resource pods. Let me know if the cleanup logic looks good!

iemAnshuman:
> @iemAnshuman @mesutoezdil Thanks for the catches! I've pushed a commit addressing the feedback:
> 
> Added cleanupLegacyLock() to strip the orphaned dsmlu.lock via JSON patch so nodes don't stay locked after an upgrade. Applied the CodeRabbit refactors (hasMLURequest() helper, removed duplicate constants, fixed test assertions). Totally agree on waiting for #2255 to merge first so we don't deadlock mixed-resource pods. Let me know if the cleanup logic looks good!

also I checked the current Cambricon device plugin code and found a separate lifecycle mismatch. in DynamicSmlu mode, its allocation, startup, and pod cleanup paths still remove only cambricon.com/dsmlu.lock; they never release hami.io/mutex.lock. HAMi doesnt release the node lock after a successful Bind, so a Cambricon only DynamicSmlu pod can leave the new shared lock behind and block later accelerator pods on that node until the lock times out at five minutes by default.

could we either keep the legacy key until the plugin supports the shared lock, or land a coordinated, owner safe plugin change first? also a focused test proving that a successful DynamicSmlu allocation releases the exact key written by LockNode() would be useful. The legacy cleanup helps with upgrades, but it doesnt fix this release path mismatch.

manmathbh:
@iemAnshuman You're right, this is a real lifecycle mismatch that can't be fixed solely on the scheduler side. nodelock.SetNodeLock and ReleaseNodeLock both hardcode hami.io/mutex.lock (the lockname parameter is accepted but unused), so there's no way to keep using cambricon.com/dsmlu.lock through the shared infra.

The Cambricon device plugin's DynamicSmlu paths need to release hami.io/mutex.lock after allocation. Since the plugin isn't in this repo, two options:

1. Keep the hand-rolled lock for now, fix the original 3 bugs (stale Node reference, no-backoff retry, informer map mutation) in-place without migrating to nodelock, then switch to shared lock later once the plugin catches up.
2. Coordinate a plugin-side PR first, add hami.io/mutex.lock release to the device plugin's DynamicSmlu cleanup, then merge this.

Preference? I lean toward option 2; it avoids keeping dead code around. Happy to help draft the plugin-side change.

mesutoezdil:
You know the rule from previous prs. 
For future:

PR #2254, opened 2026-08-01
fix(device): guard nil maps in mthreads MutateAdmission

### LINKED ISSUE #2253: [bug]: mthreads MutateAdmission panics on nil maps (nil Annotations / nil Limits)
**What happened**:

`MthreadsDevices.MutateAdmission` (`pkg/device/mthreads/device.go`) writes to
`ctr.Resources.Limits` and `p.Annotations` without checking they are non-nil.
Writing to a nil map panics in Go and crashes the mutating-webhook goroutine.

Two ordinary inputs trigger it:

1. A pod that requests more than one `mthreads.com/vgpu` but has no annotations
   (`p.Annotations == nil`) panics at the `mthreads.com/request-gpu-num` write.
2. A container that sets the vgpu count under `requests` only, with no `limits`
   block (`ctr.Resources.Limits == nil`), panics at the cores/memory write.

Because the webhook runs `MutateAdmission` for every registered device on every
pod (`pkg/scheduler/webhook.go` `Handle`), and `Handle` does not initialize
`pod.Annotations`, this takes down admission for affected mthreads pods.

`MutateAdmission` should initialize the maps before writing (as the metax
backend already does in `pkg/device/metax/sdevice.go`), so multi-GPU mthreads
pods admit normally instead of panicking.

**How to reproduce it**:

Call `MutateAdmission` with either:
- a container whose `Limits` has `mthreads.com/vgpu: 2` and a pod with
  `Annotations == nil`; or
- a container whose `Requests` (not `Limits`) has `mthreads.com/vgpu: 2`.

Both produce `panic: assignment to entry in nil map`.

**Environment**:
- HAMi version: master

imantaba:
## What this PR does

`MthreadsDevices.MutateAdmission` writes to `ctr.Resources.Limits` and
`p.Annotations` without checking they are non-nil. Writing to a nil map panics
in Go and crashes the mutating-webhook goroutine. Two ordinary inputs trigger
it:

- a pod that requests more than one `mthreads.com/vgpu` but carries no
  annotations (`p.Annotations == nil`) panics on the `request-gpu-num` write;
- a container that specifies the vgpu count under `requests` only, with no
  `limits` block (`ctr.Resources.Limits == nil`), panics on the cores/memory
  write.

Because the webhook runs `MutateAdmission` for every registered device on every
pod, this takes down admission for affected mthreads pods.

This initializes `ctr.Resources.Limits` and `p.Annotations` before writing,
matching the nil guard already used in the metax backend
(`pkg/device/metax/sdevice.go`).

## Fixes

Fixes #2253

## Testing

Added two cases to `Test_MutateAdmission` covering both inputs. Each produces
`panic: assignment to entry in nil map` before this change and passes after.
Verified red→green; full package passes under `-race`; gofmt clean.

## AI assistance disclosure

This change was developed with AI assistance (per CONTRIBUTING.md).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved admission handling when resource limits or pod annotations are not initially configured.
  * vGPU requests with multiple devices are now admitted successfully in additional request and limit scenarios.

**Note:** a requests-only extended-resource pod is eventually rejected by API validation, but mutating admission runs before validation, so the nil-`Limits` panic is still reachable.

### DIFF
--- a/pkg/device/mthreads/device.go
+++ b/pkg/device/mthreads/device.go
@@ -97,9 +97,15 @@ func (dev *MthreadsDevices) MutateAdmission(ctr *corev1.Container, p *corev1.Pod
 		if count.Value() <= 0 {
 			return false, fmt.Errorf("%s must be greater than 0", MthreadsResourceCount)
 		}
+		if ctr.Resources.Limits == nil {
+			ctr.Resources.Limits = corev1.ResourceList{}
+		}
 		if count.Value() > 1 {
 			ctr.Resources.Limits[corev1.ResourceName(MthreadsResourceCores)] = *resource.NewQuantity(count.Value()*int64(coresPerMthreadsGPU), resource.DecimalSI)
 			ctr.Resources.Limits[corev1.ResourceName(MthreadsResourceMemory)] = *resource.NewQuantity(count.Value()*int64(memoryPerMthreadsGPU), resource.DecimalSI)
+			if p.Annotations == nil {
+				p.Annotations = make(map[string]string)
+			}
 			p.Annotations["mthreads.com/request-gpu-num"] = fmt.Sprint(count.Value())
 			return ok, nil
 		}
--- a/pkg/device/mthreads/device_test.go
+++ b/pkg/device/mthreads/device_test.go
@@ -103,6 +103,27 @@ func Test_MutateAdmission(t *testing.T) {
 			want: true,
 		},
 		{
+			name: "count one requested without limits",
+			args: struct {
+				ctr *corev1.Container
+				p   *corev1.Pod
+			}{
+				ctr: &corev1.Container{
+					Resources: corev1.ResourceRequirements{
+						Requests: corev1.ResourceList{
+							"mthreads.com/vgpu": *resource.NewQuantity(1, resource.DecimalSI),
+						},
+					},
+				},
+				p: &corev1.Pod{
+					ObjectMeta: metav1.ObjectMeta{
+						Annotations: map[string]string{},
+					},
+				},
+			},
+			want: true,
+		},
+		{
 			name: "don't set to count limit",
 			args: struct {
 				ctr *corev1.Container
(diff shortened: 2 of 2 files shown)


pkg/device/mthreads/device.go
```diff
@@ -97,9 +97,15 @@ func (dev *MthreadsDevices) MutateAdmission(ctr *corev1.Container, p *corev1.Pod
 		if count.Value() <= 0 {
 			return false, fmt.Errorf("%s must be greater than 0", MthreadsResourceCount)
 		}
+		if ctr.Resources.Limits == nil {
```
mesutoezdil, line 100: fyi a requests only extended resource pod gets rejected by api validation later anyway, but mutation runs before validation so the panic is still reachable, worth a word in the desc

pkg/device/mthreads/device_test.go
```diff
@@ -64,6 +64,44 @@ func Test_MutateAdmission(t *testing.T) {
 			},
 			want: true,
 		},
+		{
```
mesutoezdil, line 67: pls also add a count=1 requests only case, the !memok branch below writes to limits too and hit the same panic

mesutoezdil (commented):
pls fix the title scope, files are under pkg/device so it should be fix(device) not fix(device-plugin).

pkg/device/mthreads/device.go
imantaba: A note added to PR description regarding this.

imantaba:
> pls fix the title scope, files are under pkg/device so it should be fix(device) not fix(device-plugin).

Updated

pkg/device/mthreads/device_test.go
imantaba: Pushed in 9c752f9

archlitchi (approved):
/lgtm

archlitchi (approved):
/lgtm

PR #2255, opened 2026-08-01
fix(nodelock): let a pod re-acquire its own still-valid node lock

### LINKED ISSUE #2243: [Bug] Multi device pods fail to bind because backends acquire the same node lock

adity1raut:
/kind bug

`lockAllDevices` (`pkg/scheduler/scheduler.go`) calls `nodelock.LockNode` once per device vendor backend a pod requests resources from (`device.GetDevices()` is iterated per registered vendor), and every vendor writes to the same shared node annotation (`hami.io/mutex.lock`).

`LockNode` had no case for "this exact pod already holds the lock": the dangling-lock check only ran when the lock's namespace/name differed from the requesting pod, so a same-pod, non-expired lock fell straight through to the contention error. Concretely: a pod requesting resources from two or more HAMi-managed vendors in the same spec (e.g. one container asking for `nvidia.com/gpu`, another for `cambricon.com/vmlu`) would have its first `LockNode` call succeed and its second call immediately contend with its own still-valid lock — making that pod permanently unschedulable.

This PR adds an explicit branch: if the existing lock's namespace/name match the calling pod and it hasn't expired, `LockNode` treats it as already acquired instead of erroring.

It also fixes `Test_LockNode`'s "node has been locked" case, which had (accidentally) used the exact same pod identity as both the lock holder and the requester — so it was asserting the buggy contention behavior as correct. That case is now a genuine third-party-contention scenario (locked by a different, still-live pod), and a new dedicated test (`TestLockNodeReentrantSamePod`) reproduces `lockAllDevices`' actual multi-vendor call pattern.

Fixes #2243

Found this while reviewing other parts of the scheduler/monitor code for unrelated PRs; it's a distinct issue from any of my other open PRs.

```release-note
Fix pods requesting resources from multiple HAMi-managed device vendors becoming permanently unschedulable due to the pod's own node lock contending with itself.
```

This PR was written with AI assistance (Claude Code), per the AI-assistance disclosure requirement in CONTRIBUTING.md.

## Summary by CodeRabbit

* **Bug Fixes**
  * Node locks can now be safely reacquired by the same pod without being reported as conflicts.
  * Improved handling of dangling or invalid node locks to avoid unnecessary contention.

* **Tests**
  * Added coverage for same-pod lock reacquisition and additional lock-holder scenarios.
  * Expanded validation for invalid and unlocked node setups.

### DIFF
--- a/pkg/util/nodelock/nodelock.go
+++ b/pkg/util/nodelock/nodelock.go
@@ -233,9 +233,18 @@ func LockNode(nodeName string, lockname string, pods *corev1.Pod) error {
 	if time.Since(lockTime) > NodeLockTimeout {
 		klog.InfoS("Node lock expired", "node", nodeName, "lockTime", lockTime, "timeout", NodeLockTimeout)
 		skipOwnerCheck = true
+	} else if ns == pods.Namespace && previousPodName == pods.Name {
+		// The lock is already held by this exact pod. lockAllDevices calls
+		// LockNode once per device vendor a pod requests resources from, so
+		// a pod requesting resources from two or more vendors (e.g. both
+		// nvidia.com/gpu and cambricon.com/vmlu) would otherwise contend
+		// with its own still-valid lock on the second call and never
+		// become schedulable. Treat this as already acquired.
+		klog.V(4).InfoS("Node lock already held by this pod, treating as acquired", "node", nodeName, "podName", pods.Name)
+		return nil
 	} else
 	// Check dangling nodeLock
-	if ns != "" && previousPodName != "" && (ns != pods.Namespace || previousPodName != pods.Name) {
+	if ns != "" && previousPodName != "" {
 		if _, err := client.GetClient().CoreV1().Pods(ns).Get(ctx, previousPodName, metav1.GetOptions{}); err != nil {
 			if !apierrors.IsNotFound(err) {
 				klog.ErrorS(err, "Failed to get pod of NodeLock", "podName", previousPodName, "namespace", ns)
--- a/pkg/util/nodelock/nodelock_test.go
+++ b/pkg/util/nodelock/nodelock_test.go
@@ -34,7 +34,7 @@ import (
 func Test_LockNode(t *testing.T) {
 	client.KubeClient = fake.NewClientset()
 	type args struct {
-		nodeName func() string
+		nodeName func(t *testing.T) string
 		lockname string
 		pods     *corev1.Pod
 	}
@@ -46,7 +46,7 @@ func Test_LockNode(t *testing.T) {
 		{
 			name: "node not found",
 			args: args{
-				nodeName: func() string {
+				nodeName: func(t *testing.T) string {
 					return "node"
 				},
 				pods: &corev1.Pod{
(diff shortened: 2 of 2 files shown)


mesutoezdil (commented):
pls add "fixes #2243" to the desc so it auto closes.

pkg/util/nodelock/nodelock.go
```diff
@@ -233,9 +233,18 @@ func LockNode(nodeName string, lockname string, pods *corev1.Pod) error {
...
 		skipOwnerCheck = true
+	} else if ns == pods.Namespace && previousPodName == pods.Name {
+		// The lock is already held by this exact pod. lockAllDevices calls
+		// LockNode once per device vendor a pod requests resources from, so
+		// a pod requesting resources from two or more vendors (e.g. both
+		// nvidia.com/gpu and cambricon.com/vmlu) would otherwise contend
+		// with its own still-valid lock on the second call and never
+		// become schedulable. Treat this as already acquired.
+		klog.V(4).InfoS("Node lock already held by this pod, treating as acquired", "node", nodeName, "podName", pods.Name)
+		return nil
```
mesutoezdil, line 244: pls also add a scheduler level test driving lockAllDevices w/ two backends sharing the lock, #2256 has a good one to port

adity1raut: Okk i can fixed it

iemAnshuman:
@adity1raut this handles taking the lock, but giving it back is still a problem. whichever device plugin finishes first deletes hami.io/mutex.lock while the other one is still working on the same pod, and the owner check passes for both since it's the same pod either way. wrote it up in #2243 under fix consideration. happy to take the release side as a separate pr if you want to keep this one to the acquire path.

Shouren (approved):
/lgtm

PR #2256, opened 2026-08-01
fix(util): make same-pod node lock acquisition idempotent

### LINKED ISSUE #2243: [Bug] Multi device pods fail to bind because backends acquire the same node lock

shivv23:
/kind bug

Makes node lock acquisition idempotent for the same pod. lockAllDevices acquires a node lock for every registered device backend, and several backends (nvidia, ascend, hygon, metax, kunlun, etc.) share the same hami.io/mutex.lock annotation through pkg/util/nodelock. When a pod requests two of those device types, the second backend sees the lock created by the first backend, which is held by the *same pod*, but LockNode treats it as contention and fails the Bind.

Fixes #2243

LockNode previously only released and re-acquired the lock when it was expired or dangling (owner pod no longer exists). A lock owned by the same pod fell through to the contention error, so a single Bind that locked the node through multiple backends would fail with 
ode ... has been locked within 5m0s: node lock contention.

The fix treats a lock already held by the same pod as an idempotent re-acquisition and returns success. Release behavior is unchanged: ReleaseNodeLock already skips locks not owned by the calling pod, so the first backend to finish does not remove the shared lock.

Validation:

- go test ./pkg/util/nodelock/... -count=1 -race
- go test ./pkg/scheduler -run Test_Bind_MultiDeviceBackendsSharingNodeLock -count=1 -v (Linux CI; the scheduler package depends on go-nvml which is Linux-only)

No.

**AI assistance disclosure**:

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved node-lock handling so the same pod can safely reacquire its existing lock.
  * Prevented unnecessary lock conflicts when multiple device backends coordinate access to the same node.

* **Tests**
  * Added coverage for shared node locking across device backends.
  * Added validation that locks are correctly associated with pods and removed after release.

### DIFF
--- a/pkg/scheduler/scheduler_test.go
+++ b/pkg/scheduler/scheduler_test.go
@@ -2091,6 +2091,74 @@ func (m *bindLockMockDevice) ReleaseNodeLock(_ *corev1.Node, _ *corev1.Pod) erro
 	return nil
 }
 
+// sharedLockMockDevice mirrors backends (e.g. nvidia, ascend, hygon, metax) that
+// all acquire the same node-lock annotation key "hami.io/mutex.lock" through the
+// shared pkg/util/nodelock helper.
+type sharedLockMockDevice struct {
+	registerMockDevice
+	vendor string
+}
+
+func (m *sharedLockMockDevice) CommonWord() string { return m.vendor }
+func (m *sharedLockMockDevice) LockNode(n *corev1.Node, p *corev1.Pod) error {
+	return nodelockutil.LockNode(n.Name, nodelockutil.NodeLockKey, p)
+}
+func (m *sharedLockMockDevice) ReleaseNodeLock(n *corev1.Node, p *corev1.Pod) error {
+	return nodelockutil.ReleaseNodeLock(n.Name, nodelockutil.NodeLockKey, p, false)
+}
+
+// Test_Bind_MultiDeviceBackendsSharingNodeLock reproduces #2243: a pod requesting
+// two device types whose backends both lock the same node annotation must not
+// contend with itself during Bind.
+func Test_Bind_MultiDeviceBackendsSharingNodeLock(t *testing.T) {
+	pod := &corev1.Pod{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "pod-multi-device", Namespace: "default", UID: types.UID("uid-multi-device"),
+		},
+	}
+	mockA := &sharedLockMockDevice{vendor: "shared-lock-a"}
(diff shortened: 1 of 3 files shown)


mesutoezdil (commented):
the disclosure only covers the desc, pls state whether the code change was ai assisted too.

mesutoezdil:
closing, #2255 was opened earlier and handles the expired own lock case correctly. pls check existing prs before opening a new one.

shivv23:
@mesutoezdil Sure sure. I won't repeat that. Thanks!

PR #2259, opened 2026-08-01
fix(scheduler): handle ResourceQuota tombstones

### LINKED ISSUE #2257: ResourceQuota tombstone deletion leaves stale device limits in scheduler cache
**What happened**:

HAMi's ResourceQuota informer delete handler only accepts a direct `*corev1.ResourceQuota`:

https://github.com/Project-HAMi/HAMi/blob/c7891ded87b71e402a36f5f29ae7d245e0de70b2/pkg/scheduler/scheduler.go#L258-L265

K8s informers may instead pass a `cache.DeletedFinalStateUnknown` tombstone when a watch disconnects and the deletion is discovered during a subsequent relist.

When this happens, `onDelQuota` fails the type assertion, logs:

```text
unknown del object type
```
and returns without calling `QuotaManager.DelQuota`.

The deleted ResourceQuota's device limit therefore remains in HAMi's private quota cache. `FitQuota` continues enforcing that stale limit, so device workloads can be rejected even though the Kubernetes ResourceQuota no longer exists.
HAMi's Pod and Node deletion handlers already unwrap `cache.DeletedFinalStateUnknown`, but the ResourceQuota deletion handler does not:
https://github.com/Project-HAMi/HAMi/blob/c7891ded87b71e402a36f5f29ae7d245e0de70b2/pkg/scheduler/scheduler.go#L174-L231

`onDelQuota` should unwrap a `cache.DeletedFinalStateUnknown` containing a `*corev1.ResourceQuota` and remove the cached device quota just like it does for a normal deletion event.
A tombstone containing an unexpected object type should be logged and ignored safely.

Add the following test as pkg/scheduler/quota_tombstone_test.go:
```go
package scheduler

import (
	"testing"

	"github.com/Project-HAMi/HAMi/pkg/device"
	corev1 "k8s.io/api/core/v1"
	"k8s.io/apimachinery/pkg/api/resource"
	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
	"k8s.io/client-go/tools/cache"
)

func TestResourceQuotaTombstoneClearsCachedLimit(t *testing.T) {
	s := NewScheduler()

	resourceName := device.GetDevices()["NVIDIA"].
		GetResourceNames().ResourceMemoryName
	if resourceName == "" {
		t.Fatal("NVIDIA memory resource name is empty")
	}

	quota := &corev1.ResourceQuota{
		ObjectMeta: metav1.ObjectMeta{
			Name:      "gpu-quota",
			Namespace: "repro",
		},
		Spec: corev1.ResourceQuotaSpec{
			Hard: corev1.ResourceList{
				corev1.ResourceName("limits." + resourceName):
					resource.MustParse("1Gi"),
			},
		},
	}

	s.quotaManager.Quotas["repro"] = &device.DeviceQuota{
		resourceName: &device.Quota{
			Limit: 1024 * 1024 * 1024,
		},
	}

	s.onDelQuota(cache.DeletedFinalStateUnknown{
		Key: "repro/gpu-quota",
		Obj: quota,
	})

	got := (*s.quotaManager.Quotas["repro"])[resourceName].Limit
	if got != 0 {
		t.Fatalf(
			"deleted ResourceQuota limit remains cached: got %d, want 0",
			got,
		)
	}
}
```

Run: 
```
go test ./pkg/scheduler \
  -run TestResourceQuotaTombstoneClearsCachedLimit \
  -count=1 -v
```

Result on current master:
```
unknown del object type
deleted ResourceQuota limit remains cached: got 1073741824, want 0
--- FAIL: TestResourceQuotaTombstoneClearsCachedLimit
```
No GPU hardware or Kubernetes cluster is required to reproduce the failure.

Kubernetes client-go explicitly documents that `OnDelete` may receive `DeletedFinalStateUnknown` if a watch closes and misses a deletion:
https://pkg.go.dev/k8s.io/client-go/tools/cache#ResourceEventHandler

Because the deleted ResourceQuota is absent from subsequent relists, ordinary informer resyncs do not invoke another deletion that could repair HAMi's separate quota cache. The stale limit can remain until the scheduler restarts or another ResourceQuota update overwrites it.
The expected fix is to use the same tombstone-unwrapping pattern already present in `onDelPod` and `onDelNode`.

- The output of `nvidia-smi -a` on your host: NA
- Your docker or containerd configuration file (e.g: `/etc/docker/daemon.json`): NA
- The hami-device-plugin container logs: NA, as it was reproduced with a unit test.
- The hami-scheduler container logs: The handler logs unknown `del object type`.
- The kubelet logs on the node (e.g: `sudo journalctl -r -u kubelet`): NA
- Any relevant kernel output lines from `dmesg`: NA

**Environment**:
- HAMi version: master at `c7891ded87b71e402a36f5f29ae7d245e0de70b2`
- nvidia driver or other AI device driver version: NA
- Docker version from `docker version`: NA
- Docker command, image and tag used: NA
- Kernel version from `uname -a`: macOS arm64; no Linux kernel dependency in this reproduction
- Others: Go `go1.26.5 darwin/arm64`

blackdragoon26:
**What type of PR is this?**
/kind bug

The ResourceQuota informer may deliver a `cache.DeletedFinalStateUnknown` when a deletion is missed while its watch is disconnected.

`onDelQuota` previously accepted only a direct `*corev1.ResourceQuota`, so it ignored tombstone deletion events and left the deleted quota's device limits in HAMi's scheduler cache.

This change unwraps ResourceQuota tombstones while preserving normal deletion handling. Tombstones containing an unexpected object type are logged and ignored safely.

A regression test verifies that processing a ResourceQuota tombstone clears the cached device limit.

Fixes #2257 

This change is scoped to the scheduler extender and does not require accelerator hardware.
Validation performed:
- `go test ./pkg/scheduler -run TestSchedulerOnDelQuotaClearsLimitFromTombstone -count=1 -v`
- `go test -short ./pkg/scheduler -count=1`
- `go test -race ./pkg/scheduler -run TestSchedulerOnDelQuotaClearsLimitFromTombstone -count=1`
- `make verify`
- `make test`

All checks passed on macOS arm64 with Go 1.26.5. with flying colors

Yes. ResourceQuota deletions delivered through informer tombstones now clear cached device limits instead of continuing to enforce stale limits.

AI assistance disclosure:
AI assistance was used to investigate the bug and formatting the PR. I reviewed the changes, reproduced the failure before the fix, and verified the implementation and test results myself.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling when resource quotas are deleted, including Kubernetes deletion tombstones.
  * Invalid or unsupported deletion events are safely ignored without causing errors or crashes.
  * Tracked resource limits are correctly cleared after quota deletion.

* **Tests**
  * Added coverage for quota deletion through tombstones and invalid deletion objects.

### DIFF
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -256,11 +256,23 @@ func (s *Scheduler) onUpdateQuota(oldObj, newObj any) {
 }
 
 func (s *Scheduler) onDelQuota(obj any) {
-	quota, ok := obj.(*corev1.ResourceQuota)
-	if !ok {
-		klog.Errorf("unknown del object type")
+	var quota *corev1.ResourceQuota
+
+	switch t := obj.(type) {
+	case *corev1.ResourceQuota:
+		quota = t
+	case cache.DeletedFinalStateUnknown:
+		var ok bool
+		quota, ok = t.Obj.(*corev1.ResourceQuota)
+		if !ok {
+			klog.Errorf("resource quota tombstone contained object of type %T", t.Obj)
+			return
+		}
+	default:
+		klog.Errorf("unknown resource quota delete object type %T", obj)
 		return
 	}
+
 	s.quotaManager.DelQuota(quota)
 }
 
--- a/pkg/scheduler/scheduler_test.go
+++ b/pkg/scheduler/scheduler_test.go
@@ -900,7 +900,37 @@ func TestSchedulerOnDelQuotaClearsLimitFromTombstone(t *testing.T) {
 }
 
 func TestSchedulerOnDelQuotaIgnoresInvalidObjects(t *testing.T) {
+	const namespace = "invalid-quota-delete-test"
+
 	s := NewScheduler()
+
+	nvidiaDevice, ok := device.GetDevices()[nvidia.NvidiaGPUDevice]
+	require.True(t, ok, "NVIDIA device should be registered")
+
+	resourceName := nvidiaDevice.GetResourceNames().ResourceMemoryName
+	require.NotEmpty(t, resourceName)
+
+	quota := &corev1.ResourceQuota{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "gpu-quota",
+			Namespace: namespace,
+		},
+		Spec: corev1.ResourceQuotaSpec{
+			Hard: corev1.ResourceList{
+				corev1.ResourceName("limits." + resourceName): *resource.NewQuantity(1024, resource.BinarySI),
+			},
+		},
+	}
+
+	s.onAddQuota(quota)
+	t.Cleanup(func() {
+		s.onDelQuota(quota)
+	})
+
+	quotas := s.quotaManager.GetResourceQuota()
+	namespaceQuota, ok := quotas[namespace]
+	require.True(t, ok)
+	expectedLimit := (*namespaceQuota)[resourceName].Limit
 
 	tests := []struct {
 		name string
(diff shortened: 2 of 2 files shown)


Shouren (approved):
/lgtm

PR #2261, opened 2026-08-01
docs: add AI assistance disclosure to the PR template

AswaniSahoo:
**What type of PR is this?**

/kind documentation

CONTRIBUTING.md requires that any use of AI assistance be disclosed in the pull request, along with the extent of it. The PR template does not ask for this, so the requirement is easy to miss — in #2046 the disclosure was added only after a reviewer asked for it.

This adds a "Was AI assistance used in this PR?" field pointing at the AI Assistance Notice, including the exception that notice already makes for trivial tab-completion.

None — raised by @mesutoezdil during review of #2046.

@mesutoezdil suggested this in #2046 and asked @spencercjh to open it; opening here since it hasn't landed yet — happy to close if one is already in flight.

The wording follows the existing AI Assistance Notice in CONTRIBUTING.md rather than introducing any new policy.

**Was AI assistance used in this PR?**:

Yes — Claude Code, to locate the CONTRIBUTING.md requirement and draft the template wording. I reviewed the result and take responsibility for its accuracy.

NONE

## Summary by CodeRabbit

* **Documentation**
  * Updated the pull request template with an AI assistance disclosure section.
  * Added guidance for reporting AI usage and AI-generated review-comment replies.
  * Added a Contribution Gates acknowledgment checkbox and an exception for trivial tab completion.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -19,4 +19,13 @@ Fixes #
 
 **Special notes for your reviewer**:
 
+**Was AI assistance used in this PR?**:
+
+<!--
+Required by the AI Assistance Notice in CONTRIBUTING.md. Disclose any use of AI
+assistance and the extent of it, e.g. docs only vs. code generation. This also
+applies to AI-generated replies to review comments. Trivial tab-completion limited
+to single keywords or short phrases does not need to be disclosed.
+-->
+
 **Does this PR introduce a user-facing change?**:
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -19,13 +19,18 @@ Fixes #
 
 **Special notes for your reviewer**:
 
-**Was AI assistance used in this PR?**:
+**AI assistance disclosure**:
+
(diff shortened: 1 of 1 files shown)


mesutoezdil (commented):
direction is right, three asks: rename the header to "AI assistance disclosure" so it matches what contributors already write and a future ci check can grep for it, put the "write None if unused" instruction in visible text w/ the CONTRIBUTING link so a blank field is not ambiguous, and add a checkbox like "- [ ] i have read the contribution gates and understand review replies must be written by me, not by an ai". nit: add the missing trailing newline while here.

spencercjh:
Agree with @mesutoezdil

AswaniSahoo:
Pushed as 7f9c861.

Renamed the field to `**AI assistance disclosure**:` so a ci check can grep that exact string.

Pulled the "write `None` if you did not use any" line out of the html comment into visible text and linked the AI Assistance Notice. A blank field should read as unfilled now, not as no ai used.

Added the checkbox, pointed at Contribution Gates rather than free text, since gate 4 already says the reply has to be written by the author. Felt better to reference the rule than restate it.

Newline fixed.

One I wasn't sure on: I kept the trivial tab completion exception in the html comment and only lifted the `None` line into visible text, to keep the visible block short.
Happy to lift that too if you'd rather have both visible.

mesutoezdil:

PR #2262, opened 2026-08-01
fix(plugin): release node lock on getAllocateResponse failure

### LINKED ISSUE #2214: [bug]: Node lock leak when getAllocateResponse fails in NvidiaDevicePlugin.Allocate
**What happened**:

In `pkg/device-plugin/nvidiadevice/nvinternal/plugin/server.go`, inside `NvidiaDevicePlugin.Allocate()`, if `plugin.getAllocateResponse()` returns an error in the non-MIG allocation path (lines 648–651), the function returns the error immediately without calling `PodAllocationFailed(nodename, current, NodeLockNvidia)`.

Because `PodAllocationFailed()` is skipped, the node lock (`hami.io/mutex.lock`) is not released, causing subsequent GPU allocation requests on that node to hang or fail.

When `plugin.getAllocateResponse()` fails, `PodAllocationFailed(nodename, current, NodeLockNvidia)` should be invoked to release the node lock before returning the error, consistent with all other error handling paths inside `Allocate()`.

1. Inspect `pkg/device-plugin/nvidiadevice/nvinternal/plugin/server.go` lines 648–651
2. Trigger a failure in getAllocateResponse() (for example, via an invalid CDI specification or configuration).
3. Observe that the node lock hami.io/mutex.lock is not released upon error return.

**Environment**:
- HAMi version: master (https://github.com/Project-HAMi/HAMi/commit/05e6c800b41c5544356682ad5b6bf19d8d6fe838)

manmathbh:
**What type of PR is this?**

/kind bug

getAllocateResponse failure path in the non-MIG Allocate loop was missing
a PodAllocationFailed call. Every other error exit in the same function
calls it, this was the only one that skipped it.

Node lock stays set on the node after the failure. Next GPU request on
that node either hangs or fails.

Fixes #2214

Only affects the non-MIG path. getAllocateResponse isn't called in the
MIG branch.

`NONE`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of failed pod resource allocations by recording the failure state when an allocation response cannot be generated.

### DIFF
--- a/pkg/device-plugin/nvidiadevice/nvinternal/plugin/server.go
+++ b/pkg/device-plugin/nvidiadevice/nvinternal/plugin/server.go
@@ -647,6 +647,7 @@ func (plugin *NvidiaDevicePlugin) Allocate(ctx context.Context, reqs *kubeletdev
 			}
 			response, err := plugin.getAllocateResponse(plugin.GetContainerDeviceStrArray(devreq))
 			if err != nil {
+				PodAllocationFailed(nodename, current, NodeLockNvidia)
 				return nil, fmt.Errorf("failed to get allocate response: %v", err)
 			}
 


mesutoezdil:
pls answer Shouren's question on #2214 first, if kubelet retries Allocate after an error then releasing the lock here may be wrong and the existing 7 call sites have the same problem, this needs a decision before adding an 8th. also cotishq had offered to fix that issue, pls sync w/ them.

manmathbh:
hey @cotishq go ahead and take this one, i hadn't realized you were planning to raise the pr for it. @Shouren that's a great point about the kubelet retry behavior breaking the single-pod assumption on #2214. i'll close this out so you guys can align on the right approach for all 8 call sites.

PR #2267, opened 2026-08-01
fix(leaderelection): require underscore separator in isHolderOf

### LINKED ISSUE #2154: [Bug] Leader election isHolderOf false positive when hostname is a prefix of another node's hostname

Norway-02:
/kind bug

`isHolderOf` in `pkg/util/leaderelection/leaderelection.go` checks whether the lease holder identity belongs to this instance using:

```go
```

The Kubernetes lease holder identity format is `hostname_<uuid>`. Because the check does not require the `_` separator, a replica with hostname `dev` will incorrectly return `true` for a lease held by `dev-server_<uuid>`.

Fixes #2154

The change is a one-character addition: `m.hostname` → `m.hostname+"_"`.

A regression test is added in the existing Ginkgo suite that directly exercises the false-positive case:

- hostname `dev` vs holder identity `dev-server_<uuid>` → must return `false`

This is scheduler-only. No device-allocation or in-container isolation path changed.

No.

---

**AI assistance disclosure:**

AI assistance was used to identify the issue, implement the fix, and draft this PR description. The change and test were reviewed and verified manually. I take responsibility for this contribution.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved leader identity validation to prevent similarly prefixed hostnames from being incorrectly treated as the local leader.
  * Added coverage for hostname-prefix edge cases.

### DIFF
--- a/pkg/util/leaderelection/leaderelection.go
+++ b/pkg/util/leaderelection/leaderelection.go
@@ -166,7 +166,7 @@ func (m *leaderManager) isHolderOf(lease *coordinationv1.Lease) bool {
 	if lease == nil || lease.Spec.HolderIdentity == nil {
 		return false
 	}
-	return strings.HasPrefix(*lease.Spec.HolderIdentity, m.hostname)
+	return strings.HasPrefix(*lease.Spec.HolderIdentity, m.hostname+"_")
 }
 
 func (m *leaderManager) isLeaseValid(now time.Time) bool {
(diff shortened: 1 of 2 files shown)


Norway-02:
@mesutoezdil @archlitchi @DSFans2014 PTAL when you have a chance — this is a one-line fix for the `isHolderOf` false positive (issue #2154), adding the `_` separator to match the Kubernetes lease holder identity format `hostname_<uuid>`. A regression test is included. Thank you!

Shouren:
It seems a duplicate PR of #2155

PR #2270, opened 2026-08-01
fix(ascend): validate module-pair count and prevent under-allocation for Ascend 910C

### LINKED ISSUE #2268: silent under-allocation and topology bypass in Ascend 910C device allocation
What happened:
When allocating Ascend 910C NPU devices, which require full physical module-pair allocation of 2 NPUs per card there are two related bugs in Fit() in pkg/device/ascend/device.go.

First is false-positive under-allocation. If a pod requests multiple devices, say 4 NPUs, and the node has candidate NPUs spread across partial modules, computeBestCombination910C() filters out the partial modules and returns only the complete pairs it can find, for example 2 NPUs. Fit() then assigns this shorter combination to tmpDevs and unconditionally returns true, meaning success, even though only 2 of the 4 requested NPUs were actually allocated.

Second, an exact-count bypass. When the raw candidate count happens to equal originReq, for example 2 NPUs requested and 2 candidate NPUs exist but on two separate cards, Fit() returns true immediately without ever calling computeBestCombination910C(), so it never validates that those 2 NPUs actually form a real module pair.

Third, a minor sort-order issue in computeBestCombination910C(): cards are sorted in ascending order by available NPU count, which contradicts the comment right above it saying full cards should be preferred. This doesn't cause allocation failures on its own, since partial cards are excluded outright by the equality check further down regardless of sort order, but it does mean that when a node has more full cards than the pod needs, the function doesn't actually prefer the ones you'd expect. Worth fixing alongside the main issue for correctness and to match the code's own stated intent.

What you expected to happen: 
Fit() should enforce that the allocated device count after module pairing strictly equals originReq. If full module pairs can't satisfy originReq, Fit() should return false with reason AllocatedCardsInsufficientRequest, instead of reporting success on a partial allocation. Module pairing validation should run for Ascend910CType requests regardless of whether the raw candidate count happens to equal originReq.

How to reproduce it:
Run this test against pkg/device/ascend:

func TestAscend910C_FitPartialAllocationBug(t *testing.T) {
dev := &Devices{config: VNPUConfig{CommonWord: Ascend910CType}}
nodeInfo := &device.NodeInfo{
Node: &corev1.Node{},
Devices: map[string][]device.DeviceInfo{
Ascend910CType: {
{ID: "dev-0", Index: 0, CustomInfo: map[string]any{"NetworkID": float64(0)}},
{ID: "dev-1", Index: 1, CustomInfo: map[string]any{"NetworkID": float64(0)}},
{ID: "dev-2", Index: 2, CustomInfo: map[string]any{"NetworkID": float64(0)}},
{ID: "dev-4", Index: 4, CustomInfo: map[string]any{"NetworkID": float64(0)}},
{ID: "dev-6", Index: 6, CustomInfo: map[string]any{"NetworkID": float64(0)}},
},
},
}
devices := []*device.DeviceUsage{
{ID: "dev-0", Index: 0, Count: 1, Totalmem: 32000, CustomInfo: map[string]any{"NetworkID": float64(0)}},
{ID: "dev-1", Index: 1, Count: 1, Totalmem: 32000, CustomInfo: map[string]any{"NetworkID": float64(0)}},
{ID: "dev-2", Index: 2, Count: 1, Totalmem: 32000, CustomInfo: map[string]any{"NetworkID": float64(0)}},
{ID: "dev-4", Index: 4, Count: 1, Totalmem: 32000, CustomInfo: map[string]any{"NetworkID": float64(0)}},
{ID: "dev-6", Index: 6, Count: 1, Totalmem: 32000, CustomInfo: map[string]any{"NetworkID": float64(0)}},
}
req := device.ContainerDeviceRequest{Nums: 4, Type: Ascend910CType}
pod := &corev1.Pod{ObjectMeta: metav1.ObjectMeta{Name: "test-pod"}}
fit, tmpDevs, _ := dev.Fit(devices, req, pod, nodeInfo, nil)
// Returns fit=true, allocated_count=2
}

Environment:
HAMi version: master (v2.9.0+)
Go version: go 1.22+

princexpoddar:
What type of PR is this?

/kind bug

Fixes Ascend 910C NPU device allocation in pkg/device/ascend/device.go.

In Fit(), verify that len(combination) == int(originReq) after calling computeBestCombination910C(). If enough complete module pairs cannot be formed, return false with AllocatedCardsInsufficientRequest instead of reporting success with a partial allocation.
Run computeBestCombination910C() for all Ascend910CType multi-device requests so the module-pair validation is not skipped when the raw candidate count already matches the requested count.
Change the sorting in computeBestCombination910C() to descending order so full cards are considered first, matching the intended behavior.
Add regression tests in pkg/device/ascend/device_910c_pairing_test.go covering partial allocation, the exact-count bypass, and a valid full-pair allocation.

Which issue(s) this PR fixes:  Fixes #2268 

Special notes for your reviewer:  Tested with go test -v ./pkg/device/ascend/....

Does this PR introduce a user-facing change?  No

AI assistance: Used Claude during development. Verified and tested the final changes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Ascend 910C device allocation to validate complete physical card pairs.
  * Requests with incomplete pairings now correctly fail instead of allocating fewer NPUs than requested.
  * Device selection now prioritizes fully available cards before partially available cards.
  * Preserved general allocation behavior for other device types.

* **Tests**
  * Added regression coverage for incomplete and complete Ascend 910C pairing scenarios.

### DIFF
--- a/pkg/device/ascend/device.go
+++ b/pkg/device/ascend/device.go
@@ -566,6 +566,25 @@ func (npu *Devices) Fit(devices []*device.DeviceUsage, request device.ContainerD
 	}
 
 	if needTopology {
+		if k.Type == Ascend910CType && originReq > 1 {
+			// Ascend 910C requires full module-pair allocation (2 NPUs per
+			// physical card). Always run the pairing filter here, even when
+			// the raw candidate count already equals originReq, because
+			// candidates satisfying the count alone may still be spread
+			// across incomplete/partial modules rather than full pairs.
+			combination := npu.computeBestCombination910C(nodeInfo, int(originReq), tmpDevs[k.Type])
+			if len(combination) != int(originReq) {
+				// Never report success on a short allocation: doing so silently
+				// under-allocates NPUs relative to what the pod requested.
+				reason[common.AllocatedCardsInsufficientRequest] = len(tmpDevs)
+				klog.V(5).InfoS(common.AllocatedCardsInsufficientRequest, "pod", klog.KObj(pod), "request", originReq, "allocated", len(combination))
+				return false, tmpDevs, common.GenReason(reason, len(devices))
+			}
+			tmpDevs[k.Type] = combination
+			klog.V(5).InfoS("device allocate success", "pod", klog.KObj(pod), "best device combination", tmpDevs)
+			return true, tmpDevs, ""
+		}
+
 		if len(tmpDevs[k.Type]) == int(originReq) {
 			klog.V(5).InfoS("device allocate success", "pod", klog.KObj(pod), "allocate device", tmpDevs)
 			return true, tmpDevs, ""
@@ -574,13 +593,7 @@ func (npu *Devices) Fit(devices []*device.DeviceUsage, request device.ContainerD
 				tmpDevs[k.Type] = device.ContainerDevices{tmpDevs[k.Type][0]}
 			} else {
 				// If requesting multiple devices, select the best combination of cards.
-				var combination device.ContainerDevices
-				if k.Type == Ascend910CType {
-					// Use topology-aware allocation for Ascend910C: only select full modules (2 NPUs per card).
-					combination = npu.computeBestCombination910C(nodeInfo, int(originReq), tmpDevs[k.Type])
-				} else {
-					combination = npu.computeBestCombination(nodeInfo, int(originReq), tmpDevs[k.Type])
-				}
+				combination := npu.computeBestCombination(nodeInfo, int(originReq), tmpDevs[k.Type])
 				tmpDevs[k.Type] = combination
 			}
 			klog.V(5).InfoS("device allocate success", "pod", klog.KObj(pod), "best device combination", tmpDevs)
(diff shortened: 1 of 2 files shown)


mesutoezdil (commented):
heads up, #2035 refactors the same Fit block and its author owns the ascend area

pkg/device/ascend/device.go
```diff
@@ -559,12 +559,31 @@ func (npu *Devices) Fit(devices []*device.DeviceUsage, request device.ContainerD
...
 		}
 	}
 
+	if k.Type == Ascend910CType && originReq > 1 {
+		// Ascend 910C requires full module-pair allocation (2 NPUs per
+		// physical card). Always run the pairing filter here, even when
+		// NetworkID is absent or needTopology is false, because
+		// candidates satisfying the count alone may still be spread
+		// across incomplete/partial modules rather than full pairs.
+		combination := npu.computeBestCombination910C(nodeInfo, int(originReq), tmpDevs[k.Type])
```
mesutoezdil, line 574: forcing the pairing on every 910c multi request changes behavior for nodes w/o module info in CustomInfo, does computeBestCombination910C return empty there and make those pods unschedulable? pls cover that case w/ a test

princexpoddar: Added a test for exactly this - TestComputeBestCombination910C_NoFullPairsReturnsEmpty. Checks both computeBestCombination910C directly (returns an empty slice, no panic, for 4 singletons on 4 separate modules with no NetworkID) and the end-to-end Fit() path (fit=false, clean rejection).

from what I could see in the current device plugin source NetworkID is always set to 0 for 910C, so this exact no-NetworkID scenario may be rare in practice but wanted the test in regardless, since Fit() shouldn't depend on that.

mesutoezdil: > Added a test for exactly this - TestComputeBestCombination910C_NoFullPairsReturnsEmpty. Checks both computeBestCombination910C directly (returns an empty slice, no panic, for 4 singletons on 4 separate modules with no NetworkID) and the end-to-end Fit() path (fit=false, clean rejection).
> 
> from what I could see in the current device plugin source NetworkID is always set to 0 for 910C, so this exact no-NetworkID scenario may be rare in practice but wanted the test in regardless, since Fit() shouldn't depend on that.

You should already know the rule from your previous PRs
a reminder for the future:

princexpoddar: apologies for not addressing it directly by me...pls I would appreciate if you could reopen the PR..Also thanks for the heads up on #2035..as i checked it doesn't touch the needTopology branch or 910C pairing block this pr modifies..they can merge with no conflicts..will rebase if it changes

PR #2272, opened 2026-08-01
fix(device): use HasPrefix and TrimPrefix in DelQuota

### LINKED ISSUE #2271: bug(quota): DelQuota uses length check instead of HasPrefix matching AddQuota
**What happened**:

While reviewing the QuotaManager logic in pkg/device/quota.go, I noticed an asymmetry between AddQuota and DelQuota when parsing resource quota keys.

In AddQuota, we explicitly check if a ResourceQuota key starts with "limits." before trimming it:

```go
if !strings.HasPrefix(idx.String(), "limits.") {
    continue
}
dn := strings.TrimPrefix(idx.String(), "limits.")
```

However, in DelQuota, instead of checking the prefix, it only checks if the key length is greater than len("limits.") (7 characters) and blindly slices off the first 7 characters:

```go
if len(idx.String()) <= len("limits.") {
    continue
}
dn := idx.String()[len("limits."):]
```

Because of this length-only check, any non-limits. key longer than 7 characters (say custom annotations or non-limit keys) will pass the check and get sliced by offset 7. If the remaining string happens to match a managed resource name (for instance 0123456nvidia.com/gpumem), DelQuota will match IsManagedQuota(dn) as true and reset quotaInfo.Limit = 0, unexpectedly corrupting the active GPU quota limits in that namespace.

DelQuota should use strings.HasPrefix and strings.TrimPrefix to mirror AddQuota, safely ignoring any quota keys that don't start with "limits." without corrupting namespace quota limits.

1. Create a ResourceQuota with a valid limit (e.g. limits.nvidia.com/gpumem: "10000"). AddQuota sets nvidia.com/gpumem limit to 10000.
2. Update or delete a ResourceQuota containing a key longer than 7 characters whose suffix after 7 characters equals a managed resource name (e.g., 0123456nvidia.com/gpumem).
3. DelQuota runs: len("0123456nvidia.com/gpumem") > 7 passes, slices [7:] to "nvidia.com/gpumem", matches IsManagedQuota, and resets quotaInfo.Limit = 0.

I can open a quick PR to fix this along with a unit test!

**Environment**:
- HAMi version: master branch (current pkg/device/quota.go)
- Others: Code logic bug in QuotaManager (pkg/device/quota.go)

SK8-infi:
**What type of PR is this?**
/kind bug

Fixes an issue in `QuotaManager.DelQuota` where resource quota keys were not validated for the "limits." prefix before slicing.

Previously, DelQuota only checked if the key length was greater than len("limits.") (7 characters) and sliced the first 7 characters. As a result, any non-limits. key longer than 7 characters whose suffix after 7 matched a managed resource (e.g., 0123456nvidia.com/gpumem) would pass IsManagedQuota and reset quotaInfo.Limit = 0, unexpectedly corrupting namespace quota limits.

This PR updates DelQuota to use strings.HasPrefix(idx.String(), "limits.") and strings.TrimPrefix(idx.String(), "limits.") to be the same as AddQuota, ensuring non-limits. keys are safely ignored.

Fixes #2271

This would be me first time contributing to HAMI. If anything is out of order or I made some mistake do let me know. Thanks.

NA

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation when removing device quota resources, preventing incorrectly formatted quota names from being processed.
  * Fixed an issue where removing quotas containing only unrelated resources could reset existing managed limits.
  * Existing managed memory limits now remain unchanged when unrelated quota resources are removed.

### DIFF
--- a/pkg/device/quota.go
+++ b/pkg/device/quota.go
@@ -228,10 +228,10 @@ func (q *QuotaManager) DelQuota(quota *corev1.ResourceQuota) {
 	for idx, val := range quota.Spec.Hard {
 		value, ok := val.AsInt64()
 		if ok {
-			if len(idx.String()) <= len("limits.") {
+			if !strings.HasPrefix(idx.String(), "limits.") {
 				continue
 			}
-			dn := idx.String()[len("limits."):]
+			dn := strings.TrimPrefix(idx.String(), "limits.")
 			if !IsManagedQuota(dn) {
 				continue
 			}
--- a/pkg/device/quota_test.go
+++ b/pkg/device/quota_test.go
@@ -252,3 +252,33 @@ func TestAddQuotaAndDelQuota(t *testing.T) {
 		t.Errorf("DelQuota: expected core limit 0, got %d", (*qm.Quotas[ns])[coreName].Limit)
 	}
 }
+
+func TestDelQuotaNonLimitsKey(t *testing.T) {
+	initTest()
+	qm := NewQuotaManager()
+	ns := "testns"
+	memName := "nvidia.com/gpumem"
+	coreName := "nvidia.com/gpucore"
+
+	rq := &corev1.ResourceQuota{}
+	rq.Namespace = ns
+	rq.Spec.Hard = corev1.ResourceList{
+		corev1.ResourceName("limits." + memName):  *resource.NewQuantity(100, resource.DecimalSI),
+		corev1.ResourceName("limits." + coreName): *resource.NewQuantity(10, resource.DecimalSI),
+	}
+	qm.AddQuota(rq)
+
+	// Quota with non-limits keys (including 7-character prefix key).
+	unrelatedQuota := &corev1.ResourceQuota{}
+	unrelatedQuota.Namespace = ns
+	unrelatedQuota.Spec.Hard = corev1.ResourceList{
+		corev1.ResourceName("requests." + memName): *resource.NewQuantity(50, resource.DecimalSI),
+		corev1.ResourceName("0123456" + memName):   *resource.NewQuantity(50, resource.DecimalSI),
+	}
+
+	// DelQuota on non-limits keys should NOT reset active quota limits.
+	qm.DelQuota(unrelatedQuota)
+	if (*qm.Quotas[ns])[memName].Limit != 100 {
+		t.Errorf("DelQuota: expected memory limit 100 after deleting non-limits quota, got %d", (*qm.Quotas[ns])[memName].Limit)
+	}
+}


Eshiv-Pandey (commented):
i mean i dont see a functional behavior change here.` IsManagedQuota()` already appears to reject any malformed dn produced by the previous logic.
can you share a case where the old implementation behaved incorrectly? otherwise this seems more like a clean code kinda change than a bug fix

SK8-infi:
> i mean i dont see a functional behavior change here.` IsManagedQuota()` already appears to reject any malformed dn produced by the previous logic. can you share a case where the old implementation behaved incorrectly? otherwise this seems more like a clean code kinda change than a bug fix

@Eshiv-Pandey 

If a key has a 7 char prefix before the resource name (thake this for ex... custom.nvidia.com/gpumem), stripping offset [7:] leaves the exact valid resource name (---> nvidia.com/gpumem). IsManagedQuota accepts it and accidentally wipes out the active quota limit for that namespace.
HasPrefix ensures keys without limits. are safely ignored.

Kind of maybe considered a very remote edge case but still better to be safe...do let me know if I am overlooking something.

Eshiv-Pandey:
makes sense. lgtm

pkg/device/quota.go
```diff
@@ -228,10 +228,10 @@ func (q *QuotaManager) DelQuota(quota *corev1.ResourceQuota) {
 	for idx, val := range quota.Spec.Hard {
 		value, ok := val.AsInt64()
 		if ok {
-			if len(idx.String()) <= len("limits.") {
+			if !strings.HasPrefix(idx.String(), "limits.") {
 				continue
 			}
-			dn := idx.String()[len("limits."):]
+			dn := strings.TrimPrefix(idx.String(), "limits.")
```
mesutoezdil, line 234: pls add a small case to quota_test.go showing a non limits key is ignored by DelQuota so this cant regress

mesutoezdil (commented):
heads up, #2172 also touches quota.go and quota_test.go so watch for conflicts.

pkg/device/quota.go
SK8-infi: Added in recent commit

SK8-infi:
> heads up, #2172 also touches quota.go and quota_test.go so watch for conflicts.
> 
> no ai assistance disclosure is present. if any ai tool was used, it must be disclosed per CONTRIBUTING.md: https://github.com/Project-HAMi/HAMi/blob/master/CONTRIBUTING.md#ai-assistance-notice

#2172 seems to be closed for now....so ig we are good to go
Added the regression test

image

As for ai tool, will keep in mind to disclose prior next time. But was not used in this one due to minor fix.

DSFans2014 (commented):
/lgtm

archlitchi:
please fix the go-lint

SK8-infi:
@archlitchi fixed the linting issue. Will take care from now on...

mesutoezdil:
> @archlitchi fixed the linting issue. Will take care from now on...

for future: pls refrain from force pushing once you have started getting reviews, it makes it much harder to tell what has changed from review to review.

SK8-infi:
> > @archlitchi fixed the linting issue. Will take care from now on...
> 
> for future: pls refrain from force pushing once you have started getting reviews, it makes it much harder to tell what has changed from review to review.

Okay sure. there was this one project I was working with..they say to keep one commit. So got confused

Shouren (approved):
/lgtm

PR #2279, opened 2026-08-02
chore: improve error handling in scheduler routes and device plugin

shinigami-777:
**What type of PR is this?**
/kind cleanup

While investigating the codebase, I ran `errcheck` manually and noticed some unhandled errors that could lead to silent failures. These slipped through CI because `errcheck` is disabled in the project's `.golangci.yaml` configuration. Furthermore, the entire `pkg/device-plugin` directory is excluded from linter checks.

This PR adds proper error checking and logging to prevent silent failures in these components.

**Changes Made**

1. `pkg/scheduler/routes/route.go`:
Added error handling for HTTP `w.Write()` calls. If writing the JSON response to the client fails (e.g., due to a dropped connection), it will now properly log a `klog.ErrorS` instead of failing silently.

2. `pkg/device-plugin/nvidiadevice/nvinternal/plugin/server.go`:
- Added proper error handling when parsing the MIG configuration via `yaml.Unmarshal`. If unmarshaling fails, the plugin now logs an error, explicitly aborts the rest of the MIG setup block, and falls back to a non-MIG configuration. This prevents the plugin from silently continuing with an invalid, zero-value migCurrent configuration.
- Added  error handling for `os.MkdirAll`, `os.RemoveAll`, and `os.Chmod` .

Yes. By adding explicit error handling, the behavior of the device plugin has shifted from "silently ignoring" these errors to "explicitly aborting" the pod's resource allocation. This makes the plugin much safer and more predictable (since it no longer allows pods to start in an incomplete or incorrectly permissioned state).

**AI assistance disclosure**:
I manually ran `errcheck` to catch the unhandled cases. The changes was made using assistance from Gemini 3.1 Pro. I understand and take responsibility for the implementation and have manually verified and tested it using `make verify` and `make test` before creating this PR.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of invalid MIG configuration data by reporting parsing errors and safely falling back to non-MIG configuration.
  * GPU allocation now detects and clearly reports failures while preparing required cache and lock directories.
  * Improved scheduler response reliability by detecting failures when sending responses.
  * Enhanced error handling helps prevent incomplete GPU allocations and improves visibility into configuration and response-delivery issues.

### DIFF
--- a/pkg/device-plugin/nvidiadevice/nvinternal/plugin/server.go
+++ b/pkg/device-plugin/nvidiadevice/nvinternal/plugin/server.go
@@ -314,15 +314,17 @@ func (plugin *NvidiaDevicePlugin) Start(kubeletSocket string) error {
 				outStr := stdout.Bytes()
 				if err := yaml.Unmarshal(outStr, &plugin.migCurrent); err != nil {
 					klog.Errorf("failed to unmarshal mig config: %v", err)
-				}
-				writeMigConfig(outStr)
-
-				HamiInitMigConfig, err := plugin.processMigConfigs(plugin.migCurrent.MigConfigs, deviceNumbers)
-				if err != nil {
-					klog.Infof("no device in node: %v", err)
+					klog.Warning("Falling back to non-MIG configuration")
 				} else {
-					plugin.migCurrent.MigConfigs["current"] = HamiInitMigConfig
-					migApplied = true
+					writeMigConfig(outStr)
+
+					HamiInitMigConfig, err := plugin.processMigConfigs(plugin.migCurrent.MigConfigs, deviceNumbers)
+					if err != nil {
+						klog.Infof("no device in node: %v", err)
+					} else {
+						plugin.migCurrent.MigConfigs["current"] = HamiInitMigConfig
+						migApplied = true
+					}
 				}
 			}
 		}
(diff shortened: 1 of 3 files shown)


pkg/device-plugin/nvidiadevice/nvinternal/plugin/server.go
```diff
@@ -312,7 +312,9 @@ func (plugin *NvidiaDevicePlugin) Start(kubeletSocket string) error {
 				klog.Warning("Falling back to non‑MIG configuration")
 			} else {
 				outStr := stdout.Bytes()
-				yaml.Unmarshal(outStr, &plugin.migCurrent)
+				if err := yaml.Unmarshal(outStr, &plugin.migCurrent); err != nil {
+					klog.Errorf("failed to unmarshal mig config: %v", err)
```
mesutoezdil, line 316: logging alone is not enough, migCurrent stays zero and the next lines still run w/ it, pls fall back to non mig like the sibling branch above

pkg/device-plugin/nvidiadevice/nvinternal/plugin/server.go
```diff
@@ -675,9 +677,15 @@ func (plugin *NvidiaDevicePlugin) Allocate(ctx context.Context, reqs *kubeletdev
 				cacheFileHostDirectory := fmt.Sprintf("%s/vgpu/containers/%s_%s", hostHookPath, current.UID, currentCtr.Name)
 				os.RemoveAll(cacheFileHostDirectory)
 
-				os.MkdirAll(cacheFileHostDirectory, 0777)
+				if err := os.MkdirAll(cacheFileHostDirectory, 0777); err != nil {
+					PodAllocationFailed(nodename, current, NodeLockNvidia)
+					return &kubeletdevicepluginv1beta1.AllocateResponse{}, fmt.Errorf("failed to create cache directory: %v", err)
+				}
 				os.Chmod(cacheFileHostDirectory, 0777)
```
mesutoezdil: os.Chmod and os.RemoveAll right next to these are still unchecked, pls handle them too or say why not since errcheck is the whole point here

mesutoezdil (commented):
the mkdir part changes behavior (silent -> abort) so pls put that in the release note field instead of "no".

pkg/device-plugin/nvidiadevice/nvinternal/plugin/server.go
shinigami-777: Yes, made it similar to the sibling branch above it.

pkg/device-plugin/nvidiadevice/nvinternal/plugin/server.go
shinigami-777: Added error handling for the os.Chmod and os.RemoveAll.

shinigami-777:
The test for pkg/device-plugin/nvidiadevice/nvinternal/plugin was failing earlier as expected due to the strict error handling for filesystem operations during pod allocation. When the plugin fails to clean up or set permissions on the cache directory or the lock directory, it aborts the allocation and returns an error.
I was earlier testing locally using sudo so missed it :(
To fix the permission denied issue, we can override the hostHookPath variable with a safe temporary directory (using t.TempDir()) at the start of Allocate tests.

FouoF:
Thanks for auditing these error paths. The PR bundles three unrelated changes—HTTP write logging, MIG YAML fallback behavior, and allocation-directory failure handling—without a shared reproducible defect. The added test setup does not exercise the newly changed failure paths, and the normal exact-head CI suite has not produced reviewable results. A broad mechanical error-handling sweep is difficult to validate and maintain, so we are closing this PR. Please submit any real failure as a separate, minimal change with a reproduction and a regression test for that exact path.

PR #1345, opened 2025-09-12
FIx CI, add 910B4-1 template and fix vGPUmonitor metrics error

archlitchi:
/kind bug

1. Fix Ci which blocks HAMi from building
2. add 910B4-1 template
3. fix vGPUmonitor metrics error

PR #2282, opened 2026-08-02
fix(vGPUmonitor): bound v0 spec metric sums to active proc slots

### LINKED ISSUE #2281: vGPUmonitor v0 shared-region spec sums device metrics over stale proc slots
## What happened

In `pkg/monitor/nvidia/v0/spec.go`, the six per-device metric accessors iterate over the **entire** fixed-size `procs [1024]shrregProcSlotT` array instead of only the live slots bounded by `procnum`:

```go
func (s Spec) DeviceMemoryContextSize(idx int) uint64 {
	v := uint64(0)
	for _, p := range s.sr.procs {           // iterates all 1024 slots
		v += p.used[idx].contextSize
	}
	return v
}
```

The affected functions are `DeviceMemoryContextSize`, `DeviceMemoryModuleSize`, `DeviceMemoryBufferSize`, `DeviceMemoryOffset`, `DeviceMemoryTotal` and `DeviceSmUtil`.

`procnum` (an `int32` field of `sharedRegionT`) is the count of currently active process slots written by `libvgpu.so`. Slots at index `>= procnum` are not cleared and can retain **stale, non-zero** values from processes that have exited. Because these accessors sum over all 1024 slots, that stale data is added into the reported totals.

## Impact

The vGPUmonitor over-reports per-device used memory and SM utilization for the v0 shared-region layout. This inflates the Prometheus metrics derived from these accessors (e.g. `hami_container_device_memory_bytes`, `hami_container_device_utilization` / used-memory gauges), producing values that can exceed the real device usage and, in the worst case, spike far above the device limit once a slot beyond `procnum` holds leftover data.

The v0 layout is reachable: `pkg/monitor/nvidia/cudevshr.go` selects it when the shared-region file size is `1197897`.

## Root cause / precedent

The v1 layout already handles this correctly — every equivalent accessor in `pkg/monitor/nvidia/v1/spec.go` bounds the loop:

```go
for _, p := range s.sr.procs[:int(s.sr.procnum)] {
```

This was fixed for v1 in commit `acc8d91` (PR #1345) but never backported to v0.

## Proposed fix

Bound all six v0 loops to `s.sr.procs[:int(s.sr.procnum)]`, mirroring v1, and add unit-test coverage that asserts a stale proc slot beyond `procnum` is excluded from the sums.

I have the fix ready and will open a PR referencing this issue.

imantaba:
**What type of PR is this?**

/kind bug

In `pkg/monitor/nvidia/v0/spec.go`, the six per-device metric accessors (`DeviceMemoryContextSize`, `DeviceMemoryModuleSize`, `DeviceMemoryBufferSize`, `DeviceMemoryOffset`, `DeviceMemoryTotal` and `DeviceSmUtil`) iterate over the entire fixed-size `procs [1024]shrregProcSlotT` array instead of only the live slots bounded by `procnum`:

```go
for _, p := range s.sr.procs {          // iterates all 1024 slots
	v += p.used[idx].contextSize
}
```

`procnum` is the count of active process slots written by `libvgpu.so`. Slots at index `>= procnum` are not cleared and can retain stale, non-zero values from processes that have exited. Summing over all 1024 slots folds that stale data into the reported per-device memory and SM-utilization totals, inflating the corresponding vGPUmonitor gauges (and in the worst case pushing a value far above the real device usage/limit once a slot beyond `procnum` holds leftover data).

The v0 layout is reachable: `pkg/monitor/nvidia/cudevshr.go` selects it when the shared-region file size is `1197897`.

The v1 layout already bounded these loops to `s.sr.procs[:int(s.sr.procnum)]` (fixed in `acc8d91`, #1345) but the fix was never backported to v0. This PR backports it to v0.

**Changes**

- Bound all six v0 accessor loops to the active proc slots, mirroring v1.
- Route both v0 and v1 through a shared-shape `activeProcs()` helper that clamps `procnum` into `[0, len(procs)]` before slicing. `procnum` comes straight from the mmap'd shared-memory region, so a corrupt value (negative or larger than the backing array) would otherwise panic with `slice bounds out of range`. This also hardens the pre-existing v1 code, which had the same exposure.
- Update the existing spec tests to set `procnum` (they previously relied on all-zero trailing slots, so they would sum `procs[:0]` = 0 once the bound is applied).
- Add a stale-slot regression case per accessor asserting that a non-zero slot beyond `procnum` is excluded from the sum, plus a clamp regression test (negative and oversized `procnum`) for both v0 and v1.

Fixes #2281

Verified red→green:
- With the loops still unbounded, the stale-slot cases return `600` instead of `200`; after the fix all pass.
- With `activeProcs()` slicing `procnum` directly (no clamp), the clamp test panics `slice bounds out of range [:-5]`; with the clamp it passes.

`go test ./pkg/monitor/nvidia/... -race` and `gofmt`/`go vet` are clean.

The clamp addresses review feedback on this PR (thanks @mesutoezdil) — folding it in here since it hardens the exact lines this PR touches, and applying it to v1 too keeps both layouts consistent.

```release-note
Fix vGPUmonitor over-reporting device memory and SM utilization for the v0 shared-region layout, caused by counting stale process slots.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved NVIDIA monitoring reliability when process-count data is invalid or corrupted.
  * Device memory and SM utilization totals now ignore inactive process slots.
  * Prevented monitoring errors and crashes caused by negative or oversized process counts.

* **Tests**
  * Added coverage for invalid process counts and excluded process slots across NVIDIA monitoring versions.

### DIFF
--- a/pkg/monitor/nvidia/v0/spec.go
+++ b/pkg/monitor/nvidia/v0/spec.go
@@ -81,9 +81,17 @@ func (s Spec) DeviceNum() int {
 	return int(s.sr.num)
 }
 
+// activeProcs returns the process slots currently in use. procnum is read from
+// the shared-memory region and may be corrupt (negative or larger than the
+// backing array); clamp it to a valid range so slicing can never panic.
+func (s Spec) activeProcs() []shrregProcSlotT {
+	n := min(max(int(s.sr.procnum), 0), len(s.sr.procs))
+	return s.sr.procs[:n]
+}
+
 func (s Spec) DeviceMemoryContextSize(idx int) uint64 {
 	v := uint64(0)
-	for _, p := range s.sr.procs[:int(s.sr.procnum)] {
+	for _, p := range s.activeProcs() {
 		v += p.used[idx].contextSize
 	}
 	return v
@@ -91,7 +99,7 @@ func (s Spec) DeviceMemoryContextSize(idx int) uint64 {
 
 func (s Spec) DeviceMemoryModuleSize(idx int) uint64 {
 	v := uint64(0)
-	for _, p := range s.sr.procs[:int(s.sr.procnum)] {
+	for _, p := range s.activeProcs() {
 		v += p.used[idx].moduleSize
 	}
 	return v
@@ -99,7 +107,7 @@ func (s Spec) DeviceMemoryModuleSize(idx int) uint64 {
 
 func (s Spec) DeviceMemoryBufferSize(idx int) uint64 {
 	v := uint64(0)
-	for _, p := range s.sr.procs[:int(s.sr.procnum)] {
+	for _, p := range s.activeProcs() {
 		v += p.used[idx].bufferSize
 	}
 	return v
@@ -107,7 +115,7 @@ func (s Spec) DeviceMemoryBufferSize(idx int) uint64 {
 
 func (s Spec) DeviceMemoryOffset(idx int) uint64 {
 	v := uint64(0)
-	for _, p := range s.sr.procs[:int(s.sr.procnum)] {
+	for _, p := range s.activeProcs() {
 		v += p.used[idx].offset
 	}
 	return v
@@ -115,7 +123,7 @@ func (s Spec) DeviceMemoryOffset(idx int) uint64 {
 
 func (s Spec) DeviceMemoryTotal(idx int) uint64 {
 	v := uint64(0)
-	for _, p := range s.sr.procs[:int(s.sr.procnum)] {
+	for _, p := range s.activeProcs() {
 		v += p.used[idx].total
 	}
 	return v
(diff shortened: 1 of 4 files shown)


mesutoezdil (commented):
thx for pr

pkg/monitor/nvidia/v0/spec.go
```diff
@@ -83,47 +83,47 @@ func (s Spec) DeviceNum() int {
 
 func (s Spec) DeviceMemoryContextSize(idx int) uint64 {
 	v := uint64(0)
-	for _, p := range s.sr.procs {
+	for _, p := range s.sr.procs[:int(s.sr.procnum)] {
```
mesutoezdil: nit: a corrupt shm procnum over 1024 or negative would panic this slice, v1 has the same exposure, maybe a clamped helper for both in a follow up

imantaba: Instead of leave it for the future fallow-up i fixed it .Both v0 and v1 now go through an `activeProcs()` helper that clamps `procnum` into `[0, len(procs)]` before slicing and as a result neither layout can panic.
Also i did updated the PR description .

archlitchi (approved):
/lgtm

archlitchi (approved):
/lgtm

PR #2285, opened 2026-08-03
fix(device): reject int32-overflowing memory requests in iluvatar and mthreads

### LINKED ISSUE #2278: [bug]: cambricon int32 overflow in GenerateResourceRequests silently drops memory request
### What happened

Requesting cambricon shared-MLU memory in Kubernetes quantity units (e.g. `cambricon.com/mlu.smlu.vmemory: 16Gi`) makes `GenerateResourceRequests` silently drop the memory request entirely.

`memnum` is computed as `int(memnums) * 256` (a 64-bit value) and then narrowed with `int32(memnum)`. `int32` overflows for any request above 8,388,607 units, and the wrap-around produces `0` for realistic Gi-sized values:

- `16Gi` -> `AsInt64()` = 17179869184 -> `17179869184 * 256` = 4398046511104 -> `int32(...)` = **0**
- `8Gi` -> `int32(...)` = **0**

Because `Memreq` becomes `0`, the `dev.Totalmem-dev.Usedmem < memreq` check in `CambriconDevices.Fit` (`pkg/device/cambricon/device.go`) passes vacuously, so the pod can be scheduled on an already-full MLU with no memory accounting, leading to oversubscription and OOM in the container.

### What you expected to happen

An oversized memory request should be rejected (e.g. `CardInsufficientMemory`) or interpreted correctly, not silently truncated to zero memory.

### Root cause

`pkg/device/cambricon/device.go`:

```go
memnum = int(memnums) * 256        // int64 math
...
Memreq: int32(memnum),             // narrows to int32, wraps to 0 when memnums > math.MaxInt32/256
```

The same overflow class was already fixed for enflame in #2145 and #2190.

### How to reproduce

Create a container with:

```yaml
limits:
  cambricon.com/mlu: "1"
  cambricon.com/mlu.smlu.vmemory: 16Gi
```

Then `GenerateResourceRequests` returns `Memreq: 0` instead of a positive bound.

### Environment
- HAMi version: master (also 2.8.x line)
- MLU: cambricon

### LINKED ISSUE #2284: bug: int32 overflow in GenerateResourceRequests silently drops memory request (iluvatar, mthreads)
**What happened**

Requesting device memory as a normal Kubernetes quantity (like `16Gi`) makes `GenerateResourceRequests` silently return `Memreq: 0` on the iluvatar and mthreads backends. The byte value from `AsInt64()` gets multiplied by a per-backend factor and then narrowed to `int32`, which wraps for realistic Gi-sized values:

- iluvatar (`pkg/device/iluvatar/device.go`): `memnum = int(memnums) * 256`, then `Memreq: int32(memnum)`
- mthreads (`pkg/device/mthreads/device.go`): `memnum = int(memnums) * 512`, then `Memreq: int32(memnum) / int32(n)`

`16Gi` -> `AsInt64()` = 17179869184 -> x256 (or x512) -> `int32(...)` = **0**.

Because `Memreq` becomes `0`, the `dev.Totalmem - dev.Usedmem < memreq` check in `Fit` passes vacuously, so the pod can be scheduled onto an already-full device with no memory accounting, leading to oversubscription and OOM in the container.

**What I expected**

An oversized memory request should be rejected, not silently truncated to zero. This is the same overflow class that was already fixed for enflame in #2145 / #2190, and the mthreads device *count* was already guarded the same way in #2134 (same function) — the memory value just wasn't covered.

**How to reproduce**

A container like:

```yaml
limits:
  mthreads.com/vgpu: "1"
  mthreads.com/sgpu-memory: 16Gi
```

`GenerateResourceRequests` returns `Nums: 1, Memreq: 0` instead of a positive bound or a rejection.

**Environment**
- HAMi version: master (also the 2.8.x / 2.9.x lines)

**Note**

The same overflow class also affects cambricon, tracked separately in #2278. I'd like to fix the iluvatar and mthreads cases here — I have the fix and regression tests ready, matching the #2134 / #2145 approach.

/kind bug

veyron-kairo:
**What type of PR is this?**
/kind bug

`GenerateResourceRequests` in the nvidia, cambricon, iluvatar, and mthreads
backends multiplies the requested device memory by a per-backend factor and
narrows the result to `int32`. A normal Gi-scale quantity like `16Gi` wraps past
`math.MaxInt32` to `Memreq: 0`, which passes the `Totalmem - Usedmem < memreq`
check in `Fit` vacuously, so the pod is scheduled onto an already-full device —
oversubscription and OOM.

This bounds the memory value before the multiply and rejects the request on
overflow instead of silently truncating to zero. It also rejects **decimal-form**
quantities (`AsInt64()` returns false for values like `16.0Gi`, which previously
fell through and left the request at zero). Mirrors the device-count guard in
mthreads (#2134) and the enflame int32 bounding in #2145 / #2190.

Fixes #2284
Fixes #2278

Covers nvidia, cambricon, iluvatar, and mthreads (nvidia added per @archlitchi's
review). Each backend has regression cases for both an int32-overflowing integer
(`16Gi`) and a decimal-form quantity (`16.0Gi`), both of which are now rejected
rather than truncated. Merged current master to resolve conflicts with the new
`MemoryFactor` const. `make verify` and the device unit tests pass locally.

**AI assistance disclosure**:
I used AI assistance (Claude Code) to help locate the overflow across backends,
resolve the merge, and draft the tests. I reviewed and verified the change,
confirmed the before/after behavior with the regression tests, ran `make verify`
locally, and take responsibility for it.

NONE

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented invalid memory requests from overflowing, truncating, or being interpreted as zero during device allocation.
  * Out-of-range, negative, and non-integer memory quantities are now rejected safely.
  * Valid memory requests continue to be processed normally.

* **Tests**
  * Added coverage for oversized and decimal-form memory requests across supported device types.
  * Verified invalid requests return empty device allocations.


mesutoezdil:
Cambricon has the same overflow class and is tracked separately in #2278, with no open PR yet. Given you already have the pattern worked out here (negative/overflow guard before the int32 narrowing), would you be open to extending this PR to cover cambricon too, or opening a follow-up PR for it?

veyron-kairo:
thanks — extended it to cambricon in bf214f2, same guard (bound before the ×256) plus a regression case, and linked #2278 so it closes on merge.

 one thing I noticed while doing this: nvidia's GenerateResourceRequests has the same int32(memnum) narrowing. for nvidia it's only reachable via a misconfigured Gi value or a large MemoryFactor since it doesn't multiply by a fixed unit, so it's a weaker case — happy to fold it in here too or leave it out to keep this focused. your call.

mesutoezdil:
cc @archlitchi wdyt, fold nvidia's same int32 narrowing into this PR too, or keep it separate?

archlitchi:
> cc @archlitchi wdyt, fold nvidia's same int32 narrowing into this PR too, or keep it separate?

i think we should narrow them into this PR too

archlitchi:
please resolve these conflicts and resolve the code-rabbit comment

veyron-kairo:
pushed the changes.

nvidia's in now, the guard bounds the value before the int32(memnum) narrowing there too. resolved the conflicts against master (kept the new MemoryFactor const and put the overflow check back on top of it). for the coderabbit comment: the guard used to sit inside the AsInt64 ok-branch, so a decimal quantity like 16.0Gi (where AsInt64 returns false) slipped past it and left the request at 0. it rejects on !parsed or out-of-range now. added tests for both 16Gi and 16.0Gi across the four backends.

one thing on the red Unit test check, that's not coming from this PR. master itself is failing the scheduler quota tests right now because my #2313 and #2347 conflict once they're both on master (each passed CI on its own base). fix is up in #2408, it's test-only. once that merges i'll re-merge master here and this should go green.

pkg/device/cambricon/device_test.go
```diff
@@ -287,6 +287,36 @@ func Test_GenerateResourceRequests(t *testing.T) {
 				Coresreq:         int32(2),
 			},
 		},
+		{
+			// A Gi-scale memory quantity, times the x256 factor, overflows
```
archlitchi: please remove these comments and all comments below

veyron-kairo: removed, 86b9848. took them out of cambricon, iluvatar and mthreads — the nvidia ones i added later didn't have any, so those were already fine.

on the red unit test: #2408 is closed now (already fixed on master), and the master merge here picked that up, so it should go green on this run.

mesutoezdil:

veyron-kairo:
@mesutoezdil , actually I write and give it to ai every time I write to polish and make the content shorter , isn't that allowed the whole thinking and the idea is mine .

if its not allowed I'll write. it in my own words this time .!

pkg/device/nvidia/device.go
```diff
@@ -20,6 +20,7 @@ import (
 	"errors"
 	"flag"
 	"fmt"
+	"math"
```
mesutoezdil, line 23: conflict here is just import order, master added sort near math. rest of the guard logic applies clean, cambricon/iluvatar/mthreads merge clean too. just rebase for this one line.

veyron-kairo:
@archlitchi , resolved the conflicts .

FouoF:
@veyron-kairo I think we should not only reject the memory requests over int32, but also remind the users that they may used invalid unit. As your example shown, this condition happens on the user input like 16Gi not an integer over int32. So I suggest update the error message to include "Memory unit is treated as MB not Byte".

mesutoezdil:
title only says iluvatar/mthreads, but nvidia and cambricon are in scope too, fix the title

archlitchi (approved):
/lgtm

PR #2295, opened 2026-08-03
feat: allow per-node memoryFactor override in device-plugin

### LINKED ISSUE #2287: When using volcano vgpu, allow setting a memory factor for each node.

When using volcanoVGPU, allow setting a memory factor for each node to improve allocation flexibility.

**Why is this needed**:

In the discussion of this issue #2187 , it was clearly pointed out that because volcano-vgpu-dp registers each memory unit as a device, it is subject to the grpc size limitation in kubelet, which results in a segmentation granularity of less than 1MB for large memory GPUs. Some problems caused by this approach have been discussed in both of these issues #2187, [#140 in volcano-vgpu-dp](https://github.com/Project-HAMi/volcano-vgpu-device-plugin/issues/140) .

I think that under the current approach, it’s reasonable to allow each node to configure the memory factor in the configmap individually and have it recognized and reported by DP. That is, when a cluster has machines with many different GPUs, machines with less than 64GB memory (like a 4-card nv3080) can have a 1MB splitting granularity, machines with less than 128GB can have 2MB, and so on, without being forced to use 11MB like machines with 640GB GPUs.

Currently, setting a uniform memory factor in device-config has some benefits, like keeping global resource semantics consistent, but it sacrifices flexibility. For smaller models, it’s better to be precise and slice on smaller cards, especially when the cluster has many resource types but very few machines of each type.

Of course, a more fundamental solution would be to change the volcano-vgpu-dp approach to be consistent with hamivgpu or ascenfvnpu, as suggested in this issue [#140 in volcano-vgpu-dp](https://github.com/Project-HAMi/volcano-vgpu-device-plugin/issues/140), but I’m not sure about the feasibility and workload of doing that (maybe it could be discussed?). So, that’s why I propose the method I mentioned above.

I noticed that responses to issues in the volcano-vgpu-dp project seem to be a bit slow, and this feature is really important to me when using Hami, so I wanted to mention it here. If necessary, I can move this issue over there.

alehhu:
Fixes #2287

### What this PR does / why we need it:
Currently, `memoryFactor` is a global setting in `NvidiaConfig`. When using Volcano vGPU DP (or in general heterogeneous clusters), different nodes might need different chunk granularities to avoid gRPC size limits on nodes with large GPUs, without forcing small GPUs to use large chunk sizes. 

This PR allows configuring `memoryfactor` on a per-node basis via the `nodeconfig` configmap array, parsing it securely and falling back to the global default if omitted.

### Implementation details:
Added `MemoryFactor` as an optional pointer (`*int32` with `omitempty`) inside `NodeDefaultConfig` to preserve backwards compatibility. In `server.go`, the specific `memoryfactor` cleanly overrides the global `sConfig.MemoryFactor` during node config initialization.

### Testing:
- `make verify` and `go test ./...` passed locally.
- *Note on Hardware Validation:* This PR only affects how the config file is parsed into structs and doesn't change the underlying device isolation/allocation logic directly, so it was tested via unit tests without needing a bare-metal multi-GPU validation.

### Release note:
```release-note
Allow setting `memoryfactor` on a per-node basis using `nodeConfiguration.config` in helm values.
```

### Known Limits:
**Scheduler vs Device Plugin Consistency:** When using the default HAMi scheduler, the MutatingAdmissionWebhook operates at the cluster level before a node is assigned, so it will continue to use the global `memoryFactor`. The per-node `memoryFactor` is only applied locally by the DP. Users relying on the HAMi scheduler must ensure their global `memoryFactor` aligns with their workloads, while Volcano users (who bypass the HAMi webhook entirely) can fully benefit from this per-node chunking.

## Summary by CodeRabbit

* **New Features**
  * Added configurable memory scaling for NVIDIA device nodes.
  * The memory factor defaults to 1 and can be customized per node.
  * Node-specific memory settings override the global configuration when provided.

### DIFF
--- a/charts/hami/values.yaml
+++ b/charts/hami/values.yaml
@@ -370,6 +370,7 @@ devicePlugin:
             "preconfigureddevicememory": 0,
             "enablenumatopology": false,
             "migstrategy": "none",
+            "memoryfactor": 1,
             "filterdevices": {
               "uuid": [],
               "index": []
--- a/pkg/device-plugin/nvidiadevice/nvinternal/plugin/server.go
+++ b/pkg/device-plugin/nvidiadevice/nvinternal/plugin/server.go
@@ -137,6 +137,9 @@ func readFromConfigFile(sConfig *nvidia.NvidiaConfig, path string) (string, erro
 			if err := mergo.Merge(&sConfig.NodeDefaultConfig, val.NodeDefaultConfig, mergo.WithOverride); err != nil {
 				return "", err
 			}
+			if val.NodeDefaultConfig.MemoryFactor != nil {
+				sConfig.MemoryFactor = *val.NodeDefaultConfig.MemoryFactor
+			}
 			if val.FilterDevice != nil && (len(val.FilterDevice.UUID) > 0 || len(val.FilterDevice.Index) > 0) {
 				nvidia.DevicePluginFilterDevice = val.FilterDevice
 			}
--- a/pkg/device-plugin/nvidiadevice/nvinternal/plugin/server_test.go
+++ b/pkg/device-plugin/nvidiadevice/nvinternal/plugin/server_test.go
@@ -629,12 +629,12 @@ func Test_pathGeneration(t *testing.T) {
 }
 
 func Test_configOverride(t *testing.T) {
-	t.Setenv("NODE_NAME", "testnode")
 	logLevel1 := nvidia.Debugs
 	logLevel2 := nvidia.Infos
 	split1 := uint(2)
 	memScale1 := 1.5
 	coreScale1 := 1.2
+	memFactor1 := int32(2)
 
 	split2 := uint(3)
 	memScale2 := 0.8
@@ -655,6 +655,7 @@ func Test_configOverride(t *testing.T) {
 					DeviceMemoryScaling: &memScale1,
 					DeviceCoreScaling:   &coreScale1,
 					LogLevel:            &logLevel1,
+					MemoryFactor:        &memFactor1,
 				},
 				Name:                         "node-1",
 				OperatingMode:                "default",
@@ -684,6 +685,9 @@ func Test_configOverride(t *testing.T) {
 	}
 	path := t.TempDir()
 	os.WriteFile(path+"/config.json", bytes, 0644)
+
+	// Test 1: testnode (no MemoryFactor override, should fall back to global value 5)
+	t.Setenv("NODE_NAME", "testnode")
 	nvconfig := nvidia.NvidiaConfig{
 		NodeDefaultConfig: nvidia.NodeDefaultConfig{
 			DeviceSplitCount:    func() *uint { v := uint(1); return &v }(),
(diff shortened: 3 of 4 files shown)


mesutoezdil:
scheduler still uses the global nvidia.MemoryFactor for FitQuota and the webhook quota check. only device plugin config is per node here. scheduler and device plugin can end up using different factors for the same node. pls cover the scheduler side too or note this as a known limit

mesutoezdil:
pls dont force push, add new commits instead. also re-check the contribution gates in CONTRIBUTING.md before pushing again

alehhu:
Hi @mesutoezdil  , sorry for the force push! It was an automatic reflex while I was fixing the tests locally. I'm a student and this is one of my first open source contributions (small fixes so far), so I'm still learning the right workflow. I will use normal commits next time.

About the scheduler issue: you're totally right. The webhook runs before the node is assigned, so it can't know the node-specific memory factor. I just updated the PR description to note this as a known limit for HAMi scheduler users. For Volcano users though, this should already solve their issue completely.
Let me know if it looks good now!

mesutoezdil:
good that you added the known limit note. one question: does volcano vgpu skip the hami mutatingwebhook completely, or does the webhook still run and just get ignored by volcano scheduling. if it still runs, fitResourceQuota still uses the global factor, so quota checks could be wrong for volcano users too, not just chunking.

alehhu:
Volcano vGPU completely skips the HAMi mutatingwebhook logic, so the global memoryFactor issue in fitResourceQuota does not affect Volcano users.

reasons:
In pkg/scheduler/webhook.go (around lines 65-69) https://github.com/Project-HAMi/HAMi/blob/master/pkg/scheduler/webhook.go#L65-L69, the webhook intercepts the
  pod creation and performs a check on the SchedulerName:

    if pod.Spec.SchedulerName != "" &&
        (pod.Spec.SchedulerName != corev1.DefaultSchedulerName || !config.ForceOverwriteDefaultScheduler) &&
        (len(config.SchedulerName) == 0 || pod.Spec.SchedulerName != config.SchedulerName) {
        klog.V(3).Infof(template+" - Pod already has different scheduler assigned", req.Namespace, req.Name, req.UID)
        return admission.Allowed("pod already has different scheduler assigned")
    }

  When a user submits a pod using Volcano, the pod.Spec.SchedulerName is explicitly set to volcano (or their custom volcano scheduler name). This means that:
  1. It is not empty.
  2. It is not the default-scheduler.
  3. It does not match the HAMi scheduler name (config.SchedulwrName).

Since all these conditions evaluate to true, the webhook logs the event and performs an early return with admission.Allowed(...).

Because the execution stops there, the fitResourceQuota(pod) function (which is located further down on line 100) is never reached or executed for Volcano pods. Therefore, the global memoryFactor value is never applied to their quota checks.

archlitchi:
i don't think a good idea to make this a 'per-node' configuration, because that will make 1 unit of 'device-memory' represent different amounts between nodes, and that makes user unable to predict how much device-memory they can actually get.

MountainGG:
@archlitchi Hi, I think it’s not really worth giving up flexibility just to make memory units semantically consistent. Right now, the dp plugin doesn’t explicitly show the semantics of the units—it’s still up to the user (or higher-level software) to read them from config files or node annotations, so reading it once or multiple times doesn’t really matter. Plus, for people, it’s easier to handle actual physical memory identified by card type, like A800-80GB. Conversion and rounding of memory units should be done by higher-level software, and then just give people a heads-up.

PR #2297, opened 2026-08-03
fix(awsneuron): guard nil CustomInfo maps

### LINKED ISSUE #2294: fix(awsneuron): nil CustomInfo write can panic in Fit/AddResourceUsage
**What happened**:

While reading the AWS Neuron backend I noticed two sites that **write** into `DeviceUsage.CustomInfo` without ensuring the map is non-nil:

1. `AddResourceUsage` (`pkg/device/awsneuron/device.go`), on a missing `AWSUsageInfo` key it does `n.CustomInfo[AWSUsageInfo] = 0`.
2. Single-card `Fit` (`Nums == 1`), on a missing key it does `dev.CustomInfo[AWSUsageInfo] = int(dev.Usedcores)`.

In Go, nil-map **reads** are safe; nil-map **writes** panic with `assignment to entry in nil map`. A minimal unit call with `CustomInfo: nil` hits that panic on both paths.

This may well be unreachable on the normal registration path, `GetNodeDevices` already allocates `CustomInfo`, so I’m mostly flagging the inconsistency rather than claiming a field crash. Same file already handles this safely elsewhere (`addCoreUsage` returns a fresh map; `graphSelect` nil-checks). Sibling class of fix: init-before-write (e.g. #2254).

I’m **not** planning to drive a fix unless maintainers think it’s worth hardening; happy to close if this is considered intentional / covered by the registration invariant.

Either `CustomInfo` is documented/guaranteed non-nil before these writes, or the write sites initialize the map when nil (same pattern as `addCoreUsage` / other backends).

Unit-only (no Neuron hardware). Pass a `DeviceUsage` with `CustomInfo: nil` into `AddResourceUsage` or single-card `Fit`, panics on the assignment above.

- Your docker or containerd configuration file (e.g: `/etc/docker/daemon.json`): N/A
- The kubelet logs on the node (e.g: `sudo journalctl -r -u kubelet`): N/A

Filed for visibility only; low urgency if `GetNodeDevices` is treated as the hard invariant.

**Environment**:
- HAMi version: current `master`
- Others: code review / unit repro only

hafzism:
**What type of PR is this?**

/kind bug

Initializes `DeviceUsage.CustomInfo` before the two AWS Neuron write paths that can receive a nil map:

- `AddResourceUsage` when recording `AWSUsageInfo`;
- the single-card `Fit` path when deriving usage from `Usedcores`.

Without these guards, either path can panic with `assignment to entry in nil map`. Focused regression tests cover both cases and verify that the existing usage values are preserved.

Fixes #2294

This is a unit-level defensive hardening change. It does not alter the normal registration path, where `GetNodeDevices` already initializes `CustomInfo`. No AWS Neuron hardware was available; validation used race-enabled unit tests for the affected backend.

Validation:

- `go test -race ./pkg/device/awsneuron -count=1`
- `make verify`
- `git diff --check`

This PR was developed with Codex assistance. I reviewed the implementation, regression coverage, and validation results before submission.

No. It prevents a defensive edge-case panic without changing normal allocation behavior.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed AWS Neuron resource allocation when device metadata is initially empty.
  * Preserved AWS Neuron usage details during resource tracking and allocation.
  * Improved successful fitting and allocation behavior for single-device requests.

### DIFF
--- a/pkg/device/awsneuron/device.go
+++ b/pkg/device/awsneuron/device.go
@@ -261,6 +261,9 @@ func (dev *AWSNeuronDevices) AddResourceUsage(pod *corev1.Pod, n *device.DeviceU
 	n.Usedcores += ctr.Usedcores
 	n.Usedmem += ctr.Usedmem
 
+	if n.CustomInfo == nil {
+		n.CustomInfo = make(map[string]any)
+	}
 	num, ok := n.CustomInfo[AWSUsageInfo]
 	if !ok || num == nil {
 		n.CustomInfo[AWSUsageInfo] = 0
@@ -393,6 +396,9 @@ func (neuron *AWSNeuronDevices) Fit(devices []*device.DeviceUsage, request devic
 	}
 	for i, v := range slices.Backward(devices) {
 		dev := v
+		if dev.CustomInfo == nil {
+			dev.CustomInfo = make(map[string]any)
+		}
 		_, ok := dev.CustomInfo[AWSUsageInfo]
 		if !ok {
 			dev.CustomInfo[AWSUsageInfo] = int(dev.Usedcores)
--- a/pkg/device/awsneuron/device_test.go
+++ b/pkg/device/awsneuron/device_test.go
@@ -402,6 +402,46 @@ func Test_GenerateResourceRequests(t *testing.T) {
 	}
 }
 
+func TestAddResourceUsageInitializesNilCustomInfo(t *testing.T) {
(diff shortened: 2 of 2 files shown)


mesutoezdil:
getnodedevices always sets a non-nil custominfo map, so this path is unreachable in the real flow. closing per the invariant the issue itself flagged.

PR #2299, opened 2026-08-03
fix(scheduler): return early when predicate body is nil

### LINKED ISSUE #2296: fix(scheduler): PredicateRoute ignores checkBody and panics on nil body
**What happened**:

`checkBody` in `pkg/scheduler/routes/route.go` correctly writes HTTP 400 when `r.Body == nil`, but `PredicateRoute` does not return after calling it. The handler continues into `io.LimitReader(r.Body, …)` / JSON decode and nil-dereferences:

`panic: runtime error: invalid memory address or nil pointer dereference` in `io.(*LimitedReader).Read`.

`TestCheckBodyNil` only exercises `checkBody` in isolation, so it does not catch the missing early return in the handler.

When the request body is nil, the handler should stop after the 400 from `checkBody` and not touch `r.Body` again.

**How to reproduce it**:

```go
req := httptest.NewRequest("POST", "/filter", nil)
req.Body = nil
w := httptest.NewRecorder()
PredicateRoute(&scheduler.Scheduler{})(w, req, nil) // panics
```
Suggested fix: make `checkBody` return a bool (or inline the check) and `return` from `PredicateRoute` on failure; add a `PredicateRoute` regression test with nil body. Bind currently does not call `checkBody` at all (separate from this bug).

**Anything else we need to know**
- Your docker or containerd configuration file (e.g:` /etc/docker/daemon.json`): N/A
- The kubelet logs on the node (e.g: `sudo journalctl -r -u kubelet`): N/A

Environment:

- HAMi version: current `master`
- Others: unit-reproducible; no cluster required

hemantrajputt:
## What changed

- Return immediately from `PredicateRoute` when the request body is nil.
- Make `checkBody` report whether the body is valid.
- Add a regression test for the nil-body handler path.

## Testing

- `go test ./pkg/scheduler/routes -run '^TestPredicateRoute_NilBody$' -count=1`
- `go test ./pkg/scheduler/routes -short -race -count=1`
- `go test ./pkg/scheduler/... -short -race -count=1`

Fixes #2296

## AI assistance

I used Codex to inspect the code, implement the fix, and run the local tests.

## Summary by CodeRabbit

* **Bug Fixes**
  * Requests with missing bodies now receive a clear HTTP 400 response.
  * Processing stops immediately after rejecting requests without a body.

* **Tests**
  * Added coverage to verify correct handling of requests with nil bodies.

### DIFF
--- a/pkg/scheduler/routes/route.go
+++ b/pkg/scheduler/routes/route.go
@@ -32,18 +32,21 @@ import (
 
 const maxRequestSize = 1024 * 1024 // 1MB limit
 
-func checkBody(w http.ResponseWriter, r *http.Request) {
+func checkBody(w http.ResponseWriter, r *http.Request) bool {
 	if r.Body == nil {
 		http.Error(w, "Please send a request body", 400)
-		return
+		return false
 	}
+	return true
 }
 
 func PredicateRoute(s *scheduler.Scheduler) httprouter.Handle {
 	klog.Infoln("Initializing Predicate Route")
 	return func(w http.ResponseWriter, r *http.Request, _ httprouter.Params) {
 		klog.V(5).Infoln("Entering Predicate Route handler")
-		checkBody(w, r)
+		if !checkBody(w, r) {
+			return
+		}
 
 		var buf bytes.Buffer
 		// Limit the body size to prevent deep nesting/resource exhaustion attacks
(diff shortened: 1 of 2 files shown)


blackdragoon26:
Looks all good to me, I may have suggestion for you as this is your first PR in HAMi, Don't force push, maintainers over here like to see how PRs evolved,  so ofc you can stack multiple commits over in a single PR, and for better commit messages please refer to https://github.com/Project-HAMi/HAMi?tab=contributing-ov-file#code-review .
Thanks, GL!

mesutoezdil:
1 small thing: commit message says sorting issue 2296 but this change has nth to do with sorting, pls use a message that describes the actual fix.

Shouren (approved):
/lgtm

PR #2306, opened 2026-08-03
fix(device): don't panic on asymmetric GPU P2P link data in calculateGPUPairScore

### LINKED ISSUE #2228: Bug: Scheduler crashes due to panic in calculateGPUPairScore on P2P link mismatch
What happened?
When the HAMi scheduler evaluates GPU P2P links in calculateGPUPairScore, it expects the links between any two GPUs to be bidirectional. If the link counts mismatch (due to a hardware misconfiguration, topology defect, or driver glitch), the scheduler triggers a panic(err). This abruptly crashes the entire HAMi scheduler pod, taking down the scheduling component completely instead of skipping or downgrading the score for the affected GPUs.

What did you expect to happen?
The scheduler should gracefully handle the link topology mismatch. Instead of panicking and crashing the pod, it should log a descriptive error indicating the P2P link mismatch and return a default score (e.g., 0) so the allocator can continue functioning with the remaining healthy devices.

How can we reproduce it (as minimally and precisely as possible)?
This occurs automatically if the HAMi scheduler encounters a node where the NVIDIA NVML library reports asymmetrical P2P links between two GPUs, which triggers the panic in calculateGPUPairScore:

go

if len(gpu0.Links[gpu1.Index]) != len(gpu1.Links[gpu0.Index]) {
    err := fmt.Errorf("internal error in bestEffort GPU allocator: all P2PLinks between 2 GPUs should be bidirectional")
    panic(err) // <--- Crash
}
Anything else we need to know?
Kubernetes infrastructure components should avoid using panic() for recoverable hardware/driver states. This is a severe reliability flaw since one misconfigured node can crash the central scheduler.

HAMi Version
Latest master branch

Brijesh-Thakkar:
/kind bug

`calculateGPUPairScore` panicked when NVML reported asymmetric P2P/NVLink data
between a GPU pair, crashing the process that called it.

Correction to the original issue title/report: this function is only called
from the `nvidia-device-plugin` DaemonSet's background registration loop
(`pkg/device-plugin/.../register.go`), gated behind `ENABLE_TOPOLOGY_SCORE=true`
(set when `gpuSchedulerPolicy: topology-aware`). It is never called from the
central HAMi scheduler (`cmd/scheduler`). The actual blast radius is the
device-plugin pod on the affected node entering CrashLoopBackOff — not a
cluster-wide scheduling outage.

Asymmetric NVLink/P2P data between two GPUs is a legitimate, persistent
hardware/driver state (e.g. one-sided NVLink lane/port failure, or a driver
re-init that completes asymmetrically across two GPUs), not a software bug in
this codebase — NVML is accurately reporting a real asymmetric topology. Since
this state is persistent, crashing and retrying never resolves it: the
device-plugin would panic on every registration cycle indefinitely
(CrashLoopBackOff), permanently losing that node's GPU reporting until someone
notices and patches the code or replaces the hardware.

This PR:
- Replaces the `panic()` in `calculateGPUPairScore` with a `klog.Warning` log
  (GPU indices/UUIDs and link counts) and a returned score of `0` for the
  affected pair, instead of aborting the whole scoring pass.
- Leaves all other GPU pairs on the same node scored normally — we don't
  discard valid topology data for the rest of the node over one degraded pair.
- Emits a `Warning` Event (`AsymmetricGPUP2PLink`) on the Node object when any
  pair on that node is asymmetric, so operators can discover the degraded
  hardware via `kubectl describe node` or standard Event-based monitoring,
  instead of the plugin failing silently or crash-looping forever.
- Deduplicates Event emission by querying the API server directly (list
  existing events for the node+reason, update `Count`/`LastTimestamp` if one
  exists within a 1h window, otherwise create) rather than using client-go's
  `record.EventRecorder`. A recorder's dedup state is in-memory and would be
  wiped on every process restart — and this exact failure mode causes
  repeated restarts, so an in-memory recorder would re-emit on every
  crash-restart cycle. Querying the API server directly makes dedup durable
  across restarts.

**Testing**:
- New unit test `Test_calculateGPUPairScore_asymmetric` reproduces the
  one-sided NVLink case and asserts no panic, score `0`, `asymmetric == true`.
- New unit test suite `TestEmitNodeWarningEvent` (fake clientset) covers: no
  existing event → Create, existing event within dedup window → Update (Count
  incremented, message updated, no duplicate object), existing event outside
  the window → Create.
- Full existing test suites for `pkg/device/nvidia`, `pkg/util`, and
  `pkg/device-plugin` pass. Two pre-existing, unrelated failures
  (`pkg/util/client` — requires in-cluster/kubeconfig env; `rm.TestNvmlMigDevice_GetPaths`
  — requires MIG hardware) were confirmed present identically on `master` via
  a `git stash` before/after comparison, and are unaffected by this change.
- `golangci-lint run` and `go vet` are clean on all changed packages.

Fixes #2228

- `calculateGPUPairScore`, `calculateGPUScore`, and `CalculateGPUScore` all
  gained an additional `bool` return value (`asymmetric` / `anyAsymmetric`) to
  propagate the mismatch signal up to the Event-emission call site. Confirmed
  via full-repo grep that `calculate_score.go:194` is the only production call
  site for `calculateGPUPairScore`, and `register.go` is the only caller of
  `CalculateGPUScore`, so this signature change has no other blast radius.
- Confirmed the downstream allocator (`computeWorstSingleCard`,
  `computeBestCombination` in `device.go`) reads scores from a plain
  `map[string]int` with no symmetry assumption — a `0` score for one direction
  of a pair is handled the same as a missing entry, so this fix is safe against
  existing allocation logic without further changes there.
- Happy to move the Event emission behind a feature flag or drop it entirely
  if maintainers prefer a smaller-scoped fix (i.e. just the panic → 0-score
  change without Event/dedup logic) — let me know and I'll split it out.

```
Fixed a bug where the nvidia-device-plugin would panic and crash-loop when
NVML reported asymmetric P2P/NVLink data between a GPU pair (e.g. due to a
one-sided hardware/driver link failure) while topology-aware scheduling was
enabled. The affected GPU pair is now scored as having no usable P2P link
instead of crashing the plugin, and a Warning Event is emitted on the Node
object so operators can detect the underlying hardware issue.
```
**AI assistance disclosure:**
This PR was developed with AI assistance (Claude Code) used for investigation,
analysis, and initial implementation, under my direct supervision and review.
I reviewed the full diff, ran and verified all tests/lint locally, and can
explain the reasoning behind every design decision.**

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented crashes when asymmetric GPU interconnects are detected.
  * GPUs with asymmetric peer-to-peer links now receive a score of zero instead of causing failures.
* **New Features**
  * Added Kubernetes Node warning events for detected asymmetric GPU connectivity.
  * Repeated warnings are consolidated and updated for one hour to reduce duplicate notifications.
* **Tests**
  * Added coverage for asymmetric topology handling and warning-event deduplication.

### DIFF
--- a/pkg/util/util.go
+++ b/pkg/util/util.go
@@ -22,6 +22,7 @@ import (
 	"flag"
 	"fmt"
 	"strings"
+	"time"
 
 	"github.com/Project-HAMi/HAMi/pkg/util/client"
 	"github.com/Project-HAMi/HAMi/pkg/util/nodelock"
@@ -279,6 +280,77 @@ func IsPodTerminating(pod *corev1.Pod) bool {
 
 func AllContainersCreated(pod *corev1.Pod) bool {
 	return len(pod.Status.ContainerStatuses) >= len(pod.Spec.Containers)
+}
+
+// EmitNodeWarningEvent emits a Warning event on the given Node with deduplication.
+func EmitNodeWarningEvent(node *corev1.Node, reason, message string, dedupWindow time.Duration) {
+	c := client.GetClient()
+	if c == nil {
+		klog.Warningf("cannot emit node event for %s: Kubernetes client not initialized", node.Name)
+		return
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+	fieldSel := fmt.Sprintf(
+		"involvedObject.kind=Node,involvedObject.name=%s,involvedObject.uid=%s,reason=%s",
+		node.Name, string(node.UID), reason,
+	)
+	existing, err := c.CoreV1().Events(corev1.NamespaceDefault).List(ctx, metav1.ListOptions{
+		FieldSelector: fieldSel,
+	})
+	if err != nil {
+		klog.Warningf("failed to list events for node %s: %v; will attempt create", node.Name, err)
+	}
+
+	now := metav1.Now()
+
+	if err == nil && len(existing.Items) > 0 {
+		// Client-side filter: the field selector is a server-side optimization; re-check
+		// here so the function is correct even against fake clients or non-compliant servers.
+		var latest *corev1.Event
+		for i := range existing.Items {
+			ev := &existing.Items[i]
+			if ev.InvolvedObject.UID != node.UID || ev.Reason != reason {
+				continue
+			}
+			if latest == nil || ev.LastTimestamp.After(latest.LastTimestamp.Time) {
+				latest = ev
+			}
+		}
+		if latest != nil && now.Sub(latest.LastTimestamp.Time) <= dedupWindow {
+			latest.Count++
+			latest.LastTimestamp = now
+			latest.Message = message
+			if _, err := c.CoreV1().Events(corev1.NamespaceDefault).Update(ctx, latest, metav1.UpdateOptions{}); err != nil {
+				klog.Warningf("failed to update node event for %s: %v", node.Name, err)
+			}
+			return
+		}
+	}
+
+	event := &corev1.Event{
+		ObjectMeta: metav1.ObjectMeta{
+			GenerateName: node.Name + "-",
+			Namespace:    corev1.NamespaceDefault,
+		},
+		InvolvedObject: corev1.ObjectReference{
+			APIVersion: "v1",
+			Kind:       "Node",
+			Name:       node.Name,
+			UID:        node.UID,
+		},
+		Reason:         reason,
+		Message:        message,
+		Type:           corev1.EventTypeWarning,
+		Count:          1,
+		FirstTimestamp: now,
+		LastTimestamp:  now,
+		Source:         corev1.EventSource{Component: "hami-device-plugin"},
+	}
+	if _, err := c.CoreV1().Events(corev1.NamespaceDefault).Create(ctx, event, metav1.CreateOptions{}); err != nil {
+		klog.Warningf("failed to create node event for %s: %v", node.Name, err)
+	}
 }
 
 func IsPodGroupMember(pod *corev1.Pod) bool {
--- a/charts/hami/templates/device-plugin/monitorrole.yaml
+++ b/charts/hami/templates/device-plugin/monitorrole.yaml
@@ -24,6 +24,14 @@ rules:
       - update
       - list
       - patch
+  - apiGroups:
+      - ""
+    resources:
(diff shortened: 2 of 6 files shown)


Eshiv-Pandey:
No ai disclosure is present pls refer to [CONTRIBUTING.md](https://github.com/Project-HAMi/HAMi/blob/master/CONTRIBUTING.md) before making any contribution.

Eshiv-Pandey:
Also dco check is failing pls sign-off your commits.

mesutoezdil:
dco check fails, missing signed-off-by. pls sign off your commit and push again.

pkg/util/util.go
```diff
@@ -281,6 +282,91 @@ func AllContainersCreated(pod *corev1.Pod) bool {
...
+		for i := range existing.Items {
+			ev := &existing.Items[i]
+			if ev.InvolvedObject.UID != node.UID || ev.Reason != reason {
+				continue
+			}
+			if latest == nil || ev.LastTimestamp.After(latest.LastTimestamp.Time) {
+				latest = ev
+			}
+		}
+		if latest != nil && now.Sub(latest.LastTimestamp.Time) <= dedupWindow {
```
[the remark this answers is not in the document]
Brijesh-Thakkar: i checked metav1.Time  and it embeds time.Time so .Sub is promoted , like both compile in same way and i verifed it against the apimachinery as a source and observed clean build , so ig no need to do this changes
Free to discuss this
Thank you

pkg/util/util.go
```diff
@@ -281,6 +282,92 @@ func AllContainersCreated(pod *corev1.Pod) bool {
 	return len(pod.Status.ContainerStatuses) >= len(pod.Spec.Containers)
 }
 
+// EmitNodeWarningEvent emits a Warning event on the given Node with deduplication.
```
archlitchi, line 285: This comment is too long, only reserve the first sentence is sufficient

Brijesh-Thakkar: Trimmed the comment down to the first sentence

Brijesh-Thakkar:
@chaunceyjiang @DSFans2014 @ouyangluwei163 
Please review this PR whenever u get time
thankss

archlitchi (approved):
/lgtm

PR #2311, opened 2026-08-03
fix(monitor): return locked snapshot from ListContainers to prevent race with Update

### LINKED ISSUE #2309: bug(vGPUmonitor): ListContainers races with Update — unlocked map iteration and possible use-after-munmap
## What happened

`ContainerLister.ListContainers()` returns the internal `containers` map directly without taking the lister mutex:

https://github.com/Project-HAMi/HAMi/blob/master/pkg/monitor/nvidia/cudevshr.go#L156-L158

Meanwhile `Update()` mutates the same map **under** the mutex, and also `syscall.Munmap`s the shared-memory backing of removed entries:

- `Update()` is called every 5 seconds from the feedback loop (`cmd/vGPUmonitor/feedback.go`)
- `ListContainers()` is iterated concurrently by the metrics collector (`cmd/vGPUmonitor/metrics.go` — `collectPodAndContainerInfo`) and by `Observe()` in the same feedback file

Two consequences:

1. **Data race**: concurrent map read (collector iterating) + map write (`Update()` adding/deleting entries). Caught by `-race`; undefined behavior in production.
2. **Use-after-munmap**: `Update()` munmaps `c.data` for stale entries and deletes them from the map, but a collector goroutine holding the same `*ContainerUsage` can still read `c.Info` — which dereferences the now-unmapped region. That is a SIGBUS crash of the whole vGPUmonitor DaemonSet pod, taking node metrics and the utilization-switch feedback loop down with it.

## Additional hardening issue in the same function

`Update()` derives the container name with `strings.Split(entry.Name(), "_")[1]`. Any directory in the containers hook path whose name has no underscore panics with index out of range — again crashing the monitor. The hook path is writable by node tooling, so a stray directory should be skipped, not fatal.

## Suggested fix

- `ListContainers()` takes the mutex and returns a copied snapshot map. All existing callers just range over the result, so a snapshot is a drop-in.
- `SplitN(name, "_", 2)` + length check; skip malformed names with a warning.

The snapshot shrinks the munmap race window to raw pointer reads of entries removed mid-scrape; full elimination would need refcounting the mmap lifetime, which seems overkill next to the current always-racy behavior. Can be noted as a follow-up.

I have a fix with unit tests (snapshot semantics + a concurrent Update/ListContainers test that fails under `-race` on the current code) ready to submit.

Aamod007:
## What type of PR is this?

/kind bug

`ListContainers()` returned the internal `containers` map without holding the lister mutex, while `Update()` — called every 5s from the feedback loop — mutates that map and `Munmap`s removed entries under the lock. The metrics collector and `Observe()` iterate the same map concurrently, which is:

1. a data race (concurrent map read/write, fails `-race`), and
2. a potential use-after-munmap SIGBUS when a collector dereferences `c.Info` for an entry whose shared-memory region `Update()` just unmapped — crashing the entire vGPUmonitor DaemonSet pod.

This PR makes `ListContainers()` copy the map under the mutex and return the snapshot. All existing callers only range over the result, so the snapshot is a drop-in replacement.

It also hardens `Update()` against container directory names without an underscore: previously `strings.Split(entry.Name(), "_")[1]` panicked with index out of range on any stray directory in the hook path; now such entries are skipped with a warning.

Note: the snapshot shrinks the munmap race window to entries removed mid-scrape; fully eliminating it would require refcounting the mmap lifetime, which I left as a possible follow-up given the current behavior is always racy.

Fixes #2309

## Tests

- `Test_ListContainers_snapshot` — snapshot is isolated from subsequent internal-map mutation
- `Test_ListContainers_concurrentUpdate` — drives `Update()` and `ListContainers()` concurrently; fails under `-race` on the previous implementation
- `Test_ContainerLister_Update` — new subtest: dirname without underscore is skipped, not panicked on

## AI assistance disclosure

Per CONTRIBUTING.md: AI tooling was used to help draft this change; I reviewed and verified the code, race analysis, and tests.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented returned NVIDIA container listings from modifying monitored state.
  * Improved safety when container updates and listings occur concurrently.
  * Improved reliability of container monitoring during active updates.

* **Tests**
  * Added regression coverage to verify listing snapshots remain independent from internal monitoring state.
  * Added concurrent-access tests to validate safe iteration while container data is updated and cached resources are cleaned up.

### DIFF
--- a/pkg/monitor/nvidia/cudevshr.go
+++ b/pkg/monitor/nvidia/cudevshr.go
@@ -19,6 +19,7 @@ package nvidia
 import (
 	"errors"
 	"fmt"
+	"maps"
 	"os"
 	"path/filepath"
 	"strings"
@@ -160,9 +161,7 @@ func (l *ContainerLister) ListContainers() map[string]*ContainerUsage {
 	l.mutex.Lock()
 	defer l.mutex.Unlock()
 	snapshot := make(map[string]*ContainerUsage, len(l.containers))
-	for k, v := range l.containers {
-		snapshot[k] = v
-	}
+	maps.Copy(snapshot, l.containers)
 	return snapshot
 }
 
--- a/pkg/monitor/nvidia/cudevshr.go
+++ b/pkg/monitor/nvidia/cudevshr.go
@@ -153,8 +153,17 @@ func (l *ContainerLister) UnLock() {
 	l.mutex.Unlock()
 }
 
+// ListContainers returns a snapshot of the current container usages. It copies
+// the internal map under the lock so that callers can safely iterate the result
+// while Update() concurrently adds or removes entries.
 func (l *ContainerLister) ListContainers() map[string]*ContainerUsage {
-	return l.containers
+	l.mutex.Lock()
+	defer l.mutex.Unlock()
+	snapshot := make(map[string]*ContainerUsage, len(l.containers))
+	for k, v := range l.containers {
+		snapshot[k] = v
+	}
+	return snapshot
 }
 
 func (l *ContainerLister) Clientset() *kubernetes.Clientset {
--- a/pkg/monitor/nvidia/cudevshr_test.go
+++ b/pkg/monitor/nvidia/cudevshr_test.go
@@ -463,3 +463,62 @@ func Test_ContainerLister_Update(t *testing.T) {
 		assert.Equal(t, len(l.containers), 0)
 	})
 }
+
+// Test_ListContainers_snapshot verifies ListContainers returns a copy that is
+// safe to iterate while Update() concurrently mutates the internal map.
+func Test_ListContainers_snapshot(t *testing.T) {
+	l := &ContainerLister{
+		containers: map[string]*ContainerUsage{
+			"a_ctr": {PodUID: "a", ContainerName: "ctr"},
+		},
+	}
+	snapshot := l.ListContainers()
+	assert.Equal(t, len(snapshot), 1)
+
+	// Mutating the internal map must not affect the snapshot.
+	l.mutex.Lock()
+	delete(l.containers, "a_ctr")
+	l.containers["b_ctr"] = &ContainerUsage{PodUID: "b", ContainerName: "ctr"}
+	l.mutex.Unlock()
+
+	_, ok := snapshot["a_ctr"]
+	assert.Equal(t, ok, true)
+	_, ok = snapshot["b_ctr"]
+	assert.Equal(t, ok, false)
+}
+
+// Test_ListContainers_concurrentUpdate drives ListContainers and Update
+// concurrently; under -race this fails on the previous implementation, which
+// returned the internal map without holding the lock.
+func Test_ListContainers_concurrentUpdate(t *testing.T) {
+	dir := t.TempDir()
+	pod := &corev1.Pod{ObjectMeta: metav1.ObjectMeta{Name: "p", Namespace: "default", UID: "uidr"}}
+	entryName := "uidr_ctr"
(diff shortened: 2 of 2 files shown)


archlitchi:
please sign-off your commit

Aamod007:
@archlitchi Done — amended the commit with sign-off and force-pushed. Thanks for the review!

archlitchi:
please resolve these conflicts, we've dealt with unavailable directory name in https://github.com/Project-HAMi/HAMi/pull/2371

imantaba:
The `TestDescribeCollectSync` 10m hang is a real deadlock from this change.
'ListContainers()' now locks `l.mutex`, but the metrics collector already holds that lock across the scrape ([metrics.go:380-383](https://github.com/Project-HAMi/HAMi/blob/master/cmd/vGPUmonitor/metrics.go#L380-L383), from #2194 ). `sync.Mutex` is not reentrant -> locks twice -> hangs. Master alone passes: only shows up rebased on current master.

Fix : 
```
// Containers returns the live internal map without locking; caller must already hold Lock().
func (l *ContainerLister) Containers() map[string]*ContainerUsage {
	return l.containers
}
```
Collector then calls `.Containers()` instead of `.ListContainers()`

imantaba:
Fix : apply this `git apply fix.diff`
```diff
diff --git a/cmd/vGPUmonitor/metrics.go b/cmd/vGPUmonitor/metrics.go
index f943a63..317dfc6 100644
--- a/cmd/vGPUmonitor/metrics.go
+++ b/cmd/vGPUmonitor/metrics.go
@@ -380,7 +380,9 @@ func (cc ClusterManagerCollector) collectPodAndContainerInfo(ch chan<- prometheu
 	cc.ClusterManager.containerLister.Lock()
 	defer cc.ClusterManager.containerLister.UnLock()
 
-	containers := cc.ClusterManager.containerLister.ListContainers()
+	// We already hold the lister lock above, so use the non-locking accessor;
+	// ListContainers() would re-lock the non-reentrant mutex and deadlock.
+	containers := cc.ClusterManager.containerLister.Containers()
 	containerMap := make(map[string][]*nvidia.ContainerUsage) // podUID -> containers
 	for _, c := range containers {
 		if c.Info != nil && c.PodUID != "" {
diff --git a/pkg/monitor/nvidia/cudevshr.go b/pkg/monitor/nvidia/cudevshr.go
index 9b5b461..b6e567c 100644
--- a/pkg/monitor/nvidia/cudevshr.go
+++ b/pkg/monitor/nvidia/cudevshr.go
@@ -165,6 +165,14 @@ func (l *ContainerLister) ListContainers() map[string]*ContainerUsage {
 	return snapshot
 }
 
+// Containers returns the live internal map without locking. The caller must
+// already hold Lock(); the metrics collector does so to keep the mmap'd
+// ContainerUsage.Info alive across a scrape, so it cannot use the locking
+// ListContainers() above without deadlocking on the non-reentrant mutex.
+func (l *ContainerLister) Containers() map[string]*ContainerUsage {
+	return l.containers
+}
+
 func (l *ContainerLister) Clientset() *kubernetes.Clientset {
 	return l.clientset
 }
```

FouoF:
Thanks for investigating the container-map race. The current implementation introduces a deterministic deadlock: the metrics collector already holds this non-reentrant mutex when it calls ListContainers(), and the new code attempts to acquire the same mutex again. The unit-test failure reflects this regression. The unrelated directory fix included in the branch has already been merged through #2371. We are closing this implementation. If the underlying mmap lifetime race still exists, please open a new, focused proposal that defines lock ownership and includes a regression test for both concurrent update safety and scrape completion.

PR #2313, opened 2026-08-03
fix(quota): honor an explicit zero limit in ResourceQuota

### LINKED ISSUE #2312: bug(quota): FitQuota treats an explicit zero limit as unlimited, so ResourceQuota "0" allows all GPU usage
**What happened**

Setting a HAMi device resource to `0` in a `ResourceQuota` (for example `limits.nvidia.com/gpumem: "0"`) — which an operator would use to block all GPU memory usage in a namespace — is silently treated as "no limit", and every pod is admitted.

In `pkg/device/quota.go`, `FitQuota` gates the check on `limit != 0`:

```go
// memory (~line 78)
    return false
}
// cores (~line 84)
if ok && coreQuota.Limit != 0 && coreQuota.Used+coresreq > coreQuota.Limit {
    return false
}
```

`AddQuota` does store the explicit `0` (`(*dp)[dn].Limit = value`), but because `FitQuota` uses `Limit != 0` to decide whether a quota is configured, an explicit zero is indistinguishable from "unset" and is treated as unlimited. A namespace the operator intended to fully block ends up with no limit at all.

**What I expected**

An explicit `0` limit should block usage (admission denied), not be ignored. Only a genuinely absent quota should mean "no limit".

**How to reproduce**

1. Create a `ResourceQuota` in a namespace with `limits.nvidia.com/gpumem: "0"`.
2. Submit a pod requesting `nvidia.com/gpumem`.
3. The pod is admitted, even though the namespace quota for that resource is zero.

**Root cause**

`Limit == 0` is overloaded to mean both "explicitly set to zero" and "no limit configured", and `FitQuota` cannot tell the two apart.

**Prior art**

This was reported and fixed before in #1904 (@iasthc), which added a `LimitSet` flag to distinguish an explicit zero from an auto-initialized entry. That PR was closed on July 17 by the activity policy without review, not on merit, so the bug is still present on master.

**Suggested fix**

Restore the "set vs unset" distinction (the `LimitSet` flag from #1904, or a sentinel) and have `FitQuota` gate on "is a limit configured" rather than `Limit != 0`. I'd be glad to take this — I have a regression test that fails on the current code and passes with the fix. I'm flagging it as an issue first to confirm the intended semantics before opening a PR, given the history on #1904.

/kind bug

veyron-kairo:
**What type of PR is this?**
/kind bug

`FitQuota` used `Limit != 0` to decide whether a quota was configured. As a
result, setting `limits.nvidia.com/gpumem: "0"` (or `gpucores: "0"`) in a
ResourceQuota — which an operator would use to block all GPU usage in a
namespace — was silently treated as "no limit", and every pod was admitted
instead of being denied.

This adds a `LimitSet` flag to `Quota` to distinguish an explicitly configured
limit (including an explicit `0`) from an entry auto-created by usage tracking.
`AddQuota` sets it, `DelQuota` clears it, and `FitQuota` now gates on `LimitSet`
instead of `Limit != 0`, so an explicit zero is honored as a hard block.
Usage-tracking entries keep `LimitSet` false and behave exactly as before.

Fixes #2312

This revives the fix from #1904 by @iasthc, which was closed by the activity
policy without review rather than on merit — full credit to them for the
original approach. I've added a regression test (`TestFitQuotaExplicitZeroBlocks`)
that admits a request under a `"0"` quota on the pre-fix code and blocks it after,
and updated `TestFitQuota` so its manually-built quotas carry `LimitSet`.
`make verify` and the `pkg/device` unit tests pass locally.

**AI assistance disclosure**:
I used AI assistance (Claude Code) while confirming the root cause .
wrote  the regression test by myself . I reviewed and verified the change, ran `make verify` locally,
and take responsibility for it.

NONE

## Summary by CodeRabbit

* **Bug Fixes**
  * Quota enforcement now correctly applies explicitly configured limits, including zero-valued memory, CPU core, and GPU memory limits.
  * Quota usage and copied quota data preserve whether a limit was intentionally configured.
  * Removing a quota now clears its configured-limit status, preventing stale restrictions from affecting future resource requests.

### DIFF
--- a/pkg/device/quota.go
+++ b/pkg/device/quota.go
@@ -27,6 +27,12 @@ import (
 type Quota struct {
 	Used  int64
 	Limit int64
+	// LimitSet distinguishes an explicitly configured limit (including an
+	// explicit 0, which blocks all usage) from an entry auto-created by usage
+	// tracking, which carries Limit 0 but means "no limit configured". FitQuota
+	// gates on this rather than Limit != 0 so that a ResourceQuota of "0" is
+	// honored as a hard block instead of being read as unlimited.
+	LimitSet bool
 }
 
 type DeviceQuota map[string]*Quota
@@ -75,13 +81,13 @@ func (q *QuotaManager) FitQuota(ns string, memreq int64, memoryFactor int32, cor
 		if memoryFactor > 1 {
 			limit = limit * int64(memoryFactor)
 		}
-		if limit != 0 && memQuota.Used+memreq > limit {
+		if memQuota.LimitSet && memQuota.Used+memreq > limit {
 			klog.V(4).InfoS("resourceMem quota not fitted", "limit", limit, "used", memQuota.Used, "alloc", memreq)
 			return false
 		}
 	}
 	coreQuota, ok := (*dq)[coreResourceName]
-	if ok && coreQuota.Limit != 0 && coreQuota.Used+coresreq > coreQuota.Limit {
+	if ok && coreQuota.LimitSet && coreQuota.Used+coresreq > coreQuota.Limit {
 		klog.V(4).InfoS("resourceCores quota not fitted", "limit", coreQuota.Limit, "used", coreQuota.Used, "alloc", coresreq)
 		return false
 	}
@@ -210,6 +216,7 @@ func (q *QuotaManager) AddQuota(quota *corev1.ResourceQuota) {
 				}
 			}
 			(*dp)[dn].Limit = value
+			(*dp)[dn].LimitSet = true
 			klog.V(4).InfoS("quota set:", "idx=", idx, "val", value)
 		}
 	}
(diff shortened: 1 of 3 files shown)


mesutoezdil:
pkg/scheduler/webhook_test.go:272-273 also builds Quota{} directly and needs LimitSet: true, that's why TestFitResourceQuota fails in CI.

veyron-kairo:
Good catch, thanks . that's the one I missed. TestFitResourceQuota builds the Quota values by hand instead of going through AddQuota, so LimitSet stayed false and FitQuota read both limits as unconfigured, which is why the two denial cases got admitted.

archlitchi:
Have you tested on your environment?

archlitchi (approved):
/lgtm

PR #2314, opened 2026-08-03
fix(vGPUmonitor): skip exited process slots in v1 spec metric aggregations

### LINKED ISSUE #2310: bug(vGPUmonitor): v1 spec aggregations ignore `status` field, inflating memory metrics for dead process slots
## Description

In `pkg/monitor/nvidia/v1/spec.go`, all aggregation functions loop through `procs[:procnum]` without checking `p.status`. This means the metrics include memory and utilization from process slots that have already exited.

### What's wrong

- `shrregProcSlotT` has a `status int32` field, which `libvgpu` uses to indicate if a process slot is alive.
- All 6 aggregation functions (`DeviceMemoryTotal`, `DeviceSmUtil`, `DeviceMemoryContextSize`, `DeviceMemoryModuleSize`, `DeviceMemoryBufferSize`, `DeviceMemoryOffset`) completely miss this check.
- When a CUDA process exits, the `used[idx]` memory values in its slot are not zeroed right away; only `status` is reset.
- The monitor adds those outdated values to the next scrape, incorrectly reporting memory as still in use.

### Impact

- `hami_vgpu_memory_used_bytes` is inflated after any process exits until the slot is reused.
- False "high memory" alerts trigger on inactive workloads.
- Operators see non-zero memory for containers that have already finished.

### Fix

Add `if p.status == 0 { continue }` inside the loop in all 6 functions in `pkg/monitor/nvidia/v1/spec.go`.

### Related

- The analogous v0 bug (iterating all 1024 slots with no `procnum` bound) is tracked in #2281, with a proposed fix in #2282 (not yet merged).
- v1 limited iteration by `procnum`, but the `status` check was never included.
- This issue is tracked under the observability category: #2126.

Nakshatra480:
## Summary

- The six per-device metric accessors in `pkg/monitor/nvidia/v1/spec.go` (`DeviceMemoryContextSize`, `DeviceMemoryModuleSize`, `DeviceMemoryBufferSize`, `DeviceMemoryOffset`, `DeviceMemoryTotal`, `DeviceSmUtil`) were iterating over `procs[:procnum]` without checking `p.status`, so slots belonging to processes that had already exited were still included in the sum.

- When a CUDA process exits, libvgpu sets `status = 0` on its slot but does not immediately zero out the memory fields. This meant the monitor was reporting inflated memory and utilization values until the slot got reused by the next process.

- Added a `procBound()` helper that clamps `procnum` to `[0, len(procs)]` before slicing, since `procnum` comes directly from the mmap'd shared-memory region and a corrupt value would otherwise cause a slice-bounds panic.

- Each aggregation function now skips slots where `p.status == 0`, so only live process slots contribute to the reported metrics.

- Updated existing tests to set `status: 1` on active slots, and added regression cases covering dead-slot exclusion, negative `procnum`, and oversized `procnum` for all six functions.

Fixes #2310

The analogous v0 bug (iterating all 1024 slots with no `procnum` bound at all) is tracked separately in #2281. This PR only touches the v1 layout.

Yes — `hami_vgpu_memory_used_bytes` and related vGPUmonitor gauges will no longer be inflated after a CUDA process exits inside a container.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved NVIDIA monitoring accuracy by excluding inactive process slots from context, module, buffer, offset, total memory, and SM utilization calculations.
  * Added safeguards for invalid process counts, including negative and oversized values, while preserving valid active-process data.

* **Tests**
  * Expanded coverage for inactive process slots and process-count boundary conditions across NVIDIA monitoring versions.
  * Corrected a buffer-size test case label.

### DIFF
--- a/pkg/monitor/nvidia/v0/spec.go
+++ b/pkg/monitor/nvidia/v0/spec.go
@@ -92,6 +92,9 @@ func (s Spec) activeProcs() []shrregProcSlotT {
 func (s Spec) DeviceMemoryContextSize(idx int) uint64 {
 	v := uint64(0)
 	for _, p := range s.activeProcs() {
+		if p.status == 0 {
+			continue
+		}
 		v += p.used[idx].contextSize
 	}
 	return v
@@ -100,6 +103,9 @@ func (s Spec) DeviceMemoryContextSize(idx int) uint64 {
 func (s Spec) DeviceMemoryModuleSize(idx int) uint64 {
 	v := uint64(0)
 	for _, p := range s.activeProcs() {
+		if p.status == 0 {
+			continue
+		}
 		v += p.used[idx].moduleSize
 	}
 	return v
@@ -108,6 +114,9 @@ func (s Spec) DeviceMemoryModuleSize(idx int) uint64 {
 func (s Spec) DeviceMemoryBufferSize(idx int) uint64 {
 	v := uint64(0)
 	for _, p := range s.activeProcs() {
+		if p.status == 0 {
+			continue
+		}
 		v += p.used[idx].bufferSize
 	}
 	return v
@@ -116,6 +125,9 @@ func (s Spec) DeviceMemoryBufferSize(idx int) uint64 {
 func (s Spec) DeviceMemoryOffset(idx int) uint64 {
 	v := uint64(0)
 	for _, p := range s.activeProcs() {
+		if p.status == 0 {
+			continue
+		}
 		v += p.used[idx].offset
 	}
 	return v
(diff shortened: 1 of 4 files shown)


mesutoezdil:
v0/spec.go has same bug, as we see no status check and no activeprocs clamp either 

dead proc slots still summed into totals.
v0 is live code, used for old cache size 1197897, see cudevshr.go line 279. why not fixed there too?

Nakshatra480:
I fixed it in v0 as well. I added the same status == 0 check to all 6 functions there too. Can you take another look when you get a chance?

archlitchi (approved):
/lgtm

Nakshatra480:
@archlitchi thanks for merging

PR #2317, opened 2026-08-03
refactor(scheduler): compute score before appending to shared NodeList

### LINKED ISSUE #2316: fix(scheduler): data race in calcScoreWithOptions — pointer published before OverrideScore finishes
## Bug Description

In `pkg/scheduler/score.go`, `calcScoreWithOptions` launches a goroutine for each candidate node to evaluate score and fitness.

When a node fits (`ctrfit == true`), the local `&score` pointer is appended to `res.NodeList` inside `fitNodesMutex`. Immediately after releasing the lock, `score.OverrideScore(snapshot, userNodePolicy)` is invoked to compute the final node score.

Because the pointer is published to shared state before `OverrideScore` finishes mutating `score.Score`, any concurrent reader accessing elements of `res.NodeList` experiences a data race.

## Location
- `pkg/scheduler/score.go` lines 179–185

## Expected Behavior
The node score should be fully computed (`score.OverrideScore(...)`) before appending `&score` to `res.NodeList`.

## Proposed Fix
Move `score.OverrideScore(...)` above the `fitNodesMutex.Lock()` block so that only fully initialized `NodeScore` objects are published.

AyushSrivastava1818:
## What type of PR is this?

/kind cleanup

This PR is a small readability and code-ordering improvement.

In `calcScoreWithOptions()`, `score.OverrideScore(snapshot, userNodePolicy)` was called after `&score` had already been appended to `res.NodeList`. During review, this ordering was initially suspected to expose a data race. However, as confirmed by @Eshiv-Pandey using the Go race detector, this is **not** a race—the existing `wg.Wait()` guarantees that `res.NodeList` is not accessed until all scoring goroutines, including `OverrideScore()`, have completed.

This change moves `score.OverrideScore(...)` before acquiring `fitNodesMutex` and publishing `&score` to `res.NodeList`. The behavior is unchanged, but the publication point now makes it explicit that a `NodeScore` is fully initialized before becoming visible through the shared slice, improving readability and making the ordering easier to reason about.

N/A — no functional bug is being fixed. This is a cleanup following the discussion in #2316.

- No functional or behavioral changes.
- Verified locally with `go test ./pkg/scheduler/... -race -count=1`.
- This change only improves the ordering of operations to make the publication invariant more explicit.

## AI assistance disclosure

I consulted ChatGPT and Claude during the initial investigation of the scoring path, which led me to incorrectly suspect a data race. Following @Eshiv-Pandey's review, I verified that the synchronization was already correct, updated the change accordingly, and confirmed the behavior locally. The final implementation, commit message, and PR description were written and verified by me, and I take full responsibility for the contribution.

```release-note
NONE
```

### DIFF
--- a/pkg/scheduler/score.go
+++ b/pkg/scheduler/score.go
@@ -177,10 +177,10 @@ func (s *Scheduler) calcScoreWithOptions(nodes *map[string]*NodeUsage, resourceR
 			}
 
 			if ctrfit {
+				score.OverrideScore(snapshot, userNodePolicy)
 				fitNodesMutex.Lock()
 				res.NodeList = append(res.NodeList, &score)
 				fitNodesMutex.Unlock()
-				score.OverrideScore(snapshot, userNodePolicy)
 				klog.V(4).InfoS(common.NodeFitPod, "pod", klog.KObj(task), "node", nodeID, "score", score.Score)
 			}
 		}(nodeID, node)


pkg/scheduler/score.go
```diff
@@ -177,10 +177,10 @@ func (s *Scheduler) calcScoreWithOptions(nodes *map[string]*NodeUsage, resourceR
 			}
 
 			if ctrfit {
+				score.OverrideScore(snapshot, userNodePolicy)
```
Eshiv-Pandey, line 180: Thanks @AyushSrivastava1818! This isn't actually a data race.

`wg.Wait()` guarantees that` score.OverrideScore()` completes before `res.NodeList` is ever read by the caller. running the Go race detector on the original codebase reports zero races:
`go test ./pkg/scheduler/... --race -count=1`

Image

but if you wish to resubmit this as code cleanup pr, feel free to do tht!

Eshiv-Pandey: also refer to [CONTRIBUTING.md](https://github.com/Project-HAMi/HAMi/blob/master/CONTRIBUTING.md) before opening a pr, or review 2-3 merged PRs for templates.

AyushSrivastava1818: Thanks for catching this and running the race detector, @Eshiv-Pandey — appreciate you checking so thoroughly. You're right: wg.Wait() gives the happens-before guarantee, so res.NodeList is never read concurrently with OverrideScore() still mutating score.Score. This isn't a real data race — I was wrong to characterize it as one.

I'll update this PR to a code-cleanup framing instead, as you suggested. Computing the full score before it's ever appended to the shared slice keeps the invariant "only fully-initialized NodeScore objects go into res.NodeList" explicit in the code, rather than relying on the reader knowing wg.Wait() makes the current ordering safe. Updating the title/description now — thanks again for the review

AyushSrivastava1818: Ok @Eshiv-Pandey will check it for sure!

archlitchi (approved):
/lgtm

PR #2328, opened 2026-08-03
fix(vGPUmonitor): bound v0 spec aggregation loops by procnum

### LINKED ISSUE #2327: bug(vGPUmonitor): v0 spec sums all 1024 proc slots instead of bounding by procnum, inflating memory metrics
What happened

In pkg/monitor/nvidia/v0/spec.go (lines 86, 94, 102, 110, 118, 126; overall span 86–129), the six aggregation methods (Total, ContextSize, ModuleSize, BufferSize, Offset, SmUtil) iterate over the entire s.sr.procs array (1024 entries) instead of limiting iteration to the active process count (procnum).

Current implementation:

for _, p := range s.sr.procs {
    v += p.used[idx].total
}

As a result, stale or uninitialized entries beyond procnum may be included in the aggregation, inflating DeviceMemoryTotal() and related metrics. These incorrect values can propagate into feedback.go and potentially cause false throttling of active GPU workloads.

The v1 implementation correctly bounds iteration:

for _, p := range s.sr.procs[:int(s.sr.procnum)] {
    v += p.used[idx].contextSize
}
What you expected to happen

The v0 aggregation methods should only iterate over active process entries (bounded by procnum), matching the v1 implementation.

How to reproduce it (as minimally and precisely as possible)
Deploy HAMi using the legacy v0 shared-memory monitor format.
Ensure procnum is smaller than the fixed-size s.sr.procs array.
Leave one or more entries beyond procnum with stale/non-zero data.
Observe that DeviceMemoryTotal() and the other affected metrics include those extra entries.
Anything else we need to know?

This issue was identified through static code review of pkg/monitor/nvidia/v0/spec.go and comparison with pkg/monitor/nvidia/v1/spec.go.

It is related to, but distinct from:

#2310 (v1 missing p.status guard)
#2281 / #2282 (v0 status check)

The root cause here is the missing procnum bound in the v0 aggregation methods.

Update

Further investigation traced procnum back to loadCache() → CastSpec(), where raw mmap'd bytes are reinterpreted without validating individual fields. Although loadCache() verifies the mapped buffer size, it does not validate that procnum is within a valid range.

Simply changing the loop to:

for _, p := range s.sr.procs[:int(s.sr.procnum)] {

could panic (slice bounds out of range) if a corrupted or malicious shared-memory region contains an invalid procnum (negative or >1024).

The fix therefore clamps procnum to [0, len(s.sr.procs)] before slicing, preventing both stale-data aggregation and slice-bound panics.

Note: pkg/monitor/nvidia/v1/spec.go has the same unguarded procnum read. That is already tracked under #2310 and is outside the scope of this issue.

Suggested fix

Clamp procnum before slicing and use the validated value in all six aggregation methods.

Environment
HAMi version: v2.9.0
nvidia driver or other AI device driver version: N/A (code-level bug)
Docker version: N/A
Docker command, image and tag used: N/A
Kernel version: N/A
Others: Identified through source code review of pkg/monitor/nvidia/v0/spec.go and comparison with pkg/monitor/nvidia/v1/spec.go.

Brijesh-Thakkar:
**What type of PR is this?**
/kind bug

The six metric-aggregation methods in `pkg/monitor/nvidia/v0/spec.go` (`DeviceMemoryContextSize`, 
`DeviceMemoryModuleSize`, `DeviceMemoryBufferSize`, `DeviceMemoryOffset`, `DeviceMemoryTotal`, 
`DeviceSmUtil`) iterated over the full 1024-slot `procs` array instead of stopping at `procnum`. 
Any slot beyond the active process count may retain stale, non-zeroed data from a previous CUDA 
process, causing `DeviceMemoryTotal()` and the five related methods to inflate reported usage. 
Inflated usage can trigger false throttling of live GPU workloads via the feedback loop in 
`feedback.go`.

This PR adds an `activeProcCount()` helper that clamps `procnum` to `[0, len(procs)]` before 
slicing. This is needed because `procnum` is read directly from raw mmap'd shared memory 
(written by `libvgpu.so` inside the container) with no field-level validation upstream — the 
existing file-size check only guarantees the buffer is the right total size, not that individual 
field values are sane. A naive fix using `procs[:int(procnum)]` would trade the stale-data bug 
for a new panic (slice bounds out of range) if `procnum` were ever corrupted to a negative value 
or a value greater than 1024. The clamp makes out-of-range values degrade safely instead.

Fixes #2327

Note:
- The `v1` implementation (`pkg/monitor/nvidia/v1/spec.go`) already bounds these loops by 
  `procnum`, but has the same unguarded-value issue — a corrupted `procnum` there would panic 
  too. That's tracked separately under #2310 (which covers a different, pre-existing v1 defect: 
  a missing `p.status` guard). This PR intentionally does not touch `v1/spec.go` to keep scope 
  minimal; flagging it here for whoever picks up #2310.
- Existing tests in `spec_test.go` had `procnum` implicitly left at its zero value, meaning they 
  weren't actually exercising the bounded-iteration path before this fix — they'd have passed 
  regardless of correctness. I added explicit `procnum` values to all affected existing cases, 
  plus:
  - A new `TestSpec_ActiveProcCount` table covering 7 boundary values (zero, normal, exact 
    capacity, negative, large negative, over-capacity, large over-capacity)
  - 6 "stale slots beyond procnum are excluded" regression cases (one per aggregation method)
  - 2 end-to-end clamp cases in `TestSpec_DeviceMemoryTotal` proving negative/overflow `procnum` 
    values don't panic
- `go vet` and `golangci-lint` both pass clean on the changed package; full `v0` test suite 
  (26 tests) passes with `-race`.
- AI-assisted: this fix was identified by claude code - code review and then the coding part was done by claude but the methodolgy to solve it was not done by it, for that, I used gemini a bit to get better solution to it and a bit of chatgpt as well so i can get best solution

```
Fix vGPUmonitor v0 spec: correctly bound process-slot aggregation by active process count 
instead of scanning all 1024 fixed slots, preventing inflated GPU memory/utilization metrics 
and false throttling caused by stale data in unused slots.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved NVIDIA monitoring accuracy by limiting process-based memory and utilization calculations to valid active processes.
  * Prevented stale process data from affecting reported metrics.
  * Added safeguards for negative or oversized process counts to avoid invalid calculations and runtime errors.

* **Tests**
  * Expanded coverage for zero, negative, bounded, and oversized process counts.

### DIFF
--- a/pkg/monitor/nvidia/v0/spec.go
+++ b/pkg/monitor/nvidia/v0/spec.go
@@ -81,9 +81,22 @@ func (s Spec) DeviceNum() int {
 	return int(s.sr.num)
 }
 
+// activeProcCount returns procnum clamped to [0, len(procs)], guarding
+// against a corrupted shared-memory region reporting a bogus value.
+func (s Spec) activeProcCount() int {
+	n := int(s.sr.procnum)
+	if n < 0 {
+		return 0
+	}
+	if n > len(s.sr.procs) {
+		return len(s.sr.procs)
+	}
+	return n
+}
+
 func (s Spec) DeviceMemoryContextSize(idx int) uint64 {
 	v := uint64(0)
-	for _, p := range s.sr.procs {
+	for _, p := range s.sr.procs[:s.activeProcCount()] {
 		v += p.used[idx].contextSize
 	}
 	return v
@@ -91,7 +104,7 @@ func (s Spec) DeviceMemoryContextSize(idx int) uint64 {
 
 func (s Spec) DeviceMemoryModuleSize(idx int) uint64 {
 	v := uint64(0)
-	for _, p := range s.sr.procs {
+	for _, p := range s.sr.procs[:s.activeProcCount()] {
 		v += p.used[idx].moduleSize
 	}
 	return v
@@ -99,7 +112,7 @@ func (s Spec) DeviceMemoryModuleSize(idx int) uint64 {
 
 func (s Spec) DeviceMemoryBufferSize(idx int) uint64 {
 	v := uint64(0)
-	for _, p := range s.sr.procs {
+	for _, p := range s.sr.procs[:s.activeProcCount()] {
 		v += p.used[idx].bufferSize
 	}
 	return v
@@ -107,7 +120,7 @@ func (s Spec) DeviceMemoryBufferSize(idx int) uint64 {
 
 func (s Spec) DeviceMemoryOffset(idx int) uint64 {
 	v := uint64(0)
-	for _, p := range s.sr.procs {
+	for _, p := range s.sr.procs[:s.activeProcCount()] {
 		v += p.used[idx].offset
 	}
 	return v
@@ -115,7 +128,7 @@ func (s Spec) DeviceMemoryOffset(idx int) uint64 {
 
 func (s Spec) DeviceMemoryTotal(idx int) uint64 {
 	v := uint64(0)
-	for _, p := range s.sr.procs {
+	for _, p := range s.sr.procs[:s.activeProcCount()] {
 		v += p.used[idx].total
 	}
 	return v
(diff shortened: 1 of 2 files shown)


archlitchi:
Done in https://github.com/Project-HAMi/HAMi/pull/2282

PR #1431, opened 2025-10-23
Improve: Replace `StrategicMergePatchType` by `MergePatchType`

luohua13:
**What type of PR is this?**
/kind bug

**What this PR does / why we need it**:

`MergePatchType` method is more stable. `StrategicMergePatchTyp`e requires the k8s resource schema field to have a corresponding key.
Strategic merge patch is not supported for custom resources or metadata.
refer to [update-api-object-kubectl-patch](https://kubernetes.io/docs/tasks/manage-kubernetes-objects/update-api-object-kubectl-patch/
)
**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

PR #2329, opened 2026-08-04
fix(cambricon): prevent node lock leaks by reading from apiserver in ReleaseNodeLock

### LINKED ISSUE #2251: Cambricon ReleaseNodeLock can return nil without removing the node lock

jianzhangbjz:
**What type of PR is this?**

/kind bug

Fix three bugs in ReleaseNodeLock:
- setNodeLock patches the apiserver without updating the caller's in-memory Node, causing ReleaseNodeLock to read stale state and leak locks
- the retry loop reused the same DeepCopy without re-fetching resourceVersion, causing 409 conflicts to never resolve
- delete on the shared informer's annotation map raced with concurrent readers

Fixes https://github.com/Project-HAMi/HAMi/issues/2251

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved node lock acquisition and release using the latest node state.
  - Added safer handling for concurrent updates, already-removed locks, stale state, and temporary retrieval failures.
  - Prevented unrelated node annotations from being affected when releasing a lock.
  - Improved error reporting when node state cannot be retrieved or updated.

- **Tests**
  - Expanded coverage for lock creation, removal, missing locks, stale state, concurrent changes, and retry scenarios.

### DIFF
--- a/pkg/device/cambricon/device.go
+++ b/pkg/device/cambricon/device.go
@@ -98,25 +98,21 @@ func (dev *CambriconDevices) CommonWord() string {
 
 func (dev *CambriconDevices) setNodeLock(node *corev1.Node) error {
 	ctx := context.Background()
-	if _, ok := node.Annotations[DsmluLockTime]; ok {
-		return fmt.Errorf("node %s is locked", node.Name)
-	}
-
 	patchedAnnotation, err := json.Marshal(
 		map[string]any{
 			"metadata": map[string]map[string]string{"annotations": {
 				DsmluLockTime: time.Now().Format(time.RFC3339),
 			}}})
 	if err != nil {
-		klog.ErrorS(err, "Failed to patch node annotation", "node", node.Name)
-		return fmt.Errorf("patch node annotation %v", err)
+		klog.ErrorS(err, "Failed to marshal node annotation", "node", node.Name)
+		return fmt.Errorf("marshal node annotation: %w", err)
 	}
 
-	_, err = client.GetClient().CoreV1().Nodes().Patch(ctx, node.Name, types.StrategicMergePatchType, patchedAnnotation, metav1.PatchOptions{})
+	_, err = client.GetClient().CoreV1().Nodes().Patch(ctx, node.Name, types.MergePatchType, patchedAnnotation, metav1.PatchOptions{})
 	for i := 0; i < retry && err != nil; i++ {
 		klog.ErrorS(err, "Failed to patch node annotation", "node", node.Name, "retry", i)
 		time.Sleep(time.Duration(rand.Intn(i+1)) * 10 * time.Millisecond)
-		_, err = client.GetClient().CoreV1().Nodes().Patch(ctx, node.Name, types.StrategicMergePatchType, patchedAnnotation, metav1.PatchOptions{})
+		_, err = client.GetClient().CoreV1().Nodes().Patch(ctx, node.Name, types.MergePatchType, patchedAnnotation, metav1.PatchOptions{})
 	}
 	if err != nil {
 		return fmt.Errorf("setNodeLock exceeds retry count %d", retry)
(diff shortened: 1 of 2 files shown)


mesutoezdil:
dco check fails, missing signed-off-by. pls sign off your commit and push again

pkg/device/cambricon/device.go
```diff
@@ -156,27 +149,48 @@ func (dev *CambriconDevices) LockNode(n *corev1.Node, p *corev1.Pod) error {
 }
 
 func (dev *CambriconDevices) ReleaseNodeLock(n *corev1.Node, p *corev1.Pod) error {
-	if n.Annotations == nil {
+	ctx := context.Background()
+	nodeName := n.Name
+
+	// Fetch the node from the apiserver to avoid relying on a stale
```
archlitchi: please remove these comments

jianzhangbjz: Done.

archlitchi (dismissed):
/lgtm

archlitchi:
please sync with master to pass the UT

jianzhangbjz:
Hi @archlitchi , I've updated it. Could you help approve it when you get a chance? Thanks!

pkg/device/cambricon/device.go
```diff
@@ -156,27 +152,44 @@ func (dev *CambriconDevices) LockNode(n *corev1.Node, p *corev1.Pod) error {
...
 		return nil
 	}
 
-	newNode := n.DeepCopy()
-	delete(newNode.Annotations, DsmluLockTime)
-	_, err := client.GetClient().CoreV1().Nodes().Update(context.Background(), newNode, metav1.UpdateOptions{})
+	nodeToUpdate := current.DeepCopy()
+	delete(nodeToUpdate.Annotations, DsmluLockTime)
+
+	_, err = client.GetClient().CoreV1().Nodes().Update(ctx, nodeToUpdate, metav1.UpdateOptions{})
```
Shouren: @jianzhangbjz I check the ClusterRole for scheduler in the helm chart and it grants only `get`, `list`, `patch` and `watch`. So I think we can't calls `Nodes().Update()` here.

mesutoezdil: clusterrole for nodes only has get list patch watch, no update. this call gets 403 in a real cluster. same issue as #2479. pls use patch instead.

mesutoezdil:
fixes line has a bad format: "fixes #https://github.com/Project-HAMi/HAMi/issues/2251". github wont auto close the issue like this. can you change it to just "fixes #2251"?

pkg/device/cambricon/device.go
jianzhangbjz: Good catch! Thanks very much!

jianzhangbjz: Hi @Shouren , I also added a CI check for this. Could you help have a look when you get a chance? Thanks! https://github.com/Project-HAMi/HAMi/pull/2567

jianzhangbjz: Hi @mesutoezdil , thanks! I have addressed it. Besides, I also created a CI test to check it automatically, see: https://github.com/Project-HAMi/HAMi/pull/2567

pkg/device/cambricon/device.go
```diff
@@ -156,27 +152,48 @@ func (dev *CambriconDevices) LockNode(n *corev1.Node, p *corev1.Pod) error {
...
+	patchData, err := json.Marshal(
+		map[string]any{
+			"metadata": map[string]map[string]any{"annotations": {
+				DsmluLockTime: nil,
+			}}})
+	if err != nil {
+		return fmt.Errorf("marshal patch for lock release: %w", err)
+	}
+
+	_, err = client.GetClient().CoreV1().Nodes().Patch(ctx, nodeName, types.StrategicMergePatchType, patchData, metav1.PatchOptions{})
```
Shouren: @jianzhangbjz Please StrategicMergePatchType with MergePatchType for Patch. We follow the updated from this PR #1431

jianzhangbjz: Thanks for the info! Updated it.

jianzhangbjz:
Hi @mesutoezdil , I've updated it. Could you help have a review? Thanks!

Shouren (approved):
/lgtm

PR #2333, opened 2026-08-04
fix(scheduler): return deep copy in nodeManager.GetNode to prevent concurrent map read/write

### LINKED ISSUE #2332: scheduler: GetNode returns unsynchronized internal map, causing potential concurrent map read/write crash
# Description

While reviewing the scheduler's node management code, I noticed that `nodeManager.GetNode()` returns a live pointer to the internal `NodeInfo` stored in the manager's map instead of returning a defensive copy. This differs from `ListNodes()`, which explicitly deep-copies each `NodeInfo` to avoid exposing shared mutable state.

## What happens

In `pkg/scheduler/nodes.go`, `GetNode()` currently returns the object stored in `m.nodes` directly:

```go
func (m *nodeManager) GetNode(nodeID string) (*device.NodeInfo, error) {
    m.mutex.RLock()
    defer m.mutex.RUnlock()

    if n, ok := m.nodes[nodeID]; ok {
        return n, nil
    }

    return &device.NodeInfo{}, fmt.Errorf("node %v not found", nodeID)
}
```

Immediately below it, `ListNodes()` performs a deep copy of every `NodeInfo` before returning it, explicitly avoiding races caused by sharing mutable state. `GetNode()` does not provide the same protection.

## Why this is a problem

`GetNode()` is used during scheduling, including from `calcScoreWithOptions()` (`pkg/scheduler/score.go`), where scoring work is performed concurrently across nodes using goroutines.

After `GetNode()` returns, the scheduler reads fields such as `nodeInfo.Devices` without holding the `nodeManager` lock.

At the same time, the scheduler's node synchronization path (`Scheduler.register()` in `pkg/scheduler/scheduler.go`) updates the same `NodeInfo` objects through methods such as `addNode()` and `rmNodeDevices()`, which modify the `Devices` map while holding `nodeManager.mutex.Lock()`.

Because `GetNode()` exposes the internal object directly, readers continue accessing the shared `Devices` map after the read lock has been released, while another goroutine may be modifying that same map.

This creates a classic concurrent access pattern:

* Goroutine A obtains a `NodeInfo` via `GetNode()` and reads `nodeInfo.Devices`.
* Goroutine B updates the same `NodeInfo.Devices` during node registration or inventory refresh.
* The read and write occur concurrently on the same Go map.

Under sufficient concurrency, this can result in Go's runtime fatal error:

```text
fatal error: concurrent map read and map write
```

Unlike a normal panic, this runtime error is unrecoverable and terminates the scheduler process.

## How to reproduce

Although timing-dependent, the race can occur when:

1. Run the scheduler while nodes are being added, updated, removed, or their device inventory is refreshed.
2. Simultaneously schedule a large number of pods so `calcScoreWithOptions()` is executing across multiple goroutines.
3. A scoring goroutine reads `nodeInfo.Devices` while the node synchronization path updates the same map.

With sufficient concurrency, the scheduler may terminate with:

```text
fatal error: concurrent map read and map write
```

## Expected behavior

`GetNode()` should not expose mutable internal state that can be modified concurrently.

Consumers should receive an isolated snapshot (or another read-safe representation) so concurrent node updates cannot race with scheduling.

## Actual behavior

`GetNode()` returns the live `NodeInfo` pointer stored in the internal map. Once the read lock is released, callers continue accessing shared mutable state while other goroutines may modify it, making concurrent map access possible.

## Suggested fix

Make `GetNode()` consistent with `ListNodes()` by returning a deep copy of `NodeInfo` (including the nested `Devices` map), or otherwise return an immutable/read-only snapshot instead of the live internal object.

This would prevent callers from observing concurrent mutations while preserving the existing synchronization guarantees provided by `nodeManager`.

AyushSrivastava1818:
## What type of PR is this?

/kind bug

`nodeManager.GetNode()` in `pkg/scheduler/nodes.go` returned a live pointer to the internal `NodeInfo` stored in the `nodes` map. In contrast, `ListNodes()` already returns deep copies to avoid exposing shared mutable state.

`GetNode()` is used from `calcScoreWithOptions()` (`pkg/scheduler/score.go`), where the returned `NodeInfo.Devices` map is accessed after the `nodeManager` read lock has been released. Meanwhile, `addNode()` and `rmNodeDevices()` (invoked from `register()`) update the same underlying `Devices` map while holding the write lock.

Returning the live object allows readers to continue accessing shared mutable state outside the lock, making concurrent map access possible.

This PR makes `GetNode()` return a deep copy of `NodeInfo`, matching the existing behavior of `ListNodes()`. Callers now receive an isolated snapshot instead of the live internal object, preventing concurrent access to the shared map.

Fixes #2332

- Makes `GetNode()` consistent with `ListNodes()`.
- No behavioral or API changes; only the returned object's ownership changes.
- Verified locally with `go test ./pkg/scheduler/... -race -count=1`.

## AI assistance disclosure

I consulted ChatGPT and Claude to analyze the scheduler code paths and compare `GetNode()` with `ListNodes()`. The implementation, debugging, validation, commit message, and final code changes were completed and verified by me. I reviewed all generated suggestions, confirmed the fix locally, and take full responsibility for the final contribution.

```release-note
NONE
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved node data isolation to prevent changes made by consumers from affecting scheduler state.
  * Preserved existing error handling when requested nodes are unavailable.

* **Tests**
  * Added coverage validating independent node data copies and missing-node behavior.

### DIFF
--- a/pkg/scheduler/nodes.go
+++ b/pkg/scheduler/nodes.go
@@ -117,9 +117,6 @@ func (m *nodeManager) GetNode(nodeID string) (*device.NodeInfo, error) {
 	if !ok {
 		return &device.NodeInfo{}, fmt.Errorf("node %v not found", nodeID)
 	}
-	// Return a deep copy to prevent concurrent map read/write races between
-	// scoring goroutines (readers) and addNode/rmNodeDevices (writers).
-	// ListNodes() already follows this same pattern for the same reason.
 	nodeInfoCopy := &device.NodeInfo{
 		ID:      n.ID,
 		Devices: make(map[string][]device.DeviceInfo, len(n.Devices)),
--- a/pkg/scheduler/nodes.go
+++ b/pkg/scheduler/nodes.go
@@ -113,10 +113,24 @@ func (m *nodeManager) rmNode(nodeID string) {
 func (m *nodeManager) GetNode(nodeID string) (*device.NodeInfo, error) {
 	m.mutex.RLock()
 	defer m.mutex.RUnlock()
-	if n, ok := m.nodes[nodeID]; ok {
-		return n, nil
+	n, ok := m.nodes[nodeID]
+	if !ok {
+		return &device.NodeInfo{}, fmt.Errorf("node %v not found", nodeID)
 	}
-	return &device.NodeInfo{}, fmt.Errorf("node %v not found", nodeID)
+	// Return a deep copy to prevent concurrent map read/write races between
+	// scoring goroutines (readers) and addNode/rmNodeDevices (writers).
+	// ListNodes() already follows this same pattern for the same reason.
+	nodeInfoCopy := &device.NodeInfo{
+		ID:      n.ID,
+		Devices: make(map[string][]device.DeviceInfo, len(n.Devices)),
+	}
+	if n.Node != nil {
+		nodeInfoCopy.Node = n.Node.DeepCopy()
+	}
+	for k, v := range n.Devices {
+		nodeInfoCopy.Devices[k] = device.DeepCopyDeviceInfos(v)
+	}
+	return nodeInfoCopy, nil
 }
 
 func (m *nodeManager) ListNodes() (map[string]*device.NodeInfo, error) {
--- a/pkg/scheduler/nodes_test.go
+++ b/pkg/scheduler/nodes_test.go
@@ -496,6 +496,17 @@ func TestGetNode_DeepCopy(t *testing.T) {
 		t.Errorf("GetNode returned a live pointer instead of a deep copy: internal Devmem was mutated to %d", internalMem)
 	}
 
+	// Mutate the returned copy's Node field.
+	got.Node.Name = "mutated-name"
+
+	m.mutex.RLock()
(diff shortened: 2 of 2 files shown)


pkg/scheduler/nodes.go
```diff
@@ -113,10 +113,24 @@ func (m *nodeManager) rmNode(nodeID string) {
...
 	m.mutex.RLock()
 	defer m.mutex.RUnlock()
-	if n, ok := m.nodes[nodeID]; ok {
-		return n, nil
+	n, ok := m.nodes[nodeID]
+	if !ok {
+		return &device.NodeInfo{}, fmt.Errorf("node %v not found", nodeID)
 	}
-	return &device.NodeInfo{}, fmt.Errorf("node %v not found", nodeID)
+	// Return a deep copy to prevent concurrent map read/write races between
```
archlitchi: please remove these comments here and in test

AyushSrivastava1818: I've removed the explanatory comments from both the implementation and the test. Thanks for the review!

archlitchi (approved):
/lgtm

PR #2337, opened 2026-08-04
fix(scheduler): keep per-type device allocations aligned to container index

### LINKED ISSUE #2335: scheduler: multi-type pods can assign a device to the wrong container
## Summary
For a pod whose containers request **different** device types (e.g. container 0 requests an NVIDIA GPU and container 1 requests a kunlun XPU), the scheduler can annotate the device onto the wrong container.

## Root cause
\calcScoreWithOptions\ in \pkg/scheduler/score.go\ fills missing per-container allocation slots in \score.Devices\ with this loop:

\\\go
// found certain deviceType, fill missing empty allocation for containers before this
for idx := range score.Devices {
    deviceType = idx
    for len(score.Devices[idx]) <= ctrid {
        emptyContainerDevices := device.ContainerDevices{}
        emptyPodSingleDevice := device.PodSingleDevice{}
        emptyPodSingleDevice = append(emptyPodSingleDevice, emptyContainerDevices)
    }
}
\\\

After \itInDevices\ appends the current container's allocation to the requested type, every **already-present** type has \len == ctrid\ (one entry per prior container) and only needs a **trailing** empty slot. The loop instead **prepends** to any type with \len <= ctrid\, which shifts the already-aligned entries right by one container. For the c0=NVIDIA / c1=kunlun example, the NVIDIA device that was aligned to container 0 ends up aligned to container 1.

\PatchAnnotations\ → \EncodePodSingleDevice\ writes per-container annotations, and \DecodePodDevices\ restores the device by container index, so the wrong container receives the device.

Secondary flaw: \deviceType = idx\ records an arbitrary map key, so a container with **no** device requests appends the empty slot to only one randomly-chosen type.

## Impact
- Wrong container gets the device; the container that requested it gets none (or an empty slot).
- For a no-device container interleaved between device-requesting containers of different types, misalignment and missing slots also occur.
- Introduced by https://github.com/Project-HAMi/HAMi/pull/2105's follow-up commit (542e737d); the related fixes #2104/#2105/#1097 only addressed the \itInDevices\ function-scope accumulator leak, not this fill loop.

## Expected
Every discovered type has exactly one \ContainerDevices\ entry per container, indexed by container order, so annotation decoding assigns each device to the container that requested it.

## Proposed fix
Extract the fill logic into \padDeviceAllocations\:
- requested types: prepend the just-appended entry by the missing leading empty slots,
- other discovered types: append a single trailing empty slot,
- a no-device container: append an empty slot to **all** discovered types.

See PR (link) with a regression test covering c0=mockA/c1=mockB and no-device containers.

shivv23:
## What this PR does
Fixes container-index misalignment of per-type device allocations in `calcScoreWithOptions` (`pkg/scheduler/score.go`).

## Why
For a pod whose containers request **different** device types (e.g. c0=NVIDIA, c1=kunlun), the fill loop prepended an empty slot to every already-discovered type, shifting previously-aligned allocations right by one container. `PatchAnnotations` → `EncodePodSingleDevice` → device-plugin `DecodePodDevices` restore devices by container index, so the wrong container receives the device. The fill loop also recorded `deviceType` from an arbitrary map key, so a no-device container padded only one randomly-chosen type.

See #2335.

## What changed
- New helper `padDeviceAllocations`:
  - types requested by the current container: prepend the missing leading empty slots (fixes late-discovery of a type after no-device containers),
  - other discovered types: append a single trailing empty slot,
  - a container with no device requests: append an empty slot to **all** discovered types.
- Regression tests (`Test_calcScore_MultiTypeContainerAlignment`) covering c0=mockA/c1=mockB, a type first discovered after a no-device container, and a no-device container between two different types.

## Verification
- The alignment algorithm was validated against the real `device` types for 10 scenarios (A–G + 3 regression cases), all pass.
- Full scheduler package cannot compile locally (Windows host; go-nvml requires `dlfcn.h`), so CI (Linux) runs the scheduler unit tests.

## AI assistance disclosure
The PR description was generated with the help of AI assistance.

Fixes #2335

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved scheduling accuracy when containers request different device types or no devices.
  * Preserved consistent device-allocation positions across all containers, preventing mismatched allocation data.

* **Tests**
  * Added coverage for multi-device-type and device-free container scheduling scenarios.

### DIFF
--- a/pkg/scheduler/score.go
+++ b/pkg/scheduler/score.go
@@ -90,6 +90,25 @@ func fitInDevices(node *NodeUsage, requests device.ContainerDeviceRequests, pod 
 	return true, ""
 }
 
+// padDeviceAllocations ensures that for every device type discovered so far,
+// devices has exactly one entry per container, indexed by container order.
+// fitInDevices appends the current container's allocation to each requested
+// type, so when earlier containers did not request a type, its entry lands at
+// the wrong index; those entries are shifted right by the missing leading empty
+// slots. Types not requested by the current container only need a trailing
+// empty slot.
+func padDeviceAllocations(devices device.PodDevices, ctrid int, requestedTypes map[string]bool) {
+	for idx := range devices {
+		for len(devices[idx]) <= ctrid {
+			if requestedTypes[idx] {
+				devices[idx] = append(device.PodSingleDevice{device.ContainerDevices{}}, devices[idx]...)
+			} else {
+				devices[idx] = append(devices[idx], device.ContainerDevices{})
+			}
+		}
+	}
+}
+
 func (s *Scheduler) calcScore(nodes *map[string]*NodeUsage, resourceReqs device.PodDeviceRequests, task *corev1.Pod, failedNodes map[string]string) (*policy.NodeScoreList, error) {
 	return s.calcScoreWithOptions(nodes, resourceReqs, task, failedNodes, true, false)
 }
@@ -135,7 +154,6 @@ func (s *Scheduler) calcScoreWithOptions(nodes *map[string]*NodeUsage, resourceR
 			// Assume the node is a fit by default. This handles pods with no device
 			// requests, which should be schedulable on any node.
 			ctrfit := true
-			deviceType := ""
 			//This loop is for different container request
 			for ctrid, n := range resourceReqs {
 				sums := 0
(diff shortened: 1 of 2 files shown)


FouoF:
There is none or rare pod requesting more than one device type.

PR #2338, opened 2026-08-04
fix(device): clamp oversized memory requests to int32 range

shivv23:
## What this PR does
Clamps memory requests that exceed `math.MaxInt32` in `GenerateResourceRequests` for the ascend, hygon, metax and nvidia backends, instead of letting the `int32` cast silently wrap.

## Why
A request above int32 range (e.g. a `Quantity` with a Gi unit, or a plain value scaled by a large memory factor) wraps to a small or negative `Memreq`. A wrapped `Memreq` of 0 defeats the `Totalmem - Usedmem < memreq` fit check, causing oversubscription and OOM. metax additionally guards the `v*1024` scaling, which can wrap int64 to a negative value.

See #2336.

## What changed
- `pkg/device/nvidia/device.go`, `pkg/device/hygon/device.go`, `pkg/device/ascend/device.go` (soft-partition path), `pkg/device/metax/sdevice.go`: if the computed memory exceeds `math.MaxInt32` (or is negative after an int64 wrap), clamp to `math.MaxInt32` and log an error.
- Unit tests for hygon, metax and ascend (`*_MemoryOverflowClamped`) verifying clamp + in-range preservation.

## Verification
- New + full package tests pass locally for hygon, metax, ascend: `go test ./pkg/device/{hygon,metax,ascend}/ -count=1`.
- The nvidia package cannot compile on this Windows host (go-nvml requires `dlfcn.h`); CI (Linux) covers it.

## AI assistance disclosure
The PR description was generated with the help of AI assistance.

Fixes #2336

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented oversized or invalid GPU and accelerator memory requests from overflowing during conversion.
  - Values outside the supported range are safely capped, while valid requests continue to behave as expected.
  - Added handling across Ascend, Hygon, MetaX, and NVIDIA device integrations.

- **Tests**
  - Added regression coverage for overflow and boundary-value memory requests across supported devices.

PR #2339, opened 2026-08-04
fix(cambricon): clamp oversized memory request before int32 conversion

### LINKED ISSUE #2278: [bug]: cambricon int32 overflow in GenerateResourceRequests silently drops memory request

shivv23:
## What this PR does
Clamps the cambricon memory request in `GenerateResourceRequests` to the int32 range instead of letting the `int32` cast silently wrap.

## Why
`memnums` is computed as a 64-bit int and then narrowed with `int32()`, which wraps to 0 for any request above `math.MaxInt32/256`. Requesting `cambricon.com/mlu.smlu.vmemory` in Gi units (e.g. 16Gi) therefore produced `Memreq=0`, so `Fit()` scheduled the pod with no memory accounting and could oversubscribe an already-full MLU. Same class of bug fixed for enflame (#2145/#2190) and addressed for the other backends in #2338.

See #2278.

## What changed
- `pkg/device/cambricon/device.go`: if the memory request exceeds `math.MaxInt32/256` (or is negative), clamp `memnum` to `math.MaxInt32` and log an error; otherwise apply the scaling.
- Unit test in `pkg/device/cambricon/device_test.go` verifying the clamp and in-range preservation.

## Verification
- New + existing cambricon package tests pass locally: `go test ./pkg/device/cambricon/ -count=1`.

## AI assistance disclosure
The PR description was generated with the help of AI assistance.

Fixes #2278

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented invalid or oversized Cambricon memory requests from overflowing.
  - Large and negative memory values are now safely capped at the supported maximum.
  - Valid memory requests continue to be converted correctly.

### DIFF
--- a/pkg/device/cambricon/device.go
+++ b/pkg/device/cambricon/device.go
@@ -21,6 +21,7 @@ import (
 	"encoding/json"
 	"flag"
 	"fmt"
+	"math"
 	"math/rand"
 	"slices"
 	"strings"
@@ -251,7 +252,12 @@ func (dev *CambriconDevices) GenerateResourceRequests(ctr *corev1.Container) dev
 				memnums, ok := mem.AsInt64()
 				klog.Infoln("mluResourceMem", mem, memnums)
 				if ok {
-					memnum = int(memnums) * 256
+					if memnums > math.MaxInt32/256 || memnums < 0 {
+						klog.ErrorS(nil, "cambricon memory request is out of int32 range, clamping to max int32", "container", ctr.Name, "request", memnums)
+						memnum = math.MaxInt32
+					} else {
+						memnum = int(memnums * 256)
+					}
 				}
 			}
 			corenum := int32(100)
--- a/pkg/device/cambricon/device_test.go
+++ b/pkg/device/cambricon/device_test.go
@@ -19,6 +19,7 @@ package cambricon
 import (
 	"context"
 	"flag"
+	"math"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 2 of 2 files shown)


mesutoezdil:
duplicate of #2285, which already clamps/rejects this exact overflow in cambricon's GenerateResourceRequests for the same issue #2278. closing.

pls check the existing prs before new one

PR #2347, opened 2026-08-04
fix(scheduler): enforce ResourceQuota for every device backend

### LINKED ISSUE #2157: [Bug] ResourceQuota enforcement only applies to NVIDIA devices, not other accelerator backends

Lakshya77089:
**What type of PR is this?**

/kind bug

`fitResourceQuota` in the admission webhook only ever looked at NVIDIA:

```go
// Only supports NVIDIA
    continue
}
```

So a namespace `ResourceQuota` on `cambricon.com/mlu.smlu.vmemory`,
`huawei.com/Ascend910B-memory`, `hygon.com/dcumem` or any other vendor resource
was accepted by the apiserver and then ignored. Pods were admitted no matter how
much of that quota the namespace had already used. `QuotaManager` itself is
device agnostic and works fine when called directly, so the gap is entirely in
the wiring.

Two commits:

1. `refactor(device): name the vmemory scaling constants` — no behaviour change,
   pulled out so the second diff is readable.
2. `fix(scheduler): enforce ResourceQuota for every device backend` — the fix.

**The unit problem, and why the loop could not just be widened**

This is where the earlier attempt in #2218 came unstuck, and @mesutoezdil's
review there is what shaped this patch. Several backends scale the memory value
from the pod spec before recording it as used:

| backend | scale |
|---|---|
| nvidia, ascend, hygon | configured `memoryFactor` |
| cambricon, iluvatar | 256 |
| mthreads | 512 |
| metax sgpu | 1024 |

Recorded usage is in scaled units. A `ResourceQuota` limit is written in
unscaled ones. Compare them directly and a pod asking for 50 units of MLU
vmemory against a 100 unit limit is measured as 12800 against 100 and denied,
even though it is well inside quota. Hardcoding `memoryFactor = 1` for
non-NVIDIA backends, which is what #2218 did, produces exactly that.

So each backend now reports its own scale through a new
`ResourceNames.MemoryFactor`, and `FitQuota` raises the limit by it — the same
thing it already did for NVIDIA. Backends that do not scale leave it zero.

The webhook also stops re-reading the container spec by hand and asks the
backend instead:

```go
req := dev.GenerateResourceRequests(&pod.Spec.Containers[i])
memoryReq += int64(req.Memreq) * int64(req.Nums)
coresReq  += int64(req.Coresreq) * int64(req.Nums)
```

That is the same call the scheduler makes on the Fit path, so admission and the
scheduler now work from identical numbers, including per-backend defaults and
Ascend's template rounding. It also drops about 25 lines of duplicated
limits/requests parsing.

**Tests**

`TestFitResourceQuotaNonNvidia`, `TestFitResourceQuotaCountsEveryDevice` and
`TestFitResourceQuotaAscendMemoryFactor` in `pkg/scheduler/webhook_test.go`.
They cover deny and allow for MLU memory, MLU cores, DCU memory with
`memoryFactor: 2`, Ascend with `memoryFactor: 4`, multi-device requests, and a
namespace with no quota. All five deny cases fail on master and pass with this
change.

`TestFitResourceQuotaAscendMemoryFactor` asserts the backend actually registered
before running, because the existing `TestFitResourceQuota/request_ascend` case
passes vacuously today — Ascend is behind `--enable-ascend` and was never in
`DevicesMap` for that test.

Fixes #2157

Two behaviour changes worth a second look:

- NVIDIA memory quota now counts `defaultMemory` when a pod asks for
  `nvidia.com/gpu` without `nvidia.com/gpumem`. The old code only counted memory
  the pod named explicitly, which under-counted against what the scheduler
  actually records. Default config has `defaultMemory: 0`, so most deployments
  see no difference.
- Backends that default cores when none are requested (cambricon defaults to
  100) now have those cores counted. Same reasoning: that is what ends up in
  `Usedcores`.

Deliberately left out, happy to add if you would rather see them here:

- `FitQuota` on the scheduler `Fit()` path is still NVIDIA only. This PR fixes
  admission. Doing `Fit()` properly means touching every backend's allocation
  loop and I did not want to bury the webhook fix in that. Tracked separately in
  #2363, which also covers the admission-to-scheduling race @mesutoezdil raised
  below.
- Metax reads a bare `vmemory` value as Gi but a suffixed one as bytes. The 1024
  factor is right for the bare form, which is what a `ResourceQuota` uses. A
  quota written as `16Gi` would still be wrong, but that is `AddQuota`'s
  handling of suffixed quantities and predates this change.
- Init containers are not counted, same as before. That overlaps #1667.

Not validated on real hardware. The change is confined to the scheduler
extender's admission path and is covered by unit tests, which CONTRIBUTING
allows for scheduler-scoped changes.

```release-note
ResourceQuota limits on non-NVIDIA accelerator memory and core resources are now
enforced at admission instead of being silently ignored.
```

---

AI assistance disclosure: this change was developed with Claude Code — codebase
exploration, working through the per-backend unit scaling, the tests, and this
description. Flagging the extent up front per CONTRIBUTING.

### DIFF
--- a/pkg/device/ascend/device.go
+++ b/pkg/device/ascend/device.go
@@ -416,6 +416,7 @@ func (dev *Devices) GetResourceNames() device.ResourceNames {
 		ResourceCountName:  dev.config.ResourceName,
 		ResourceMemoryName: dev.config.ResourceMemoryName,
 		ResourceCoreName:   dev.config.ResourceCoreName,
+		MemoryFactor:       dev.config.MemoryFactor,
 	}
 }
 
--- a/pkg/device/cambricon/device.go
+++ b/pkg/device/cambricon/device.go
@@ -429,5 +429,6 @@ func (dev *CambriconDevices) GetResourceNames() device.ResourceNames {
 		ResourceCountName:  MLUResourceCount,
 		ResourceMemoryName: MLUResourceMemory,
 		ResourceCoreName:   MLUResourceCores,
+		MemoryFactor:       MemoryFactor,
 	}
 }
--- a/pkg/device/cambricon/device.go
+++ b/pkg/device/cambricon/device.go
@@ -54,6 +54,9 @@ const (
 	DsmluProfile          = "CAMBRICON_DSMLU_PROFILE"
 	DsmluResourceAssigned = "CAMBRICON_DSMLU_ASSIGNED"
 	retry                 = 5
+	// MemoryFactor converts the vmemory unit used in the pod spec into the MiB
+	// HAMi accounts internally. One cambricon.com/mlu.smlu.vmemory is 256 MiB.
+	MemoryFactor = 256
 )
 
 var (
@@ -198,7 +201,7 @@ func (dev *CambriconDevices) GetNodeDevices(n corev1.Node) ([]*device.DeviceInfo
 			Index:        uint(i),
 			ID:           n.Name + "-cambricon-mlu-" + fmt.Sprint(i),
 			Count:        100,
-			Devmem:       int32(memoryTotal * 256 * 100 / cards),
+			Devmem:       int32(memoryTotal * MemoryFactor * 100 / cards),
 			Devcore:      100,
 			Type:         CambriconMLUDevice,
 			Numa:         0,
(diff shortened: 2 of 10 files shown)


mesutoezdil:
fit() path is still nvidia only for quota, non-nvidia has a race window between admission and scheduling. pls open a followup issue instead of letting fixes #2157 close it fully.

archlitchi (approved):
/lgtm

PR #2250, opened 2026-07-31
test: run vGPUmonitor unit tests

iemAnshuman:
/kind cleanup
**What this PR does / why we need it**:
`make test` only includes packages under `pkg`, so the existing `vGPUmonitor` tests are skipped. This adds `cmd/vGPUmonitor` to the same race-enabled test command.
**Which issue(s) this PR fixes**:
None.
**Special notes for your reviewer**:
No runtime code changes. The focused test, full unit suite, lint, license check, and import check pass.
**Does this PR introduce a user-facing change?**:
NONE.
AI disclosure: I used an AI tool for codebase navigation and validation. I reviewed the change.

## Summary by CodeRabbit

* **Tests**
  * Expanded test coverage checks to include both package and command-line code.
  * Retained short test execution, race detection, single-run behavior, and atomic coverage reporting.
  * Coverage results now provide broader validation across the project’s core packages and executable components without changing the existing test workflow.

PR #2350, opened 2026-08-04
test(vGPUmonitor): cover the container metrics collection path

veyron-kairo:
**What type of PR is this?**
/kind test

`cmd/vGPUmonitor` had almost no coverage of its metrics collector (~6% of the
package). `collectContainerMetrics` — which emits the per-container GPU memory,
utilization, and last-kernel metrics — had none. This adds unit tests for it:

- a healthy device, asserting the value and the `namespace/pod/container/vdevice_index/device_uuid` labels of every emitted metric;
- the `hami_container_last_kernel_elapsed_seconds` metric only being emitted once a kernel has actually run (`LastKernelTime > 0`), and computed as `nowSec - lastKernel`;
- the error paths: nil `ContainerUsage`, nil `Info`, and a UUID shorter than 40 chars;
- an invalid-UTF-8 UUID being skipped (so a not-yet-initialised shared-memory slot doesn't take down the scrape).

Rather than add a second mock, it extends the existing `stubInfo` in
`feedback_test.go` with the memory/utilization fields the collector reads; the
existing feedback tests are unaffected (the new fields default to zero).

This raises `cmd/vGPUmonitor` coverage from ~6% to ~18%. No runtime code changes.

None.

`make verify` and the full unit suite pass locally. These run in CI now that
`cmd/vGPUmonitor` was added to the test command in #2250.

**AI assistance disclosure**:
I used AI assistance (Claude Code) to understand what tests is needed and then took assistance from codex while writing these tests. I reviewed and verified them, confirmed they exercise the intended paths, ran `make verify`
locally, and take responsibility for the change.

NONE

## Summary by CodeRabbit

* **Tests**
  * Expanded coverage for multi-device blocking and priority checks under contention and incomplete device data.
  * Added validation for container GPU metrics, including labels, metric values, kernel timing, inactive kernels, invalid input, and malformed device identifiers.

### DIFF
--- a/cmd/vGPUmonitor/feedback_test.go
+++ b/cmd/vGPUmonitor/feedback_test.go
@@ -31,8 +31,22 @@ const stubDeviceMax = 16
 // count and the trailing slots read back invalid, so the check functions walk the
 // same 16 slots they do in production.
 type stubInfo struct {
-	priority int
-	uuids    []string
+	priority   int
+	uuids      []string
+	total      []uint64
+	limit      []uint64
+	ctxSize    []uint64
+	modSize    []uint64
+	bufSize    []uint64
+	smUtil     []uint64
+	lastKernel int64
+}
+
+func slot(v []uint64, i int) uint64 {
+	if i >= 0 && i < len(v) {
+		return v[i]
+	}
+	return 0
 }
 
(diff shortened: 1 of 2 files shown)


archlitchi (approved):
/lgtm

PR #2353, opened 2026-08-04
fix(device-plugin): return error on failed Send in ListAndWatch

### LINKED ISSUE #2352: device-plugin: ListAndWatch ignores gRPC stream Send() errors
## What happened

In `pkg/device-plugin/nvidiadevice/nvinternal/plugin/server.go`, `ListAndWatch()` calls `s.Send(...)` twice to stream device updates to kubelet, but both calls ignore the returned error:

```go
s.Send(&kubeletdevicepluginv1beta1.ListAndWatchResponse{Devices: plugin.apiDevices()})
```

## Why it matters

`Send()` returns an error if the server can no longer write to the gRPC stream (for example, if the client disconnects or the stream is cancelled).

The current implementation discards that error and continues executing the loop. As a result, subsequent device health updates continue attempting to write to a stream that has already failed instead of terminating the RPC handler.

Returning the error immediately follows the standard gRPC server-stream pattern and allows the failed stream to be cleaned up instead of repeatedly attempting writes on a broken stream.

## How to reproduce

1. Start `ListAndWatch()` and allow it to successfully send an initial response.
2. Terminate or interrupt the client-side gRPC stream (for example, by cancelling the context or disconnecting the client).
3. Trigger another device health update.
4. Observe that `Send()` returns an error, but the error is ignored and `ListAndWatch()` continues running.

## Expected behavior

If `Send()` returns an error, `ListAndWatch()` should return that error and terminate the current stream.

## Actual behavior

`Send()` errors are silently ignored and the handler continues attempting to send future updates on the failed stream.

## Suggested fix

Replace both `Send()` calls with:

```go
if err := s.Send(&kubeletdevicepluginv1beta1.ListAndWatchResponse{
    Devices: plugin.apiDevices(),
}); err != nil {
    return err
}
```

I'll open a PR with the proposed fix shortly.

AyushSrivastava1818:
## What type of PR is this?

/kind bug

`ListAndWatch()` (`pkg/device-plugin/nvidiadevice/nvinternal/plugin/server.go`) discarded the error returned by `s.Send()` at both call sites.

If the gRPC stream to kubelet becomes unavailable (for example, due to a client disconnect or stream cancellation), `Send()` returns an error. The current implementation ignored that error and continued running, causing subsequent device health updates to continue attempting writes on a failed stream.

This PR logs and returns the `Send()` error at both call sites, allowing the current `ListAndWatch()` RPC to terminate instead of silently ignoring stream failures. This follows the standard gRPC server-stream error handling pattern already used elsewhere in the codebase.

Fixes #2352 

- No behavioral changes to the successful (`Send()` succeeds) path.
- Added `TestListAndWatch_SendError` covering both `Send()` call sites:
  - Initial `Send()` failure.
  - Health-update `Send()` failure after an initial successful send.
- The tests verify that `ListAndWatch()` returns the propagated error and stops sending additional updates after the failure.
- Verified locally with:
  - `go test ./pkg/device-plugin/nvidiadevice/nvinternal/plugin/... -v`
  - `make verify`

## AI assistance disclosure

I consulted ChatGPT and Claude to identify the issue and discuss the implementation approach. The implementation, tests, validation, commit message, and final code changes were completed and verified by me. I reviewed all generated suggestions, ran the relevant tests and `make verify` locally, and take full responsibility for the final contribution.

```release-note
NONE
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of device status reporting by detecting communication failures with the cluster.
  * Prevented failed initial or health-update messages from being silently ignored.
  * Ensured device monitoring responds appropriately when status updates cannot be delivered.

* **Tests**
  * Added coverage for communication failures during initial and subsequent device status updates.

### DIFF
--- a/pkg/device-plugin/nvidiadevice/nvinternal/plugin/server.go
+++ b/pkg/device-plugin/nvidiadevice/nvinternal/plugin/server.go
@@ -454,7 +454,11 @@ func (plugin *NvidiaDevicePlugin) GetDevicePluginOptions(context.Context, *kubel
 
 // ListAndWatch lists devices and update that list according to the health status
 func (plugin *NvidiaDevicePlugin) ListAndWatch(e *kubeletdevicepluginv1beta1.Empty, s kubeletdevicepluginv1beta1.DevicePlugin_ListAndWatchServer) error {
-	s.Send(&kubeletdevicepluginv1beta1.ListAndWatchResponse{Devices: plugin.apiDevices()})
+	err := s.Send(&kubeletdevicepluginv1beta1.ListAndWatchResponse{Devices: plugin.apiDevices()})
+	if err != nil {
+		klog.Errorf("Failed to send ListAndWatch response: %v", err)
+		return err
+	}
 
 	for {
 		select {
@@ -464,7 +468,11 @@ func (plugin *NvidiaDevicePlugin) ListAndWatch(e *kubeletdevicepluginv1beta1.Emp
 			// FIXME: there is no way to recover from the Unhealthy state.
 			d.Health = kubeletdevicepluginv1beta1.Unhealthy
 			klog.Infof("'%s' device marked unhealthy: %s", plugin.rm.Resource(), d.ID)
-			s.Send(&kubeletdevicepluginv1beta1.ListAndWatchResponse{Devices: plugin.apiDevices()})
+			err := s.Send(&kubeletdevicepluginv1beta1.ListAndWatchResponse{Devices: plugin.apiDevices()})
+			if err != nil {
+				klog.Errorf("Failed to send ListAndWatch response: %v", err)
+				return err
+			}
 		}
 	}
 }
--- a/pkg/device-plugin/nvidiadevice/nvinternal/plugin/server.go
+++ b/pkg/device-plugin/nvidiadevice/nvinternal/plugin/server.go
@@ -468,10 +468,9 @@ func (plugin *NvidiaDevicePlugin) ListAndWatch(e *kubeletdevicepluginv1beta1.Emp
 			// FIXME: there is no way to recover from the Unhealthy state.
 			d.Health = kubeletdevicepluginv1beta1.Unhealthy
 			klog.Infof("'%s' device marked unhealthy: %s", plugin.rm.Resource(), d.ID)
-			err := s.Send(&kubeletdevicepluginv1beta1.ListAndWatchResponse{Devices: plugin.apiDevices()})
-			if err != nil {
-				klog.Errorf("Failed to send ListAndWatch response: %v", err)
-				return err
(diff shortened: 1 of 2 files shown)


pkg/device-plugin/nvidiadevice/nvinternal/plugin/server.go
```diff
@@ -464,7 +468,11 @@ func (plugin *NvidiaDevicePlugin) ListAndWatch(e *kubeletdevicepluginv1beta1.Emp
 			// FIXME: there is no way to recover from the Unhealthy state.
 			d.Health = kubeletdevicepluginv1beta1.Unhealthy
 			klog.Infof("'%s' device marked unhealthy: %s", plugin.rm.Resource(), d.ID)
-			s.Send(&kubeletdevicepluginv1beta1.ListAndWatchResponse{Devices: plugin.apiDevices()})
+			err := s.Send(&kubeletdevicepluginv1beta1.ListAndWatchResponse{Devices: plugin.apiDevices()})
+			if err != nil {
+				klog.Errorf("Failed to send ListAndWatch response: %v", err)
+				return err
```
Shouren: @AyushSrivastava1818 I check the [implementation ](https://github.com/NVIDIA/k8s-device-plugin/blob/c648e14098589a4a917796596bc4f96908b54433/internal/plugin/server.go#L281) of NVIIDA Device Plugin, it returns nil in `case d := <-plugin.health`. 
Can you tell me why you choose to return err here?

Shouren (changes requested):

pkg/device-plugin/nvidiadevice/nvinternal/plugin/server.go
AyushSrivastava1818: Thanks for checking the upstream implementation. My reasoning was that if Send() returns an error, the gRPC stream is no longer usable, so returning the error would terminate the current ListAndWatch() RPC instead of continuing to send on a failed stream. That seemed consistent with the usual gRPC server-stream pattern.

I wasn't aware that the NVIDIA Device Plugin intentionally returns nil in the health-update path. If HAMi aims to stay aligned with the upstream behavior here, I'm happy to update the implementation accordingly.

Shouren: @AyushSrivastava1818 I prefer to staying aligned with the upstream.

AyushSrivastava1818: @Shouren Updated health-update Send() failures now log and return nil (matching upstream), while the initial Send() still returns the error.

Shouren (approved):
/lgtm

archlitchi (approved):
/lgtm

PR #2367, opened 2026-08-04
 Missing Test for InitContainers & EphemeralContainers GPU Quota Checks

aniket866:
### Path Under Test
*   [webhook.go:L131-L178](file:///c:/Users/Hp/HAMi/pkg/scheduler/webhook.go) (`fitResourceQuota`)

### Explanation & Bug Gap
The admission webhook's quota controller only checks resource limits inside `pod.Spec.Containers`. If a pod runs an `InitContainer` requesting GPUs, the quota checks are bypassed entirely. The current tests in `webhook_test.go` only verify resource configurations within regular containers.

### Why Add This Test
To verify that init containers and ephemeral containers cannot bypass namespace GPU quota constraints, ensuring proper resource isolation inside multi-tenant clusters.

### Mermaid Diagram
```mermaid
graph TD
    A[Test Case: InitContainer with GPUs] --> B[fitResourceQuota]
    B --> C{Scan Containers, InitContainers}
    C --> D[Aggregate Total GPU Core/Memory]
    D --> E{Exceeds Quota?}
    E -- Yes --> F[Assert: Webhook Denies Admission]
    E -- No --> G[Assert: Webhook Allows Admission]
```

### Proposed Test Implementation
Create a test inside `webhook_test.go` that mocks an admission request for a Pod where `Spec.InitContainers` requests a GPU count or memory limit exceeding the namespace resource quota. Assert that the admission response has `Allowed: false` and mentions exceeding the quota.

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented crashes when GPUs report asymmetric peer-to-peer link information.
  * Added a warning and safely returns a fallback score of zero for affected GPU pairs.

* **Documentation**
  * Added a changelog entry documenting the fix in version 2.9.1.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -408,3 +408,8 @@ Add "NVIDIA_VISIBLE_DEVICES=none" to none-gpu tasks
 - Fix initialization error when using tensor parallelism on vLLM above 0.18
 - Fix multiple device typos
 
+## v2.9.1 - 2026-08-05
+
+**Bug fixes**
+- Fix panic on asymmetric NVML GPU Link topologies by logging a warning and returning a fallback score of 0
+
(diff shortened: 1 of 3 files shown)


mesutoezdil:
pls read and understand what are you doing with codes first

PR #2368, opened 2026-08-04
Missing Test for Node Discovery Handshake on Malformed/Empty Annotations

aniket866:
### Path Under Test
*   [devices.go:L272-L336](file:///c:/Users/Hp/HAMi/pkg/device/devices.go) (`DecodeNodeDevices`)

### Explanation & Bug Gap
The node handshake parser `DecodeNodeDevices` splits node device annotations using the `:` delimiter. If a node registers with a malformed annotation or lacks the `:` split symbol entirely, the function immediately returns an error. There are no tests validating the discovery loop's response to malformed or empty node annotations.

### Why Add This Test
To ensure that a malformed handshake or invalid annotation on a single node does not disrupt the main scheduler registration ticker or cause node discovery to crash or hang.

### Mermaid Diagram
```mermaid
graph TD
    A[Test Case: Malformed Node Annotation] --> B[DecodeNodeDevices]
    B --> C{Separator : missing?}
    C -- Yes --> D[Assert: Return error gracefully, do not panic]
    C -- No --> E[Process device info]
```

### Proposed Test Implementation
Add a unit test in `devices_test.go` that feeds malformed string inputs (e.g. empty strings, strings without commas, or strings missing the `:` delimiter) to `DecodeNodeDevices` and asserts that it returns errors gracefully without panicking.

---

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of empty, delimiter-only, comma-only, and malformed device discovery data.
  - Invalid input now produces clear parsing errors without returning incorrect device results.

- **Tests**
  - Expanded coverage for node discovery handshake parsing across additional edge cases.

- **Documentation**
  - Added changelog entry highlighting the improved parsing test coverage.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -407,4 +407,6 @@ Add "NVIDIA_VISIBLE_DEVICES=none" to none-gpu tasks
 **Bug fixes**
 - Fix initialization error when using tensor parallelism on vLLM above 0.18
 - Fix multiple device typos
+- Add unit test coverage for node discovery handshake parsing with malformed or empty annotations
+
 
--- a/pkg/device/devices_test.go
+++ b/pkg/device/devices_test.go
@@ -462,7 +462,62 @@ func Test_DecodeNodeDevices(t *testing.T) {
 		}
 	}{
 		{
-			name: "args is invalid",
+			name: "args is empty",
+			args: "",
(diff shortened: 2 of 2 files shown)


archlitchi (approved):
/lgtm

PR #2241, opened 2026-07-31
fix(ascend): skip unhealthy devices in Fit()

Eshiv-Pandey:
**What type of PR is this?**

/kind bug

**What this PR does / why we need it:**

The Ascend device plugin writes `Health: false` into node annotations when `GetDeviceHealth()` reports an unhealthy NPU. The scheduler decodes this via` UnMarshalNodeDevices() `and `uildNodeUsage()` copies it to `DeviceUsage.Health.` However, `Ascend.Fit()` never checks dev.Health, so unhealthy NPUs remain eligible for allocation.

NVIDIA's `Fit()` already skips unhealthy devices with `common.CardNotHealth`. This PR adds the same check to the Ascend backend, placed at the top of the device loop before` checkType()` matching the NVIDIA pattern exactly, **in short adding consistency here :)** 

**Which issue(s) this PR fixes:**

None filed.

**Special notes for your reviewer:**

- Scoped to Ascend only ... no changes to serialization, common helpers, or other backends.
- The Ascend device plugin always initializes `Health `for every device (manager.go: Health: health == 0), so this check is safe.

**Does this PR introduce a user-facing change?**

No. Unhealthy Ascend NPUs are now skipped during scheduling instead of being allocated.

**AI assistance disclosure:**

I used an AI tool for codebase navigation and review support.

## Bug Fixes

* Unhealthy devices are now excluded from allocation.
* Device selection reports a clear health-related rejection reason when a device is skipped.
* Added logging to improve visibility into devices rejected due to health status.
* Added coverage to verify that unhealthy devices are not allocated and receive the appropriate rejection reason.

PR #2369, opened 2026-08-04
fix (ascend) - validate module pair count and prevent under allocation for ascend 910C

### LINKED ISSUE #2268: silent under-allocation and topology bypass in Ascend 910C device allocation

princexpoddar:
**What type of PR is this?**
/kind bug

while going through ascend 910c device allocation logic in pkg/device/ascend/device.go...I found a issue in how multi NPU requests are validated during scheduling:
-when a pod requested multiple NPUs ( lets say 4 NPUs) on a node with 1 full card pair(2 NPUs) and other partial single NPU available... computeBestCombination910C filtered out partial cards and returned only the full pair found(2NPUs) and fit() assigned this result to tmpDevs and unconditionally returned true,thus causing under allocation of 2NPUs for a 4 NPU request while reporting success
-when the raw candidature NPU count happened to equal originReq (lets say asking for 2 NPUs on a node with 2 available NPUs on separate cards), Fit() returned true early without invoking computeBestCombination910C() whenever needTopology was false or NetworkID was missing..thus allowing candidate NPUs on separate cards to bypass module pairing validation completely
-In computeBestCombination910C() cards were sorted in ascending order by available NPU count which contradicted the comment directly above it stating full cards should be preferred first

To resolve these, 
I have updated Fit() to verify that len(combination) == int(originReq) after module filtering and if sufficient complete module pairs cannot satsfy the request it now returns fit=false with AllocatedCardsInsufficientRequest instead of reporting success on a partial allocation

Also ensured module-pair validation always runs for multi-device Ascend910CType requests regardless of whether NetworkID is present

Updated common.GenReason parameter passing and failure tracking so Kubernetes logs and reason strings accurately report requested versus allocated NPU counts.

Changed card sorting in computeBestCombination910C() to descending order so fully available cards are evaluated before partially available ones. 

Added regression tests in pkg/device/ascend/device_910c_pairing_test.go covering 
-partial allocation rejection (TestAscend910C_FitPartialAllocationBug)
-candidate-count bypass prevention (TestAscend910C_FitExactCountBypassBug)
-successful full-pair allocation (TestAscend910C_FitFullPairSucceeds)
-missing NetworkID handling (TestAscend910C_FitWithoutNetworkID_ValidatesPairing)
-empty combination handling when no candidate NPUs form a complete module (TestComputeBestCombination910C_NoFullPairsReturnsEmpty).

Fixes #2268

- go test -v ./pkg/device/ascend/.... All unit tests pass cleanly.
- updated unit tests to explicitly set Health =true to align with the recently merged upstream health checks (#2241)
- AI assistance - Used Claude assistance

No

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Ascend 910C allocation to require complete two-NPU module pairs.
  * Prevented partial or insufficient allocations, including when topology information is unavailable.
  * Prioritized cards with more available NPUs while maintaining valid pairing requirements.

* **Tests**
  * Added regression coverage for partial allocations, exact-count requests, missing network metadata, valid pair allocation, and unavailable complete pairs.

### DIFF
--- a/pkg/device/ascend/device.go
+++ b/pkg/device/ascend/device.go
@@ -560,15 +560,9 @@ func (npu *Devices) Fit(devices []*device.DeviceUsage, request device.ContainerD
 	}
 
 	if k.Type == Ascend910CType && originReq > 1 {
-		// Ascend 910C requires full module-pair allocation (2 NPUs per
-		// physical card). Always run the pairing filter here, even when
-		// NetworkID is absent or needTopology is false, because
-		// candidates satisfying the count alone may still be spread
-		// across incomplete/partial modules rather than full pairs.
+		// Ascend 910C requires full module-pair allocation (2 NPUs per physical card).
 		combination := npu.computeBestCombination910C(nodeInfo, int(originReq), tmpDevs[k.Type])
 		if len(combination) != int(originReq) {
-			// Never report success on a short allocation: doing so silently
-			// under-allocates NPUs relative to what the pod requested.
 			reason[common.AllocatedCardsInsufficientRequest] = len(combination)
 			klog.V(5).InfoS(common.AllocatedCardsInsufficientRequest, "pod", klog.KObj(pod), "request", originReq, "allocated", len(combination))
 			return false, tmpDevs, common.GenReason(reason, int(originReq))
@@ -661,7 +655,6 @@ func (npudev *Devices) computeBestCombination(nodeInfo *device.NodeInfo, reqNum 
 }
 
 func (npudev *Devices) computeBestCombination910C(nodeInfo *device.NodeInfo, reqNum int, containerDevices device.ContainerDevices) device.ContainerDevices {
-	// Build a mapping from NPU index to device object for quick lookup.
 	indexToDevice := make(map[int]device.ContainerDevice)
 	var npuIndices []int
 	for _, dev := range containerDevices {
@@ -673,25 +666,17 @@ func (npudev *Devices) computeBestCombination910C(nodeInfo *device.NodeInfo, req
 	// Each physical card hosts exactly 2 NPUs (Ascend 910C module design).
 	const MaxCardNPUNum = 2
 
-	// Group NPU indices by the module and Sort
 	cardTopology := make(map[int][]int)
 	for _, idx := range npuIndices {
 		cardId := idx / MaxCardNPUNum
 		cardTopology[cardId] = append(cardTopology[cardId], idx)
 	}
 
-	// Convert the card topology map into a slice for sorting.
 	cardTopSlice := make([][]int, 0, len(cardTopology))
 	for _, card := range cardTopology {
 		cardTopSlice = append(cardTopSlice, card)
 	}
 
-	// Sort cards by the number of available NPUs in descending order, so that
-	// full cards are considered before partial ones. Note: partial cards are
-	// still excluded outright below via the MaxCardNPUNum equality check, so
-	// this ordering only affects which full cards get picked first when more
-	// full cards are available than requested — it does not by itself change
-	// whether a request succeeds or fails.
 	sort.Slice(cardTopSlice, func(i, j int) bool {
 		return len(cardTopSlice[i]) > len(cardTopSlice[j])
 	})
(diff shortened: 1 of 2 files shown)


pkg/device/ascend/device.go
```diff
@@ -553,12 +553,31 @@ func (npu *Devices) Fit(devices []*device.DeviceUsage, request device.ContainerD
...
 		}
-		if k.Nums == 0 && !needTopology {
+		if k.Nums == 0 && !needTopology && (k.Type != Ascend910CType || originReq <= 1) {
 			klog.V(4).InfoS("device allocate success", "pod", klog.KObj(pod), "allocate device", tmpDevs)
 			return true, tmpDevs, ""
 		}
 	}
 
+	if k.Type == Ascend910CType && originReq > 1 {
+		// Ascend 910C requires full module-pair allocation (2 NPUs per
```
archlitchi: please remove all uncessary comments here and in test files

princexpoddar: done...i removed the unnecessary comments while just keeping the essential ones

archlitchi (approved):
/lgtm

Issue #2363, opened 2026-08-04
Quota is only re-checked in Fit() for NVIDIA, leaving a race window for every other backend


Lakshya77089:
**What happened**:

Raised as a follow-up to #2347, from @mesutoezdil's review there.

`FitQuota` is called from a backend's `Fit()` path in exactly one place —
`pkg/device/nvidia/device.go:809`, via the package-local `fitQuota` helper. No
other backend does this. Grep over the tree finds three call sites total: the
definition in `pkg/device/quota.go`, the admission webhook, and NVIDIA's
`Fit()`.

That leaves a window between admission and scheduling for every non-NVIDIA
backend.

Namespace usage is only incremented by `QuotaManager.AddUsage`, which runs at
Filter time, not at admission. So a burst of pods created back-to-back all
reach the webhook before any of them has been scheduled, all read the same
value of `Used`, and all pass. NVIDIA catches this on the way through `Fit()`,
because by the time the second pod is scored the first one's usage has been
recorded. Cambricon, Ascend, Hygon, Iluvatar, mthreads, Metax and the rest have
no second gate, so the whole burst schedules and the namespace ends up over its
limit.

#2347 fixed the case where a non-NVIDIA quota was ignored outright. This is the
remaining half.

**What you expected to happen**:

A namespace `ResourceQuota` on any vendor's memory or core resource should hold
under concurrent pod creation, not only when pods arrive one at a time.

**How to reproduce it (as minimally and precisely as possible)**:

1. Register a non-NVIDIA backend, e.g. cambricon.
2. Set a namespace `ResourceQuota` on `limits.cambricon.com/mlu.smlu.vmemory`
   sized to fit roughly two pods.
3. Create a Deployment with five replicas requesting that resource, so all five
   pods are admitted before any is scheduled.
4. All five pass admission and schedule. Recorded usage ends up above the limit.

Repeating with `nvidia.com/gpumem` and an equivalent quota denies the excess
replicas, because `Fit()` re-checks.

**Anything else we need to know?**:

Found by code review while working on #2347, not on a cluster, so no hardware
logs. Reproducible with a fake clientset in unit tests.

Two things worth settling before implementation:

- Whether every backend should grow its own `fitQuota` helper, or whether the
  accumulate-then-check logic in `pkg/device/nvidia/device.go:725-744` should
  move to `pkg/device` so each `Fit()` is a single call. The NVIDIA helper is
[...]

PR #2377, opened 2026-08-05
fix(device): re-check ResourceQuota in Fit() for non-NVIDIA backends

Lakshya77089:
**What type of PR is this?**

/kind bug

Follow-up to #2347, from @mesutoezdil's review there. #2347 made the admission
webhook enforce `ResourceQuota` for every backend. Admission alone is not enough.

Namespace usage is only recorded by `QuotaManager.AddUsage` at Filter time, not
at admission. So a burst of pods created together all reach the webhook before
any of them has been scheduled, all read the same value of `Used`, and all pass.
NVIDIA caught this on the way through `Fit()` — by the time the second pod is
scored, the first one's usage has been recorded. No other backend had that
second gate, so the whole burst schedules and the namespace ends up over its
limit.

Reproduces as a five-replica Deployment against a quota sized for two pods: on
`nvidia.com/gpumem` the extra replicas are denied, on
`cambricon.com/mlu.smlu.vmemory` all five schedule.

`fitQuota` was the only implementation and it lived in `pkg/device/nvidia`,
hardcoded to `NvidiaGPUDevice` and `MemoryFactor`. Both of those are now
per-backend values after #2347, so the helper moves to `pkg/device` as
`FitQuotaWithPodDevices` and takes them as arguments. NVIDIA keeps its
one-line wrapper and behaves exactly as before.

The five backends that share NVIDIA's `Fit()` shape — ascend, cambricon, hygon,
iluvatar, mthreads — call it at the same point NVIDIA always has: after `memreq`
is resolved, before the per-card capacity checks.

```go
if !device.FitQuotaWithPodDevices(tmpDevs, allocated, pod.Namespace, k.Type,
    int64(memreq), int64(k.Coresreq), cam.GetResourceNames().MemoryFactor) {
    reason[common.ResourceQuotaNotFit]++
    klog.V(3).InfoS(common.ResourceQuotaNotFit, "pod", klog.KObj(pod), "memreq", memreq, "coresreq", k.Coresreq)
    continue
}
```

The helper counts cards already picked for this pod, both the ones in `tmpDevs`
and the ones in `allocated`, so a multi-card request is weighed as a whole
rather than waved through one card at a time.

**Tests**

`TestFitQuotaWithPodDevices` and `TestFitQuotaWithPodDevicesNilAllocated` in
`pkg/device/quota_test.go` cover the accumulation across `tmpDevs` and
`allocated`, cores exceeding on their own, the memory factor raising the limit,
and a namespace with no quota.

`TestDevices_FitResourceQuota` in `pkg/device/cambricon/device_test.go` is the
one that matters — it drives the real `Fit()` and asserts a card with plenty of
free memory is still refused when the namespace is out of quota, with
`ResourceQuotaNotFit` in the reason. It fails on master and passes here.

Refs #2363

Deliberately `Refs` rather than `Fixes`, because #2363 also covers the backends
below.

Scope. amd, awsneuron, enflame, kunlun vdevice and metax sgpu also expose
quota-capable resources, but their `Fit()` is shaped differently — no shared
`memreq` resolution block to hang the check on. Wiring them up means reading
each allocation loop on its own terms, and I would rather not bury five
one-line insertions and five bespoke ones in the same diff. #2363 stays open
for them.

The bigger question from #2363 is still open and I have not tried to answer it
here: reserving quota at admission instead of at Filter would close the window
for every backend at once, rather than adding a gate per backend. That is a
larger change and wants your input before anyone writes it. This PR is the
incremental step, and the shared helper is needed either way.

One behaviour note: a pod that previously scheduled while its namespace was
over quota will now be refused on these five backends, with
`ResourceQuotaNotFit` in the filter reason. That is the fix working, but it is
a visible change for anyone who had been relying on the quota not being
enforced.

Not validated on real hardware. The change is in the scheduler extender's Fit
path and is covered by unit tests, which CONTRIBUTING allows for
scheduler-scoped changes.

```release-note
Namespace ResourceQuota is now re-checked during scheduling for Ascend,
Cambricon, Hygon, Iluvatar and Mthreads devices, closing a window where pods
created concurrently could all pass admission and collectively exceed the limit.
```

---

AI assistance disclosure: this change was developed with Claude Code — codebase
exploration, the shared helper, and the tests. Flagging the extent up front per
CONTRIBUTING.

## Summary by CodeRabbit

* **Bug Fixes**
  * Device scheduling now respects namespace-level memory and core quotas across supported device types.
  * Requests exceeding available quotas are rejected with clearer diagnostic information, while other eligible devices continue to be evaluated.
  * Quota calculations account for resources already selected or allocated across containers and finalized device combinations, including topology-aware selections.
  * Namespaces without configured quotas remain unrestricted.

### DIFF
--- a/pkg/device/ascend/device.go
+++ b/pkg/device/ascend/device.go
@@ -512,6 +512,12 @@ func (npu *Devices) Fit(devices []*device.DeviceUsage, request device.ContainerD
 			//This incurs an issue
 			memreq = dev.Totalmem * k.MemPercentagereq / 100
 		}
+		if !needTopology && !device.FitQuotaWithPodDevices(tmpDevs, allocated, pod.Namespace, k.Type,
+			int64(memreq), int64(k.Coresreq), npu.GetResourceNames().MemoryFactor) {
+			reason[common.ResourceQuotaNotFit]++
+			klog.V(3).InfoS(common.ResourceQuotaNotFit, "pod", klog.KObj(pod), "memreq", memreq, "coresreq", k.Coresreq)
+			continue
+		}
 		if dev.Totalmem-dev.Usedmem < memreq {
 			reason[common.CardInsufficientMemory]++
 			klog.V(5).InfoS(common.CardInsufficientMemory, "pod", klog.KObj(pod), "device", dev.ID, "device index", i, "device total memory", dev.Totalmem, "device used memory", dev.Usedmem, "request memory", memreq)
@@ -575,6 +581,9 @@ func (npu *Devices) Fit(devices []*device.DeviceUsage, request device.ContainerD
 
 	if needTopology {
 		if len(tmpDevs[k.Type]) == int(originReq) {
+			if !npu.fitQuotaForSelection(tmpDevs, allocated, pod, k.Type, reason) {
+				return false, tmpDevs, common.GenReason(reason, len(devices))
+			}
 			klog.V(5).InfoS("device allocate success", "pod", klog.KObj(pod), "allocate device", tmpDevs)
 			return true, tmpDevs, ""
 		} else if len(tmpDevs[k.Type]) > int(originReq) {
@@ -584,6 +593,9 @@ func (npu *Devices) Fit(devices []*device.DeviceUsage, request device.ContainerD
 				// If requesting multiple devices, select the best combination of cards.
 				combination := npu.computeBestCombination(nodeInfo, int(originReq), tmpDevs[k.Type])
 				tmpDevs[k.Type] = combination
+			}
+			if !npu.fitQuotaForSelection(tmpDevs, allocated, pod, k.Type, reason) {
+				return false, tmpDevs, common.GenReason(reason, len(devices))
 			}
 			klog.V(5).InfoS("device allocate success", "pod", klog.KObj(pod), "best device combination", tmpDevs)
 			return true, tmpDevs, ""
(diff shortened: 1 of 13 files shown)


archlitchi:

pkg/device/ascend/device_test.go
```diff
@@ -2457,3 +2458,156 @@ func TestDevices_AddResourceUsage(t *testing.T) {
...
+			Index:      uint(i),
+			Health:     true,
+			CustomInfo: map[string]any{"NetworkID": float64(1)},
+		})
+	}
+
+	qm := device.NewQuotaManager()
+	t.Cleanup(func() { delete(qm.Quotas, "ascend-topo") })
+
+	// Room for the two cards actually requested (32768) but not for all four
```
archlitchi: Please remove these comments

pkg/device/ascend/device.go
```diff
@@ -512,6 +512,17 @@ func (npu *Devices) Fit(devices []*device.DeviceUsage, request device.ContainerD
 			//This incurs an issue
 			memreq = dev.Totalmem * k.MemPercentagereq / 100
 		}
+		// In topology mode k.Nums is not decremented, so this loop gathers every
```
archlitchi: please remove all comments inside method body here and in other files

archlitchi:
please sync with master to pass the UT

Lakshya77089:
Thanks for the review — both done.

Removed the comments inside method bodies, 25 lines across `ascend/device.go` and the ascend, cambricon, hygon, iluvatar and mthreads test files. I kept the doc comments directly above declarations, since those are the usual Go convention; say the word if you would rather those went too and I will strip them as well.

While rebasing I also had to update the quota fixtures in my tests. #2313 changed `FitQuota` to gate on `Quota.LimitSet` rather than `Limit != 0`, so a `Quota` built as a struct literal instead of through `AddQuota` now reads as "no limit configured" and every deny case silently passed. Same thing that had `TestFitResourceQuotaNonNvidia` and two others failing on master until you fixed those fixtures this morning.

Rebased onto c5e99b6 and CI is green.

Lakshya77089:
@archlitchi gentle ping on this one when you have a moment. The in-body comments you asked about are removed, and I've rebased onto current master — CI is green.

pkg/device/quota.go
```diff
@@ -94,6 +94,34 @@ func (q *QuotaManager) FitQuota(ns string, memreq int64, memoryFactor int32, cor
...
+	}
+	if allocated != nil {
+		for _, containerDevices := range (*allocated)[deviceName] {
+			for _, val := range containerDevices {
+				mem += int64(val.Usedmem)
+				core += int64(val.Usedcores)
+			}
+		}
+	}
+	return GetLocalCache().FitQuota(ns, mem, memoryFactor, core, deviceName)
```
Shouren, line 123: @Lakshya77089 I think we shouldn't remove the log message in the previouse code.

Shouren (changes requested):

Lakshya77089:
@Shouren good catch — that was an accidental drop when the helper moved out of the nvidia package, not intentional. Restored it in `FitQuotaWithPodDevices` so every backend now logs the accumulated mem/cores the same way nvidia did. Pushed.

mesutoezdil:

PR #2380, opened 2026-08-05
fix(device-plugin): resolve stale annotation in multi-container Allocate

### LINKED ISSUE #1050: CUDA_DISABLE_CONTROL=true has no effect — HAMI still hijacks GPU behavior
**Description:**

Hi,

I’m currently running a container with the following setup:

* GPU: NVIDIA L40S
* Driver version: 570.124.06
* CUDA version: 12.8

Even when I set `CUDA_DISABLE_CONTROL=true`, I’m seeing that `hami` still hijacks control of the GPU and outputs messages like:

```
[HAMI-core Msg]: Initializing.....
[HAMI-core Msg]: Calling exit handler
```

As shown in `nvidia-smi`, HAMI seems to remain active:

```
root@model-d6fb5dbc5-d2t7p:/workspace# env | grep CUDA_DISABLE_CONTROL
CUDA_DISABLE_CONTROL=true
root@model-d6fb5dbc5-d2t7p:/workspace# nvidia-smi
[HAMI-core Msg(417:139803620223872:libvgpu.c:836)]: Initializing.....
Wed May 14 21:23:40 2025
+-----------------------------------------------------------------------------------------+
| NVIDIA-SMI 570.124.06             Driver Version: 570.124.06     CUDA Version: 12.8     |
|-----------------------------------------+------------------------+----------------------+
| GPU  Name                 Persistence-M | Bus-Id          Disp.A | Volatile Uncorr. ECC |
| Fan  Temp   Perf          Pwr:Usage/Cap |           Memory-Usage | GPU-Util  Compute M. |
|                                         |                        |               MIG M. |
|=========================================+========================+======================|
|   0  NVIDIA L40S                    On  |   00000000:3F:00.0 Off |                    0 |
| N/A   42C    P0             84W /  350W |     970MiB /  46068MiB |      0%      Default |
|                                         |                        |                  N/A |
+-----------------------------------------+------------------------+----------------------+

+-----------------------------------------------------------------------------------------+
| Processes:                                                                              |
|  GPU   GI   CI              PID   Type   Process name                        GPU Memory |
|        ID   ID                                                               Usage      |
|=========================================================================================|
+-----------------------------------------------------------------------------------------+
[HAMI-core Msg(417:139803620223872:multiprocess_memory_limit.c:501)]: Calling exit handler 417
```

**Question:**
Is there a proper way to fully disable HAMI behavior inside containers?
Setting `CUDA_DISABLE_CONTROL=true` does not appear to work.

Thanks in advance for your help!

**Environment**:
- HAMi version: 2.5.1
- Kubernetes version: v1.28.10+rke2r1
- Others:

peachest:
## What type of PR is this?

/kind bug

A Pod with 2 containers both requesting split GPU. If the first container does NOT set `CUDA_DISABLE_CONTROL` and the second does, the expected behavior is: the first container's libvgpu.so interception takes effect, the second does NOT. But the actual behavior: the second container's `CUDA_DISABLE_CONTROL=true` is ignored — libvgpu.so stays active.

### Root Cause

The `Allocate` loop decodes the pod annotation on every iteration via `GetNextDeviceRequest`, but `eraseNextDeviceTypeFromAnnotation` only patches the API server without updating the in-memory `pod.Annotations` map. When kubelet passes multiple `ContainerRequests` in a single `Allocate` call, every iteration re-reads the original annotation, always returning the first container's devices and env vars.

This is complementary to PR #1299 which fixed the **scheduler side** (correct annotation indexing for multi-container pods). The **device plugin side** (`Allocate` stale annotation) was not fixed.

### Fix

Replace the per-iteration decode+erase pattern with a pop-based design:

- `decodePodSingleDevice(dtype, pod)` — decode the annotation once before the loop
- `popNextContainerDevices(pod, podSingleDev)` — mutate the in-memory slice in place and return the resolved container directly
- `patchErasedAnnotation(pod, dtype, podSingleDev)` — patch the API server exactly once after the loop

This also reduces API server pressure: one patch per `Allocate` call instead of one per container.

Fixes #1050

- this fix align to my previous implementation for ascend-device-plugin Allocate method. Now both device-plugin use the same pop-erase in memory-patch mechanism
- `make verify` passes (staticcheck warnings are pre-existing in `cmd/nvlink-diag/main.go`, not introduced by this PR)
- 14 regression tests added in `alloc_refactor_test.go`
- `GetNextDeviceRequest` is kept as exported (dead code now) to avoid breaking API consumers; can be cleaned up in a follow-up

```release-note
Fix: multi-container Pods now correctly honor `CUDA_DISABLE_CONTROL=true` per-container. Previously, the second container's `CUDA_DISABLE_CONTROL` was silently ignored because the Allocate function re-read stale pod annotations on each iteration.
```

---

> AI Assistance: The code and tests in this PR were primarily authored by  pi agent+glm5.2. The solution design, code review, and test planning were directed by the contributor. All code has been reviewed and understood by the contributor.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved NVIDIA GPU allocation for single- and multi-container workloads.
  * Improved handling of MIG, virtual, replica, and initialization-container assignments.
  * Added clearer errors for invalid requests, allocation mismatches, and annotation failures.
  * Preserved allocation state when processing fails, supporting safer retries.
  * Ensured device allocation annotations are updated consistently after successful processing.

* **Tests**
  * Expanded coverage for allocation behavior, annotation updates, CUDA controls, error handling, and preferred MIG allocation.

### DIFF
--- a/pkg/device-plugin/nvidiadevice/nvinternal/plugin/alloc_refactor_test.go
+++ b/pkg/device-plugin/nvidiadevice/nvinternal/plugin/alloc_refactor_test.go
@@ -18,6 +18,7 @@ package plugin
 
 import (
 	"context"
+	"errors"
 	"os"
 	"testing"
 
@@ -31,6 +32,7 @@ import (
 	clienttesting "k8s.io/client-go/testing"
 
 	"github.com/Project-HAMi/HAMi/pkg/device"
+	"github.com/Project-HAMi/HAMi/pkg/device-plugin/nvidiadevice/nvinternal/rm"
 	"github.com/Project-HAMi/HAMi/pkg/device/nvidia"
 	"github.com/Project-HAMi/HAMi/pkg/util/client"
 )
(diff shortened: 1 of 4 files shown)


pkg/device-plugin/nvidiadevice/nvinternal/plugin/server.go
```diff
@@ -727,6 +732,17 @@ func (plugin *NvidiaDevicePlugin) Allocate(ctx context.Context, reqs *kubeletdev
 			responses.ContainerResponses = append(responses.ContainerResponses, response)
 		}
 	}
+
+	// Patch the annotation with the in-memory erased podSingleDev exactly once.
```
archlitchi: please remove these comments here and above

peachest: Comments removed in c1499ffe. Thank you for the review!

pkg/device-plugin/nvidiadevice/nvinternal/plugin/alloc_refactor_test.go
```diff
@@ -0,0 +1,801 @@
...
+	request := &kubeletdevicepluginv1beta1.AllocateRequest{
+		ContainerRequests: []*kubeletdevicepluginv1beta1.ContainerAllocateRequest{
+			{DevicesIds: []string{"MIG-aaa-0"}},
+		},
+	}
+
+	response, err := plugin.Allocate(context.Background(), request)
+	require.NoError(t, err)
+	require.Len(t, response.ContainerResponses, 1)
+}
```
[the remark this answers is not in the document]
peachest: Thanks for the analysis. After investigating, we decided **not** to add `popNextContainerDevices` to the MIG branch, for the following reasons:

MIG branch does not use annotation data.** The MIG branch obtains device IDs directly from `req.DevicesIds` (kubelet-provided) and passes them to `getAllocateResponse`. It does not read UUID/Usedmem/Usedcores from the annotation.  

Also the original code also skipped annotation for MIG,  our refactor preserves this behavior.

peachest:
unit test fail  due to master branch fail the unit test

archlitchi:
please resolve these conflicts and sync with master to pass the CI

archlitchi (approved):
/lgtm

PR #2386, opened 2026-08-05
fix(scheduler): apply ResourceQuota updates without dropping the limit

### LINKED ISSUE #2385: ResourceQuota is unenforced while an update is applied: onUpdateQuota zeroes limits before restoring them
**What happened**:

`Scheduler.onUpdateQuota` in `pkg/scheduler/scheduler.go` handles a ResourceQuota
update by deleting the old one and adding the new one:

```go
func (s *Scheduler) onUpdateQuota(oldObj, newObj any) {
	s.onDelQuota(oldObj)
	s.onAddQuota(newObj)
}
```

`QuotaManager.DelQuota` sets `Limit = 0` for each managed key, takes and releases
the manager's write lock, and returns. `AddQuota` then takes the write lock again
to restore the limits. The lock is free between the two calls.

`FitQuota` treats a zero limit as "no limit":

```go
```

So any `FitQuota` running in that gap — from admission or from a backend's
`Fit()`, both on separate goroutines — sees `Limit == 0` and skips the check
entirely. The namespace is unenforced for the duration.

This is not a rare interleaving. The kube-controller-manager's quota controller
rewrites `status.used` on essentially every pod create and delete in the
namespace, so `UpdateFunc` fires continuously in any active namespace. The
informer's 1 hour resync fires it again for every ResourceQuota even when
nothing changed. Note `AddQuota` only reads `Spec.Hard`, so a status-only update
is semantically a no-op — it zeroes and restores the limits for nothing.

A ResourceQuota update should never leave a window in which the quota is not
enforced. Applying an update should be atomic with respect to `FitQuota`.

Drive `DelQuota`/`AddQuota` on one goroutine while calling `FitQuota` on
another, with the namespace already at its limit so every answer must be false:

```go
qm.AddQuota(q)                                   // limits.nvidia.com/gpumem: 1000
(*qm.Quotas[ns])["nvidia.com/gpumem"].Used = 1000 // already full

go func() { for { qm.DelQuota(q); qm.AddQuota(q) } }()   // the informer thread

for i := 0; i < 200000; i++ {
    if FitQuota(ns, 1, 1, 0, "NVIDIA") {          // must never be true
        wrongAdmits++
    }
}
```

Result on master:

```
FitQuota wrongly admitted 108503 times out of 200000 while quota updates were in flight
```

Roughly half of all quota checks pass while updates are in flight. On a real
cluster the window is smaller than this synthetic loop, but it opens on every
pod event in the namespace, which is exactly when pods are being scheduled.

Found by code review and confirmed with the loop above, not on a cluster, so no
cluster logs. It reproduces in a unit test.

The fix I have in mind is to apply the two halves under a single acquisition of
the manager's lock, something like a `QuotaManager.UpdateQuota(old, new)` that
does the delete and the add without releasing in between, with `onUpdateQuota`
calling that instead of the two handlers. Happy to send that, or to take a
different shape if you would rather the manager not grow another method.

Two nearby things this touches, neither of which I want to fold in here:

- #2312 changes what a zero limit means. If an explicit `0` becomes "deny
  everything", then `DelQuota` writing `Limit = 0` to mean "removed" becomes
  ambiguous in the other direction, and the two want reconciling.
- `DelQuota` never removes the map entry, it only zeroes the limit, so a
  namespace that once had a quota keeps its entry and its `Used` accounting
  forever. That is a leak rather than a correctness bug and probably deserves
  its own issue.

**Environment**:
- HAMi version: master

Lakshya77089:
**What type of PR is this?**

/kind bug

`Scheduler.onUpdateQuota` applied a ResourceQuota change by deleting the old one
and adding the new one:

```go
func (s *Scheduler) onUpdateQuota(oldObj, newObj any) {
	s.onDelQuota(oldObj)
	s.onAddQuota(newObj)
}
```

`DelQuota` zeroes each managed limit, releases the manager's write lock and
returns. `AddQuota` then takes the lock again to restore them. The lock is free
in between, and `FitQuota` treats a zero limit as no limit:

```go
```

So a quota check landing in that gap, from admission or from a backend's
`Fit()`, both on their own goroutines, skips the comparison and admits the pod.
The namespace is unenforced for the duration.

This is not a rare interleaving. The kube-controller-manager quota controller
rewrites `status.used` on essentially every pod create and delete in the
namespace, so `UpdateFunc` fires continuously in any active namespace, which is
exactly when pods are being scheduled. The informer's 1 hour resync fires it
again for every ResourceQuota even when nothing changed. `AddQuota` only reads
`Spec.Hard`, so those status-only updates zero and restore the limits for
nothing.

**The fix**

`QuotaManager.UpdateQuota(old, new)` does both halves under one acquisition of
the lock. The add and delete bodies move into `addQuotaLocked` / `delQuotaLocked`
so all three entry points share them, and the repeated `limits.` prefix and
`IsManagedQuota` check becomes `managedQuotaName`.

On the scheduler side `onUpdateQuota` calls it, and the informer-object
unwrapping that `onDelQuota` already did, tombstone included, moves into
`asResourceQuota` so both handlers use it. `onDelQuota` previously logged
"unknown resource quota delete object type" for an update too; it now says
"object" since it is shared.

**Tests**

`TestUpdateQuotaKeepsLimitEnforced` drives updates on one goroutine while a
second asks `FitQuota` 50000 times, with the namespace already at its limit so
every answer must be false. Against the old del-then-add it records **25001**
wrongly admitted checks; with this change, zero. `TestUpdateQuota` covers the
plain semantics: a raised limit is applied, a key the new quota drops is
released, and `Used` survives the update.

Fixes #2385

`DelQuota` writing `Limit = 0` to mean "removed" is left as it is. #2312 is
changing what an explicit zero means, and if it lands then the sentinel and a
real zero limit want reconciling. That is a semantic decision for that issue,
not something to settle here, and this PR is correct either way, since the point
is that no reader ever observes the intermediate state.

Also left alone: `DelQuota` never removes the map entry, so a namespace that
once had a quota keeps its entry and its `Used` accounting for the process
lifetime. That is a slow leak rather than a correctness problem and I did not
want to fold it in. Noted in #2385.

Not validated on real hardware. The change is confined to the scheduler
extender's quota bookkeeping and is covered by unit tests, which CONTRIBUTING
allows for scheduler-scoped changes.

```release-note
Fixed namespace ResourceQuota going briefly unenforced every time the quota
object was updated, which in an active namespace meant on nearly every pod
create and delete.
```

---

AI assistance disclosure: this change was developed with Claude Code — spotting
the window, the fix, and the tests. Flagging the extent up front per
CONTRIBUTING.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved quota updates to apply limit changes atomically, preventing temporary incorrect admission decisions.
  * Preserved existing resource usage when quota limits change.
  * Correctly removed resources that are no longer included in updated quotas.
  * Improved handling of quota add and delete events, including deleted-object notifications.

* **Tests**
  * Added coverage for concurrent quota updates and enforcement of limits during changes.

### DIFF
--- a/pkg/device/quota.go
+++ b/pkg/device/quota.go
@@ -188,21 +188,49 @@ func IsManagedQuota(quotaName string) bool {
 func (q *QuotaManager) AddQuota(quota *corev1.ResourceQuota) {
 	q.mutex.Lock()
 	defer q.mutex.Unlock()
+	q.addQuotaLocked(quota)
+	q.logQuotasLocked()
+}
+
+func (q *QuotaManager) DelQuota(quota *corev1.ResourceQuota) {
+	q.mutex.Lock()
+	defer q.mutex.Unlock()
+	q.delQuotaLocked(quota)
+	q.logQuotasLocked()
+}
+
+// UpdateQuota swaps oldQuota's limits for newQuota's without releasing the lock
+// in between. Doing it as DelQuota followed by AddQuota would leave the limits
+// zeroed for the gap between the two, and FitQuota reads a zero limit as no
+// limit, so a check landing in that gap is waved through. The kube quota
+// controller rewrites status.used on every pod event, so updates arrive
+// constantly while pods are being scheduled.
+func (q *QuotaManager) UpdateQuota(oldQuota, newQuota *corev1.ResourceQuota) {
+	q.mutex.Lock()
+	defer q.mutex.Unlock()
+	if oldQuota != nil {
+		q.delQuotaLocked(oldQuota)
+	}
+	if newQuota != nil {
+		q.addQuotaLocked(newQuota)
+	}
+	q.logQuotasLocked()
+}
+
+// addQuotaLocked requires q.mutex to be held.
+func (q *QuotaManager) addQuotaLocked(quota *corev1.ResourceQuota) {
 	for idx, val := range quota.Spec.Hard {
 		value, ok := val.AsInt64()
 		if ok {
-			if !strings.HasPrefix(idx.String(), "limits.") {
-				continue
-			}
-			dn := strings.TrimPrefix(idx.String(), "limits.")
-			if !IsManagedQuota(dn) {
+			dn, ok := managedQuotaName(idx)
+			if !ok {
 				continue
 			}
 			if q.Quotas[quota.Namespace] == nil {
 				q.Quotas[quota.Namespace] = &DeviceQuota{}
 			}
 			dp := q.Quotas[quota.Namespace]
-			_, ok := (*dp)[dn]
+			_, ok = (*dp)[dn]
 			if !ok {
 				(*dp)[dn] = &Quota{
 					Used:  0,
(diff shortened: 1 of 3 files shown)


archlitchi (approved):
/lgtm

PR #2388, opened 2026-08-05
fix(hygon,metax): guard int32 narrowing in GenerateResourceRequests

### LINKED ISSUE #2383: bug: int32 overflow in GenerateResourceRequests silently drops memory request (hygon, metax-sgpu)
## What happened

GenerateResourceRequests in the hygon and metax-sgpu backends converts the requested device memory into an int32 Memreq without bounds checking. A request whose value exceeds math.MaxInt32 silently wraps to a small or negative number.

### Affected backends

- **hygon** (pkg/device/hygon/device.go, lines 185-215): memnums = memnums * int64(MemoryFactor) then memnum = int(memnums) then Memreq: int32(memnum) — no overflow guard before the int32 cast
- **metax-sgpu** (pkg/device/metax/sdevice.go, lines 241-268): mem = v * MemoryFactor then Memreq: int32(mem) — no overflow guard before the int32 cast. Additionally, Nums: int32(count) and Coresreq: int32(core) also lack bounds checks.

### Root cause

A Kubernetes Quantity with a Gi unit (16Gi) yields AsInt64() == 17179869184, which exceeds the int32 range even before the memory factor is applied. For example, on the hygon backend with MemoryFactor: 2:

`
16Gi -> AsInt64() = 17179869184
17179869184 * 2 = 34359738368
int(34359738368) = 34359738368
int32(34359738368) = 0  (wrapped)
`

Because `Memreq` overflows to an incorrect `int32` value (which may be zero or another wrapped value depending on the input), the `dev.Totalmem - dev.Usedmem < memreq` check in `Fit` can incorrectly succeed. As a result, the scheduler may treat the device as having sufficient memory and schedule a pod onto an already overcommitted device, leading to incorrect resource accounting and potential runtime failures.

Oversized memory requests should be rejected instead of silently overflowing. `GenerateResourceRequests` should return an empty `ContainerDeviceRequest{}` when the computed memory request exceeds `math.MaxInt32`, matching the overflow protection already implemented in other device backends.

## How to reproduce it

A container such as:

```yaml
limits:
  hygon.com/dcunum: "1"
  hygon.com/dcumem: 16Gi
```

causes `GenerateResourceRequests` to overflow the computed memory request when converting it to `int32`, producing an incorrect `Memreq` value instead of rejecting the request.

## Anything else we need to know

This is the same overflow class that was already addressed for:

- **enflame** in #2145 / #2190
- **mthreads** and **iluvatar** in #2285
- The **AMD** backend already performs equivalent bounds checks (`pkg/device/amd/device.go`, lines 223–226).

Issue #2336 identified **hygon** and **metax-sgpu** as affected, but it was closed before fixes for these two backends were merged. PR #2285 covers **cambricon**, **iluvatar**, and **mthreads** only.

This issue follows the same validation pattern introduced in #2285.

Eshiv-Pandey:
**What type of PR is this?**
/kind bug

`GenerateResourceRequests` in the hygon and metax-sgpu device backends converts user-controlled `int64` resource limits (device count, memory, cores) into `int32` fields on `ContainerDeviceRequest` (`Nums`, `Memreq`, `Coresreq`) without bounds checking. A resource limit exceeding `math.MaxInt32` silently wraps to an incorrect small or negative value that flows into `Fit()`, causing the scheduler to incorrectly assess device availability and potentially schedule pods onto overcommitted devices.

This PR adds range guards that return an empty `ContainerDeviceRequest{}` (the existing rejection sentinel) before any `int32()` narrowing, matching the overflow protection already implemented in the enflame, iluvatar, mthreads, and amd backends.

Fixes #2383

- The hygon fix guards three conversion points: device count, memory pre-factor, and memory post-factor (after `MemoryFactor` multiplication).
- The metax-sgpu fix guards three conversion points: device count, core, and memory.
- Test coverage includes both pre-factor overflow (`16Gi` exceeds `MaxInt32` directly) and post-factor overflow (raw value `3000000` fits in int32, but `3000000 * 1024` overflows), plus an out-of-range memory case for metax-sgpu.
- Follows the same validation pattern introduced in #2285.

```
Oversized device resource requests (e.g. hygon.com/dcumem: 16Gi, or metax-tech.com/vmemory exceeding the int32 range) are now correctly rejected instead of silently wrapping to an incorrect memory value.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Added validation for device counts, memory, and core resource requests to reject negative or oversized values.
  * Prevented integer overflow when converting memory requests, including values affected by memory scaling.
  * Invalid resource requests now return an empty device request instead of producing incorrect results.
* **Tests**
  * Added regression coverage for oversized and overflow-prone resource requests.

### DIFF
--- a/pkg/device/metax/sdevice.go
+++ b/pkg/device/metax/sdevice.go
@@ -259,9 +259,6 @@ func (sdev *MetaxSDevices) GenerateResourceRequests(ctr *corev1.Container) devic
 			if hasUnit {
 				mem = v / 1024 / 1024
 			} else {
-				// Guard the multiplication itself: v * MemoryFactor can overflow
-				// int64 and wrap to a small value that slips past the range check
-				// below, so reject before multiplying.
 				if v < 0 || v > int64(math.MaxInt32)/int64(MemoryFactor) {
 					klog.ErrorS(nil, "metax sgpu device memory request is out of range", "container", ctr.Name, "request", memQuantity.String())
 					return device.ContainerDeviceRequest{}
--- a/pkg/device/metax/sdevice.go
+++ b/pkg/device/metax/sdevice.go
@@ -21,6 +21,7 @@ import (
 	"errors"
 	"fmt"
 	"maps"
+	"math"
 	"slices"
 	"sort"
 	"strconv"
@@ -229,11 +230,19 @@ func (sdev *MetaxSDevices) GenerateResourceRequests(ctr *corev1.Container) devic
 			ctr.Name, MetaxResourceNameVCount)
 		return device.ContainerDeviceRequest{}
 	}
+	if count <= 0 || count > math.MaxInt32 {
+		klog.ErrorS(nil, "metax sgpu device count request is out of range", "container", ctr.Name, "request", count)
+		return device.ContainerDeviceRequest{}
+	}
 
 	core := int64(100)
 	coreQuantity, ok := ctr.Resources.Limits[corev1.ResourceName(MetaxResourceNameVCore)]
 	if ok {
 		if v, ok := coreQuantity.AsInt64(); ok {
+			if v < 0 || v > math.MaxInt32 {
+				klog.ErrorS(nil, "metax sgpu device core request is out of range", "container", ctr.Name, "request", coreQuantity.String())
+				return device.ContainerDeviceRequest{}
+			}
 			core = v
 		}
 	}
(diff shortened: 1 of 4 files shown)


pkg/device/metax/sdevice.go
```diff
@@ -250,7 +259,18 @@ func (sdev *MetaxSDevices) GenerateResourceRequests(ctr *corev1.Container) devic
 			if hasUnit {
 				mem = v / 1024 / 1024
 			} else {
-				mem = v * MemoryFactor
+				// Guard the multiplication itself: v * MemoryFactor can overflow
```
archlitchi: please remove these comments here

Eshiv-Pandey: Done

archlitchi (approved):
/lgtm

PR #2390, opened 2026-08-05
fix(scheduler): make ResourceQuota update atomic to prevent enforcement gap

### LINKED ISSUE #2385: ResourceQuota is unenforced while an update is applied: onUpdateQuota zeroes limits before restoring them

Nitish08-08:
### Problem

onUpdateQuota calls DelQuota then AddQuota as two separate lock
acquisitions. Between the two calls, FitQuota sees Limit == 0 and
treats it as "no limit", wrongly admitting pods. This fires on every
pod event in the namespace since the quota controller rewrites
status.used on each pod create/delete, plus the 1-hour informer resync.

### Solution

Add QuotaManager.UpdateQuota(old, new) that performs both operations
under a single lock, eliminating the enforcement window. Update
onUpdateQuota to call it instead of the separate handlers.

### Verification

    go test ./pkg/device/... -run "TestUpdateQuota" -short --race -count=1 -v

TestUpdateQuotaAtomicity drives concurrent FitQuota (200k calls) while
UpdateQuota races. Zero wrong admits confirms the fix.

Closes #2385

## Summary by CodeRabbit

* **New Features**
  * Resource quota updates are now applied atomically, preserving existing usage while updating limits.
  * Invalid quota updates safely fall back to removing the previous quota.
* **Bug Fixes**
  * Prevented quota checks from temporarily allowing admissions when usage has reached a limit during concurrent updates.

### DIFF
--- a/pkg/device/quota.go
+++ b/pkg/device/quota.go
@@ -253,6 +253,74 @@ func (q *QuotaManager) DelQuota(quota *corev1.ResourceQuota) {
 	}
 }
 
+// UpdateQuota atomically removes old quota limits and applies new ones under
+// a single lock acquisition. This prevents a window where FitQuota sees zeroed
+// limits (treated as "no limit") between a DelQuota and AddQuota pair.
+func (q *QuotaManager) UpdateQuota(oldQuota, newQuota *corev1.ResourceQuota) {
+	q.mutex.Lock()
+	defer q.mutex.Unlock()
+
+	// Remove old limits
+	if oldQuota != nil {
+		for idx, val := range oldQuota.Spec.Hard {
+			_, ok := val.AsInt64()
+			if !ok {
+				continue
+			}
+			if !strings.HasPrefix(idx.String(), "limits.") {
+				continue
+			}
+			dn := strings.TrimPrefix(idx.String(), "limits.")
+			if !IsManagedQuota(dn) {
+				continue
+			}
+			if dq, ok := q.Quotas[oldQuota.Namespace]; ok {
+				if quotaInfo, ok := (*dq)[dn]; ok {
+					quotaInfo.Limit = 0
+				}
+			}
+		}
+	}
+
+	// Apply new limits
+	if newQuota != nil {
+		for idx, val := range newQuota.Spec.Hard {
+			value, ok := val.AsInt64()
+			if !ok {
+				continue
+			}
+			if !strings.HasPrefix(idx.String(), "limits.") {
+				continue
+			}
+			dn := strings.TrimPrefix(idx.String(), "limits.")
+			if !IsManagedQuota(dn) {
+				continue
+			}
+			if q.Quotas[newQuota.Namespace] == nil {
+				q.Quotas[newQuota.Namespace] = &DeviceQuota{}
+			}
+			dp := q.Quotas[newQuota.Namespace]
+			_, ok := (*dp)[dn]
+			if !ok {
+				(*dp)[dn] = &Quota{
+					Used:  0,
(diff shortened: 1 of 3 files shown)


SK8-infi (commented):
Hii @Nitish08-08 this seems to be a duplicate of #2386 . Can you confirm?
Usually duplicates are closed by maintainers.

archlitchi:
Duplicated with PR: https://github.com/Project-HAMi/HAMi/pull/2386

PR #2393, opened 2026-08-06
fix(device-plugin): skip devices that do not support event based healthchecking

Wangmin362:
**What type of PR is this?**

/kind bug

`nvmlDeviceRegisterEvents` returning `ERROR_NOT_SUPPORTED` means the device cannot report health through the event mechanism, not that it is faulty. The two separate `if`s let that return code fall through into the second branch, so the device is handed to `ListAndWatch` as `Unhealthy` and stops being schedulable. That state is not reversible - `ListAndWatch` carries a `FIXME: there is no way to recover from the Unhealthy state` - and a restart only repeats the same failed registration. The logs contradict themselves: one line says the device is "too old to support healthchecking", the next one marks it unhealthy.

A `switch` makes the two cases exclusive. NVIDIA fixed the same block upstream in k8s-device-plugin (commit `8cd1447`, "Fix healthchecking on old devices").

NONE

Added a test covering the three `RegisterEvents` outcomes. Checked both ways: with `health.go` reverted to master the test fails on the `ERROR_NOT_SUPPORTED` assertion, with the fix the package passes.

I have no card that answers `ERROR_NOT_SUPPORTED` for the mask HAMi asks for, so this is not reproduced end to end on a node. I did probe NVML on our test cluster to confirm the return code is reachable in practice: on a GeForce RTX 4090 D (driver 595.71.05) `nvmlDeviceRegisterEvents` returns `ERROR_NOT_SUPPORTED` for the ECC event bits, and on a Tesla V100 (driver 550.54.15) for a bit outside its supported mask.

This PR was written primarily by Claude Code.

Devices that do not support event based health checking are no longer reported as unhealthy, so they stay schedulable.

## Summary by CodeRabbit

* **Bug Fixes**
  * Devices remain healthy when event registration is unsupported by the available hardware or driver.
  * Other event registration failures continue to be reported and mark the device unhealthy.

* **Tests**
  * Added coverage for successful, unsupported, and failed event registration scenarios.

### DIFF
--- a/pkg/device-plugin/nvidiadevice/nvinternal/rm/health.go
+++ b/pkg/device-plugin/nvidiadevice/nvinternal/rm/health.go
@@ -118,10 +118,10 @@ func (r *nvmlResourceManager) checkHealth(stop <-chan interface{}, devices Devic
 		}
 
 		ret = gpu.RegisterEvents(eventMask&supportedEvents, eventSet)
-		if ret == nvml.ERROR_NOT_SUPPORTED {
+		switch {
+		case ret == nvml.ERROR_NOT_SUPPORTED:
 			klog.Warningf("Device %v is too old to support healthchecking.", d.ID)
-		}
-		if ret != nvml.SUCCESS {
+		case ret != nvml.SUCCESS:
 			klog.Infof("Marking device %v as unhealthy: %v", d.ID, ret)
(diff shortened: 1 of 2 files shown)


archlitchi (approved):
/lgtm

PR #2395, opened 2026-08-06
fix(amd): empty nouse-gputype annotation should not exclude every device

adity1raut:
## What this fixes

`checkAMDType` in `pkg/device/amd/device.go` decides whether an AMD card is eligible for a pod based on the `amd.com/use-gputype` / `amd.com/nouse-gputype` annotations. The in-use check is guarded against an empty annotation value:

```go
if inuse, ok := annos[AMDInUse]; ok {
```

but the no-use check was not:

```go
if noUse, ok := annos[AMDNoUse]; ok {
    noUseTypes := strings.Split(noUse, ",")
    if slices.ContainsFunc(noUseTypes, func(noUseType string) bool {
        return strings.Contains(cardType, strings.ToUpper(strings.TrimSpace(noUseType)))
    }) {
        return false
    }
}
```

If `amd.com/nouse-gputype` is present but set to an empty string, `strings.Split("", ",")` returns `[""]`, and `strings.Contains(cardType, "")` is always `true` in Go (every string contains the empty string). So `ContainsFunc` returns `true`, and the function returns `false` for every card — an empty exclusion annotation ends up excluding *all* AMD devices instead of excluding none, and the pod becomes unschedulable.

The shared `device.CheckType` helper used by the other vendor backends (nvidia, hygon, ...) already guards against this with `ok && strings.TrimSpace(value) != ""`. The AMD backend has its own local copy of this logic and was missing that guard on the no-use branch.

## Fix

Add the same `ok && strings.TrimSpace(...) != ""` guard to the no-use branch, mirroring the existing guard on the in-use branch and the shared helper.

## Testing

- Added `TestCheckAMDType` covering empty/non-empty use and no-use annotations, including the regression case. Verified it fails against the old code and passes with the fix.
- `go test ./pkg/device/amd/... -race -count=1` passes.
- This is scheduler-extender filtering logic (runs during `Fit()`), not device-plugin/container isolation code, so per the contribution guide's testing policy, unit tests are the appropriate validation here — I don't have AMD GPU hardware to validate against.

## AI disclosure

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved AMD device filtering so empty or whitespace-only usage annotations no longer incorrectly exclude devices.
  * Comma-separated filters now ignore blank entries and trim surrounding whitespace for more predictable matching.

* **Tests**
  * Added coverage for AMD filtering edge cases.
  * Improved scheduler cleanup and quota validation test reliability.

### DIFF
--- a/pkg/device/amd/device.go
+++ b/pkg/device/amd/device.go
@@ -164,7 +164,7 @@ func (dev *AMDDevices) NodeCleanUp(nn string) error {
 
 func checkAMDType(annos map[string]string, cardType string) bool {
 	cardType = strings.ToUpper(cardType)
-	if inuse, ok := annos[AMDInUse]; ok {
+	if inuse, ok := annos[AMDInUse]; ok && strings.TrimSpace(inuse) != "" {
 		useTypes := strings.Split(inuse, ",")
 		if !slices.ContainsFunc(useTypes, func(useType string) bool {
 			return strings.Contains(cardType, strings.ToUpper(strings.TrimSpace(useType)))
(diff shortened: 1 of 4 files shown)


archlitchi:
please resolve the code-rabbit comment

archlitchi (approved):
/lgtm

archlitchi (approved):
/lgtm

PR #2397, opened 2026-08-06
fix(device): enforce ResourceQuota in Fit() for all backends

### LINKED ISSUE #2363: Quota is only re-checked in Fit() for NVIDIA, leaving a race window for every other backend

Nitish08-08:
### Problem

Namespace ResourceQuota enforcement has a race window for every backend
except NVIDIA.

The admission webhook records no device usage — that happens at Filter
time via QuotaManager.AddUsage. So when a burst of pods is created
back-to-back, they all reach the webhook before any has been scheduled,
all read the same Used value, and all pass.

NVIDIA catches this because its Fit() method calls a private fitQuota
helper that re-checks quota after each device assignment. The second
pod in the burst sees the first pod's recorded usage and is correctly
denied when over limit.

Every other backend — cambricon, ascend, hygon, iluvatar, mthreads,
metax, amd, enflame, kunlun, awsneuron, vastai, biren — has no such
check. The whole burst schedules and the namespace ends up over its
limit.

This was identified during review of #2347, which fixed the case where
non-NVIDIA quotas were ignored outright. This is the remaining half.

### Reproducer

1. Register a non-NVIDIA backend (e.g. cambricon).
2. Set a namespace ResourceQuota on
   limits.cambricon.com/mlu.smlu.vmemory sized for ~2 pods.
3. Create a Deployment with 5 replicas requesting that resource.
4. All 5 pods pass admission and schedule. Usage exceeds the limit.
5. Repeat with nvidia.com/gpumem — excess replicas are correctly denied.

### Solution

Extract the accumulate-then-check logic from nvidia's private fitQuota
helper into a shared FitQuotaForDevice function that any backend can
call. Every backend's Fit() method now invokes this shared function
before assigning a device, closing the race.

The shared function:
- Sums memory and core usage from the current allocation round (tmpDevs)
  and from previously allocated containers in the pod (allocated)
- Delegates to QuotaManager.FitQuota with the backend's MemoryFactor

### Files changed

| File | Change |
|------|--------|
| pkg/device/quota.go | New FitQuotaForDevice function |
| pkg/device/quota_test.go | TestFitQuotaForDevice (8 test cases) |
| pkg/device/nvidia/device.go | Removed private fitQuota, use shared function |
| pkg/device/cambricon/device.go | Add FitQuotaForDevice call in Fit() |
| pkg/device/ascend/device.go | Add FitQuotaForDevice call in Fit() |
| pkg/device/hygon/device.go | Add FitQuotaForDevice call in Fit() |
| pkg/device/iluvatar/device.go | Add FitQuotaForDevice call in Fit() |
| pkg/device/mthreads/device.go | Add FitQuotaForDevice call in Fit() |
| pkg/device/metax/device.go | Add FitQuotaForDevice call in Fit() |
| pkg/device/metax/sdevice.go | Add FitQuotaForDevice call in Fit() |
| pkg/device/amd/device.go | Add FitQuotaForDevice call in Fit() |
| pkg/device/enflame/device.go | Add FitQuotaForDevice call in Fit() |
| pkg/device/enflame/gcu.go | Add FitQuotaForDevice call in Fit() |
| pkg/device/kunlun/device.go | Add FitQuotaForDevice call in Fit() |
| pkg/device/kunlun/vdevice.go | Add FitQuotaForDevice call in Fit() |
| pkg/device/awsneuron/device.go | Add FitQuotaForDevice call in Fit() |
| pkg/device/vastai/device.go | Add FitQuotaForDevice call in Fit() |
| pkg/device/biren/device.go | Add FitQuotaForDevice call in Fit() |

### How to verify

    go test ./pkg/device/... -run TestFitQuotaForDevice -short --race -count=1 -v

Closes #2363

## Summary by CodeRabbit

* **New Features**
  * Added resource-quota validation for device allocation across supported accelerator types.
  * Allocation now accounts for requested memory and cores alongside existing device capacity checks.
  * When a candidate exceeds quota, it is skipped so other eligible devices can still be considered.
* **Bug Fixes**
  * Improved allocation failure reporting with quota-specific reasons and diagnostic logging.
* **Tests**
  * Added coverage for quota checks involving temporary allocations, existing allocations, missing quotas, and namespaces.

### DIFF
--- a/pkg/device/amd/device.go
+++ b/pkg/device/amd/device.go
@@ -312,6 +312,11 @@ func (amddevice *AMDDevices) Fit(devices []*device.DeviceUsage, request device.C
 		if memReq <= 0 && dev.Totalmem > 0 {
 			memReq = dev.Totalmem
 		}
+		if !device.FitQuotaForDevice(tmpDevs, allocated, pod.Namespace, int64(memReq), int64(k.Coresreq), AMDDevice, amddevice.GetResourceNames()) {
+			reason[common.ResourceQuotaNotFit]++
+			klog.V(3).InfoS(common.ResourceQuotaNotFit, "pod", pod.Name, "memreq", memReq, "coresreq", k.Coresreq)
+			continue
+		}
 		if dev.Totalmem-dev.Usedmem < memReq {
 			reason[common.CardInsufficientMemory]++
 			klog.V(5).InfoS(common.CardInsufficientMemory, "pod", klog.KObj(pod), "device", dev.ID, "device total memory", dev.Totalmem, "device used memory", dev.Usedmem, "request memory", memReq)
--- a/pkg/device/ascend/device.go
+++ b/pkg/device/ascend/device.go
@@ -512,6 +512,11 @@ func (npu *Devices) Fit(devices []*device.DeviceUsage, request device.ContainerD
 			//This incurs an issue
 			memreq = dev.Totalmem * k.MemPercentagereq / 100
 		}
+		if !device.FitQuotaForDevice(tmpDevs, allocated, pod.Namespace, int64(memreq), int64(k.Coresreq), k.Type, npu.GetResourceNames()) {
+			reason[common.ResourceQuotaNotFit]++
+			klog.V(3).InfoS(common.ResourceQuotaNotFit, "pod", pod.Name, "memreq", memreq, "coresreq", k.Coresreq)
+			continue
+		}
 		if dev.Totalmem-dev.Usedmem < memreq {
 			reason[common.CardInsufficientMemory]++
 			klog.V(5).InfoS(common.CardInsufficientMemory, "pod", klog.KObj(pod), "device", dev.ID, "device index", i, "device total memory", dev.Totalmem, "device used memory", dev.Usedmem, "request memory", memreq)
--- a/pkg/device/awsneuron/device.go
+++ b/pkg/device/awsneuron/device.go
@@ -428,6 +428,12 @@ func (neuron *AWSNeuronDevices) Fit(devices []*device.DeviceUsage, request devic
 			continue
 		}
 
+		if !device.FitQuotaForDevice(tmpDevs, allocated, pod.Namespace, int64(k.Memreq), int64(k.Coresreq), AWSNeuronDevice, neuron.GetResourceNames()) {
+			reason[common.ResourceQuotaNotFit]++
+			klog.V(3).InfoS(common.ResourceQuotaNotFit, "pod", pod.Name, "memreq", k.Memreq, "coresreq", k.Coresreq)
+			continue
+		}
+
 		if countMaskAvailable(dev.Totalcore)-countMaskAvailable(dev.Usedcores) < k.Coresreq {
 			reason[common.CardInsufficientCore]++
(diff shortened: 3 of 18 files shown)


Lakshya77089:
There's overlap here with #2377, which is open against the same issue and covers ascend, cambricon, hygon, iluvatar and mthreads. Flagging that early so we don't both spend time on the same five.

One concrete thing worth checking in your ascend change. CodeRabbit raised this on #2377 and it applies to the same insertion point here: in topology mode `k.Nums` is not decremented, so that loop keeps collecting candidate cards and the `originReq` subset is only chosen afterwards by `computeBestCombination`. Charging quota per candidate therefore counts cards the final combination may never use, and once the running total passes the limit the remaining candidates are rejected, which shrinks the pool the combination is picked from. Your insert looks like it sits at that same spot without a `needTopology` guard.

#2377 handles it by skipping the per-card check when `needTopology` is set and validating the chosen selection once instead. Whichever PR ends up carrying this, it probably wants that guard.

Your coverage of amd, awsneuron, biren, enflame, kunlun, metax sgpu and vastai goes further than #2377 does — I left those out because their `Fit()` shapes differ. Happy to defer on the wider set if maintainers would rather take one PR; just wanted the topology point on the record either way.

PR #2343, opened 2026-08-04
fix(scheduler): standardize metric descriptor label key from node to node_name

swastikCommits:
**What type of PR is this?**
/kind bug

**What this PR does / why we need it**:
In `cmd/scheduler/metrics.go`, standard Prometheus metric descriptors currently export the node label as `node`, whereas `cmd/vGPUmonitor/metrics.go` exports `node_name`. 
This label key inconsistency prevents operators from performing Prometheus cross-component joins such as:
```
hami_gpu_memory_allocated_bytes * on(node_name) hami_host_gpu_utilization_ratio
```

Here we standardize scheduler metric descriptors to use the `node_name` label key while preserving legacy mode descriptors for backward compatibility. It also injects a mock metrics provider into unit tests and adds assertions verifying both the presence of node_name and the absence of the old node label.

**Which issue(s) this PR fixes**:
Fixes #2161 

**AI Assistance disclosure**:
I used AI assistance to analyze codebase patterns and structure tests, but all changes were manually inspected, written, and verified.

**Special notes for your reviewer**:
 **_Supersedes #2170_**

**Does this PR introduce a user-facing change?**:
```release-note
Standardize scheduler metric descriptor label key from node to node_name to match vGPUmonitor
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved scheduler metrics stability when metric providers, collectors, or node usage data are unavailable.
  * Prevented quota and scheduled-pod metrics from being emitted when their corresponding managers are unavailable.
  * Preserved filtering of container metrics with empty GPU identifiers.

* **Metrics**
  * Renamed the node label to `node_name`.
  * Added shared-container counts to GPU overview metrics.
  * Enhanced GPU memory lookup logging for container metrics.

PR #2398, opened 2026-08-06
feat(monitor): add node label to host GPU metrics

ipsitapp8:
## AI Assistance Disclosure

This PR was implemented with assistance from Kiro (Claude Sonnet 4.5) and Claude Code. The AI helped with:
- Code generation for adding the node label to metrics
- Test case creation
- Understanding the existing codebase patterns

I have reviewed all changes, understand how they work, and validated them with local tests using WSL2.

---

## Problem

Scheduler metrics include a `node` label, but vGPUmonitor host metrics
(`hami_host_gpu_memory_used_bytes`, `hami_host_gpu_utilization_ratio`)
only have `device_index`, `device_uuid`, and `device_type`.

This breaks PromQL joins between allocation and runtime metrics, making
it impossible to create dual-axis panels showing allocated vs. used memory
on multi-node clusters.

Related to #2126 (Gap #2 from observability gap analysis).

## Changes

- Add `node_name` label to `hami_host_gpu_memory_used_bytes` and
  `hami_host_gpu_utilization_ratio` descriptors
- Read `NODE_NAME` from env in `collectGPUInfo()` (matches existing
  pattern in `collectPodAndContainerInfo()`)
- Thread `nodeName` through `collectGPUDeviceMetrics` →
  `collectGPUMemoryMetrics` → `collectGPUUtilizationMetrics`
- Update both new and legacy metric descriptors for consistency

The label is named `node_name` rather than `node`, per review feedback,
to match the naming direction proposed in #2343 (which would standardize
the scheduler's `node` label to `node_name` too, for consistency across
components).

## Testing

Added `TestHostMetricsIncludeNodeLabel` to verify:
- Host metrics have 4 labels including `node_name`
- `node_name` label contains the correct value from NODE_NAME env

All tests pass (run via WSL2/Ubuntu):
```
=== RUN   TestDescribeCollectSync
--- PASS: TestDescribeCollectSync (0.00s)
=== RUN   TestHostMetricsIncludeNodeLabel
--- PASS: TestHostMetricsIncludeNodeLabel (0.00s)
PASS
ok      github.com/Project-HAMi/HAMi/cmd/vGPUmonitor    0.020s
```

Tests show expected NVML errors (no GPU drivers in test environment), but this
is normal—the change affects only metric label schema, not NVML interaction logic.

No GPU hardware available for live multi-node validation. If a maintainer can
test against a real cluster, I'm happy to iterate based on actual output.

## Example

**Before** (no node label on host metrics — join fails on multi-node):
```promql
# Error: multiple matches for device_uuid across different nodes
hami_gpu_memory_allocated_bytes * on(device_uuid) hami_host_gpu_memory_used_bytes
```

**After** (label added, key aligned with the `node_name` direction proposed in #2343):
```promql
hami_host_gpu_memory_used_bytes{node_name="<node>"}
```

Note: scheduler metrics (`hami_gpu_memory_allocated_bytes`) still use the
label key `node` on master — #2343, which would have renamed it to
`node_name`, was closed without merging. Until that lands, a same-key join
across components needs a bridge:
```promql
hami_gpu_memory_allocated_bytes
  * on(node, device_uuid) group_left()
  label_replace(hami_host_gpu_memory_used_bytes, "node", "$1", "node_name", "(.*)")
```

## Breaking Change?

No. This adds a label, which is backward-compatible. Existing queries that don't reference `node_name` will continue to work.

## Summary by CodeRabbit

* **New Features**
  * Host GPU metrics now include the node name for improved identification.
  * Legacy host metrics include a `nodeid` label.
  * Metrics use the configured node name, or `unknown` when unavailable.

* **Bug Fixes**
  * Improved labeling consistency across GPU device, memory, utilization, and legacy metrics.

### DIFF
--- a/cmd/vGPUmonitor/metrics.go
+++ b/cmd/vGPUmonitor/metrics.go
@@ -61,13 +61,13 @@ var (
 	hostGPUdesc = prometheus.NewDesc(
 		"hami_host_gpu_memory_used_bytes",
 		"GPU device memory usage in bytes",
-		[]string{"node", "device_index", "device_uuid", "device_type"}, nil,
+		[]string{"node_name", "device_index", "device_uuid", "device_type"}, nil,
 	)
 
 	hostGPUUtilizationdesc = prometheus.NewDesc(
 		"hami_host_gpu_utilization_ratio",
 		"GPU core utilization ratio (0-100)",
-		[]string{"node", "device_index", "device_uuid", "device_type"}, nil,
+		[]string{"node_name", "device_index", "device_uuid", "device_type"}, nil,
 	)
 
 	ctrvGPUdesc = prometheus.NewDesc(
--- a/cmd/vGPUmonitor/metrics.go
+++ b/cmd/vGPUmonitor/metrics.go
@@ -61,13 +61,13 @@ var (
 	hostGPUdesc = prometheus.NewDesc(
 		"hami_host_gpu_memory_used_bytes",
 		"GPU device memory usage in bytes",
-		[]string{"device_index", "device_uuid", "device_type"}, nil,
+		[]string{"node", "device_index", "device_uuid", "device_type"}, nil,
 	)
 
 	hostGPUUtilizationdesc = prometheus.NewDesc(
 		"hami_host_gpu_utilization_ratio",
 		"GPU core utilization ratio (0-100)",
-		[]string{"device_index", "device_uuid", "device_type"}, nil,
+		[]string{"node", "device_index", "device_uuid", "device_type"}, nil,
 	)
 
 	ctrvGPUdesc = prometheus.NewDesc(
@@ -136,12 +136,12 @@ func initLegacyDescriptors() {
 	legacyHostGPUdesc = prometheus.NewDesc(
 		"HostGPUMemoryUsage",
 		"GPU device memory usage",
-		[]string{"deviceidx", "deviceuuid", "devicetype"}, nil,
+		[]string{"nodeid", "deviceidx", "deviceuuid", "devicetype"}, nil,
 	)
 	legacyHostGPUUtilizationdesc = prometheus.NewDesc(
 		"HostCoreUtilization",
 		"GPU core utilization",
-		[]string{"deviceidx", "deviceuuid", "devicetype"}, nil,
+		[]string{"nodeid", "deviceidx", "deviceuuid", "devicetype"}, nil,
 	)
 	legacyCtrvGPUdesc = prometheus.NewDesc(
 		"vGPU_device_memory_usage_in_bytes",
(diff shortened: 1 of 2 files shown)


cmd/vGPUmonitor/metrics.go
```diff
@@ -61,13 +61,13 @@ var (
 	hostGPUdesc = prometheus.NewDesc(
 		"hami_host_gpu_memory_used_bytes",
 		"GPU device memory usage in bytes",
-		[]string{"device_index", "device_uuid", "device_type"}, nil,
+		[]string{"node", "device_index", "device_uuid", "device_type"}, nil,
```
archlitchi: please change 'node' to 'node_name' in order to be compatible with https://github.com/Project-HAMi/HAMi/pull/2343

ipsitapp8:
@archlitchi The latest commit (e205fb5) is pushed, addressing the `node_name` label rename and adding table-driven test coverage for legacy/unset-NODE_NAME cases. The `CI Workflow` and `CodeQL` runs are stuck in `action_required` and need a maintainer to approve them before `Compile`/`lint`/`Unit test` can run. Could you approve the workflow run when you get a chance?

cmd/vGPUmonitor/metrics.go
```diff
@@ -239,6 +239,12 @@ func (cc ClusterManagerCollector) Collect(ch chan<- prometheus.Metric) {
 }
```
mesutoezdil, line 239: collectpodandcontainerinfo errors out when node_name is empty. this one falls back to unknown instead. why not do the same here, for consistency?

ipsitapp8: Good catch on the inconsistency, it's intentional, because the two functions use `NODE_NAME` for different things.

In `collectPodAndContainerInfo` the node name is a **functional input**: it's the selector passed to the pod lister (`metrics.go:374`). With an empty value the lister would match pods whose assigned-node annotation is empty, i.e. silently wrong data. Erroring out is the only correct behavior there.

In `collectGPUInfo` the node name is **only a label value**. The underlying NVML reads don't depend on it at all. Since `Collect` just logs the error and moves on (`metrics.go:221-224`), returning an error would mean an unset `NODE_NAME` drops `hami_host_gpu_memory_used_bytes` and `hami_host_gpu_utilization_ratio` from every scrape, metrics that work fine today without any node dependency. That would turn a label-only, backward-compatible change into a regression for anyone deploying vGPUmonitor outside the Helm chart (the chart does inject `NODE_NAME` in `daemonsetnvidia.yaml`, but manual/vendored manifests exist).

I also chose the literal `"unknown"` over an empty string on purpose: Prometheus treats an empty label value as absent, so `on(node_name)` joins would fail in a confusing way, whereas `node_name="unknown"` keeps the series shape stable and makes the misconfiguration visible in a query. This matches the existing precedent in `pkg/version/version.go`, where unset build metadata is exposed as `"unknown"` on `hami_build_info`.

The warning log at `metrics.go:244` surfaces the misconfiguration, and `TestHostMetricsIncludeNodeLabel` covers the unset case for both the new and legacy descriptors.

If you'd rather have strict consistency and prefer that host metrics disappear when `NODE_NAME` isn't set, I'm happy to flip it.

mesutoezdil:
Answers must be written by human being. 

ipsitapp8:
@mesutoezdil The mistake was from my side! however I understood the workflow and wrote the reply by myself.. If you want you want give a chance for this
collectPodAndContainerInfo uses the node name to filter data so node name must be exact while the other one uses nodename as a label on data and if any machine is misconfigured, it would return an error and all metrics of the GPU will be lost even when the GPU is working fine.

PR #2399, opened 2026-08-06
fix(scheduler): stop leaked test informers to prevent data races

### LINKED ISSUE #2389: Flaky: data race on global device.SupportDevices between leaked test informers and Test_onAddPod_BadDeviceAnnotation
**What happened**:

`go test ./pkg/scheduler/ --race` fails intermittently on master with a data race
on the global `device.SupportDevices` map. It surfaces as
`Test_onAddPod_BadDeviceAnnotation` failing, but that test is only the victim,
not the cause.

```
WARNING: DATA RACE
Write at 0x00c0000ce480 by goroutine 217:
  runtime.mapassign_faststr()
  github.com/Project-HAMi/HAMi/pkg/scheduler.Test_onAddPod_BadDeviceAnnotation()
      pkg/scheduler/scheduler_test.go:2065

Previous read at 0x00c0000ce480 by goroutine 105:
  runtime.mapIterStart()
  github.com/Project-HAMi/HAMi/pkg/device.DecodePodDevices()
      pkg/device/devices.go:493
  github.com/Project-HAMi/HAMi/pkg/scheduler.(*Scheduler).onAddPod()
      pkg/scheduler/scheduler.go:160
  github.com/Project-HAMi/HAMi/pkg/scheduler.(*Scheduler).onUpdatePod()
      pkg/scheduler/scheduler.go:171
  k8s.io/client-go/tools/cache.(*processorListener).run()
```

The write is the test's own setup line:

```go
device.SupportDevices["TEST"] = "hami.io/test-allocated"
```

The read is a **shared informer goroutine still running from an earlier test**.
Several tests in this package call `informerFactory.Start(s.stopCh)` and never
close `stopCh`, so the listener goroutines outlive the test that created them
and keep dispatching pod events into `onAddPod` for the rest of the binary's
life. `onAddPod` reads `device.SupportDevices`, which has no mutex, while a
later test writes to it.

Because the race detector fails whichever test is running when it fires, the
blame lands on `Test_onAddPod_BadDeviceAnnotation` rather than on the leaked
informer.

`go test ./pkg/scheduler/ --race` should be deterministic. CI runs with `--race`,
so this is a real source of spurious red builds, and a flaky failure here trains
people to re-run rather than read the output.

On a clean checkout of master, run the package with the race detector in a loop:

```
for i in $(seq 1 12); do go test ./pkg/scheduler/ -count=1 --race; done
```

It reproduced on attempt 3 of 12 for me on master. Without `--race` it does not
fail, which is why it is easy to miss locally: the failure needs the detector to
observe the unsynchronised access.

Two separable problems, and I would rather ask before picking:

1. The leaked informers. The tests that call `informerFactory.Start(s.stopCh)`
   never close `stopCh`, so those goroutines run until the binary exits. Closing
   it in a `t.Cleanup` in each affected test would confine each informer to its
   own test.

2. The unsynchronised global. `device.SupportDevices` and `device.InRequestDevices`
   are plain maps written by every `Init*Device` and read from `DecodePodDevices`
   and `EncodePodDevices`. In production they are populated once by
   `config.InitDevicesWithConfig` before informers start, so the ordering is safe
   there and this is not a production data race as far as I can see. But nothing
   in the types says so, and a test that touches them concurrently with a running
   informer has no way to be correct today.

Fixing 1 removes the flake. Fixing 2 makes the invariant explicit rather than
incidental. Happy to send either, or both as separate PRs, but I did not want to
guess which shape you would want.

Found while investigating an intermittent failure in my own PR; it reproduces on
master with nothing of mine applied.

**Environment**:
- HAMi version: master
- Others: unit tests only, `go test --race`, go 1.26

v0idheaven:
## What this PR does

Fixes #2389

Tests that start informers via `informerFactory.Start(s.stopCh)` but never close `stopCh` leave goroutines running after the test completes. These leaked informer goroutines keep dispatching pod events into `onAddPod`, which reads `device.SupportDevices` a package-level `map[string]string` with no mutex, causing unsynchronised concurrent map access and intermittent `-race` failures.

The root cause:
- The **write** is in the test's own setup: `device.SupportDevices["test"] = "hami.io/test-allocated"`
- The **read** is a shared informer goroutine still running from an earlier test
- Because tests call `informerFactory.Start(s.stopCh)` and never close `stopCh`, the listener goroutines outlive the test that created them and keep dispatching pod events into `onAddPod` for the rest of the binary's life

## Changes

Added `t.Cleanup(func() { close(s.stopCh) })` to every test that starts an informer without a corresponding stop:

- `Test_getPodUsage`
- `Test_Filter`
- `TestRegisterSkipsCleanupForUntrackedVendor`
- `Test_ResourceQuota`
- `Test_Bind_DelPodOnGetPodFailure`
- `Test_Bind_DelPodOnGetNodeFailure`

Also removed redundant `s.stopCh = make(chan struct{})` reassignments in `Test_RegisterFromNodeAnnotations` and `Test_RegisterFromNodeAnnotations_NIL`, `NewScheduler()` already initialises the channel, so recreating it orphans the original.

## How to verify

```bash
for i in $(seq 1 12); do go test ./pkg/scheduler/ -count 1 -race; done
```

Should pass consistently. Before this fix it fails on attempt 3-4 out of 12 on average (reproduced with `go test ./pkg/scheduler/ --race` on master).

## Summary by CodeRabbit

* **Tests**
  * Improved automatic cleanup in scheduler tests.
  * Enhanced test reliability by consistently shutting down informer resources and scheduler stop channels.
  * Simplified cleanup handling for bind lock retry scenarios.

### DIFF
--- a/pkg/scheduler/scheduler_test.go
+++ b/pkg/scheduler/scheduler_test.go
@@ -365,6 +365,7 @@ test case matrix.
 
 func Test_getPodUsage(t *testing.T) {
 	s := NewScheduler()
+	t.Cleanup(s.Stop)
 	client.KubeClient = fake.NewClientset()
 	s.kubeClient = client.KubeClient
 	informerFactory := informers.NewSharedInformerFactoryWithOptions(client.KubeClient, time.Hour*1)
@@ -476,6 +477,7 @@ test case matrix.
 */
 func Test_Filter(t *testing.T) {
 	s := NewScheduler()
+	t.Cleanup(s.Stop)
 	client.KubeClient = fake.NewClientset()
 	s.kubeClient = client.KubeClient
 	informerFactory := informers.NewSharedInformerFactoryWithOptions(client.KubeClient, time.Hour*1)
@@ -1042,7 +1044,6 @@ func Test_RegisterFromNodeAnnotations(t *testing.T) {
 			name: "test node handshake annotations layout",
 			Scheduler: func() *Scheduler {
 				s := NewScheduler()
-				s.stopCh = make(chan struct{})
 				s.nodeNotify = make(chan struct{})
 				client.KubeClient = fake.NewClientset()
 				s.kubeClient = client.KubeClient
@@ -1136,7 +1137,6 @@ func Test_RegisterFromNodeAnnotations_NIL(t *testing.T) {
 	// Define a helper function to create a scheduler with a node that has nil annotations.
 	createSchedulerWithNilAnnotations := func() *Scheduler {
 		s := NewScheduler()
-		s.stopCh = make(chan struct{})
 		s.nodeNotify = make(chan struct{})
 
 		client.KubeClient = fake.NewClientset()
@@ -1309,7 +1309,6 @@ func TestRegisterSkipsCleanupForUntrackedVendor(t *testing.T) {
 	}
 
 	s := NewScheduler()
-	s.stopCh = make(chan struct{})
 	client.KubeClient = fake.NewClientset()
 	s.kubeClient = client.KubeClient
 
@@ -1346,6 +1345,7 @@ func TestRegisterSkipsCleanupForUntrackedVendor(t *testing.T) {
 
 	informerFactory.Start(s.stopCh)
 	informerFactory.WaitForCacheSync(s.stopCh)
+	t.Cleanup(s.Stop)
 
 	s.addNode("node-1", &device.NodeInfo{
 		ID:   "node-1",
@@ -1379,6 +1379,7 @@ func TestRegisterSkipsCleanupForUntrackedVendor(t *testing.T) {
 
 func Test_ResourceQuota(t *testing.T) {
 	s := NewScheduler()
+	t.Cleanup(s.Stop)
 	client.KubeClient = fake.NewClientset()
 	s.kubeClient = client.KubeClient
 	informerFactory := informers.NewSharedInformerFactoryWithOptions(client.KubeClient, time.Hour*1)
@@ -2094,6 +2095,7 @@ func Test_Bind_DelPodOnGetPodFailure(t *testing.T) {
 	}
 
 	s := NewScheduler()
+	t.Cleanup(s.Stop)
 	scheme := runtime.NewScheme()
 	_ = corev1.AddToScheme(scheme)
 	s.eventRecorder = record.NewBroadcaster().NewRecorder(scheme, corev1.EventSource{})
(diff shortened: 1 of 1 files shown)


v0idheaven:
Updated per review feedback from @coderabbitai.

Changed from 	.Cleanup(func() { close(s.stopCh) }) to 	.Cleanup(informerFactory.Shutdown).

**Why this is better:**

- close(s.stopCh) only *signals* goroutines to stop but returns immediately
- informerFactory.Shutdown() *waits* until all event-handler goroutines have actually exited

This ensures the informer goroutines are fully drained before the test function returns, preventing any possibility of them outliving the test and racing against subsequent tests.

Eshiv-Pandey:
@v0idheaven wlcm to the community. pls sign off your commits dco check is failing.

Eshiv-Pandey (commented):
this still regresses. `Shutdown()` only waits for goroutines it doesnt stop them. The stop signal is close(s.stopCh), which these six tests never call, so `Shutdown()` blocks until the test times out. Like` setupBindLockRetryTest`, use:

> t.Cleanup(func() {
>     close(s.stopCh)
>     informerFactory.Shutdown()
> })
> 
a 12x-race loop should hang on the first run with the current changes.

v0idheaven:
Thanks for the review @Eshiv-Pandey! Updated the fix:

- Replaced all informerFactory.Shutdown() usages with 	.Cleanup(s.Stop) s.Stop() calls close(s.stopCh) which is the actual stop signal the informer goroutines watch, so they exit cleanly without any deadlock risk
- Also squashed all commits into one and added the DCO sign-off (Signed-off-by)

Let me know if anything else needs changing!

archlitchi (approved):
/lgtm

archlitchi:
please sync with master to pass the CI

pkg/scheduler/scheduler_test.go
```diff
@@ -476,6 +477,7 @@ test case matrix.
 */
 func Test_Filter(t *testing.T) {
 	s := NewScheduler()
+	t.Cleanup(s.Stop)
```
Shouren, line 480: @v0idheaven I test this PR with the following commands
```bash
go test ./pkg/scheduler/ \
      -run '^(Test_Filter|Test_onAddPod_BadDeviceAnnotation)$' \
      -count=1 -race
```
And here comes an error
```bash
E0809 23:37:49.760989  973437 scheduler.go:162] "failed to decode pod devices" err="pod annotation format error, missing fields, do not use nodeName in task spec" pod="default/bad-anno-pod"
--- FAIL: Test_onAddPod_BadDeviceAnnotation (0.00s)
    testing.go:1712: race detected during execution of test
```
Is this an expected error?

Shouren (changes requested):

mesutoezdil:

PR #2406, opened 2026-08-06
fix(device-plugin): guard Allocate() against empty DevicesIds to prevent process-crashing panic

### LINKED ISSUE #2405: device-plugin: Allocate() panics with index-out-of-range when a ContainerRequest has empty DevicesIds
**What happened**:

In the NVIDIA device plugin's `Allocate()` gRPC handler, the code indexes into `req.DevicesIds[0]` without first checking that the slice is non-empty:

```go
func (plugin *NvidiaDevicePlugin) Allocate(ctx context.Context, reqs *kubeletdevicepluginv1beta1.AllocateRequest) (*kubeletdevicepluginv1beta1.AllocateResponse, error) {
    ...
    for idx, req := range reqs.ContainerRequests {
        // If the devices being allocated are replicas, then (conditionally)
        // error out if more than one resource is being allocated.

        if strings.Contains(req.DevicesIds[0], "MIG") {
```

If any `ContainerAllocateRequest` in the batch arrives with an empty `DevicesIds` slice, this line panics with `index out of range [0] with length 0`.

**Blast radius**: I checked and there is no panic-recovery interceptor registered on this gRPC server (no `recover()`, `UnaryInterceptor`, or `StreamInterceptor` anywhere under `pkg/device-plugin/`). An unrecovered panic in a gRPC handler crashes the entire process — not just the one RPC. Since a single `NvidiaDevicePlugin` process serves all GPU allocation for an entire node, this would kill the device-plugin pod and disrupt every pending GPU pod on that node, not just the one triggering the bad request.

`Allocate()` should validate that `req.DevicesIds` is non-empty before indexing into it, and return a clean gRPC error (consistent with how other device plugins in this repo already handle this case — e.g. the XPU device plugin returns `rpc error: code = Unknown desc = empty id from pod annotation` rather than crashing, as seen in #1548) instead of panicking and taking down the whole plugin process.

This was found via code inspection, not a live cluster repro. To reproduce:

1. Deploy the HAMi NVIDIA device plugin normally.
2. Craft (or otherwise cause kubelet to send) an `AllocateRequest` where `ContainerRequests` contains an entry with `DevicesIds: []` (empty slice) — e.g. by dialing the plugin's registered Unix socket under `kubeletdevicepluginv1beta1.DevicePluginPath` directly and sending a malformed request, or via any upstream kubelet behavior change that could produce this.
3. Observe the plugin process panics and crashes at `server.go`, line with `req.DevicesIds[0]`.

Kubelet's device manager is expected to always populate `DevicesIds` with at least one entry per container request under normal operation, so this isn't hit in the common path today. This is a defensive-coding gap at an external RPC boundary rather than a currently-observed crash — the same class of gap flagged and fixed in the recent `ListAndWatch()` `Send()` error-handling work (#2353), just on the `Allocate()` side.

- Relevant code: `pkg/device-plugin/nvidiadevice/nvinternal/plugin/server.go`, `Allocate()`, the line `if strings.Contains(req.DevicesIds[0], "MIG") {`
- Confirmed no other check earlier in the loop guards `len(req.DevicesIds)`.
- Confirmed no panic-recovery middleware is registered on the gRPC server in this package.
- Proposed fix:

```go
for idx, req := range reqs.ContainerRequests {
    if len(req.DevicesIds) == 0 {
        PodAllocationFailed(nodename, current, NodeLockNvidia)
        return nil, fmt.Errorf("invalid allocation request: container request %d has no DevicesIds", idx)
    }
    if strings.Contains(req.DevicesIds[0], "MIG") {
        ...
```

**Environment**:

- HAMi version: master branch (code inspection)
- Others: Found via static code review, not a device-specific runtime issue.

AyushSrivastava1818:
### Summary
In `pkg/device-plugin/nvidiadevice/nvinternal/plugin/server.go`, `Allocate()` iterates over `reqs.ContainerRequests` and accesses `req.DevicesIds[0]` without checking if `DevicesIds` is empty. If a request with an empty `DevicesIds` slice is received, Go panics with `runtime error: index out of range [0]`. Because the gRPC server has no panic-recovery interceptor, this crashes the entire `nvidia-device-plugin` process — affecting every pod on the node waiting on GPU allocation, not just the one triggering the bad request.

This PR adds a guard at the top of the `ContainerRequests` loop:
- Calls `PodAllocationFailed(nodename, current, NodeLockNvidia)` to clean up node locks and pod status annotations.
- Returns `nil, fmt.Errorf("invalid allocation request: container request %d has no DevicesIds", idx)` for a clean gRPC error instead of a crash.

Fixes #2405

### Changes
- **`pkg/device-plugin/nvidiadevice/nvinternal/plugin/server.go`**: add `len(req.DevicesIds) == 0` check before `req.DevicesIds[0]` is accessed.
- **`pkg/device-plugin/nvidiadevice/nvinternal/plugin/server_test.go`**: add table-driven unit test `TestAllocate_EmptyDevicesIdsAndRegression` covering:
  1. Empty `DevicesIds` → non-nil error, no panic, `PodAllocationFailed` invoked (asserted directly, not just the error return).
  2. Existing non-MIG allocation path still succeeds unchanged.
  3. Existing MIG allocation path still succeeds unchanged.

### Test Matrix & Verification Status
- **Local (Windows)**: `go test` and `make verify` cannot compile `pkg/device-plugin/nvidiadevice/nvinternal/plugin` locally due to pre-existing CGo / Linux-only dependencies (`go-nvml`, `nvidia-container-toolkit`, `opencontainers/cgroups` using `syscall.Mmap` / `unix.Mkdev`). This occurs identically on an unmodified `master` checkout.
- **CI**: full compilation, `go test`, and `make verify` will run on Linux runners.

### AI Assistance Disclosure
I used AI assistance (Claude) to identify this panic risk, implement the guard, and draft the regression test. I reviewed the diff, ran the test suite where locally possible, and take full responsibility for the final change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented allocation requests with no device IDs from causing failures.
  * Added clear validation errors identifying the affected request.
  * Ensured allocation locks are released when validation fails.
  * Preserved successful allocation behavior for standard and MIG devices.

* **Tests**
  * Added regression coverage for empty device requests, failure handling, and successful standard and MIG allocations.

### DIFF
--- a/pkg/device-plugin/nvidiadevice/nvinternal/plugin/server.go
+++ b/pkg/device-plugin/nvidiadevice/nvinternal/plugin/server.go
@@ -602,6 +602,11 @@ func (plugin *NvidiaDevicePlugin) Allocate(ctx context.Context, reqs *kubeletdev
 	klog.Infof("Allocate pod name is %s/%s, annotation is %+v", current.Namespace, current.Name, current.Annotations)
 
 	for idx, req := range reqs.ContainerRequests {
+		if len(req.DevicesIds) == 0 {
+			PodAllocationFailed(nodename, current, NodeLockNvidia)
+			return nil, fmt.Errorf("invalid allocation request: container request %d has no DevicesIds", idx)
+		}
+
 		// If the devices being allocated are replicas, then (conditionally)
 		// error out if more than one resource is being allocated.
 
--- a/pkg/device-plugin/nvidiadevice/nvinternal/plugin/server_test.go
+++ b/pkg/device-plugin/nvidiadevice/nvinternal/plugin/server_test.go
@@ -1456,6 +1456,8 @@ func TestAllocate_EmptyDevicesIdsAndRegression(t *testing.T) {
 			require.Equal(t, tc.expectSuccess, successCalled, "PodAllocationTrySuccess invoked mismatch")
 		})
 	}
+}
+
 type mockListAndWatchServer struct {
 	grpc.ServerStream
 	sendErrs []error
--- a/pkg/device-plugin/nvidiadevice/nvinternal/plugin/server_test.go
+++ b/pkg/device-plugin/nvidiadevice/nvinternal/plugin/server_test.go
@@ -1308,3 +1308,152 @@ func TestMigCurrentConfigsNeverNil(t *testing.T) {
 		require.False(t, cfg.MigEnabled)
 	}
 }
+
+func TestAllocate_EmptyDevicesIdsAndRegression(t *testing.T) {
+	deviceListStrategies, _ := v1.NewDeviceListStrategies([]string{"envvar"})
+	deviceIDStrategy := v1.DeviceIDStrategyUUID
+	memScale := 1.0
+	logLevel := nvidia.Error
+	expectedNodeName := "node-a"
+	t.Setenv(util.NodeNameEnvName, expectedNodeName)
+
+	mockRM := &rm.ResourceManagerMock{
+		DevicesFunc: func() rm.Devices {
+			return rm.Devices{
+				"MIG-12345": &rm.Device{},
+			}
+		},
+		ResourceFunc: func() v1.ResourceName {
+			return v1.ResourceName("nvidia.com/mig-1g.5gb")
+		},
+	}
+
+	plugin := &NvidiaDevicePlugin{
+		rm: mockRM,
+		config: &nvidia.DeviceConfig{
+			Config: &v1.Config{
+				Flags: v1.Flags{
+					CommandLineFlags: v1.CommandLineFlags{
+						Plugin: &v1.PluginCommandLineFlags{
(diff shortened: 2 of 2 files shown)


archlitchi:
please resolve this conflicts

AyushSrivastava1818:
@archlitchi resolved the conflicts take a look whenever you are free!

FouoF:
Thanks for considering defensive handling here. No supported kubelet path or production reproduction reaching an Allocate request with an empty DevicesIds list has been shown; the supplied reproduction requires sending a malformed gRPC request directly. The current PR adds a large test harness for a very small guard and is already conflicting with master. The maintenance cost is not justified by the available evidence, so we are closing this PR. A new change would need a real kubelet/runtime reproduction or a confirmed protocol requirement.

PR #2408, opened 2026-08-06
fix(test): set LimitSet on quota fixtures added in #2347

veyron-kairo:
**What type of PR is this?**
/kind bug

master is red on the scheduler unit tests right now. #2313 changed FitQuota to gate on LimitSet instead of `Limit != 0`, so an explicit ResourceQuota of "0" is honored as a hard block. The quota tests added in #2347 build `device.Quota` values directly with only `Limit` set, so once both landed on master those limits read as "not configured" and the over-quota cases got admitted instead of denied. AddQuota sets LimitSet in the real path, so this sets it in the fixtures to match.

Both PRs passed CI on their own base and only conflict once they're on master together, which is why neither run caught it.

Tests this brings back to green:
- TestFitResourceQuotaNonNvidia (mlu and dcu subcases)
- TestFitResourceQuotaCountsEveryDevice
- TestFitResourceQuotaAscendMemoryFactor

none filed, this is a master CI regression

test-only, no production code touched. `go test ./pkg/scheduler/...` passes locally. #2313 was mine, so this is cleaning up after it.

**AI assistance disclosure**:
I used AI assistance (Claude Code) to bisect which merge combination broke master and to confirm the fix. I read and ran the tests myself.

NONE

## Summary by CodeRabbit

* **Tests**
  * Updated quota test fixtures to explicitly mark Cambricon, Hygon, multi-MLU, and Ascend limits as active.
  * Improved test coverage reliability for scheduler webhook quota scenarios.

### DIFF
--- a/pkg/scheduler/webhook_test.go
+++ b/pkg/scheduler/webhook_test.go
@@ -486,13 +486,13 @@ func TestFitResourceQuotaNonNvidia(t *testing.T) {
 	// One MLU vmemory unit is 256 MiB, so a limit of 100 units leaves room for
 	// 25600 MiB. Comparing the request against the raw 100 would deny every pod.
 	qm.Quotas["mlu-mem"] = &device.DeviceQuota{
-		"cambricon.com/mlu.smlu.vmemory": &device.Quota{Used: 0, Limit: 100},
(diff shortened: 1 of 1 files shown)


archlitchi:
Already fixed

Issue #2379, opened 2026-08-05
test: add a shared conformance test suite for the `device.Devices` interface


im-Toqeer-506:
## Summary

Every hardware backend in HAMi implements the same `device.Devices` interface
(`pkg/device/devices.go:36`), but each one is tested in isolation. This proposes a
single **table-driven conformance suite** — `pkg/device/conformance_test.go` — that
runs one shared set of contract assertions against *all* backends at once, so a
backend that violates the interface contract fails CI immediately instead of
shipping a bug we then fix vendor-by-vendor.

## Motivation

HAMi has **14 device backends** (`nvidia`, `cambricon`, `hygon`, `iluvatar`,
`metax`, `mthreads`, `ascend`, `amd`, `enflame`, `kunlun`, `awsneuron`, `vastai`,
`biren`, `common`), all satisfying the same interface — yet test depth is very
uneven:

| Backend | `func Test…` count |
|---|---|
| `nvidia` | 56 |
| `ascend` | 20 |
| `enflame` | 26 (device + gcu) |
| `metax` | 27 (across 4 files) |
| `amd` | 6 |
| `kunlun/device` | 6 |

Because there is no shared contract test, **the same class of bug keeps being
rediscovered one backend at a time:**

- **`int32` overflow in `GenerateResourceRequests`** silently drops the memory
  request. Fixed for some backends in #2336, but still **open** for iluvatar and
  mthreads (#2284) and cambricon (#2278). The unchecked conversion is still live in
  `cambricon/device.go:280`, `mthreads/device.go:254`, `iluvatar/device.go:243`,
  while `enflame` already had to grow its own private `clampToInt32` helper
  (`enflame/device.go:663`).
- **nil-map / nil-pointer panics** on the admission and `Fit` paths — see merged
  fixes #2254 (`guard nil maps in mthreads MutateAdmission`) and #2294
  (`nil CustomInfo write can panic in Fit/AddResourceUsage`).

A conformance suite turns these from *"find it again in the next backend"* into
*"the backend fails CI the moment it breaks the contract,"* and gives authors of
future backends (Cambricon 580, Teco DPU, … per the v2.10 roadmap #1889) an
executable definition of correct.

## Proposed solution

1. Add a small **test-only registry** of constructed backends, reusing the fake
   configs each package already uses in its `*_test.go`
   (e.g. `nvidia.InitNvidiaDevice(NvidiaConfig{…})`).
[...]

PR #2412, opened 2026-08-06
Feature

im-Toqeer-506:
**What type of PR is this?**

/kind feature
/kind failing-test

All 14 HAMi device backends implement the same `device.Devices` interface
(`pkg/device/devices.go`), but each is tested in isolation — so the same class of
contract violation keeps being rediscovered and fixed one backend at a time
(int32 overflow in `GenerateResourceRequests`: #2278/#2284/#2336; nil-map / nil-pointer
panics on the admission and Fit paths: #2254/#2294).

This PR lands **phase 1** of a shared, table-driven conformance suite that runs one
set of contract assertions against every constructible backend, so a regression in
any of them fails CI immediately instead of shipping.

Included:
- A test-only backend **registry** (`conformanceCases()`) that constructs 14 backend
  variants from plain in-memory configs mirroring the production resource names wired
  in `pkg/scheduler/config.InitDevicesWithConfig`.
- Shared invariants:
  - `GetResourceNames()` returns at least one non-empty name (else the backend is
    unreachable by the scheduler).
  - A container requesting none of a backend's resources yields `Nums == 0`.
  - `Fit(nil, …)` and `Fit([]*DeviceUsage{}, …)` return `false` without panicking.
  - `MutateAdmission` does not panic on a pod that requests none of the resources.
  - **No negative `Nums`/`Memreq`/`Coresreq`** from `GenerateResourceRequests` — the
    int32-overflow guard.
- The overflow invariant surfaced a **previously-untracked overflow in `metax-sgpu`**,
  which this PR fixes inline (reject out-of-range memory before the unchecked
  `int32(mem)` cast) — exactly the value the suite is meant to deliver.
- Backends with dedicated open issues (`cambricon` #2278, `mthreads` #2284) go on an
  explicit, commented skip list linked to their tracking issues, so the gap is visible,
  not silent.

Kept additive and phased per the issue: `ascend` and `iluvatar` (slice / enable-flag
constructors) and further invariants (scoring monotonicity, `PatchAnnotations`
round-trip, lock idempotency) are documented as follow-up.

Fixes # https://github.com/Project-HAMi/HAMi/issues/2379

- The overflow invariant has teeth: with the skip list disabled, `cambricon` and
  `mthreads` fail and `metax-sgpu` passes (confirming the metax fix is exercised).
- Overflow "case 2" is gated on `MemoryFactor > 0`, so it targets the
  scaling-multiplication overflow that the bug reports describe; backends that don't
  scale still get the in-range non-negativity check (case 1).
- The suite lives in the external `device_test` package on purpose — the backend
  sub-packages import `pkg/device`, so an internal test importing them back would
  create an import cycle.
- Validation: `go test -race ./pkg/device/...` and `./pkg/scheduler/...` all pass;
  `gofmt`, `goimports` (local-prefix), and `go vet` are clean.

```release-note
NONE


## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented invalid device resource requests when memory values exceed supported limits.
  * Invalid requests now fail safely instead of producing negative or corrupted resource values.

* **Tests**
  * Added comprehensive conformance coverage for registered device backends, including empty devices, unrelated containers, admission safety, resource names, and overflow handling.

### DIFF
--- a/pkg/device/conformance_test.go
+++ b/pkg/device/conformance_test.go
@@ -25,15 +25,10 @@ limitations under the License.
 // It lives in the external device_test package on purpose: the backend
 // sub-packages import github.com/Project-HAMi/HAMi/pkg/device, so an internal
 // (package device) test that imported them back would create an import cycle.
-//
-// The invariants asserted here are intentionally the ones that should hold for
-// every backend today. The int32-overflow invariant for GenerateResourceRequests
-// is deliberately left out of this first pass because the underlying fixes are
-// still open for several backends (#2278, #2284); it is tracked as follow-up
-// work so this suite stays green while those land.
 package device_test
 
 import (
+	"math"
 	"testing"
 
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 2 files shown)


archlitchi:
You need to edit the title, to reflect the feature you intend to implement

PR #2413, opened 2026-08-06
cleanup(metax): decouple MetaxSDevices.ScoreNode from the scheduler policy string

### LINKED ISSUE #2404: MetaxSDevices.ScoreNode couples device scoring to the scheduler policy string

Refactor `MetaxSDevices.ScoreNode` so that device-level scoring no longer depends on the cluster scheduling policy string, resolving the existing `TODO` in the code.

### Current implementation

In `pkg/device/metax/sdevice.go`, `ScoreNode` currently changes the score based on the scheduler policy:

```go
func (sdev *MetaxSDevices) ScoreNode(node *corev1.Node, podDevices device.PodSingleDevice, previous []*device.DeviceUsage, policy string) float32 {
    // TODO: score should not depend on policy

    // we have to give it a smaller value because of Spread policy
    weight := 10000
    if policy == string(util.NodeSchedulerPolicySpread) {
        weight = -10000
    }

    ...
}
```

The device layer is responsible not only for calculating the device score, but also for determining how that score should behave under different scheduling policies.

### Proposed approach

Refactor the scoring flow so that:

- `MetaxSDevices.ScoreNode` returns a normalized, policy-independent score.
- The shared scheduling policy layer (`pkg/scheduler/policy/`) is responsible for applying the Spread/Binpack weighting.
- The policy string is no longer used to determine the score sign inside the MetaX device implementation.
- The existing `TODO` can be removed.

This would make the MetaX backend consistent with the scoring contract used by the other device backends.

---

## Why is this needed?

The current implementation tightly couples device scoring with scheduler policy.

Specifically:

- The MetaX backend changes its score based on the scheduler policy string.
- The `±10000` weight acts as a hidden policy implementation detail inside the device layer.
- Future scheduling policies would require updating MetaX-specific code, making the implementation more difficult to maintain.
- The dependency is already acknowledged in the source code by the existing `// TODO: score should not depend on policy` comment.

Separating these responsibilities would make the architecture easier to understand and maintain:

- Device backends should only calculate device scores.
- Scheduling policies should decide how those scores are interpreted (Spread, Binpack, or future policies).

This also removes the embedded magic-number sign flip (`±10000`) from the MetaX implementation.

---

## Expected outcome

Acceptance would include:

- `ScoreNode` no longer changes its score based on the scheduler policy.
- Scheduler policy logic is handled by the shared policy layer.
- Spread and Binpack produce the same node ranking as today.
- Existing behavior is preserved through unit tests.
- The existing `TODO` is removed.

---

asadjan4611:
**What type of PR is this?**

/kind cleanup

`MetaxSDevices.ScoreNode` previously changed its result based on the scheduler
policy string and carried a `TODO: score should not depend on policy`. It also
hard-coded a `±10000` magic number in the device layer whose only purpose was to
flip the score's sign so the best node would win under both the Binpack and
Spread sort directions. This coupled a device backend to scheduler-policy
concerns that belong in the shared policy layer.

This PR decouples them:

- `ScoreNode` now returns a **policy-independent** score following a
  "higher score is a better node" convention, and the `TODO` is removed.
- A new optional marker interface `device.PolicyNeutralScorer` lets a backend
  declare that its `ScoreNode` result is policy-independent.
- The shared policy layer (`policy.OverrideScore`) applies the weight (`10000`)
  and the Spread-policy sign inversion (`-10000`) in one place, **only** for
  backends that opt in via the marker. Policy handling now lives in a single
  location, so future policies no longer require backend-specific changes.

The final weighted scores are numerically identical to the previous behavior
under both policies, so node ranking is unchanged. Backends that genuinely need
the policy string (e.g. `Metax-GPU`, which reads different annotations per
policy) do **not** implement the marker and are completely unaffected.

Fixes #2404

- **No behavior change / ranking preserved.** Binpack weight stays `+10000` and
  Spread stays `-10000`, applied in `OverrideScore` instead of inside
  `ScoreNode`. For a raw device score `s`, the node's contribution is identical
  to before: `+10000·s` (binpack) and `-10000·s` (spread).
- **Opt-in, minimal blast radius.** Only `MetaxSDevices` implements
  `PolicyNeutralScorer`. The other 12 device backends and the sibling
  `Metax-GPU` backend are untouched (verified by grep + build).
- **Tests added:**
  - `TestScoreNodePolicyIndependence` — asserts `ScoreNode` returns the same
    value for `binpack`, `spread`, and `""` across the online, topology-aware,
    and no-hint paths.
  - `TestMetaxSDevicesImplementsPolicyNeutralScorer` — asserts the marker is wired.
  - Two new `TestOverrideScore` cases — binpack (`+600000`) and spread
    (`-600000`) for the same input, proving ranking preservation.
- **Local checks:** `go build ./...`, `go vet`, race-tested modified packages,
  `gofmt`/`goimports`, license headers, and `golangci-lint v2.12.2` (0 issues)
  all pass.

```release-note
NONE
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved device-aware node scoring for MetaX-SGPU workloads.
  * Ensured topology and online status are evaluated consistently across scheduler policies.
  * Preserved expected placement behavior for both Binpack and Spread scheduling through policy-aware score weighting.

* **Tests**
  * Added coverage validating consistent scores across policies.
  * Added tests confirming correct score weighting for Binpack and Spread scheduling.

### DIFF
--- a/pkg/device/devices.go
+++ b/pkg/device/devices.go
@@ -52,18 +52,6 @@ type MigPlacement struct {
 	Size  uint32 `json:"size"`
 }
 
-// PolicyNeutralScorer is an optional interface a device backend may implement
-// to declare that its ScoreNode result is policy-independent and follows a
-// "higher score is a better node" convention. Backends that implement it must
-// not inspect the scheduler policy inside ScoreNode; instead, the shared
-// scheduler policy layer weights the returned score and adapts it to the active
-// scheduling policy (for example, inverting it under the Spread policy). This
-// keeps policy handling in one place so new policies do not require
-// backend-specific changes.
-type PolicyNeutralScorer interface {
-	PolicyNeutralScore()
-}
-
 // Deprecated scheduler-template types are retained for source compatibility
 // with non-NVIDIA device implementations and tests. Dynamic NVIDIA MIG does
 // not populate or consume them.
--- a/pkg/device/devices.go
+++ b/pkg/device/devices.go
@@ -52,6 +52,18 @@ type MigPlacement struct {
 	Size  uint32 `json:"size"`
 }
 
+// PolicyNeutralScorer is an optional interface a device backend may implement
+// to declare that its ScoreNode result is policy-independent and follows a
+// "higher score is a better node" convention. Backends that implement it must
+// not inspect the scheduler policy inside ScoreNode; instead, the shared
+// scheduler policy layer weights the returned score and adapts it to the active
+// scheduling policy (for example, inverting it under the Spread policy). This
+// keeps policy handling in one place so new policies do not require
+// backend-specific changes.
+type PolicyNeutralScorer interface {
+	PolicyNeutralScore()
+}
+
 // Deprecated scheduler-template types are retained for source compatibility
 // with non-NVIDIA device implementations and tests. Dynamic NVIDIA MIG does
 // not populate or consume them.
--- a/pkg/scheduler/policy/node_policy.go
+++ b/pkg/scheduler/policy/node_policy.go
@@ -61,7 +61,23 @@ func (ns *NodeScore) OverrideScore(previous []*device.DeviceUsage, policy string
 			klog.V(4).Infof("Skip scoring for device type %s: not registered", idx)
 			continue
 		}
-		devScore += dev.ScoreNode(ns.Node, val, previous, policy)
+
+		score := dev.ScoreNode(ns.Node, val, previous, policy)
+
+		// Backends implementing PolicyNeutralScorer return policy-independent
+		// "higher is better" scores. For the Spread policy (which selects the
+		// lowest score), we invert the score to preserve the ranking.
+		// We also apply a weight of 10000 to ensure device scores dominate
+		// the base node score.
+		if _, ok := dev.(device.PolicyNeutralScorer); ok {
+			weight := float32(10000)
+			if policy == util.NodeSchedulerPolicySpread.String() {
+				weight = -10000
+			}
+			score = weight * score
+		}
+
+		devScore += score
 	}
 	ns.Score += devScore
 	klog.V(2).Infof("node %s default score is %f, computer override score is %f", ns.NodeID, ns.Score-devScore, ns.Score)
--- a/pkg/scheduler/policy/node_policy.go
+++ b/pkg/scheduler/policy/node_policy.go
@@ -53,6 +53,17 @@ func (l NodeScoreList) Less(i, j int) bool {
 	return l.NodeList[i].Score < l.NodeList[j].Score
 }
 
+// policyNeutralScorer is an optional interface a device backend may implement to
+// declare that its ScoreNode result is policy-independent and follows a "higher
+// score is a better node" convention. Backends that implement it must not
+// inspect the scheduler policy inside ScoreNode; instead OverrideScore weights
+// the returned score and adapts it to the active scheduling policy (for example,
+// inverting it under the Spread policy). This keeps policy handling in one place
+// so new policies do not require backend-specific changes.
+type policyNeutralScorer interface {
+	PolicyNeutralScore()
+}
+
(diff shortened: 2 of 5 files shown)


pkg/scheduler/policy/node_policy_test.go
```diff
@@ -300,6 +300,106 @@ func TestOverrideScore(t *testing.T) {
...
+					"Metax-SGPU": device.PodSingleDevice{
+						device.ContainerDevices{
+							{
+								UUID:      "GPU-3",
+								Usedcores: 100,
+								CustomInfo: map[string]any{
+									"LinkZone": int32(1),
+									"Pod.Annotations": map[string]string{
+										"metax-tech.com/sgpu-topology-aware": "true",
+									},
```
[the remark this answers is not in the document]
asadjan4611: `metax-tech.com/sgpu-topology-aware` is a pre-existing, user-facing pod annotation (defined in `protocol.go`, documented in `examples/metax/sgpu/enable_topology_aware.yaml`) that predates this PR. It follows the same vendor-domain convention as every other backend (amd.com/…, cambricon.com/…, etc.), and its producer is the user's pod spec, so it can't be renamed without a breaking change. This PR (#2404) only decouples ScoreNode from the policy string and reuses the existing constant in test fixtures. A prefix migration would be a separate breaking change with a deprecation path, out of scope here.

archlitchi:
The title of this PR seems unrelated to the content

asadjan4611:
@archlitchi  thanks for noticing i have update the title can you please check it

pkg/device/devices.go
```diff
@@ -49,6 +49,18 @@ type Devices interface {
...
 
+// PolicyNeutralScorer is an optional interface a device backend may implement
+// to declare that its ScoreNode result is policy-independent and follows a
+// "higher score is a better node" convention. Backends that implement it must
+// not inspect the scheduler policy inside ScoreNode; instead, the shared
+// scheduler policy layer weights the returned score and adapts it to the active
+// scheduling policy (for example, inverting it under the Spread policy). This
+// keeps policy handling in one place so new policies do not require
+// backend-specific changes.
+type PolicyNeutralScorer interface {
```
Shouren: @asadjan4611 I see little value to advertise a public abstratction that currently has one consumer and one implementer. I suggest that make it private in `node_policy.go` to its consumer and still keep the capability.

Shouren (commented):
@asadjan4611 Please rebase the master to fix the CI Workflow action.

pkg/device/devices.go
asadjan4611: Resolve conversation

asadjan4611:
@Shouren  can you please review it now

pkg/scheduler/policy/node_policy.go
```diff
@@ -53,11 +53,37 @@ func (l NodeScoreList) Less(i, j int) bool {
...
+type policyNeutralScorer interface {
+	PolicyNeutralScore()
+}
+
 func (ns *NodeScore) OverrideScore(previous []*device.DeviceUsage, policy string) {
 	// current user having request resource
 	devScore := float32(0)
 	for idx, val := range ns.Devices {
-		devScore += device.GetDevices()[idx].ScoreNode(ns.Node, val, previous, policy)
+		score := device.GetDevices()[idx].ScoreNode(ns.Node, val, previous, policy)
```
mesutoezdil: #1773 merged today, it added a nil check for unregistered devices in this loop. your change drops that check and conflicts now. pls rebase and keep the nil check too.

asadjan4611: fixed

asadjan4611:
@mesutoezdil can you please review it now.

asadjan4611:
@mesutoezdil sorry for disturbance 
can you please help me about this PR  _https://github.com/Project-HAMi/HAMi/pull/2445_
can i now work  on it 
cc @FouoF

Shouren (approved):
/lgtm

PR #2416, opened 2026-08-06
fix(ascend): guard nil Requests in MutateAdmission

### LINKED ISSUE #2415: [bug]: ascend MutateAdmission panics on nil Requests (assignment to nil map)
### What happened

`Devices.MutateAdmission` in the ascend backend writes the trimmed memory
quantity into `ctr.Resources.Requests` unconditionally:

https://github.com/Project-HAMi/HAMi/blob/87d9795adc4d73042efe1751351c8d04488270cf/pkg/device/ascend/device.go#L181-L182

A container that declares only `limits` (no `requests` block) reaches this
line with `ctr.Resources.Requests == nil`, and the write panics with
`assignment to entry in nil map` inside the admission webhook.

The early return at the top of the function guarantees `Limits` is non-nil
(the NPU count was found in it), but nothing guarantees `Requests` exists.

### Reproduction

Minimal unit-level repro — this panics on current master (87d9795):

```go
dev := ... // any ascend VNPUConfig, e.g. the fixture used in Test_MutateAdmission
ctr := corev1.Container{
    Resources: corev1.ResourceRequirements{
        Limits: corev1.ResourceList{
            "huawei.com/Ascend910A":        resource.MustParse("1"),
            "huawei.com/Ascend910A-memory": resource.MustParse("8738"),
        },
        // no Requests block
    },
}
dev.MutateAdmission(&ctr, &corev1.Pod{})   // panic: assignment to entry in nil map
```

Pod-level: any pod requesting an Ascend NPU with only `resources.limits` set.

Worth noting: the existing test suite skirts this — in `Test_MutateAdmission`,
the only case that reaches the write ("resourcememoryname not within the
template scope, but smaller than MemoryCapacity") is also the only case whose
fixture explicitly sets a `Requests` block. Every nil-Requests case exits on
an earlier error path.

### Same class, already fixed elsewhere

- #2253 / #2254 fixed the identical nil-map class in the mthreads backend
  (merged Aug 3).
- The kunlun backend already guards this **exact same write** (trimmed memory
  into Requests):
  https://github.com/Project-HAMi/HAMi/blob/87d9795adc4d73042efe1751351c8d04488270cf/pkg/device/kunlun/vdevice.go#L86-L89

A sweep over `pkg/device/` shows ascend is the last backend with an unguarded
`Resources.Requests` write in `MutateAdmission`.

### Environment

- HAMi master @ 87d9795

I have a fix ready (guard mirroring kunlun/vdevice + regression tests) and
will open a PR.

keshav9926:
### What this PR does / why we need it

`MutateAdmission` in the ascend backend writes the trimmed memory quantity
into `ctr.Resources.Requests` unconditionally. When a container declares only
`limits` (no `requests` block), `Requests` is a nil map and the write panics
(`assignment to entry in nil map`) inside the admission webhook.

This initializes `Requests` before the write, mirroring the guard the kunlun
backend already uses for the identical write
(`pkg/device/kunlun/vdevice.go`), and the pattern merged for mthreads in
#2254. After this, no backend under `pkg/device/` writes to
`Resources.Requests` unguarded.

Also adds `Test_MutateAdmission_NilRequests`: two regression cases (memory
limit without a requests block; count-only without a requests block), both of
which panic on master today.

### Which issue(s) this PR fixes

Fixes #2415

### Special notes for your reviewer

The existing `Test_MutateAdmission` never hits this path with nil Requests:
the only case reaching the write is the only fixture that explicitly sets a
`Requests` block. The new test covers exactly the gap.

### AI assistance disclosure

This PR was developed with AI assistance (Claude); the bug was found by an
automated audit of unguarded map writes across `pkg/device/`, and the fix and
tests were human-reviewed before submission.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed pod admission when resource requests are missing but memory limits are specified.
  * Memory requests are now populated correctly without errors or crashes, including when default card allocation applies.
  * Improved handling for pods that define only memory limits.

* **Tests**
  * Added coverage for admission scenarios with absent resource requests.
  * Verified both explicit memory limits and default card allocation behavior.

### DIFF
--- a/pkg/device/ascend/device.go
+++ b/pkg/device/ascend/device.go
@@ -178,6 +178,11 @@ func (dev *Devices) MutateAdmission(ctr *corev1.Container, p *corev1.Pod) (bool,
 			return true, errors.New("vNPU not supported for multiple devices")
 		}
 	}
+	// Requests may be nil when the pod declares only limits; writing to a
+	// nil map panics.
+	if ctr.Resources.Requests == nil {
+		ctr.Resources.Requests = corev1.ResourceList{}
+	}
 	ctr.Resources.Limits[corev1.ResourceName(dev.config.ResourceMemoryName)] = resource.MustParse(fmt.Sprint(trimMem))
 	ctr.Resources.Requests[corev1.ResourceName(dev.config.ResourceMemoryName)] = resource.MustParse(fmt.Sprint(trimMem))
 
--- a/pkg/device/ascend/device_test.go
+++ b/pkg/device/ascend/device_test.go
@@ -707,6 +707,86 @@ func Test_MutateAdmission(t *testing.T) {
 	}
 }
 
+func Test_MutateAdmission_NilRequests(t *testing.T) {
+	// Regression test: a pod that declares only limits (no requests block)
+	// must not panic when MutateAdmission writes the trimmed memory request.
+	tests := []struct {
+		name       string
+		ctr        corev1.Container
+		wantMemory string
+	}{
+		{
+			name: "memory limit set, requests block absent",
+			ctr: corev1.Container{
+				Resources: corev1.ResourceRequirements{
+					Limits: corev1.ResourceList{
+						"huawei.com/Ascend910A":        resource.MustParse("1"),
+						"huawei.com/Ascend910A-memory": resource.MustParse("8738"),
+					},
+				},
+			},
+			wantMemory: "8738",
+		},
+		{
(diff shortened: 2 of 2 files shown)


archlitchi:
please sync with master to pass the CI

keshav9926:
Done — rebased onto current master (now based on c5e99b6). Thanks for taking a look @archlitchi!

Shouren (approved):
/lgtm

PR #2420, opened 2026-08-06
fix(device-plugin): update device cache only after successful annotation patch

### LINKED ISSUE #2418: Title: device-plugin: RegisterInAnnotation() caches the new device digest before the node patch succeeds, permanently skipping retries after a transient failure
**What happened**:

`RegisterInAnnotation()` (`pkg/device-plugin/nvidiadevice/nvinternal/plugin/register.go`,
lines 193–245) updates `plugin.deviceCache` to the new device digest *before*
confirming the node annotation patch actually succeeded:

```go
encodeddevices := device.MarshalNodeDevices(*devices)
if encodeddevices == plugin.deviceCache {
    klog.V(3).Info("Device info unchanged, skipping annotation update")
    return false, nil
}
plugin.deviceCache = encodeddevices   // <-- written unconditionally, before the patch below

...
err = util.PatchNodeAnnotations(node, annos)
if err != nil {
    klog.Errorln("patch node error", err.Error())
}
return true, err
```

If `PatchNodeAnnotations` fails (API server timeout, conflict, rate limit, etc.), the
function returns an error — but `plugin.deviceCache` has already been permanently
updated to the value that was *supposed* to be patched.

On the next call, if the device list hasn't changed again, `RegisterInAnnotation`
should retry the patch. Instead it hits the early-return: `encodeddevices ==
plugin.deviceCache` is now true (even though the patch never actually landed on the
node object), so it logs "Device info unchanged, skipping annotation update" and
returns `false, nil` — silently giving up forever.

`deviceCache` is a single string field on `NvidiaDevicePlugin` (`server.go` line
113), initialized once to `""` at construction (`server.go` line 201), and is
written/read *only* inside `RegisterInAnnotation()` — no other code path resets it
after a failed patch.

1. Cause a single `PatchNodeAnnotations` call to fail transiently (e.g. simulate an
   API server 409/timeout on the node patch, or run a fault-injection proxy in front
   of the API server).
2. Observe `WatchAndRegister()`'s loop
   (`pkg/device-plugin/nvidiadevice/nvinternal/plugin/register.go`) logs the patch
   error once, sleeps `errorSleepInterval` (5s), and retries.
3. On the retry (assuming the device list hasn't changed in those 5s, which is
   virtually guaranteed), observe it now logs "Device info unchanged, skipping
   annotation update" and returns no error — even though the node's actual annotation
   was never updated.
4. The node's `hami.io/node-nvidia-register`-style annotation remains stale
   indefinitely, until the underlying device list itself changes again (which may
   never happen on a static node).

Proposed fix — only update `plugin.deviceCache` after `PatchNodeAnnotations` returns
successfully:

```go
err = util.PatchNodeAnnotations(node, annos)
if err != nil {
    klog.Errorln("patch node error", err.Error())
    return true, err
}
plugin.deviceCache = encodeddevices
return true, nil
```

**Environment**:
- HAMi version: master @ 90b2ede1d0575eeeb0966b383580660969519a11
- Others: found via static code review of
  `pkg/device-plugin/nvidiadevice/nvinternal/plugin/register.go`

AyushSrivastava1818:
## What type of PR is this?

/kind bug

This PR fixes a retry issue in `RegisterInAnnotation()` caused by updating `plugin.deviceCache` before `PatchNodeAnnotations()` completed successfully.

Previously, `plugin.deviceCache` was assigned to the newly encoded device information before attempting to patch the Kubernetes node annotations. If `PatchNodeAnnotations()` failed (for example due to a transient API server timeout, network interruption, or rate limiting), the cache still reflected the new state even though the node annotations were never updated.

On the next iteration of `WatchAndRegister()`, the newly generated device information matched the cached value, causing `RegisterInAnnotation()` to conclude that nothing had changed and skip the annotation update. As a result, retries were prevented until the discovered device list changed again, potentially leaving node GPU annotations stale or missing.

This PR resolves the issue by:

* Updating `plugin.deviceCache` only after `PatchNodeAnnotations()` succeeds.
* Leaving the cache unchanged when patching fails so the existing retry loop can retry the annotation update on the next iteration.
* Returning `(true, err)` on patch failure to indicate that registration was attempted but did not complete successfully.
* Introducing package-level function variables (`getNode`, `patchNodeAnnotations`, and `getAPIDevices`) to enable isolated unit testing without requiring Kubernetes API servers or NVML.
* Adding `TestRegisterInAnnotation_DeviceCacheOrder` to verify that the cache is only updated after a successful annotation patch and that retries continue after failures.

Fixes #2418

* No behavior changes on successful registration.
* Ensures transient failures while patching node annotations do not prevent future retry attempts.
* Verified locally with the added unit tests covering both successful and failed annotation updates.

## AI assistance disclosure

I consulted ChatGPT during the initial investigation and discussion of the retry behavior. The final implementation, unit tests, validation, commit message, and PR description were completed, reviewed, and verified by me. I take full responsibility for the final contribution.

```text
NONE
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved device registration reliability by updating cached device information only after annotation changes succeed.
  * Prevented failed annotation updates from leaving stale or inconsistent device state.
  * Avoided reporting changes when registration is repeated without device differences.

* **Tests**
  * Added coverage for device-cache behavior across failed, successful, and repeated registration attempts.

### DIFF
--- a/pkg/device-plugin/nvidiadevice/nvinternal/plugin/register.go
+++ b/pkg/device-plugin/nvidiadevice/nvinternal/plugin/register.go
@@ -49,6 +49,12 @@ import (
 	"github.com/Project-HAMi/HAMi/pkg/util"
 )
 
+var (
+	getNode              = util.GetNode
+	patchNodeAnnotations = util.PatchNodeAnnotations
+	getAPIDevices        = (*NvidiaDevicePlugin).getAPIDevices
+)
+
 // int8Slice wraps an []int8 with more functions.
 type int8Slice []int8
 
@@ -191,14 +197,14 @@ func (plugin *NvidiaDevicePlugin) getAPIDevices() *[]*device.DeviceInfo {
 // RegisterInAnnotation scans devices and patches node annotations.
 // Returns (changed, error) where changed indicates whether the annotation was actually updated.
 func (plugin *NvidiaDevicePlugin) RegisterInAnnotation() (bool, error) {
-	devices := plugin.getAPIDevices()
+	devices := getAPIDevices(plugin)
 
 	// Log compact summary at V(3); full details at V(5)
 	klog.V(3).Infof("Discovered %d device(s) for registration", len(*devices))
 	klog.V(5).InfoS("Device details", "devices", devices)
 
 	annos := make(map[string]string)
-	node, err := util.GetNode(util.NodeName)
+	node, err := getNode(util.NodeName)
 	if err != nil {
 		klog.Errorln("get node error", err.Error())
 		return false, err
@@ -208,7 +214,6 @@ func (plugin *NvidiaDevicePlugin) RegisterInAnnotation() (bool, error) {
 		klog.V(3).Info("Device info unchanged, skipping annotation update")
 		return false, nil
 	}
-	plugin.deviceCache = encodeddevices
 
 	var data []byte
 	if os.Getenv("ENABLE_TOPOLOGY_SCORE") == "true" {
@@ -236,12 +241,13 @@ func (plugin *NvidiaDevicePlugin) RegisterInAnnotation() (bool, error) {
 	}
 	klog.Infof("Updating node annotations with %d device(s)", len(*devices))
 	klog.V(3).Infof("Annotation content: %v", annos)
-	err = util.PatchNodeAnnotations(node, annos)
-
+	err = patchNodeAnnotations(node, annos)
 	if err != nil {
 		klog.Errorln("patch node error", err.Error())
+		return true, err
 	}
-	return true, err
+	plugin.deviceCache = encodeddevices
+	return true, nil
 }
 
 func (plugin *NvidiaDevicePlugin) WatchAndRegister(disableNVML <-chan bool, ackDisableWatchAndRegister chan<- bool) {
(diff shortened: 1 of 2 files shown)


pkg/device-plugin/nvidiadevice/nvinternal/plugin/register.go
```diff
@@ -49,6 +49,12 @@ import (
 	"github.com/Project-HAMi/HAMi/pkg/util"
 )
 
+var (
```
FouoF, line 52: why do we need this wrapper?

AyushSrivastava1818: Testability seam-RegisterInAnnotation() previously called util.GetNode/util.PatchNodeAnnotations/plugin.getAPIDevices directly, which meant any test had to hit a real API server and real NVML. Wrapping them in package vars lets TestRegisterInAnnotation_DeviceCacheOrder stub each one: fake devices from getAPIDevices, an empty node from getNode, and-the part that actually matters for this fix- patchNodeAnnotations returns an injected error on the first call and nil on the second, so the test can assert deviceCache stays untouched on failure and only updates after a real success. That's the exact bug from #2418 (cache written before the patch confirmed), and there's no way to force that failure branch deterministically without this seam. Same pattern as the existing getPendingPod wrapper in server.go.

mesutoezdil:

PR #2430, opened 2026-08-07
fix(device-plugin): validate slice bounds in GetMigUUIDFromSmiOutput

### LINKED ISSUE #2359: bug(device-plugin): GetMigUUIDFromSmiOutput assumes fixed nvidia-smi output format and may panic on malformed input
## What happened

`GetMigUUIDFromSmiOutput()` (`pkg/device-plugin/nvidiadevice/nvinternal/plugin/util.go`) parses the output of `nvidia-smi` by indexing directly into the slices returned by `strings.Split()` without validating their length.

For example:

```go
num := strings.Split(val, "Device")[1]
...
outputStr := strings.Split(val, ":")[2]
```

If the expected delimiter is missing or the parsed line does not contain the expected number of segments, these direct index operations will panic with an `index out of range` error.

## Why it matters

The parser currently assumes every matching line always follows the expected `nvidia-smi` output format.

Although the normal output follows this format, defensive parsing should not rely on unchecked slice indexing. If the parser ever encounters malformed, truncated, or otherwise unexpected input, the device-plugin will panic instead of handling the parse failure gracefully.

## How to reproduce

Provide `GetMigUUIDFromSmiOutput()` with input containing `"MIG"` but missing one of the expected delimiters.

For example, a line that does not contain `"Device"` or does not contain enough `":"`-separated fields will cause:

```text
panic: runtime error: index out of range
```

## Expected behavior

Unexpected or malformed input should be handled gracefully by validating the parsed slices before indexing into them.

## Actual behavior

The function indexes directly into the slices returned by `strings.Split()`, which can panic if the expected format is not present.

## Suggested fix

Validate the result of `strings.Split()` before indexing.

For example:

```go
parts := strings.Split(val, "Device")
if len(parts) < 2 {
    klog.Warningf("unexpected MIG output format: %q", val)
    continue
}
```

and similarly for the `":"` split.

I'll open a PR with the proposed fix shortly.

Nitish08-08:
Closes #2359

GetMigUUIDFromSmiOutput indexes directly into strings.Split() results without checking length. Malformed nvidia-smi output missing the 'Device' delimiter or sufficient colon-separated fields causes an index out of range panic.

Add length validation before each slice access and replace klog.Fatal with Warningf+continue so a single malformed line does not crash the device-plugin.

Also add unit tests covering valid output, empty input, missing delimiters, non-numeric indices, and uuid mismatches.

**What type of PR is this?**

Fixes #

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of malformed NVIDIA MIG device output.
  * Continues scanning when individual device entries contain invalid or incomplete data.
  * Prevents parsing errors from causing incorrect failures when retrieving MIG UUIDs.

* **Tests**
  * Added coverage for valid extraction, instance selection, mismatched GPU UUIDs, malformed output, invalid device indices, and missing instances.

### DIFF
--- a/pkg/device-plugin/nvidiadevice/nvinternal/plugin/util.go
+++ b/pkg/device-plugin/nvidiadevice/nvinternal/plugin/util.go
@@ -161,15 +161,26 @@ func GetMigUUIDFromSmiOutput(output string, uuid string, idx int) string {
 			continue
 		}
 		klog.Infoln("inspecting", val)
-		num := strings.Split(val, "Device")[1]
+		deviceParts := strings.Split(val, "Device")
+		if len(deviceParts) < 2 {
+			klog.Warningf("unexpected MIG output format, missing Device delimiter: %q", val)
+			continue
+		}
+		num := deviceParts[1]
 		num = strings.Split(num, ":")[0]
 		num = strings.TrimSpace(num)
 		index, err := strconv.Atoi(num)
 		if err != nil {
-			klog.Fatal("atoi failed num=", num)
+			klog.Warningf("failed to parse MIG device index %q: %v", num, err)
+			continue
 		}
 		if index == idx {
-			outputStr := strings.Split(val, ":")[2]
+			colonParts := strings.Split(val, ":")
+			if len(colonParts) < 3 {
+				klog.Warningf("unexpected MIG output format, missing colon fields: %q", val)
+				continue
+			}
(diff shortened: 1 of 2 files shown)


archlitchi:
Already fixed in https://github.com/Project-HAMi/HAMi/pull/2246

PR #2436, opened 2026-08-07
fix(scheduler): don't cache a nil NodeUsage when node is missing from snapshot

adity1raut:
## Summary

`getNodesUsage()` (`pkg/scheduler/scheduler.go`) snapshots all nodes via `ListNodes()` into `overallnodeMap`, then for each candidate node calls `GetNode(nodeID)` and did:

```go
cachenodeMap[node.ID] = overallnodeMap[node.ID]
```

with no check that the key exists. Two ways this goes wrong:

1. **Deterministic:** `ListNodes()` silently skips any node whose `Node` field is nil (`nodes.go:138`), but `GetNode()` applies no such filter and still returns success (`nodes.go:113-131`). So a node accepted by `GetNode()` but filtered out of the `ListNodes()` snapshot gets cached as a **nil `*NodeUsage`**, with nothing recorded in `failedNodes`.
2. **Racy:** the same nil-map outcome is reachable if a node registers in the window between the `ListNodes()` snapshot and the later per-node `GetNode()` call.

That `cachenodeMap` feeds `calcScore` → `calcScoreWithOptions`, which ranges over the map and calls `viewStatus(*node)` unconditionally per node — a nil entry panics the scoring goroutine. Nothing in this package recovers goroutine panics, so the panic takes down the whole scheduler process (a live filter-request denial-of-service, not just a bad score).

## Fix

Check the snapshot lookup explicitly; if the node is missing from the snapshot, record it in `failedNodes` instead of caching `nil`.

## Test plan

- [x] Added `Test_getNodesUsage_NodeMissingFromSnapshotIsNotCachedAsNil`, which reproduces the bug deterministically via the `ListNodes()`/`GetNode()` nil-`Node` divergence (no need to race goroutines) and asserts the node is neither cached nor silently dropped.
- [x] `go test ./pkg/scheduler/... -short --race -count=1` passes, including existing `Test_getNodesUsage*` tests.
- [x] `golangci-lint run ./pkg/scheduler/...` passes (0 issues).
- [x] `hack/verify-license.sh` and `hack/verify-import-aliases.sh` pass.
- Scoped to the scheduler extender's in-memory node cache; no device hardware involved, so unit testing applies per the hardware-validation gate in CONTRIBUTING.md.

## AI assistance disclosure

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved scheduling accuracy as pods terminate or complete initialization.
  * Removed stale device allocations and vendor information when devices are no longer available.
  * Improved NVIDIA node health cleanup when nodes are deleted.
  * Made device selection consistent and safely recover when a lock cannot be acquired.
  * Improved handling of MIG reservations and unavailable node usage information.

* **Tests**
  * Added regression coverage for stale allocations, initialization usage changes, device lock recovery, and missing node data.

### DIFF
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -849,10 +849,6 @@ func (s *Scheduler) getNodesUsage(nodes *[]string, task *corev1.Pod) (*map[strin
 		}
 		usage, ok := overallnodeMap[node.ID]
 		if !ok {
-			// GetNode() succeeded but the node is absent from the overallnodeMap snapshot
-			// (e.g. registered between the ListNodes() snapshot and this lookup, or filtered
-			// out of ListNodes() by a nil Node). Treat it as unavailable instead of caching a
-			// nil *NodeUsage, which would later be dereferenced unconditionally by the scorer.
 			klog.V(5).InfoS("node usage not found in snapshot", "node", nodeID)
 			failedNodes[nodeID] = "node usage unavailable"
 			continue
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -847,7 +847,17 @@ func (s *Scheduler) getNodesUsage(nodes *[]string, task *corev1.Pod) (*map[strin
 			failedNodes[nodeID] = "node unregistered"
 			continue
 		}
-		cachenodeMap[node.ID] = overallnodeMap[node.ID]
+		usage, ok := overallnodeMap[node.ID]
+		if !ok {
+			// GetNode() succeeded but the node is absent from the overallnodeMap snapshot
+			// (e.g. registered between the ListNodes() snapshot and this lookup, or filtered
+			// out of ListNodes() by a nil Node). Treat it as unavailable instead of caching a
+			// nil *NodeUsage, which would later be dereferenced unconditionally by the scorer.
+			klog.V(5).InfoS("node usage not found in snapshot", "node", nodeID)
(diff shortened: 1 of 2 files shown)


pkg/scheduler/scheduler.go
```diff
@@ -729,7 +729,17 @@ func (s *Scheduler) getNodesUsage(nodes *[]string, task *corev1.Pod) (*map[strin
 			failedNodes[nodeID] = "node unregistered"
 			continue
 		}
-		cachenodeMap[node.ID] = overallnodeMap[node.ID]
+		usage, ok := overallnodeMap[node.ID]
+		if !ok {
+			// GetNode() succeeded but the node is absent from the overallnodeMap snapshot
```
archlitchi: remove these comments

adity1raut: Removed in a9fdcc0.

pkg/scheduler/scheduler.go
```diff
@@ -729,7 +729,13 @@ func (s *Scheduler) getNodesUsage(nodes *[]string, task *corev1.Pod) (*map[strin
 			failedNodes[nodeID] = "node unregistered"
 			continue
 		}
-		cachenodeMap[node.ID] = overallnodeMap[node.ID]
+		usage, ok := overallnodeMap[node.ID]
```
mesutoezdil, line 850: after rebase, 4 old mig tests in scheduler_test.go fail, they are stale against #2378, not from your change. pls rebase and update those too.

archlitchi (approved):
/lgtm

PR #982, opened 2025-04-06
fix: Add error handling for nvml.Init in NvidiaDevicePlugin

yxxhero:
**What type of PR is this?**
/kind bug

**What this PR does / why we need it**:
This pull request includes several changes to improve error handling and add additional logging in the Nvidia device plugin. The most important changes involve initializing the NVML library with error checks and modifying the `GetDeviceNums` function to return an error.

Improvements to error handling:

* [`pkg/device-plugin/nvidiadevice/nvinternal/plugin/register.go`](diffhunk://#diff-f01e6dee63f1e8bca06b5508aeaee9563a45f2e0c58c8dd4ef917c346f1552c0L113-R116): Added error handling for `nvml.Init()` in the `getAPIDevices` method. If initialization fails, an error is logged and a panic is triggered.
* [`pkg/device-plugin/nvidiadevice/nvinternal/plugin/util.go`](diffhunk://#diff-8b63c6d64d45d3fd4890aafbcf25428f0747080081d88dc8d6d73cee29e31dd5L95-R99): Updated `GetIndexAndTypeFromUUID`, `GetMigUUIDFromIndex`, and `GetMigUUIDFromSmiOutput` methods to include error handling for `nvml.Init()`. If initialization fails, an error is logged and a panic is triggered. [[1]](diffhunk://#diff-8b63c6d64d45d3fd4890aafbcf25428f0747080081d88dc8d6d73cee29e31dd5L95-R99) [[2]](diffhunk://#diff-8b63c6d64d45d3fd4890aafbcf25428f0747080081d88dc8d6d73cee29e31dd5L148-R155) [[3]](diffhunk://#diff-8b63c6d64d45d3fd4890aafbcf25428f0747080081d88dc8d6d73cee29e31dd5L178-R195)
* [`pkg/device-plugin/nvidiadevice/nvinternal/plugin/util.go`](diffhunk://#diff-8b63c6d64d45d3fd4890aafbcf25428f0747080081d88dc8d6d73cee29e31dd5L178-R195): Modified the `GetDeviceNums` function to return an error if `nvml.Init()` or `nvml.DeviceGetCount()` fails, and added logging for these errors.

Additional logging:

[...]

PR #2437, opened 2026-08-07
fix(device-plugin): replace panic with graceful error handling in NVM…

nishantbkl3345-ship-it:
…L operations

- Replace panic(0) with error returns in getAPIDevices()
- Replace panic(0) with sentinel values in GetIndexAndTypeFromUUID()
- Replace panic(0) with empty string returns in GetMigUUIDFromIndex()
- Add error validation in GetContainerDeviceStrArray() callers
- Add structured error logging with nvml.ErrorString()

This prevents device plugin crashes when NVML operations fail due to GPU hardware failures, driver issues, or permission problems.

Completes panic removal work from #982, #2043, #1964

**What type of PR is this?**

Fixes #

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved resilience when NVIDIA device management services fail to initialize or return errors.
  * Invalid or unavailable devices are now skipped instead of causing the device plugin to crash.
  * MIG device lookup failures are handled gracefully, with clear error logging and safe fallback behavior.


FouoF:
Thanks for the proposed error handling. This PR bundles several NVML and registration changes while overlapping the MIG UUID parsing path already being handled by #2246, and it is now conflicting with that work. The parsing implementations also choose different failure semantics: this PR can report success after skipping a requested MIG device and produce fewer devices than requested, while #2246 follows the fail-closed behavior selected during maintainer review. We are closing the current bundled implementation. Please move any still-uncovered parsing inputs or tests to #2246 and propose unrelated error paths separately with their own reproductions.

PR #2442, opened 2026-08-07
fix(scheduler): persist per-node failure reasons as pod annotation wh…

### LINKED ISSUE #2286: Make GPU scheduling failures diagnosable: keep and surface the per-node reasons the scheduler already computes
**What I'd like to add**

When a GPU pod can't be scheduled, it's hard to find out why. The scheduler actually knows the exact reason per node — I can see it building the `failureReason` map in `pkg/scheduler/score.go` with values like `CardInsufficientMemory`, `CardTypeMismatch`, `NumaNotFit`, and `CardTimeSlicingExhausted` — but most of that never reaches the user in a usable form. In practice the pod just sits in `Pending` and the operator ends up reading scheduler logs to guess what happened.

Three things make it hard today:

- The failure reason is only recorded when *every* node fails. In `score.go` the block that records it is guarded by `len(res.NodeList) == 0`, so if some nodes fit but not the ones with the right card type or enough memory, no reason is recorded at all.
- What does get recorded is a Kubernetes Event, which is ephemeral (events age out) and is emitted once per reason type, so it's noisy and often already gone by the time someone looks.
- Because HAMi is a scheduler extender, it doesn't own the pod's `PodScheduled` condition, so the clean "0/N nodes are available: ..." summary the default scheduler gives never carries the GPU-specific reasons.

**Why it's useful**

"My GPU pod is stuck Pending and I don't know why" is one of the most common operator questions, and HAMi already computes the answer — it just discards most of it. Making that answer durable and complete would remove a lot of guesswork and log-diving, especially on busy multi-tenant clusters where the reason is often a specific card type or memory shortfall on the nodes the user expected.

**What I'm proposing (rough, happy to adjust with mentors)**

1. Record the aggregated failure reason even when only some nodes fail, not only when all of them do.
2. Persist the last scheduling decision / failure reason onto the pod as an annotation, so it survives after the event ages out and can be read back later.
3. A small surfacing step, for example a `why <pod>` command, that turns that into plain language: "no node has a free MR-V100 with 16Gi; 3 nodes failed CardTypeMismatch, 2 failed CardInsufficientMemory." Where this lives is up to the maintainers — it could sit in the CLI discussed in #1638, or start purely as the scheduler-side improvement in points 1 and 2 with no new binary.

I'd like to work on this. I've been reading the filter/score path fairly closely and it looks like a well-bounded piece. I'm happy to write a short design doc first if that's the preferred flow, and to scope it down to whatever the maintainers think fits.

/kind feature
/kind design

Nitish08-08:
…en scheduling fails

When a GPU pod cannot be scheduled, the scheduler already computes the exact per-node failure reasons (CardInsufficientMemory, CardTypeMismatch, etc.) but only records them as ephemeral Kubernetes events when ALL nodes fail. Operators frequently cannot find these events because they age out, leaving pods stuck in Pending with no diagnostic information.

This change:
- Always builds the failureReason map in calcScoreWithOptions regardless of whether some nodes fit
- Returns the failureReason map from calcScore and calcScoreWithOptions
- Persists the aggregated failure reasons as a hami.io/scheduling-failure-reasons pod annotation when no node fits
- Adds formatFailureReasons helper for human-readable formatting

Example annotation value:
  '3 nodes CardInsufficientMemory(node1,node2,node3); 1 nodes CardTypeMismatch(node4)'

Fixes #2286

**What type of PR is this?**

Fixes #

## Summary by CodeRabbit

* **New Features**
  * Pods that cannot be scheduled now receive an annotation summarizing why nodes were rejected.
  * Failure details are consistently formatted and sorted for easier troubleshooting.

* **Bug Fixes**
  * Scheduling failures now preserve device-specific rejection reasons, including when some nodes remain viable.
  * Failure-reporting issues no longer change the scheduling result.

### DIFF
--- a/pkg/scheduler/score.go
+++ b/pkg/scheduler/score.go
@@ -90,11 +90,11 @@ func fitInDevices(node *NodeUsage, requests device.ContainerDeviceRequests, pod 
 	return true, ""
 }
 
-func (s *Scheduler) calcScore(nodes *map[string]*NodeUsage, resourceReqs device.PodDeviceRequests, task *corev1.Pod, failedNodes map[string]string) (*policy.NodeScoreList, error) {
+func (s *Scheduler) calcScore(nodes *map[string]*NodeUsage, resourceReqs device.PodDeviceRequests, task *corev1.Pod, failedNodes map[string]string) (*policy.NodeScoreList, map[string][]string, error) {
 	return s.calcScoreWithOptions(nodes, resourceReqs, task, failedNodes, true, false)
 }
 
-func (s *Scheduler) calcScoreWithOptions(nodes *map[string]*NodeUsage, resourceReqs device.PodDeviceRequests, task *corev1.Pod, failedNodes map[string]string, recordEvents bool, detailedFailureReason bool) (*policy.NodeScoreList, error) {
+func (s *Scheduler) calcScoreWithOptions(nodes *map[string]*NodeUsage, resourceReqs device.PodDeviceRequests, task *corev1.Pod, failedNodes map[string]string, recordEvents bool, detailedFailureReason bool) (*policy.NodeScoreList, map[string][]string, error) {
 	userNodePolicy := config.NodeSchedulerPolicy
 	if task.GetAnnotations() != nil {
 		if value, ok := task.GetAnnotations()[util.NodeSchedulerPolicyAnnotationKey]; ok {
@@ -188,7 +188,7 @@ func (s *Scheduler) calcScoreWithOptions(nodes *map[string]*NodeUsage, resourceR
 	wg.Wait()
 	close(errCh)
 
-	// only pod scheduler failure will record failure event
+	// Record failure events for all-failed case (existing behavior)
 	if recordEvents && len(res.NodeList) == 0 {
 		for reasonType, failureNodes := range failureReason {
 			sort.Strings(failureNodes)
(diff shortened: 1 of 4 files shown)


FouoF:
I replied at https://github.com/Project-HAMi/HAMi/issues/2286 about this pr motivation. PTAL @Nitish08-08

pkg/scheduler/score.go
```diff
@@ -201,5 +201,24 @@ func (s *Scheduler) calcScoreWithOptions(nodes *map[string]*NodeUsage, resourceR
 	for e := range errCh {
 		errorsSlice = append(errorsSlice, e)
 	}
-	return &res, utilerrors.NewAggregate(errorsSlice)
+	return &res, failureReason, utilerrors.NewAggregate(errorsSlice)
```
mesutoezdil, line 204: #1773 merged today, it rewrote calcscorewithoptions into a new scorenode() helper. this signature change conflicts hard now, almost the whole function. also fouof still waits for your reply on the pr motivation question. pls answer that first, then rebase.

Nitish08-08:
Thanks @mesutoezdil and @FouoF for the review.

On the motivation question (issue #2286 — make GPU scheduling failures diagnosable): when a GPU pod is stuck Pending, the scheduler extender already computes a precise per-node failure reason (CardInsufficientMemory, CardTypeMismatch, NumaNotFit, CardTimeSlicingExhausted, ...) in calcScoreWithOptions, but almost none of it reaches the operator:
1. The aggregated reason is only written when *every* node fails (len(res.NodeList) == 0), so a partial failure (e.g. the cards the user wants are only mis-typed on some nodes) records nothing.
2. What is written is a Kubernetes Event, which is ephemeral (events age out) and emitted once per reason type.

On recording timing, I agree with your read: a pod is only truly unschedulable once all nodes have failed, so persisting the reasons at that point reflects the final decision, not a partial view. This PR does exactly that — the per-node reasons are recorded at filter time only after no node fits, instead of being dropped.

Why an annotation rather than relying on events: in a busy multi-tenant cluster the reason a pod is stuck is usually a specific card type or memory shortfall on the nodes the user expected, and the event is often already gone by the time someone looks. Persisting the deterministic, complete reason set on the pod annotation (sorted reason types and node names so it is stable across retries) keeps it readable without log-diving, and needs no new binary/CLI — this implements points 1+2 of the issue on the scheduler-extender side only.

mesutoezdil:
Answers must be written by human being. 

Issue #2084, opened 2026-07-17
[BUG] Scheduler infinite loop in PlatternMIG on safecast.Convert failure


AdeshDeshmukh:
**What happened**:

While reviewing the MIG scheduling path, I found that PlatternMIG() at pkg/device/devices.go:515-518 can enter an infinite loop.

If safecast.Convert[int32](templateIdx) fails (e.g., integer overflow), the continue statement skips the count++ increment. Since count never reaches int(val.Count), the loop spins forever, hanging the scheduler goroutine and blocking all pod scheduling.

**What you expected to happen**:

The count variable should increment regardless of the safecast conversion result, allowing the loop to terminate normally.

**How to reproduce it (as minimally and precisely as possible)**:

Static code analysis finding. The bug is visible at pkg/device/devices.go:515-518:

for count < int(val.Count) {
    n.Index, err = safecast.Convert[int32](templateIdx)
    if err != nil {
        continue  // BUG: skips count++, infinite loop
    }
    n.UsageList = append(n.UsageList, MigTemplateUsage{...})
    count++
}

This function is called from three critical scheduling paths:
- pkg/scheduler/scheduler.go:649
- pkg/device/nvidia/device.go:613
- pkg/device/nvidia/device.go:680

**Anything else we need to know?**:

Static analysis finding — no reproduction steps apply. Proposed fix: move count++ before the continue statement.

**Environment**:

- HAMi version: HEAD of main

AdeshDeshmukh:
Happy to open a PR for this if the approach looks right, or adapt if you prefer a different error-handling pattern.

saiyam1814:
Can you share the reproducible steps for this ? Also the hardware this was tested?

AdeshDeshmukh:
Hey @saiyam1814 , this was a static code analysis finding , so there aren't any runtime repro steps or specific hardware tested.

I found it just by reading the loop in devices.go — if safecast.Convert ever returns an error (like an integer overflow) , the continue statement skips count++, so the loop never terminates. Since it's a pure logic bug, I didn't run it on a physical GPU node.

Let me know if you're okay!

AdeshDeshmukh:
Just wanted to add more context to my earlier reply —

The bug is at devices.go:515. If templateIdx is bigger than math.MaxInt32 (about 2.1 billion), safecast.Convert[int32] returns an error. The continue skips count++, so the loop just spins forever.

I put together a quick test to confirm:

[...]

PR #2443, opened 2026-08-07
fix(device): prevent infinite loop in PlatternMIG on safecast.Convert error

Nitish08-08:
… error

The inner loop used 'continue' on conversion error, which returned to the loop condition without incrementing 'count'. Since templateIdx is constant within the inner loop, a failed conversion would fail every iteration, causing an infinite loop.

Changed 'continue' to 'break' to exit the inner loop on error.

**What type of PR is this?**

Fixes #

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed template processing so it stops safely when a device value cannot be converted to the expected format.

### DIFF
--- a/pkg/device/devices.go
+++ b/pkg/device/devices.go
@@ -520,7 +520,7 @@ func PlatternMIG(n *MigInUse, templates []Geometry, templateIdx int) {
 		for count < int(val.Count) {
 			n.Index, err = safecast.Convert[int32](templateIdx)
 			if err != nil {
-				continue
+				break
 			}
 			n.UsageList = append(n.UsageList, MigTemplateUsage{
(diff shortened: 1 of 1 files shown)


FouoF:
Thanks for looking into this path. We traced templateIdx through the current master: it has already been validated or successfully used as a slice index before the reported conversion failure could occur. safecast.Convert[int32] would only fail here with more than MaxInt32 templates, which is not a reachable production state. This is the same underlying reachability issue that led to #2084 being closed. Because there is no reproducible failure and the proposed guard protects an unreachable condition, we are closing this PR. Please reopen with a minimal input that fails on the old code if a reachable case exists.

PR #2445, opened 2026-08-07
cleanup(metax): decouple Metax-GPU ScoreNode from the scheduler policy string

### LINKED ISSUE #2439: Decouple MetaxDevices.ScoreNode (Metax-GPU) from the scheduler policy string

```go
    sum := 0
        sum += len(dev)
    }

    res := float32(0)
        if ok {
        }
        if ok {
        }
    }
    return res
}
```

**Why is this needed**:

**Evidence**

```
5
0
```

**Proposed solution**

2. Add the marker method:

```go
```

**Affected components**

**Tests required**

**Acceptance criteria**

**Duplicate check**

asadjan4611:
**What type of PR is this?**

/kind cleanup

`MetaxDevices.ScoreNode` (the `Metax-GPU` backend) chose its result based on the scheduler policy string: it read the `metax-tech.com/gpu.topology.losses` annotation under Binpack and `metax-tech.com/gpu.topology.scores` under Spread,encoding the policy's sort direction directly in the device layer. This is the same anti-pattern #2404 removed from `MetaxSDevices` (`Metax-SGPU`).

This PR makes `ScoreNode` return a single policy-independent, "higher is a better node" score, and has `MetaxDevices` implement the existing `device.PolicyNeutralScorer` marker. The shared scheduler policy layer
(`OverrideScore`) then applies the weight and the Spread sign inversion in one place. `OverrideScore` itself is unchanged, and no other backend is affected.

Fixes #2439

- Depends on #2413 (issue #2404). This branch is stacked on that one because it  builds on the `PolicyNeutralScorer` marker and `OverrideScore` weighting  introduced there. Please merge #2413 first; this PR will then narrow to a
  single commit.
- Behavioral note: Binpack previously read `losses` while Spread read `scores`.  Both policies now prefer `scores` and fall back to `losses`. Spread behavior   is preserved exactly; Binpack now uses the consistent `scores` signal when a
  node publishes both annotations.
- Node ranking is preserved across both Binpack and Spread (the sign inversion  keeps the best node winning under each sort direction).
- Validation: race-detector tests for `pkg/device/metax` and   `pkg/scheduler/policy`, `go vet`, `goimports` (local prefix), and license headers all pass.

**I used AI Assistance for verification of test cases and for the PR description and also verified for the solution implementations.** 

```NONE```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Metax device scoring by prioritizing explicit scores and using topology loss data as a fallback.
  * Correctly distinguishes missing, malformed, and zero-valued topology data.
  * Ensures consistent node scoring across Binpack, Spread, and other scheduling policies.
  * Preserves expected node ordering for Metax GPU scheduling.
* **Tests**
  * Expanded coverage for topology annotation parsing, multi-GPU lookup, scoring precedence, and policy-independent behavior.

### DIFF
--- a/pkg/device/metax/device.go
+++ b/pkg/device/metax/device.go
@@ -39,6 +39,12 @@ const (
 	MetaxGPUCommonWord   = "Metax-GPU"
 	MetaxAnnotationLoss  = "metax-tech.com/gpu.topology.losses"
 	MetaxAnnotationScore = "metax-tech.com/gpu.topology.scores"
+
+	// metaxTopologyLossBase converts a topology "loss" (where lower is better)
+	// onto the same "higher is better" scale as the scores annotation, via
+	// metaxTopologyLossBase - loss. It must stay larger than the largest loss a
+	// node can publish so the converted score remains positive.
+	metaxTopologyLossBase = 2000
 )
 
 var (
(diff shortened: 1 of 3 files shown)


FouoF:
Thanks for working on the MetaX scoring code. This PR is stacked on #2413 and includes that PR's full diff, so it cannot be reviewed as an independent cleanup. More importantly, it changes the MetaX-GPU binpack signal from losses to scores, while the stated acceptance condition is that node ordering remain unchanged. That is a behavioral change rather than a decoupling-only cleanup. We are closing this PR. Please first finish or replace #2413, then propose any independent cleanup from the current master with tests that prove ordering equivalence.

asadjan4611:
@FouoF @archlitchi  can you please review my PR

mesutoezdil:

Pls read the rules.

PR #2449, opened 2026-08-07
fix(device-plugin): replace panic with graceful degradation in NVML init

nishantbkl3345-ship-it:
When NVML initialization fails (driver not loaded, permission issues, or hardware problems), the device plugin now logs the error and returns an empty device list instead of crashing with panic(0).

This allows the device plugin to continue running and retry on the next registration cycle, enabling self-healing without manual intervention.

Improves operational stability by preventing total cluster GPU unavailability when a single node's NVIDIA driver encounters issues.

Ref: #982, #2043, #1964 (related panic-fix work)

**What type of PR is this?**

Fixes #

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling when NVIDIA Management Library initialization fails.
  * The device plugin now logs the error and continues without reporting devices instead of crashing.
  * Device registration retries are preserved when node annotation updates fail.
  * Devices can be detected again after the library becomes available.

* **Documentation**
  * Added documentation covering the failure scenario, reproduction steps, expected behavior, impact, recovery, validation, and graceful-degradation approach.

### DIFF
--- a/PR_DESCRIPTION_NVML_INIT.md
+++ b/PR_DESCRIPTION_NVML_INIT.md
@@ -21,7 +21,7 @@ Replace `panic(0)` with graceful degradation when NVML initialization fails in `
 ## Problem Statement
 When NVML initialization fails (driver not loaded, permission issues, hardware problems), the device plugin crashes with `panic(0)`, making **all GPUs on the node unavailable** until manual intervention.
 
-**Current code** (`register.go:97`):
+**Before this fix** (`register.go:97`):
 ```go
 if nvret := nvmlInit(); nvret != nvml.SUCCESS {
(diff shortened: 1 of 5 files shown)


FouoF:
Thanks for investigating NVML initialization. On failure, this implementation can register an empty device list while leaving the old topology-score annotation in place; an annotation patch failure is still reported as changed=true. It also introduces a package-level mutable nvmlInit seam and bundles more than 700 lines of repository-root process documentation around a small code change, while overlapping #2437 and lacking complete hosted CI. The partial-degradation contract is unsafe and the scope is not reviewable as a focused fix, so we are closing this PR.

PR #2451, opened 2026-08-07
fix(device-plugin): re-check lock file state to prevent vGPUmonitor hang

### LINKED ISSUE #2450: vGPUmonitor: MIG lock file create/remove events can race through a saturating channel, permanently hanging watchAndFeedback
## What happened

`watchLockFile()` in `pkg/device-plugin/nvidiadevice/nvinternal/plugin/lock.go` reports MIG apply lock-file transitions over a single `chan bool` with capacity **1**, using **non-blocking sends** for both lock creation and removal events:

```go
func watchLockFile(file string) (chan bool, error) {
    sigChan := make(chan bool, 1)

    ...

    if event.Has(fsnotify.Create) {
        select {
        case sigChan <- true:
        default:
        }
    }

    if event.Has(fsnotify.Remove) {
        select {
        case sigChan <- false:
        default:
        }
    }
}
```

Because both notifications share the same single-slot channel and both use `select { ... default: }`, any event that arrives while the buffer is occupied is silently discarded.

The consumer in `cmd/vGPUmonitor/main.go` assumes that every received "lock created" notification will eventually be followed by a corresponding "lock removed" notification:

```go
if errors.Is(err, errTemporaryClosed) {
    <-lockChannel
    continue
}
```

After `watchAndFeedback()` returns `errTemporaryClosed`, the goroutine blocks waiting for the next value from `lockChannel`, expecting it to represent lock removal.

## Race condition

The following interleaving can permanently wedge the feedback loop:

1. The MIG apply lock file is created.
2. `watchLockFile()` successfully enqueues `true`.
3. Before the consumer drains the channel, the lock file is removed.
4. `watchLockFile()` attempts to enqueue `false`.
5. The channel buffer is still full, so the non-blocking send falls through the `default` branch and the remove notification is silently discarded.
6. The consumer later receives the stale `true`, returns `errTemporaryClosed`, and blocks waiting for another value from `lockChannel`.
7. Since the only remove notification was already dropped, no further value is delivered and the goroutine remains blocked indefinitely.

There is no recovery path because the consumer does not re-check the filesystem after entering the wait state—it relies entirely on receiving the queued "lock removed" notification.

## Impact

Once this happens, the vGPU monitor goroutine stops permanently.

As a result:

- GPU utilization feedback stops updating.
- `SetUtilizationSwitch()` is no longer executed.
- Compute-utilization enforcement becomes stale.
- Metrics are no longer refreshed.
- Recovery requires restarting the device-plugin process or pod.

The failure is silent: no panic or error is emitted, but the feedback loop never resumes.

## Why this is plausible

This race depends only on normal goroutine scheduling and the current drop-on-full channel semantics.

The MIG apply lock file is intentionally short-lived. `fsnotify` commonly delivers `Create` and `Remove` events back-to-back, while the producer and consumer execute independently. If the consumer has not yet drained the buffered `Create` notification when the `Remove` event arrives, the latter can be discarded by the current non-blocking send.

No malformed requests or unsupported execution paths are required.

## Root cause

Two distinct state transitions ("lock acquired" and "lock released") are encoded as boolean values on a single buffered channel with capacity one.

Because sends are intentionally lossy (`select { case ... default: }`), dropping a `false` notification is unrecoverable: the consumer has no other mechanism to determine that the lock has already been removed.

## Suggested fix

A state-based notification mechanism would avoid this race entirely.

One possible approach is:

- Replace `chan bool` with a notification channel (for example, `chan struct{}`).
- Notify only that the lock state has changed.
- Have the consumer determine the current lock state by checking the existence of the lock file (`os.Stat`) after waking.

This makes notifications idempotent and removes any dependency on receiving every intermediate event.

Alternative approaches, such as using separate channels for create/remove events or replacing the non-blocking send with a blocking send coordinated by context cancellation, would also eliminate the lost-event condition.

AyushSrivastava1818:
## Summary

`WatchLockFile()` previously reported MIG apply lock file transitions over a single buffered `chan bool`, using non-blocking sends for both lock creation and removal events. During a rapid create/remove sequence, the "lock removed" notification could be dropped if the channel buffer was already occupied. The consumer would then wait indefinitely for a notification that had already been discarded, permanently stalling the vGPU monitor feedback loop until the process was restarted.

This PR changes the notification mechanism from **event-based state propagation** to **state-based synchronization**. The notification channel is now used only to signal that the lock state may have changed, while the filesystem (`os.Stat`) becomes the single source of truth for determining whether the MIG apply lock currently exists.

* `pkg/device-plugin/nvidiadevice/nvinternal/plugin/lock.go`: replaced `chan bool` with a notification-only `chan struct{}`, added `IsMigApplyLockExist()`, and closed the notification channel on watcher shutdown.
* `cmd/vGPUmonitor/feedback.go`: added an initial lock-state check and re-evaluates the filesystem after receiving notifications instead of relying on queued channel values.
* `cmd/vGPUmonitor/main.go`: replaced the direct `<-lockChannel` wait with a state-check loop that waits until the lock file is actually removed before restarting `watchAndFeedback()`.
* `pkg/device-plugin/nvidiadevice/nvinternal/plugin/lock_test.go`: updated existing tests for the new notification API and added a `RapidCreateAndRemove` regression test covering rapid lock file transitions.

Fixes #2450 

The notification channel is no longer treated as the source of truth. It acts only as a wake-up signal, while `IsMigApplyLockExist()` determines the current lock state from the filesystem. Even if notifications are coalesced or dropped due to non-blocking sends, consumers always re-check the filesystem before proceeding, preventing the feedback loop from blocking indefinitely because of a lost notification.

**Testing**

* Updated existing `WatchLockFile` unit tests.
* Added `RapidCreateAndRemove` regression coverage.
* `go test ./pkg/device-plugin/nvidiadevice/nvinternal/plugin/... -count=1`
* Verified the modified packages build successfully (`GOOS=linux`).
* Ran `go vet` on the modified packages with no new issues reported.

Yes. The vGPU monitor no longer risks becoming permanently stalled due to a lost lock-file notification during rapid MIG apply operations. Metrics collection and utilization feedback now recover correctly even when intermediate filesystem notifications are coalesced or dropped.

**AI Disclosure:**

AI assistance was used for code inspection, concurrency analysis, and drafting the regression tests and PR description. The implementation, testing, and final changes were reviewed before submission.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of MIG apply locks during startup and runtime.
  - Monitoring now waits for locks to be removed instead of stopping prematurely.
  - Added safer behavior when lock notifications close, are unavailable, or arrive after the lock is removed.
  - Improved responsiveness to rapid lock creation and removal events.
  - Monitoring now exits cleanly when its operating context is canceled.
  - Added fail-closed handling for lock-state detection errors.

### DIFF
--- a/cmd/vGPUmonitor/feedback.go
+++ b/cmd/vGPUmonitor/feedback.go
@@ -31,6 +31,19 @@ import (
 
 var errTemporaryClosed = errors.New("temporary closed")
 
+// migLockExistFn is a package-level testability seam: production code always
+// calls plugin.IsMigApplyLockExist(); tests may override this var (restoring
+// it via t.Cleanup) to exercise lock-present/absent paths without real
+// filesystem infrastructure.
+var migLockExistFn = plugin.IsMigApplyLockExist
+
+// nvmlInitFn and nvmlShutdownFn are testability seams for the NVML lifecycle
+// calls inside watchAndFeedback. Tests swap these to stubs that return
+// nvml.SUCCESS / do nothing, so the lock-check and notification paths can
+// be exercised without real GPU hardware.
+var nvmlInitFn = func() nvml.Return { return nvml.Init() }
+var nvmlShutdownFn = func() { nvml.Shutdown() }
+
 //type hostGPUPid struct {
 //	hostGPUPid int
 //	mtime      uint64
@@ -136,12 +149,12 @@ func Observe(lister *nvidia.ContainerLister) {
 
 func watchAndFeedback(ctx context.Context, lister *nvidia.ContainerLister, migLockSignal <-chan struct{}) error {
 	klog.Info("Starting watchAndFeedback")
-	if nvret := nvml.Init(); nvret != nvml.SUCCESS {
+	if nvret := nvmlInitFn(); nvret != nvml.SUCCESS {
 		return fmt.Errorf("failed to initialize NVML: %s", nvml.ErrorString(nvret))
 	}
-	defer nvml.Shutdown()
+	defer nvmlShutdownFn()
 
-	if plugin.IsMigApplyLockExist() {
+	if migLockExistFn() {
 		klog.Info("MIG apply lock file already exists at startup")
 		return errTemporaryClosed
 	}
@@ -159,7 +172,7 @@ func watchAndFeedback(ctx context.Context, lister *nvidia.ContainerLister, migLo
 				klog.Info("MIG lock signal channel closed")
 				return nil
 			}
-			if plugin.IsMigApplyLockExist() {
+			if migLockExistFn() {
 				klog.Info("Received MIG apply lock file event, lock is held")
 				return errTemporaryClosed
 			}
--- a/cmd/vGPUmonitor/feedback.go
+++ b/cmd/vGPUmonitor/feedback.go
@@ -25,6 +25,7 @@ import (
 	"github.com/NVIDIA/go-nvml/pkg/nvml"
 	"k8s.io/klog/v2"
 
+	"github.com/Project-HAMi/HAMi/pkg/device-plugin/nvidiadevice/nvinternal/plugin"
 	"github.com/Project-HAMi/HAMi/pkg/monitor/nvidia"
 )
 
@@ -133,12 +134,17 @@ func Observe(lister *nvidia.ContainerLister) {
 	}
 }
 
-func watchAndFeedback(ctx context.Context, lister *nvidia.ContainerLister, migLockSignal <-chan bool) error {
+func watchAndFeedback(ctx context.Context, lister *nvidia.ContainerLister, migLockSignal <-chan struct{}) error {
 	klog.Info("Starting watchAndFeedback")
 	if nvret := nvml.Init(); nvret != nvml.SUCCESS {
 		return fmt.Errorf("failed to initialize NVML: %s", nvml.ErrorString(nvret))
 	}
 	defer nvml.Shutdown()
+
+	if plugin.IsMigApplyLockExist() {
+		klog.Info("MIG apply lock file already exists at startup")
+		return errTemporaryClosed
+	}
 
 	ticker := time.NewTicker(time.Second * 5)
 	defer ticker.Stop()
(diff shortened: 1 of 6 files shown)


AyushSrivastava1818:
Hi @archlitchi @chaunceyjiang @peachest,

PR #2451 (`fix/watch-lockfile-concurrency-bug`) is ready for maintainer review:

- **CI Status**: All checks are green (Compile, Lint, Unit Test, e2e, CodeQL, DCO, Dependency Quality, License Compliance, Security Analysis).
- **Review Threads**: Both CodeRabbit review threads (`RapidCreateAndRemove` test coordination in `lock_test.go` and `CoalescedBurst_FinalStateConverges` test stability in `feedback_test.go`) were addressed in commit `807e611` and marked resolved.

Could you please review and apply `/lgtm` / `/approve` when you have a moment? Thank you!

FouoF:
Why not use a blocking send or expand the buffer size?

AyushSrivastava1818:
@FouoF Considered both, went with neither — here's the reasoning:

Blocking send — ties the watcher goroutine to the consumer's readiness. Worse, the channel is explicitly closed on watcher shutdown (defer close(sigChan)), so any blocking send racing that close either panics (send on a closed channel) or, if it lands just before the close, blocks forever with nothing left to read it. A blocking send here isn't just slower, it's a crash/deadlock risk built into the shutdown path.

Bigger buffer — only raises the burst size needed to reproduce #2450, it doesn't remove the failure mode, and it doesn't fix staleness: even with zero drops, a queued event can be read after the real state has already moved past it. You can see the same tension even in fsnotify's own error channel — when watcher.Errors overflows, we don't try to force every error through either; we just fire a non-blocking coalesced wake-up and let the consumer re-check state (lines 138–151). Same principle, applied consistently everywhere in this file.

So the channel is now purely a wake-up signal (chan struct{}, buffered 1, always non-blocking send/default), and IsMigApplyLockExist() is the single source of truth on every wake. However many notifications coalesce, drop, or race the shutdown close, the consumer converges to the actual current filesystem state — so correctness doesn't depend on the channel being lossless or ordered at all, which neither a bigger buffer nor a blocking send can guarantee.

RapidCreateAndRemove in lock_test.go covers the burst case #2450 was actually failing on.

mesutoezdil:

PR #2453, opened 2026-08-07
fix(device-plugin): replace panic with graceful error handling in Get…

nishantbkl3345-ship-it:
…IndexAndTypeFromUUID and GetMigUUIDFromIndex

Replace 7 panic(0) calls with graceful error handling using sentinel return values:
- GetIndexAndTypeFromUUID: returns ("", -1) on any NVML error
- GetMigUUIDFromIndex: returns "" on any NVML error

Both functions are called by GetContainerDeviceStrArray which already checks for these sentinel values and skips failed devices gracefully.

Panic conditions replaced:
1. GetIndexAndTypeFromUUID - NVML init failure
2. GetIndexAndTypeFromUUID - DeviceGetHandleByUUID failure
3. GetIndexAndTypeFromUUID - GetName failure
4. GetIndexAndTypeFromUUID - GetIndex failure
5. GetMigUUIDFromIndex - NVML init failure
6. GetMigUUIDFromIndex - DeviceGetHandleByUUID failure
7. GetMigUUIDFromIndex - GetUUID failure

Enhanced error logging with structured logging (klog.ErrorS) including:
- Human-readable error strings via nvml.ErrorString()
- Context fields (uuid, index, returnCode)
- Proper error wrapping with fmt.Errorf()

This prevents device plugin crashes when NVML queries fail due to:
- Driver not loaded or unavailable
- GPU in bad state
- Invalid/stale UUID references
- Transient hardware issues

**What type of PR is this?**

Fixes #

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved resilience when GPU and MIG device information cannot be retrieved.
  * Prevented unexpected process termination by returning safe fallback values for device lookup failures.
  * Added clearer structured error reporting for these conditions.
  * Prevented invalid device identifiers from being exposed when hardware information is incomplete or unavailable.
  * Improved reliability of device discovery in environments with transient or unsupported GPU management errors.

### DIFF
--- a/pkg/device-plugin/nvidiadevice/nvinternal/plugin/util.go
+++ b/pkg/device-plugin/nvidiadevice/nvinternal/plugin/util.go
@@ -150,24 +150,28 @@ func patchErasedAnnotation(pod *corev1.Pod, dtype string, podSingleDev device.Po
 func GetIndexAndTypeFromUUID(uuid string) (string, int) {
 	defer nvml.Shutdown()
 	if nvret := nvml.Init(); nvret != nvml.SUCCESS {
-		klog.Errorln("nvml Init err: ", nvret)
-		panic(0)
+		klog.ErrorS(fmt.Errorf("nvml init failed: %s", nvml.ErrorString(nvret)), 
+			"Failed to initialize NVML in GetIndexAndTypeFromUUID", "uuid", uuid, "returnCode", nvret)
+		return "", -1
 	}
 	originuuid := strings.Split(uuid, "[")[0]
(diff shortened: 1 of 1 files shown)


FouoF:
Thanks for replacing the panic. The current fallback is not safe: after the NVML lookup returns the sentinel, GetContainerDeviceStrArray skips that item and can let Allocate return success with fewer MIG devices than requested. This overlaps the same author's #2437 direction and has no regression test or hosted CI proving fail-closed allocation. We are closing this PR. Please contribute any valid input case to the canonical #2246 path, which preserves requested-device count and error propagation.

PR #2455, opened 2026-08-07
fix(metrics): correct core_allocated_ratio to emit a true ratio inste…

### LINKED ISSUE #2454: [bug]: core_allocated_ratio metrics export raw percentages instead of ratios, breaking operator alerts
**What happened**:
In `cmd/scheduler/metrics.go`, the metrics `hami_gpu_core_limit_ratio` and `hami_gpu_core_allocated_ratio` are currently emitting raw core counts (e.g., `50`, `100`) rather than a mathematical `0.0` to `1.0` ratio. 

**Blast radius**: 
Because the metric names explicitly advertise a `_ratio` suffix, operators rely on these to build standard Prometheus alerts (e.g., `hami_gpu_core_allocated_ratio > 0.8`). Under the current implementation, these alerts immediately fire on any GPU exceeding 1% core allocation because the metric exports `1.0` instead of `0.01`. This effectively breaks observability and alerting for shared-core GPU deployments.

The metric should output a true mathematical ratio (i.e. `Usedcores / Totalcores`) constrained between `0.0` and `1.0`, keeping it mathematically consistent with `hami_node_gpu_memory_allocated_ratio`.

This was identified via code inspection of `cmd/scheduler/metrics.go`, specifically around the metric generation loop where `coreLimit` and `coreAllocated` are passed directly to `sendMetric` without division.

To reproduce live:
1. Run the HAMi scheduler with metrics enabled.
2. Hit the `/metrics` endpoint (port 9395).
3. Observe that `hami_gpu_core_allocated_ratio` yields integer/percentage values exceeding 1.0.

**Proposed fix:**
Modify `Collect()` to divide `coreAllocated` by `coreLimit` before passing it to `sendMetric`. To preserve backward compatibility for legacy dashboards, the `--legacy-metrics` branch of the code should be left untouched. 

*(Note: I have a PR ready to submit for this!)*

**Environment**:
- HAMi version: master branch (code inspection)
- Others: Found via static code review while auditing metrics for #2126.

Rickydama3:
**What type of PR is this?**
/kind bug

This PR fixes a bug in the scheduler's Prometheus metrics exporter where the core allocation ratios were being exported as raw percentages (e.g., `50` or `100`) rather than true mathematical ratios (e.g., `0.5` or `1.0`). 

**Root Cause:**
In `cmd/scheduler/metrics.go`, `hami_gpu_core_limit_ratio` and `hami_gpu_core_allocated_ratio` were being passed `coreLimit` (typically 100) and `coreAllocated` (e.g., 50) directly from the device struct, completely bypassing the division required to create a ratio. 

**Fix:**
- Explicitly calculate `allocatedRatio = coreAllocated / coreLimit`.
- Hardcode `limitRatio` to `1.0` (as it represents 100% of the limit).
- This is fixed for both Node-level metrics and Container-level metrics (`ctrvGPUdeviceAllocatedCoreDesc`).
- Legacy metrics were intentionally left untouched so we do not break backward compatibility for older deployments/dashboards expecting percentage values.

Fixes #2454

This resolves a silent issue where operator alerts like `hami_gpu_core_allocated_ratio > 0.8` would immediately trigger on any GPU exceeding 1% core allocation. Found via static code review while auditing metrics for #2126.

```release-note
Fix: core_allocated_ratio metric exports percentage instead of ratio

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved GPU resource metrics across all device types with consistent percentage-based limits and allocation ratios.
  * Added safeguards for zero or invalid memory values to prevent misleading results.
  * Improved per-container device matching for identifiers containing MIG UUID suffixes.
  * Added normalized core allocation metrics based on each device’s total capacity.
  * Preserved existing raw-value metrics for compatibility.
  * Updated metric reporting to provide more consistent and reliable values across devices.

### CODE REVIEW

File: cmd/scheduler/metrics.go [hunk 1]
```diff
@@ -210,7 +210,11 @@ func (cc ClusterManagerCollector) Collect(ch chan<- prometheus.Metric) {
 			if err := sendMetric(ch, nodevGPUMemoryLimitDesc, prometheus.GaugeValue, float64(devs.Device.Totalmem)*float64(1024)*float64(1024), nodeID, devs.Device.ID, fmt.Sprint(devs.Device.Index), devs.Device.Type); err != nil {
 				klog.V(4).Infof("Failed to send nodevGPUMemoryLimitDesc metric: %v", err)
 			}
-			if err := sendMetric(ch, nodevGPUCoreLimitDesc, prometheus.GaugeValue, coreLimit, nodeID, devs.Device.ID, fmt.Sprint(devs.Device.Index), devs.Device.Type); err != nil {
+			limitRatio := 1.0
```
mesutoezdil: #2527 merged today. it made these metrics use 0-100 percent on purpose, see normalizeamdcoremetrics. this pr changes them to 0-1 instead. that reverts #2527 and breaks amd normalization. also conflicts now, pls rebase and pick one scale, not both.

Rickydama3:
"Closing this PR since PR #2527 was merged and resolves the core allocation ratio issue using the project's preferred 0-100% scale!"

PR #2464, opened 2026-08-08
Fix/patch erased annotation nil map

### LINKED ISSUE #2462: fix(device-plugin): nil pod.Annotations map write panics in patchErasedAnnotation
## Summary

`patchErasedAnnotation` in `pkg/device-plugin/nvidiadevice/nvinternal/plugin/util.go`
writes directly to `pod.Annotations[annoKey]` without checking whether the map is
`nil` first. If `pod.Annotations` is `nil` at the time this function runs, the
write panics with `assignment to entry in nil map`, crashing the device-plugin
daemon on that node.

```go
func patchErasedAnnotation(pod *corev1.Pod, dtype string, podSingleDev device.PodSingleDevice) error {
	klog.V(5).Infof("After erase annotation, remaining devices: %v", podSingleDev)
	encoded := device.EncodePodSingleDevice(podSingleDev)
	annoKey := device.InRequestDevices[dtype]
	newAnnos := map[string]string{annoKey: encoded}
	if err := util.PatchPodAnnotations(pod, newAnnos); err != nil {
		return err
	}
	pod.Annotations[annoKey] = encoded   // <-- no nil check
	return nil
}
```

`util.PatchPodAnnotations` (`pkg/util/util.go:163`) only issues a server-side
merge patch against the Kubernetes API — it never mutates the caller's local
`pod.Annotations` map. So the nil-ness of `pod.Annotations` going into this
function is untouched by that call, and the following line is unguarded.

This is called from the kubelet-facing `Allocate()` gRPC handler
(`server.go:739`), so a panic here takes down GPU allocation for the whole
node until the plugin pod restarts.

## Why this matters

The identical write pattern appears twice more elsewhere in the codebase, and
both of those correctly guard against a nil map first:

- `pkg/device/metax/sdevice.go:110-113`
- `pkg/device/mthreads/device.go:110-112`

```go
if p.Annotations == nil {
    p.Annotations = make(map[string]string)
}
p.Annotations[key] = value
```

`patchErasedAnnotation` is the one place in the repo that skips this guard.

## Suggested fix

```go
if pod.Annotations == nil {
	pod.Annotations = make(map[string]string)
}
pod.Annotations[annoKey] = encoded
```

## Scope note

I also noticed `pkg/monitor/nvidia/cudevshr.go:202` discards the error from
`syscall.Munmap(c.data)` during stale-container cleanup in `Update()`. Low
severity (best-effort cleanup path), but happy to fix it in the same PR if
that's preferred, or leave it for a separate issue.

## Environment

- Found via manual code review, not yet reproduced against a live cluster
  (could not confirm whether `pod.Annotations` is ever actually nil by the
  time `Allocate()` runs in the current scheduler→plugin flow — the
  scheduler typically annotates pods before this point, so this may be a
  latent/defensive fix rather than an actively firing bug).

AyushSrivastava1818:
Fixes #2462 

## What

Guard `patchErasedAnnotation()` against a nil `pod.Annotations` map before
writing to it, and have `eraseNextDeviceTypeFromAnnotation()` delegate to
this new helper instead of duplicating the same unguarded write.

## Why

`util.PatchPodAnnotations()` only performs a server-side merge patch via the
Kubernetes API client — it does not mutate the local `pod.Annotations` map.
When `pod.Annotations` is `nil` at the time `patchErasedAnnotation` runs,
writing directly to `pod.Annotations[annoKey]` panics with `"assignment to
entry in nil map"`, crashing the device-plugin daemon.

`eraseNextDeviceTypeFromAnnotation` previously called
`util.PatchPodAnnotations` directly with the same gap. It didn't have its
own separate local-map write before this change — the fix here is
delegating it to the shared, guarded helper rather than fixing a second
independent bug.

This mirrors the existing nil-map guard pattern already used in
`pkg/device/metax/sdevice.go` and `pkg/device/mthreads/device.go`.

## Testing

Added to `pkg/device-plugin/nvidiadevice/nvinternal/plugin/util_test.go`:

- `TestPatchErasedAnnotation`
  - `NilAnnotationsMapInitializedWithoutPanic` — exercises `patchErasedAnnotation` against a pod with `Annotations: nil` via a fake clientset; asserts no panic, `pod.Annotations` is initialized non-nil, and the encoded value matches.
  - `ExistingAnnotationsPreserved` — asserts pre-existing annotation keys are not overwritten.
- `TestEraseNextDeviceTypeFromAnnotation`
  - `NilAnnotationsDoesNotPanic` — calls `eraseNextDeviceTypeFromAnnotation` directly with `Annotations: nil`, asserts no panic.
  - `WithAnnotationsErasesFirstContainer` — verifies the standard erasure path still works.

Verified against the full package (Linux/WSL2, Go 1.26.5, CGO enabled — not
native Windows, where third-party NVML/cgroups deps fail to build):

```
$ gofmt -l ./pkg/device-plugin/nvidiadevice/nvinternal/plugin/util.go ./pkg/device-plugin/nvidiadevice/nvinternal/plugin/util_test.go
(no output)

$ go vet ./pkg/device-plugin/nvidiadevice/nvinternal/plugin/...
(no output)

$ go test ./pkg/device-plugin/nvidiadevice/nvinternal/plugin/... -race -count=1 -v -run "TestPatchErasedAnnotation|TestEraseNextDeviceTypeFromAnnotation"
--- PASS: TestPatchErasedAnnotation (0.01s)
    --- PASS: TestPatchErasedAnnotation/NilAnnotationsMapInitializedWithoutPanic (0.01s)
    --- PASS: TestPatchErasedAnnotation/ExistingAnnotationsPreserved (0.00s)
--- PASS: TestEraseNextDeviceTypeFromAnnotation (0.00s)
    --- PASS: TestEraseNextDeviceTypeFromAnnotation/NilAnnotationsDoesNotPanic (0.00s)
    --- PASS: TestEraseNextDeviceTypeFromAnnotation/WithAnnotationsErasesFirstContainer (0.00s)
PASS
ok  	github.com/Project-HAMi/HAMi/pkg/device-plugin/nvidiadevice/nvinternal/plugin	1.068s

$ golangci-lint run ./pkg/device-plugin/nvidiadevice/nvinternal/plugin/...
0 issues.
```

Branch rebased onto `upstream/master` (`87d9795`) for a clean merge.

## AI Assistance Disclosure

The implementation and test changes in this PR were generated with
assistance from Claude. I reviewed the generated diff against the actual
source in `util.go` before submitting, and re-ran verification myself after
catching that an initial `go vet` invocation was incorrectly scoped to two
files instead of the full package.

## Out of scope

`cudevshr.go`'s `Munmap` error handling was noticed during this work but is
intentionally left out of scope for this PR.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved device annotation updates to preserve existing metadata while recording remaining device information.
  * Prevented errors when annotations are initially absent or uninitialized.
  * Ensured device-type removal correctly updates pod state and reports missing annotations as expected.
  * Persisted updated device information reliably after annotation changes.

* **Tests**
  * Added coverage for annotation creation, preservation, encoding, persistence, missing-annotation handling, and successful device-type removal.

### DIFF
--- a/pkg/device-plugin/nvidiadevice/nvinternal/plugin/util.go
+++ b/pkg/device-plugin/nvidiadevice/nvinternal/plugin/util.go
@@ -92,6 +92,21 @@ func GetNextDeviceRequest(dtype string, p corev1.Pod) (corev1.Container, device.
 	return corev1.Container{}, res, errors.New("device request not found")
 }
 
+func patchErasedAnnotation(pod *corev1.Pod, dtype string, podSingleDev device.PodSingleDevice) error {
+	klog.V(5).Infof("After erase annotation, remaining devices: %v", podSingleDev)
+	encoded := device.EncodePodSingleDevice(podSingleDev)
+	annoKey := device.InRequestDevices[dtype]
+	newAnnos := map[string]string{annoKey: encoded}
+	if err := util.PatchPodAnnotations(pod, newAnnos); err != nil {
+		return err
+	}
+	if pod.Annotations == nil {
+		pod.Annotations = make(map[string]string)
+	}
+	pod.Annotations[annoKey] = encoded
+	return nil
+}
+
 var eraseNextDeviceTypeFromAnnotation = func(dtype string, p corev1.Pod) error {
 	pdevices, err := device.DecodePodDevices(device.InRequestDevices, p.Annotations)
 	if err != nil {
@@ -115,10 +130,7 @@ var eraseNextDeviceTypeFromAnnotation = func(dtype string, p corev1.Pod) error {
 			}
 		}
 	}
-	klog.Infoln("After erase res=", res)
-	newannos := make(map[string]string)
-	newannos[device.InRequestDevices[dtype]] = device.EncodePodSingleDevice(res)
-	return util.PatchPodAnnotations(&p, newannos)
+	return patchErasedAnnotation(&p, dtype, res)
 }
 
 func GetIndexAndTypeFromUUID(uuid string) (string, int) {
--- a/pkg/device-plugin/nvidiadevice/nvinternal/plugin/util.go
+++ b/pkg/device-plugin/nvidiadevice/nvinternal/plugin/util.go
@@ -107,7 +107,6 @@ func patchErasedAnnotation(pod *corev1.Pod, dtype string, podSingleDev device.Po
 	return nil
 }
 
-var eraseNextDeviceTypeFromAnnotation = func(dtype string, p corev1.Pod) error {
 // decodePodSingleDevice decodes the pod's device allocation annotation for the
 // given device type into a PodSingleDevice slice. Each element corresponds to
 // one container (init containers first, then regular containers).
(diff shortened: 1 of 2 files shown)


FouoF:
Thanks for considering the nil-map case. On current master, patchErasedAnnotation is reached only after decodePodSingleDevice has successfully decoded a non-empty NVIDIA allocation annotation, which creates the map being written. The test constructs a Pod state that the production call chain cannot produce. With no reachable panic or old-code failure, the guard has no demonstrated project value, so we are closing this PR.

PR #2467, opened 2026-08-08
fix(scheduler): keep per-type device allocations aligned to container index

### LINKED ISSUE #2335: scheduler: multi-type pods can assign a device to the wrong container

adity1raut:
## What this PR does
Fixes container-index misalignment of per-type device allocations in `calcScoreWithOptions` (`pkg/scheduler/score.go`).

## Why
For a pod whose containers request **different** device types (e.g. container 0 requests an NVIDIA GPU, container 1 requests a second vendor's device), the "fill missing empty allocation" step padded missing per-container slots by **prepending** an empty slot to **every already-discovered device type**, not just the type the current container had just requested:

```go
for idx := range score.Devices {
    deviceType = idx
    for len(score.Devices[idx]) <= ctrid {
        ...
    }
}
```

Once a second device type is discovered at container N, this loop also revisits the *already-correctly-aligned* first type and prepends to it, shifting an earlier container's real allocation into a later container's slot. `PatchAnnotations` → `EncodePodSingleDevice` writes that corrupted list into the pod's annotations verbatim, and the device plugin's `Allocate()` reads it back by container index — so a container can end up being granted a device that the scheduler actually assigned to a *different* container.

A secondary flaw: the no-device-container path (`sums == 0`) only padded one arbitrarily-chosen type (`deviceType`, whatever the last map-iteration happened to set), not every discovered type, so interleaving a no-device container between two different device types also broke alignment.

This was previously reported as #2335, with a fix proposed in #2337 — but #2337 was closed without being merged, so the bug is still present on current master (`3616313`). I hit the same issue independently while reviewing the scheduler package and am resubmitting a fix with a regression test.

## What changed
- Track the device type(s) the *current* container actually requested (from `n`, via each request's `.Type` field — not the map's annotation-name keys).
- Only that type gets the leading-pad treatment (to keep the just-appended real entry aligned at index `ctrid`); every other already-discovered type gets a single trailing empty slot instead of being shifted.
- The no-device-container path now pads *every* discovered type with a trailing empty slot, not just one.
- Added `Test_calcScore_MultiTypeContainerAlignment` in `pkg/scheduler/score_test.go`, covering a 2-container pod where container 0 requests mock type A and container 1 requests mock type B, asserting each type's allocation stays at the correct container index.

## Verification
- `go build ./...`
- `go test ./pkg/scheduler/... -short --race -count=1` (all pass, including the new regression test, which fails against the pre-fix code)
- `go vet ./pkg/scheduler/...`
- `golangci-lint run ./pkg/scheduler/...` (0 issues)
- `hack/verify-import-aliases.sh` (pass)

Fixes #2335

## AI assistance disclosure
This bug was found and the fix + regression test were drafted with AI assistance (Claude Code) under my review. I read and verified the root cause, the fix, and the test, and take responsibility for their correctness.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved device allocation scoring across containers requesting different device types.
  * Ensured containers without device requests and those requesting specific types retain correctly aligned allocation slots.
  * Prevented incorrect device placement and scoring outcomes in multi-device-type scheduling scenarios.

* **Tests**
  * Added coverage for multi-type container allocation alignment and successful node scoring.

### DIFF
--- a/pkg/scheduler/score.go
+++ b/pkg/scheduler/score.go
@@ -135,7 +135,6 @@ func (s *Scheduler) calcScoreWithOptions(nodes *map[string]*NodeUsage, resourceR
 			// Assume the node is a fit by default. This handles pods with no device
 			// requests, which should be schedulable on any node.
 			ctrfit := true
-			deviceType := ""
 			//This loop is for different container request
 			for ctrid, n := range resourceReqs {
 				sums := 0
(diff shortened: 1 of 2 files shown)


FouoF:
There still lacking one pod requesting different vendors device in real world. Close for low priority.

Issue #2264, opened 2026-08-01
HAMi scheduler selects a single node during Filter, preventing kube-scheduler scoring plugins from influencing placement


blackdragoon26:
**What happened**:
HAMi performs both node feasibility filtering and node scoring inside the scheduler extender's `Filter` operation.

When multiple nodes can satisfy a Pod's device request, HAMi:

1. Calculates an internal score for every feasible node.
2. Sorts the feasible nodes.
3. Selects the highest-scoring node.
4. Patches the selected node and device allocation into the Pod annotations.
5. Returns only that single node to kube-scheduler.

The relevant code is:

https://github.com/Project-HAMi/HAMi/blob/c7891ded87b71e402a36f5f29ae7d245e0de70b2/pkg/scheduler/scheduler.go#L940-L987

```go
nodeScores, err := s.calcScore(nodeUsage, resourceReqs, args.Pod, failedNodes)
...
sort.Sort(nodeScores)
m := (*nodeScores).NodeList[len((*nodeScores).NodeList)-1]
...
res := extenderv1.ExtenderFilterResult{
    NodeNames: &[]string{m.NodeID},
}
return &res, nil
```
The generated kube-scheduler configuration registers filterVerb and bindVerb, but does not register a prioritizeVerb:
https://github.com/Project-HAMi/HAMi/blob/c7891ded87b71e402a36f5f29ae7d245e0de70b2/charts/hami/templates/scheduler/configmap.yaml#L37-L43
```yaml
filterVerb: filter
bindVerb: bind
nodeCacheCapable: true
weight: 1
```
Consequently, when kube-scheduler reaches its normal scoring phase, only the node already selected by HAMi remains. Other feasible nodes cannot be selected by Kubernetes scoring plugins.

This makes soft Kubernetes placement preferences ineffective for HAMi-managed workloads when they disagree with HAMi's internal node score, including:
- `nodeAffinity.preferredDuringSchedulingIgnoredDuringExecution`
- Preferred Pod affinity and anti-affinity
- `topologySpreadConstraints` using `whenUnsatisfiable: ScheduleAnyway`
- `ImageLocality`
- Resource-balancing Score plugins
- Custom scheduler Score plugins configured by cluster administrators

Hard scheduling constraints are not affected because kube-scheduler applies those during filtering. The problem specifically concerns scoring and soft placement preferences.
This also differs from HAMi's documented scheduler design, which states that Filter returns the available nodes and Score selects the final node:
[...]

PR #2470, opened 2026-08-08
fix(scheduler): serialize Filter device selection to prevent double GPU allocation

### LINKED ISSUE #2232: 并发创建pod情况下，存在pod 中卡id的注解和实际pod中使用卡的id不一致问题
**What happened**:
多并发创建pod,如多副本deployment
多并发创建pod的情况下，卡id注解应和pod中实际分配的一致
多次创建多副本deployment

**Environment**:
- HAMi version:2.8.3
- nvidia driver or other AI device driver version:580
- Others:

Image

yxxhero:
**What type of PR is this?**
/kind bug

Fixes #2232 — *并发创建 pod 情况下，存在 pod 中卡 id 的注解和实际 pod 中使用卡的 id 不一致问题* (under concurrent pod creation, the GPU id annotated on a pod is inconsistent with the GPU id the pod actually uses).

The reported consequence is duplicate card allocation: the shared card gets over-allocated (causing OOM) while other idle cards cannot be handed out.

### Root cause

The scheduler extender is served over HTTP with **one goroutine per request** and provides no cross-request serialization guarantee. In `Filter`, the device-selection critical section was not atomic:

1. read node usage (`getNodesUsage`)
2. pick the best node/device (`calcScore`)
3. commit the reservation to the pod cache (`podManager.AddPod`)
4. publish the allocation onto the pod annotation (`PatchPodAnnotations`)

Under concurrent pod creation (e.g. a multi-replica Deployment), two `Filter` calls can both read the same free GPU in step 1 *before* either commits it in step 3, so both pods are annotated with the **same card id**.

In the normal kube-scheduler flow scheduling cycles are already serialized, which is why the problem only surfaces under concurrent/re-entrant extender requests.

### Fix

Add a `filterLock` and move the in-memory critical section (steps 1–3) into a `selectAndCommitDevice` helper that owns the lock through a single `defer Unlock`, so two pods can never reserve the same device. The lock is released before step 4 (`PatchPodAnnotations`), so concurrent Filters are never blocked on the apiserver round trip; the on-failure rollback runs outside the lock and is safe because the pod/quota managers have their own locks. The simulation path (`filterSimulation`) does not reserve devices and is unchanged.

This keeps the change minimal — it does **not** move reservation from `Filter` to `Bind` (the larger redesign from #2273 is tracked via the #2264 design discussion).

### Test

- `Test_Filter_ConcurrentNoDoubleAllocation` fires 8 pods at `Filter` simultaneously and asserts each is annotated with a **distinct** device id. Without the lock it reliably reports `device device-7 double-allocated to pod pod-0 and pod pod-2`; with the lock it passes (`-race`).
- `Test_Filter_RollbackOnPatchFailure` verifies the cache reservation is rolled back when the annotation patch fails.

### Validation

Scoped to the scheduler extender, so per CONTRIBUTING's hardware-validation gate this is validated with unit tests (including a `-race` regression test) rather than physical GPU hardware. `make verify` (license, import-aliases, golangci-lint) and `go test -race ./pkg/scheduler/...` pass locally.

Fixes #2232

- A single `defer Unlock` in the helper means any early return added there in future can never leak the lock.
- A per-pod-UID lock was considered but not added: kube-scheduler runs one scheduling cycle at a time, so a pod UID is not filtered concurrently; the reproduced race is between *different* pods, which the global `filterLock` closes.

```release-note
Fixed a race in the scheduler extender's Filter that could assign the same GPU to multiple pods created concurrently (e.g. a multi-replica Deployment), causing duplicate card allocation and OOM. Device allocation in Filter is now serialized.
```

**AI assistance disclosure:**
This PR was written with AI assistance for codebase analysis, implementation and test generation. I reviewed, built and tested the change myself (`make verify`, `go test -race`) and can explain the locking semantics. Per CONTRIBUTING.md, disclosure belongs in the PR description only — no AI co-author trailers are used.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented concurrent scheduling operations from assigning the same GPU to multiple pods.
  * Improved consistency of device usage tracking, scoring, reservations, cache updates, and pod annotations during live scheduling.
  * Released tentative GPU reservations when pod annotation updates fail.
  * Reduced the risk of conflicting GPU allocations when multiple pods are scheduled simultaneously.

* **Tests**
  * Added coverage verifying distinct GPU assignments during concurrent scheduling.
  * Added coverage confirming failed annotation updates return an error and remove tentative reservations.

### DIFF
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -78,6 +78,10 @@ type Scheduler struct {
 
 	lock   sync.RWMutex
 	synced bool
+
+	// filterLock serializes device selection in Filter so concurrent requests
+	// cannot reserve the same device (issue #2232).
+	filterLock sync.Mutex
 }
 
 func NewScheduler() *Scheduler {
@@ -953,31 +957,70 @@ func (s *Scheduler) Filter(args extenderv1.ExtenderArgs) (*extenderv1.ExtenderFi
 		"pod", klog.KObj(args.Pod),
 		"reason", "request does not contain full nodes",
 		"nodeNamesLen", nodeNamesLen(args.NodeNames))
+	// selectAndCommitDevice holds filterLock during selection; the annotation
+	// patch below runs outside it (issue #2232).
+	selection, err := s.selectAndCommitDevice(args, resourceReqs)
+	if err != nil {
+		s.recordScheduleFilterResultEvent(args.Pod, EventReasonFilteringFailed, "", err)
+		return nil, err
+	}
+	if selection.chosen == nil {
+		s.recordScheduleFilterResultEvent(args.Pod, EventReasonFilteringFailed, "", fmt.Errorf("no available node, %d nodes do not meet", len(*args.NodeNames)))
+		return &extenderv1.ExtenderFilterResult{
+			FailedNodes: selection.failedNodes,
+		}, nil
+	}
+	m := selection.chosen
+	// Patch the annotation outside the lock; roll back the cache on failure.
+	if err = util.PatchPodAnnotations(args.Pod, selection.annotations); err != nil {
+		s.recordScheduleFilterResultEvent(args.Pod, EventReasonFilteringFailed, "", err)
+		if selection.added {
+			s.quotaManager.RmUsage(args.Pod, m.Devices)
+		}
+		s.podManager.DelPod(args.Pod)
+		return nil, err
+	}
+	successMsg := genSuccessMsg(len(*args.NodeNames), m.NodeID, selection.nodeList)
+	s.recordScheduleFilterResultEvent(args.Pod, EventReasonFilteringSucceed, successMsg, nil)
+	res := extenderv1.ExtenderFilterResult{NodeNames: &[]string{m.NodeID}}
+	return &res, nil
+}
+
+// filterSelection holds a Filter reservation; chosen is nil when no node fits
+// (failedNodes then carries the per-node reasons).
+type filterSelection struct {
+	chosen      *policy.NodeScore
+	annotations map[string]string
+	added       bool
+	nodeList    []*policy.NodeScore
+	failedNodes map[string]string
+}
+
+// selectAndCommitDevice selects a device and commits it to the cache under
+// filterLock. It does not publish the pod annotation; the caller patches it.
+func (s *Scheduler) selectAndCommitDevice(args extenderv1.ExtenderArgs, resourceReqs device.PodDeviceRequests) (*filterSelection, error) {
+	s.filterLock.Lock()
+	defer s.filterLock.Unlock()
+
+	selection := &filterSelection{}
 	if pi, ok := s.podManager.TakeAndDeletePod(args.Pod); ok {
 		s.quotaManager.RmUsage(args.Pod, pi.Devices)
 	}
 	nodeUsage, _, failedNodes, err := s.getNodesUsage(args.NodeNames, args.Pod)
 	if err != nil {
-		s.recordScheduleFilterResultEvent(args.Pod, EventReasonFilteringFailed, "", err)
 		return nil, err
 	}
 	if len(failedNodes) != 0 {
-		klog.V(5).InfoS("Nodes failed during usage retrieval",
-			"nodes", failedNodes)
+		klog.V(5).InfoS("Nodes failed during usage retrieval", "nodes", failedNodes)
 	}
 	nodeScores, err := s.calcScore(nodeUsage, resourceReqs, args.Pod, failedNodes)
 	if err != nil {
-		err := fmt.Errorf("calcScore failed %v for pod %v", err, args.Pod.Name)
-		s.recordScheduleFilterResultEvent(args.Pod, EventReasonFilteringFailed, "", err)
-		return nil, err
+		return nil, fmt.Errorf("calcScore failed %v for pod %v", err, args.Pod.Name)
 	}
 	if len((*nodeScores).NodeList) == 0 {
-		klog.V(4).InfoS("No available nodes meet the required scores",
-			"pod", args.Pod.Name)
-		s.recordScheduleFilterResultEvent(args.Pod, EventReasonFilteringFailed, "", fmt.Errorf("no available node, %d nodes do not meet", len(*args.NodeNames)))
-		return &extenderv1.ExtenderFilterResult{
-			FailedNodes: failedNodes,
-		}, nil
+		klog.V(4).InfoS("No available nodes meet the required scores", "pod", args.Pod.Name)
+		selection.failedNodes = failedNodes
+		return selection, nil
 	}
 	klog.V(4).Infoln("nodeScores_len=", len((*nodeScores).NodeList))
 	sort.Sort(nodeScores)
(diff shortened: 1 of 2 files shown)


maishivamhoo123:
@yxxhero please read the contribution.md before raising a pr https://github.com/Project-HAMi/HAMi/blob/master/CONTRIBUTING.md   and please disclose if you use any AI  Assistance .

pkg/scheduler/scheduler.go
```diff
@@ -78,6 +78,16 @@ type Scheduler struct {
 
 	lock   sync.RWMutex
 	synced bool
+
+	// filterLock serializes the device-selection critical section in Filter
```
maishivamhoo123: Please write short comments . otherwise the comments will be more than the code. and please read https://github.com/Project-HAMi/HAMi/blob/master/CONTRIBUTING.md before raising the pr.

pkg/scheduler/scheduler.go
```diff
@@ -953,31 +963,81 @@ func (s *Scheduler) Filter(args extenderv1.ExtenderArgs) (*extenderv1.ExtenderFi
 		"pod", klog.KObj(args.Pod),
 		"reason", "request does not contain full nodes",
 		"nodeNamesLen", nodeNamesLen(args.NodeNames))
+	// Device selection and cache commit are serialized inside a helper that
```
maishivamhoo123: here also.

pkg/scheduler/scheduler.go
yxxhero: Done — trimmed to one line:
`// filterLock serializes device selection in Filter so concurrent requests cannot reserve the same device (issue #2232).`

pkg/scheduler/scheduler.go
yxxhero: Done here too — the call-site and `selectAndCommitDevice` comments are now 1–2 lines each.

pkg/scheduler/scheduler.go
```diff
@@ -953,31 +963,81 @@ func (s *Scheduler) Filter(args extenderv1.ExtenderArgs) (*extenderv1.ExtenderFi
...
+	m := selection.chosen
+	// Publish the allocation onto the pod annotation. This is a network call
+	// done outside the lock; on failure roll back the cache reservation.
+	if err = util.PatchPodAnnotations(args.Pod, selection.annotations); err != nil {
+		s.recordScheduleFilterResultEvent(args.Pod, EventReasonFilteringFailed, "", err)
+		if selection.added {
+			s.quotaManager.RmUsage(args.Pod, m.Devices)
+		}
+		s.podManager.DelPod(args.Pod)
+		return nil, err
```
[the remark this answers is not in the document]
yxxhero: Intentionally not adding a per-pod lock. kube-scheduler runs one scheduling cycle at a time, so a pod UID is never filtered concurrently; the reproduced #2232 race is between *different* pods, which the global `filterLock` closes. A per-pod lock map would add unbounded per-UID state plus cleanup wiring for a scenario that does not arise in practice. The rollback path itself is now covered by `Test_Filter_RollbackOnPatchFailure`.

pkg/scheduler/config/config.go
```diff
@@ -278,24 +278,25 @@ func validateConfig(config *Config) error {
 	return fmt.Errorf("all configurations are empty")
 }
 
-func InitDevices() {
+func InitDevices() error {
 	if len(device.DevicesMap) > 0 {
 		klog.Info("Devices are already initialized, skipping initialization")
-		return
+		return nil
```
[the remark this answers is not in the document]
yxxhero: Out of scope for this PR — `pkg/scheduler/config/config.go` is not touched here. Happy to open a separate PR for the `InitDevicesWithConfig` partial-registration cleanup if the maintainers want it.

yxxhero:
@maishivamhoo123 thanks — I've read CONTRIBUTING.md.

- **Hardware validation gate:** this change is scoped to the scheduler extender, so per the gate it's validated with unit tests (incl. a `-race` regression test) rather than physical GPUs. `make verify` (license, import-aliases, golangci-lint) and `go test -race ./pkg/scheduler/...` pass locally.
- **AI assistance:** disclosed in the PR description — AI was used for codebase analysis, implementation and test generation; I reviewed, built and tested the change myself and can explain the locking semantics. No AI co-author trailers are used (gate 5).
- **Comments:** shortened the `filterLock` / `Filter` / `selectAndCommitDevice` comments to 1–2 lines each (replies on the two inline threads).

FouoF:
We can fix it or request the user use k8s >= 1.23 which is old enough. Need more discussion about lt.

mesutoezdil:

spencercjh:
Independent black-box evidence supports this fix: the race is straightforward to reproduce against the real Extender HTTP path, not only with a fake-client test.

In an isolated k3d cluster I registered one mock GPU with `count: 10`, created 12 distinct GPU Pods pinned to that physical UUID, and added a scheduling gate so kube-scheduler could not serialize or bind them first. I then used each Pod's actual Kubernetes JSON to issue 12 concurrent `/filter` requests, each with only that node in `nodeNames`.

All 12 requests succeeded and all 12 Pods received the same physical-device allocation annotation. The Scheduler's live metric afterwards was:

```text
hami_gpu_shared_count{device_uuid="GPU-MOCK-AGENT-0"} 12
```

This exceeds the mock card's `count: 10`; the associated namespace quota was also 12 × the requested memory/core amount. That is the same read-score-then-reserve TOCTOU window described in this PR. A global `filterLock` around selection and in-memory commit is therefore a directly justified minimal correction. The scheduling gate is only a test harness control; it prevents normal kube-scheduler serialization from hiding concurrent/re-entrant Extender behavior.

yxxhero:
@spencercjh so why close this PR?

spencercjh:
@yxxhero I think the HAMi community is currently experiencing a significant impact from AI, and several owners are finding it quite unbearable. Maybe you’ve been flagged as spam.

yxxhero:
@FouoF @mesutoezdil Thanks for the thorough validation and the detailed black-box test! I want to clarify that the approach in this PR wasn’t just AI-generated boilerplate. It was the result of careful deliberation regarding the concurrency issues in the Extender path. The filterLock was specifically chosen as a deliberate, minimal fix to address the TOCTOU window, and I’m glad to see it holds up under your rigorous testing.

yxxhero:
@spencercjh thanks so much.

mesutoezdil:
> @FouoF @mesutoezdil Thanks for the thorough validation and the detailed black-box test! I want to clarify that the approach in this PR wasn’t just AI-generated boilerplate. It was the result of careful deliberation regarding the concurrency issues in the Extender path. The filterLock was specifically chosen as a deliberate, minimal fix to address the TOCTOU window, and I’m glad to see it holds up under your rigorous testing.

thx for your message.
we have some rules. one of them is: 

"https://github.com/Project-HAMi/HAMi/blob/master/CONTRIBUTING.md#contribution-gates
4. Review replies. The reply you post must be written by you and must address the specific point raised. Verbatim or canned AI replies, or replies that do not engage the comment, lead to the PR being closed."

This rule is in place so that maintainers can communicate with a person and receive a direct, short, and clear answer to the specific question.

Bcs no one's native language is English, and no one has time to read long llm texts.

as i remember the reason was your this answer: "https://github.com/Project-HAMi/HAMi/pull/2470#issuecomment-5225560473"

would you like to open new PR?

yxxhero:
@spencercjh hi. any better ideas for this issue? if so please open an new PR or I open new PR?  thanks so much.

mesutoezdil:
> This is not specific to an older Kubernetes release. The reproducer makes concurrent real HTTP calls to the Scheduler Extender Filter endpoint after holding the Pods behind scheduling gates; the race is entirely inside the Scheduler's own read-current-usage → choose-device → add-temporary-allocation sequence.
> 
> Kubernetes 1.23 and later do not provide an atomic transaction or a cross-request serialization guarantee for an Extender's Filter handlers. Concurrent scheduling requests can still reach those handlers, so upgrading Kubernetes alone cannot close this window. The Scheduler must make the capacity check and temporary reservation atomic (for example, with the proposed Filter lock or an equivalent reservation primitive), and retain a concurrent regression test.

I don't know why you're sharing long LLM responses as answers.
Again, reminder..
Answers must be written by human being.
4. Review replies. The reply you post must be written by you and must address the specific point raised. Verbatim or canned AI replies, or replies that do not engage the comment, lead to the PR being closed.

spencercjh:
@mesutoezdil I'm sorry for taking up your time. I'll follow the community guidelines.

PR #2474, opened 2026-08-08
fix(util): deterministic pod selection in GetPendingPod and NVIDIA Allocate (fixes #2188)

### LINKED ISSUE #2188: bug(util): GetPendingPod non-deterministically selects pending pods during device allocation fallback
# problem statement 

In `pkg/util/util.go`, the `GetPendingPod` function is called by the device plugin server (`NvidiaDevicePlugin.Allocate()` 
and `GetPreferredAllocation()`) to identify which pod Kubelet is currently allocating GPU devices for on a given node:
```go
// util.go lines 75–117
func GetPendingPod(ctx context.Context, node string) (*corev1.Pod, error) {
	pod, err := GetAllocatePodByNode(ctx, node)
	if err != nil {
		return nil, err
	}
	if pod != nil {
		return pod, nil
	}
	// Fallback: list all pods for this node
	selector := fmt.Sprintf("spec.nodeName=%s", node)
	podListOptions := metav1.ListOptions{
		FieldSelector: selector,
	}
	podlist, err := client.GetClient().CoreV1().Pods("").List(ctx, podListOptions)
	if err != nil {
		return nil, err
	}

	for _, p := range podlist.Items {
		if p.Status.Phase != corev1.PodPending {
			continue
		}
		if _, ok := p.Annotations[BindTimeAnnotations]; !ok {
			continue
		}
		if phase, ok := p.Annotations[DeviceBindPhase]; !ok {
			continue
		} else {
			if phase != DeviceBindAllocating && phase != DeviceBindSuccess {
				continue
			}
		}

		if n, ok := p.Annotations[AssignedNodeAnnotations]; !ok {
			continue
		} else {
			if strings.Compare(n, node) == 0 {
				return &p, nil
			}
		}
	}
	return nil, fmt.Errorf("no binding pod found on node %s", node)
}

harshakumar25:
## What's the problem?

When multiple GPU pods are pending on the same node, `GetPendingPod()` was
returning whichever pod happened to come first in the Kubernetes API list
response. Since Kubernetes doesn't guarantee list ordering, the device plugin
could randomly pick the wrong pod and assign GPUs to it.

## What changed?

**Smarter pod selection in the fallback path:**
- Instead of grabbing the first match, we now collect all candidates and pick
  deterministically — preferring pods in `allocating` phase over `success`,
  then newest bind-time, then alphabetical name as a final tiebreaker.

**UUID-based pod identity (primary path):**
- Before even using the fallback, `Allocate()` now tries to match the pod
  directly by the GPU UUIDs kubelet is requesting. This is unambiguous since
  physical GPU UUIDs are globally unique.

**`GetPreferredAllocation()` gets the same fix:**
- Previously it always used the unreliable fallback. Now it tries UUID matching
  first, just like `Allocate()`.

**Allocation serialized with a mutex:**
- A new `allocateMu` lock prevents two concurrent `Allocate()` calls for the
  same pod's containers from racing on the annotation.

## Tests

Added 5 new test cases covering: multi-container annotation parsing, phase
preference, bind-time ordering, equal-timestamp tiebreaker, and UUID-based
matching. All packages pass with `-race`.

```bash
go test ./pkg/util/... ./pkg/device-plugin/nvidiadevice/... -race
# All PASS ✅


## Summary by CodeRabbit

* **Bug Fixes**
  * Improved GPU allocation reliability by matching pending workloads to requested device IDs.
  * Added fallback handling when device-based matching is unavailable.
  * Prevented allocation races during pod lookup and annotation updates.
  * Made pod selection deterministic, prioritizing allocating workloads and the most recent successful allocation.

* **Tests**
  * Expanded coverage for device matching, multi-container annotations, and deterministic selection.

### DIFF
--- a/pkg/device-plugin/nvidiadevice/nvinternal/plugin/server.go
+++ b/pkg/device-plugin/nvidiadevice/nvinternal/plugin/server.go
@@ -100,6 +100,7 @@ type NvidiaDevicePlugin struct {
 	schedulerConfig      nvidia.NvidiaConfig
 
 	applyMutex                 sync.Mutex
+	allocateMu                 sync.Mutex // serialises pod-lookup + annotation-erase in Allocate()
 	disableHealthChecks        chan bool
 	ackDisableHealthChecks     chan bool
 	disableWatchAndRegister    chan bool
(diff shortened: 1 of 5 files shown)


FouoF:
kubelet can not para real GPU UUID, so the fix not work.

PR #2479, opened 2026-08-08
fix(cambricon): release node lock with Patch instead of Update

### LINKED ISSUE #2475: Cambricon ReleaseNodeLock calls Nodes().Update(), which the scheduler ServiceAccount is not permitted to do
## What happened

`pkg/device/cambricon` has its own node-lock implementation, separate from `pkg/util/nodelock`. Acquire uses `Patch`, release uses `Update`:

https://github.com/Project-HAMi/HAMi/blob/3616313cb4d30841a6162421969d7ab9463931aa/pkg/device/cambricon/device.go#L158-L180

The scheduler ServiceAccount cannot update nodes. The chart grants `get, list, patch, watch`:

https://github.com/Project-HAMi/HAMi/blob/3616313cb4d30841a6162421969d7ab9463931aa/charts/hami/templates/scheduler/clusterrole.yaml#L15-L17

and the two other bindings it carries do not add it either: `system:volume-scheduler` covers PVs, and `system:kube-scheduler` grants nodes only `get, list, watch`. That was deliberate, `update` was removed from the node rule in b7a952f as the last step of #1156, once #1192 moved `pkg/util/nodelock` to `Patch`. The cambricon backend keeps its own lock and was missed.

So release fails with 403 five times, gives up, and the annotation stays. Then `LockNode` for the next MLU pod sees the expired lock, tries to release it first, gets the same 403, and returns the error before it ever reaches `setNodeLock`.

Actual: `cambricon.com/dsmlu.lock` is never removed. After the 2 minute expiry the node is permanently unschedulable for MLU pods until an operator deletes the annotation by hand.

Expected: release works, the same way it does for every other backend through `pkg/util/nodelock`.

## How to reproduce it

No MLU needed, the deny is what RBAC would return. Drop into `pkg/device/cambricon`:

```go
func TestReleaseNodeLockUnderSchedulerRBAC(t *testing.T) {
	ctx := context.Background()
	dev := InitMLUDevice(CambriconConfig{
		ResourceCountName:  "cambricon.com/mlu",
		ResourceMemoryName: "cambricon.com/mlu.smlu.vmemory",
		ResourceCoreName:   "cambricon.com/mlu.smlu.vcore",
	})
	node := &corev1.Node{ObjectMeta: metav1.ObjectMeta{Name: "node-1", Annotations: map[string]string{}}}
	pod := &corev1.Pod{
		ObjectMeta: metav1.ObjectMeta{Name: "p", Namespace: "default"},
		Spec: corev1.PodSpec{Containers: []corev1.Container{{Name: "c",
			Resources: corev1.ResourceRequirements{Limits: corev1.ResourceList{
				corev1.ResourceName(MLUResourceCount): resource.MustParse("1"),
			}}}}},
	}

	cs := fake.NewClientset(node)
	cs.PrependReactor("update", "nodes", func(a k8stesting.Action) (bool, runtime.Object, error) {
		return true, nil, apierrors.NewForbidden(schema.GroupResource{Resource: "nodes"}, "node-1",
			fmt.Errorf(`User "system:serviceaccount:kube-system:hami-scheduler" cannot update resource "nodes"`))
	})
	client.KubeClient = cs
	t.Cleanup(func() { client.KubeClient = nil })

	if err := dev.LockNode(node, pod); err != nil {
		t.Fatalf("LockNode: %v", err)
	}
	got, _ := cs.CoreV1().Nodes().Get(ctx, "node-1", metav1.GetOptions{})
	t.Logf("after LockNode, annotation = %q", got.Annotations[DsmluLockTime])
	t.Logf("ReleaseNodeLock = %v", dev.ReleaseNodeLock(got, pod))
	got, _ = cs.CoreV1().Nodes().Get(ctx, "node-1", metav1.GetOptions{})
	t.Logf("annotation still = %q", got.Annotations[DsmluLockTime])

	expired := got.DeepCopy()
	expired.Annotations[DsmluLockTime] = time.Now().Add(-3 * time.Minute).Format(time.RFC3339)
	t.Logf("LockNode after expiry = %v", dev.LockNode(expired, pod))
}
```

Output on `3616313`:

```
after LockNode, annotation = "2026-08-08T10:39:52Z"
ReleaseNodeLock = releaseNodeLock exceeds retry count 5
annotation still = "2026-08-08T10:39:52Z"
LockNode after expiry = releaseNodeLock exceeds retry count 5
```

## Anything else we need to know

Found while auditing the scheduler's API calls against the ClusterRole. `Nodes().Update()` appears in exactly one place in the tree, this one. Not reproduced on real MLU hardware, I do not have one, and the existing unit tests do not catch it because the fake clientset does not enforce RBAC.

The fix that matches what the project already decided is to make this use `Patch`, the way #1192 did for `pkg/util/nodelock`, rather than adding the verb back to the chart.

Worth coordinating with #2329, which rewrites this same function for three other bugs and, as far as I can tell from the thread, keeps `Update`.

## Environment

- HAMi version: `master @ 3616313cb4d30841a6162421969d7ab9463931aa`
- Chart: `charts/hami` at the same commit
- Reproduced with unit tests, no accelerator hardware

nabrahma:
**What type of PR is this?**

/kind bug

Fixes #2475

### Summary

`ReleaseNodeLock` in the cambricon backend cleared `cambricon.com/dsmlu.lock` with `Nodes().Update()`. The hami-scheduler ServiceAccount is granted `get, list, patch, watch` on nodes and not `update`, so the call is rejected, the retry loop burns its five attempts and the annotation stays on the node.

**Impact:**

- The lock is never released, so it survives past the 2 minute expiry.
- `LockNode` releases before re-locking once expired. It hits the same rejection and returns before reaching `setNodeLock`, so the node stops accepting MLU pods.
- Recovery needs an operator to delete the annotation by hand.

### Fix

Release with a strategic merge patch setting the annotation to null. `setNodeLock` a few lines above already patches, and `pkg/util/nodelock` does the same, so this brings the pair in line with the rest of the tree. A merge patch also carries no resourceVersion, so the retry loop retries instead of replaying one stale object.

### Test Plan

- `Test_ReleaseNodeLockUsesPatch` drives the real path with a fake client that rejects `update` on nodes the way RBAC does, then asserts the annotation is gone from the apiserver and the node can be locked again. Fails on master, passes here.
- The existing `Test_ReleaseNodeLock` only covers the two early returns and never reaches the client, which is why this went unnoticed.
- `go test ./pkg/device/... -short --race -count=1` passes
- `go test ./pkg/scheduler/... -short --race -count=1` passes
- `go vet`, `gofmt`, `goimports -local` clean
- Not run on MLU hardware, I do not have any. The failing path is scheduler side.

### Notes for the reviewer

Overlaps #2329, which rewrites this function for the three causes in #2251. Different root cause, so I kept the diff to the two call sites. Happy to rebase once that lands.

I left the `delete` on the caller's node alone even though it is the informer race from #2251, since #2329 already covers it and I did not want two PRs editing the same line.

### AI Assistance Disclosure

```release-note
Fix Cambricon MLU node locks never being released, which left nodes unable to schedule MLU pods until the lock annotation was removed manually.
```

## Summary by CodeRabbit

* **Tests**
  * Added coverage verifying that device node locks can be released through a patch operation when standard node updates are unavailable.
  * Confirmed lock annotations are removed correctly and that the node can be locked again afterward.

### DIFF
--- a/pkg/device/cambricon/device_test.go
+++ b/pkg/device/cambricon/device_test.go
@@ -1403,3 +1403,52 @@ func TestDevices_AddResourceUsage(t *testing.T) {
 		})
 	}
 }
+
+// Test_ReleaseNodeLockUsesPatch pins the verb used to clear the lock. The
+// scheduler ServiceAccount holds get/list/patch/watch on nodes and not update,
+// so a release that reaches for Update is rejected and the lock is never
+// cleared, leaving the node unschedulable for MLU past the expiry. The fake
+// clientset does not enforce RBAC, so the deny has to be injected.
+func Test_ReleaseNodeLockUsesPatch(t *testing.T) {
+	dev := InitMLUDevice(CambriconConfig{
+		ResourceCountName:  "cambricon.com/mlu",
+		ResourceMemoryName: "cambricon.com/mlu.smlu.vmemory",
+		ResourceCoreName:   "cambricon.com/mlu.smlu.vcore",
+	})
+
+	ctx := context.Background()
+	node := &corev1.Node{
+		ObjectMeta: metav1.ObjectMeta{Name: "node-verb-01", Annotations: map[string]string{}},
+	}
+	pod := &corev1.Pod{
+		ObjectMeta: metav1.ObjectMeta{Name: "pod-verb-01", Namespace: "default"},
+		Spec: corev1.PodSpec{Containers: []corev1.Container{{
+			Name: "ctr",
+			Resources: corev1.ResourceRequirements{Limits: corev1.ResourceList{
+				corev1.ResourceName(MLUResourceCount): resource.MustParse("1"),
+			}},
+		}}},
+	}
+
+	clientset := fake.NewClientset(node)
+	clientset.PrependReactor("update", "nodes", func(action k8stesting.Action) (bool, runtime.Object, error) {
+		return true, nil, apierrors.NewForbidden(
+			schema.GroupResource{Resource: "nodes"}, node.Name,
+			fmt.Errorf(`cannot update resource "nodes" in API group "" at the cluster scope`))
+	})
+	client.KubeClient = clientset
+	t.Cleanup(func() { client.KubeClient = nil })
+
+	assert.NoError(t, dev.LockNode(node, pod))
+	locked, err := clientset.CoreV1().Nodes().Get(ctx, node.Name, metav1.GetOptions{})
+	assert.NoError(t, err)
+	assert.NotEmpty(t, locked.Annotations[DsmluLockTime], "setNodeLock should have recorded the lock")
+
+	assert.NoError(t, dev.ReleaseNodeLock(locked, pod), "release must not need the update verb")
+
+	released, err := clientset.CoreV1().Nodes().Get(ctx, node.Name, metav1.GetOptions{})
+	assert.NoError(t, err)
+	assert.NotContains(t, released.Annotations, DsmluLockTime, "lock annotation must be gone from the apiserver")
+
+	assert.NoError(t, dev.LockNode(released, pod), "node should be lockable again after release")
+}


pkg/device/cambricon/device.go
```diff
@@ -164,13 +164,18 @@ func (dev *CambriconDevices) ReleaseNodeLock(n *corev1.Node, p *corev1.Pod) erro
 		return nil
 	}
 
-	newNode := n.DeepCopy()
-	delete(newNode.Annotations, DsmluLockTime)
-	_, err := client.GetClient().CoreV1().Nodes().Update(context.Background(), newNode, metav1.UpdateOptions{})
+	ctx := context.Background()
+	// Patch, not Update: the scheduler ServiceAccount is granted only
```
archlitchi: please remove these comments

nabrahma:
Rebased on master. #2329 landed the same fix in the meantime, so the code change is gone and only the regression test remains. It denies the update verb with a reactor, which no existing test does, so it would catch an implementation that reached for Update while keeping the retry structure. Comments removed as requested. Happy to close this if the test isn't worth carrying.

PR #2487, opened 2026-08-08
fix(scheduler): serialize pod quota updates to avoid double counting

### LINKED ISSUE #2478: Race Condition Between Filter() and onAddPod() Causes Double Counting
| Field | Details |
|-------|---------|
| **Location** | `pkg/scheduler/scheduler.go` |
| **Description** | In `Filter()`, the pod is deleted from the pod manager (line 956), then device usage is recalculated (line 959), then the pod is re-added (line 998). Meanwhile, the pod informer's `onAddPod()` (line 138) can fire concurrently and also call `s.podManager.AddPod()`. Since the PodManager uses UID-based keys, if `onAddPod` fires between line 956 and line 998, the pod could be added twice with different device allocations, leading to double-counted resource usage in the quota manager. |
| **Reason** | No mutex protects the Filter → delete → recalculate → re-add cycle against informer callbacks. The PodManager has its own internal lock, but the *sequence* of operations across Filter and informer is not atomic. |

```mermaid
sequenceDiagram
    participant Filter as Filter()
    participant Informer as onAddPod()
    participant PM as PodManager
    participant QM as QuotaManager
    Filter->>PM: TakeAndDeletePod(pod)
    Note over Filter: Pod removed from cache
    Informer->>PM: AddPod(pod, nodeID, devices_old)
    PM-->>Informer: added=true
    Informer->>QM: AddUsage(pod, devices_old)
    Filter->>PM: AddPod(pod, nodeID, devices_new)
    PM-->>Filter: added=false already exists
    Note over QM: devices_old usage counted but never removed!
```

Lcos-000:
**What does this PR do?**

Fixes #2478.

Synchronizes pod device/quota accounting between `Scheduler.Filter()` and the pod informer callbacks. A dedicated `podMu` mutex now serializes the `TakeAndDeletePod → recompute → AddPod` cycle in `Filter()` and all informer handlers (`onAddPod`, `onDelPod`, `cleanupStalePodAllocation`) that mutate `PodManager` + `QuotaManager`.

**Why is this fix needed?**

Without the lock, `onAddPod` could fire between `Filter()`'s `TakeAndDeletePod` (line 956) and its subsequent `AddPod` (line 998). Since the `PodManager` keys pods by UID:

1. `Filter` deletes the pod from the cache.
2. `onAddPod` re-adds it with stale devices and calls `AddUsage(stale_devices)`.
3. `Filter`'s later `AddPod(new_devices)` returns `added=false`, so the new usage is never recorded while the stale usage stays counted.

This causes double-counted / leaked quota usage that is never cleaned up.

**Description**

- Add `Scheduler.podMu sync.Mutex`.
- Hold it across the whole `Filter()` live-path usage window.
- Acquire it in `onAddPod`, `onDelPod`, and `cleanupStalePodAllocation`.

**Validation**

- `go test ./pkg/scheduler/... -short -count=1` passes.
- `go test -race ./pkg/scheduler/... -short -count=1` passes.

**AI assistance disclosure**

This PR was written primarily by opencode (an AI coding assistant). Per CONTRIBUTING.md, AI assistance is disclosed here.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved scheduling consistency when pods are added, deleted, or cleaned up.
  * Prevented stale resource allocations from affecting quota and usage calculations.
  * Reduced the risk of conflicting updates during concurrent pod events.

### DIFF
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -78,6 +78,12 @@ type Scheduler struct {
 
 	lock   sync.RWMutex
 	synced bool
+
+	// podMu serializes podAdd/podDelete/podRe-Add cycles between Filter()
+	// and the pod informer callbacks. Without it, onAddPod() could fire
+	// between Filter()'s TakeAndDeletePod and its subsequent AddPod,
+	// double-counting quota usage.
+	podMu sync.Mutex
 }
 
 func NewScheduler() *Scheduler {
@@ -146,6 +152,9 @@ func (s *Scheduler) onAddPod(obj any) {
 	if !ok {
 		return
 	}
+	s.podMu.Lock()
+	defer s.podMu.Unlock()
+
 	if util.IsPodInTerminatedState(pod) {
 		if pi, ok := s.podManager.TakeAndDeletePod(pod); ok {
 			s.quotaManager.RmUsage(pod, pi.Devices)
@@ -195,6 +204,8 @@ func (s *Scheduler) onDelPod(obj any) {
 	if !ok {
 		return
 	}
+	s.podMu.Lock()
+	defer s.podMu.Unlock()
 	if pi, ok := s.podManager.TakeAndDeletePod(pod); ok {
 		s.quotaManager.RmUsage(pod, pi.Devices)
 	}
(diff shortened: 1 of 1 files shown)


mesutoezdil:
duplicated

PR #2491, opened 2026-08-08
fix(scheduler): use Warningf for device health warning in register

aryansri05:
klog.Warning() does not support key-value pairs — it concatenates
all arguments via fmt.Sprint(), so the current call on line 432 of
scheduler.go produces garbled log output like:

```
Device is unhealthy, cleaning up nodenodeNamenode-1deviceVendornvidia.com/gpu
```

Switched to klog.Warningf with format specifiers so the node name and
device vendor are readable in the logs during device health check cleanup.

Before:
```go
klog.Warning("Device is unhealthy, cleaning up node", "nodeName", val.Name, "deviceVendor", devhandsk)
```

After:
```go
klog.Warningf("Device is unhealthy, cleaning up node %s for device vendor %s", val.Name, devhandsk)
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved warning messages for unhealthy device cleanup by including the affected node name and device vendor.


mesutoezdil:

PR #2501, opened 2026-08-09
fix(device): return deep copied pods in GetScheduledPods to prevent data race

### LINKED ISSUE #2471: Data race: GetScheduledPods shares PodInfo pointers with the metrics collector while AddPod rewrites them
**What happened**:

`PodManager.GetScheduledPods` copies the map but hands back the same `*PodInfo`
pointers it stores:

```go
// Return a shallow copy of the pods map to avoid race conditions.
// This prevents a "concurrent map iteration and map write" fatal error.
podsCopy := make(map[k8stypes.UID]*PodInfo, podCount)
maps.Copy(podsCopy, m.pods)
return podsCopy, nil
```

That protects the outer map, but every `*PodInfo` in it is still shared with the
manager, and `AddPod` writes to those structs in place:

```go
m.pods[pod.UID].Devices = devices   // pods.go:72
```

`UpdatePod` does the same with `pi.Pod = pod`.

The only caller is the Prometheus collector in `cmd/scheduler/metrics.go`, which
runs on the scrape goroutine and ranges over exactly those fields:

```go
schedpods, _ := cc.metricsProvider.GetPodManager().GetScheduledPods()
for _, val := range schedpods {
    for _, podSingleDevice := range val.Devices {
```

Meanwhile the informer goroutine calls `onAddPod` → `AddPod` on every pod add and
update. The pointer handoff is synchronised, but the collector reads `val.Devices`
and `val.Namespace` after it has released the read lock, so the reads are
unsynchronised against those writes.

`go test --race` reports it on a clean checkout of master:

```
WARNING: DATA RACE
Write at 0x00c000243a18 by goroutine 10:
  github.com/Project-HAMi/HAMi/pkg/device.(*PodManager).AddPod()
      pkg/device/pods.go:72

Previous read at 0x00c000243a18 by goroutine 11:
  ... ranging val.Devices, as cmd/scheduler/metrics.go does
```

followed by a second report on the devices map itself, `mapIterStart` against
`mapassign_faststr`.

This is the same shape as #2333, which made `nodeManager.GetNode` return a deep
copy for the node cache. `ListPodsInfo` in the same file already deep copies via
`PodInfo.DeepCopy()`; `GetScheduledPods` and `GetPod` do not.

A scrape should not be able to observe a `PodInfo` while the informer is
rewriting it. Anything handed outside the manager's lock should be a copy, as
`ListPodsInfo` already does.

Drive `AddPod` for one UID on one goroutine while a second goroutine does what
the collector does:

```go
pm.AddPod(pod, "node1", devs(1))

go func() { for { pm.AddPod(pod, "node1", devs(i)) } }()          // informer

schedpods, _ := pm.GetScheduledPods()                              // collector
for _, val := range schedpods {
    for _, podSingleDevice := range val.Devices {
        ...
    }
}
```

Run with `--race`. It reports on the first iteration for me on master.

In production the window is smaller, but scrapes are periodic and pod
adds/updates are continuous, so it is reachable rather than theoretical. The
consequence is the race detector's usual one: a torn read of the map header, or
a scrape observing a half-swapped device list.

`GetPod` has the same problem — it returns the stored `*PodInfo` directly. It has
no non-test callers today, so nothing is broken by it, but it is the same trap
waiting for the next caller.

The fix I have in mind is to deep copy in `GetScheduledPods`, reusing the
existing `PodInfo.DeepCopy()`, and the same for `GetPod`. That does mean a
deep copy per scrape, which `ListPodsInfo` already pays. If that cost is a
concern for large clusters, the alternative is for the collector to hold the
read lock while it builds its metric values, which is cheaper but keeps the lock
longer. Happy to send either — I did not want to pick without asking.

**Environment**:
- HAMi version: master
- Others: scheduler + metrics only, reproducible in a unit test under `--race`

aryansri05:
Fixes #2471

Currently, `GetScheduledPods()` returns a shallow copy of the pods map via `maps.Copy`. While this prevents concurrent map iteration and map write panics on the map itself, the map values are pointers to `*PodInfo`. As a result, the underlying struct pointers are still shared across threads.

When the metrics collector reads the scheduled pods concurrently, it races with scheduling routines (like `AddPod` or `UpdatePod`) mutating these same structs.

This PR fixes the data race by iterating and calling `v.DeepCopy()` on the `PodInfo` pointers, returning fully independent structs.

Before:
```go
podsCopy := make(map[k8stypes.UID]*PodInfo, podCount)
maps.Copy(podsCopy, m.pods)
```

After:
```go
podsCopy := make(map[k8stypes.UID]*PodInfo, podCount)
for k, v := range m.pods {
	podsCopy[k] = v.DeepCopy()
}
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved scheduled pod data isolation by returning independent copies, preventing unintended changes to the original pod information.


mesutoezdil:

aryansri05:
Apologies for not following the contribution guidelines — I should have disclosed AI assistance upfront and checked for existing PRs before opening this. Won't happen again.

PR #2502, opened 2026-08-09
fix(server): preserve per-container device index in Kunlun/Mthreads P…

### LINKED ISSUE #2463: kunlun/mthreads: PatchAnnotations overwrites per-container device index for multi-container pods
**What happened**:

```go
	if ok && len(devlist) > 0 {
		for _, dp := range devlist {
			value := ""
			for _, val := range dp {
			}
			if len(value) > 0 {
			}
		}
	}
	return *annoinput
}
```

```
```

Same for mthreads:

```
```

**Environment**:
- Others: N/A

Samarth1-glitch:
## What type of PR is this?

/kind bug

`PatchAnnotations` in both the Kunlun and Mthreads device backends writes the vendor-specific device-index annotation (`BAIDU_COM_DEVICE_IDX`, `mthreads.com/gpu-index`) inside a loop that assigns to the same map key on every iteration. For a pod with two or more containers, only the last container's device index survives — earlier containers' indices are silently lost.

This PR fixes both backends to join each container's index using the package's existing `OnePodMultiContainerSplitSymbol` (`;`) delimiter, so every container's index is preserved. Single-container output is unchanged — no trailing delimiter is added — to avoid changing the existing annotation format for anything that already parses a bare index (e.g. `"0"` stays `"0"`, it does not become `"0;"`).

The generic `hami.io/*-allocated` annotation (via `EncodePodSingleDevice`) was already correct for all containers and is untouched by this PR.

Fixes #2463

## Design note for reviewers

The join-with-`;`-no-trailing-delimiter format was a judgment call to keep single-container behavior byte-for-byte identical to today. If you'd prefer this instead match `EncodePodSingleDevice`'s convention of a trailing delimiter after every entry (i.e. `"0;"` for single-container too), happy to change it — flagging the decision explicitly rather than assuming.

## Special notes for reviewers

- Added `TestKunlunDevices_PatchAnnotations` in `pkg/device/kunlun/device_test.go` — this backend previously had no test coverage for `PatchAnnotations` at all.
- Added a multi-container case to the existing `Test_PatchAnnotations` in `pkg/device/mthreads/device_test.go`, alongside the existing single-container case (unchanged).
- Mthreads' `MthreadsPredicateTime` / `MthreadsAssignedNode` annotations are still set redundantly per-container inside the loop (pre-existing behavior, not index-keyed so not part of this bug) — left as-is to keep this fix scoped to the reported issue.

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved GPU device indices for each container when multiple containers share an allocation.
  * Maintained existing annotation formats for single-container allocations.
  * Continued correctly recording device metadata, assigned nodes, and allocation timestamps.

* **Tests**
  * Added and updated coverage to verify device annotations and per-container index preservation across supported GPU types.

### DIFF
--- a/pkg/device/kunlun/device.go
+++ b/pkg/device/kunlun/device.go
@@ -87,20 +87,28 @@ func (dev *KunlunDevices) GetNodeDevices(n corev1.Node) ([]*device.DeviceInfo, e
 	return nodedevices, nil
 }
 
+// PatchAnnotations writes the generic hami.io/*-allocated annotation (all
+// containers, correct) and the vendor-specific BAIDU_COM_DEVICE_IDX
+// annotation. The vendor key encodes one index-list per container, joined
+// with the same OnePodMultiContainerSplitSymbol (";") used elsewhere in the
+// package, so multi-container pods no longer lose earlier containers'
+// indices to the last container's value. Single-container output is
+// unchanged (no trailing delimiter) to preserve the existing annotation
+// contract for any external consumer already parsing a bare index.
 func (dev *KunlunDevices) PatchAnnotations(pod *corev1.Pod, annoinput *map[string]string, pd device.PodDevices) map[string]string {
 	devlist, ok := pd[KunlunGPUDevice]
 	if ok && len(devlist) > 0 {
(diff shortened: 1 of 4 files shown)


mesutoezdil:

PR #2503, opened 2026-08-09
fix(monitor): add v0.MinSize and fix file-size discriminator for v0 cache

Nakshatra480:
## Summary

The v0 cache-file detector in `loadCache` compared `info.Size() == 1197897`,
but `unsafe.Sizeof(v0.sharedRegionT{}) == 1197896` an off-by-one that
caused any real libvgpu v0 cache file to fall through to the
`"unknown cache file size"` error, silently leaving v0 containers
unmonitored.

Additionally, v0 had no `MinSize()` function, unlike v1, so the detection
threshold was a bare magic number with no programmatic anchor. A future
field addition or removal in the v0 struct would shift the size silently
with no CI signal.

**Three changes:**

- Add `MinSize()` to `pkg/monitor/nvidia/v0/spec.go`, mirroring v1's
  equivalent. Returns `int(unsafe.Sizeof(sharedRegionT{}))`.

- Broaden the detector in `cudevshr.go` from `info.Size() == 1197897` to
  `info.Size() >= int64(v0.MinSize()) && info.Size() < int64(v1.MinSize())`.
  This accepts both the struct-derived size (1197896) and the empirical
  legacy size (1197897), and distinguishes the result from a v1 file
  without relying on the version-header bytes, which map to `smInitFlag`
  in the v0 layout rather than a real version number.

- Add `TestV0SharedRegionTLayoutSize` to `v0/spec_test.go` and add a
  second loadCache test for the struct-derived size in `cudevshr_test.go`,
  so any future struct change breaks CI instead of silently shifting the
  detection boundary.

Part of #2126 (LFX observability hardening — v0 cache-file loading)

Yes nodes running older libvgpu (v0 shared-memory format, ~1.2 MB cache
files) that previously got `unknown cache file size` errors will now have
their containers correctly loaded and monitored.

The `1197897` constant in the existing test is kept as the "empirical legacy
size" subtest to confirm the range check accepts it. The new "struct-derived
size" subtest covers `1197896`. Both pass.

**AI Disclosure:**
AI assistance was used for code inspection and draft formatting

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved NVIDIA cache file detection for older v0 cache files across the supported minimum-size range.
  * Preserved existing detection behavior for v1 cache files.

* **Reliability**
  * Added safeguards to detect unexpected changes in the NVIDIA v0 cache layout size.

### DIFF
--- a/pkg/monitor/nvidia/cudevshr.go
+++ b/pkg/monitor/nvidia/cudevshr.go
@@ -278,7 +278,7 @@ func loadCache(fpath string) (*ContainerUsage, error) {
 		_ = syscall.Munmap(usage.data)
 		return nil, fmt.Errorf("cache file magic flag not matched")
 	}
-	if info.Size() == 1197897 {
+	if info.Size() >= int64(v0.MinSize()) && info.Size() < int64(v1.MinSize()) {
 		klog.Infoln("casting......v0")
 		usage.Info = v0.CastSpec(usage.data)
 	} else if head.majorVersion == 1 && info.Size() >= int64(v1.MinSize()) {
--- a/pkg/monitor/nvidia/cudevshr_test.go
+++ b/pkg/monitor/nvidia/cudevshr_test.go
@@ -35,6 +35,7 @@ import (
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/cache"
 
+	v0 "github.com/Project-HAMi/HAMi/pkg/monitor/nvidia/v0"
 	v1 "github.com/Project-HAMi/HAMi/pkg/monitor/nvidia/v1"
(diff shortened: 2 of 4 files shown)


FouoF:
Can you make an e2e test to validate v0 cache file can be resolved by newer vgpu-monitor after this change?

Nakshatra480:
> Can you make an e2e test to validate v0 cache file can be resolved by newer vgpu-monitor after this change?

Added an e2e test case in `cudevshr_test.go` that constructs a v0 cache file with real field data and verifies `DeviceNum` and `DeviceMemoryLimit` round-trip correctly through `loadCache`.

FouoF (approved):

Nakshatra480:
@FouoF Thanks for merging ☺️

PR #2504, opened 2026-08-09
fix(scheduler): copy node info into nodeManager instead of storing the caller's

Lakshya77089:
**What type of PR is this?**

/kind bug

`nodeManager.addNode` stored the `*device.NodeInfo` it was handed, including the
`*corev1.Node` inside it:

```go
} else {
    m.nodes[nodeID] = nodeInfo
}
```

and on the update path assigned the caller's slices and node straight in:

```go
m.nodes[nodeID].Devices[vendor] = nodeInfo.Devices[vendor]
m.nodes[nodeID].Node = nodeInfo.Node
```

The only caller is `RegisterFromNodeAnnotations`, which builds `nodeInfo` around a
node taken from `s.nodeLister.List(...)` — an object owned by the shared informer.

That would be harmless if nothing wrote to it, but something does. `Bind` fetches
the same node out of the same cache, and on the release path cambricon's
`ReleaseNodeLock` deletes an annotation from it:

```go
delete(n.Annotations, DsmluLockTime)   // pkg/device/cambricon/device.go:178
```

Meanwhile `GetNode` and `ListNodes` deep copy the stored node on the Filter and
Bind goroutines. `go test ./pkg/scheduler/ --race` reports the read inside
`(*Node).DeepCopyInto` against that delete.

#2333 fixed the read side of this cache by making `GetNode` return a deep copy.
The write side kept a reference to the shared object, so the copy it makes can
still run while the object is being modified.

**The fix**

Copy `Node` and the device lists when storing, on both the insert and the update
path, reusing `device.DeepCopyDeviceInfos` which `GetNode` and `ListNodes`
already use.

**Tests**

`TestAddNodeCopiesSharedNodeObject` drives the annotation delete on one goroutine
while another calls `GetNode`/`ListNodes`, which is the shape of the real
interleaving. `TestAddNodeStoresDetachedCopy` mutates everything the caller still
holds, on both the insert and the update path, and asserts the manager is
unaffected. Both fail on master with `DATA RACE`, both pass here.

None filed — raising the fix directly.

Worth being explicit about one thing, since it changes how the bug reads: shared
informers replace objects in their store rather than mutating them in place, so
holding a lister pointer is not automatically a race. What makes this one real is
that HAMi mutates the object itself, in `ReleaseNodeLock`. #2329 is open against
that mutation. If it lands, this race goes quiet, but the manager would still be
holding a reference it does not own, and the next writer brings it back. The two
changes are worth having independently.

`addNode` now costs a deep copy per node per registration cycle. That is the same
cost `GetNode` already pays per call, and registration is far less frequent than
scheduling, so it should not be noticeable.

Not validated on real hardware. The change is confined to the scheduler's
in-memory node cache and is covered by unit tests, which CONTRIBUTING allows for
scheduler-scoped changes.

```release-note
NONE
```

---

AI assistance disclosure: this change was developed with Claude Code — finding
the race, the fix, and the tests. Flagging the extent up front per CONTRIBUTING.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when registering and updating nodes by preventing external changes to node metadata and device information from unexpectedly altering stored state.
  * Added safe handling for missing node data.

* **Tests**
  * Added coverage for concurrent registration and updates, including protection against shared-data mutations.

### DIFF
--- a/pkg/scheduler/nodes.go
+++ b/pkg/scheduler/nodes.go
@@ -68,6 +68,11 @@ func newNodeManager() *nodeManager {
 	}
 }
 
+// addNode copies what it is given. RegisterFromNodeAnnotations builds nodeInfo
+// around a *corev1.Node taken from nodeLister.List, which belongs to the shared
+// informer, and a backend such as cambricon writes to that same object when it
+// releases a node lock. Storing the pointer would leave GetNode and ListNodes
+// deep-copying it while that write happens.
 func (m *nodeManager) addNode(nodeID string, nodeInfo *device.NodeInfo) {
 	if nodeInfo == nil || len(nodeInfo.Devices) == 0 {
 		return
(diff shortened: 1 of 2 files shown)


FouoF (approved):

FouoF (approved):
Re-approve to refresh GitHub review:approved search index for Tide.

PR #2505, opened 2026-08-09
feat(monitor): add DeviceProcessCount metric to expose tracked GPU process count

Nakshatra480:
## Summary

Add `hami_container_device_process_count`, a per-device Prometheus gauge
that reports the number of process slots in a container's shared-memory
region that have `status != 0` and a nonzero memory footprint on the
queried virtual device.

**Why this matters:** When a child process or a process reached through SSH
bypasses libvgpu's `LD_PRELOAD` hook it never gets a shared-memory slot
allocated. The count therefore drops below the real number of CUDA processes
in the container. Operators and alert rules can use this metric to detect
GPU memory isolation failures early before a memory overrun occurs by
comparing `hami_container_device_process_count` against the expected number
of tracked GPU processes.

## Changes

| File | Change |
|---|---|
| `pkg/monitor/nvidia/cudevshr.go` | Add `DeviceProcessCount(idx int) int` to `UsageInfo` interface |
| `pkg/monitor/nvidia/v0/spec.go` | Implement: counts slots where `status != 0` and `total+contextSize+moduleSize > 0` for device `idx` |
| `pkg/monitor/nvidia/v1/spec.go` | Identical implementation for v1 layout |
| `pkg/monitor/nvidia/v0/spec_test.go` | 5 unit tests: no processes, one active, device-specific exclusion, dead slot, corrupt procnum |
| `pkg/monitor/nvidia/v1/spec_test.go` | Same 5 cases for v1 |
| `cmd/vGPUmonitor/feedback_test.go` | Add stub method to satisfy updated interface |
| `cmd/vGPUmonitor/metrics.go` | Register `ctrDeviceProcessCountDesc`, emit in `collectContainerMetrics` |
| `cmd/vGPUmonitor/metrics_container_test.go` | Update expected metric counts and add count to want map |

## Example output

`hami_container_device_process_count{container="worker",device_uuid="GPU-abc123...", namespace="team-a",pod="trainer-0",vdevice_index="0"} 2`

If a child process escapes slot tracking, this drops to 1 while the real
GPU process count remains 2, the isolation failure becomes immediately
observable in Grafana.

Part of #2126 (LFX observability hardening)

**AI Disclosure:**
AI assistance was used for code inspection and draft formatting; all logic,
test cases, and verification were manually reviewed and validated.

## Summary by CodeRabbit

- **New Features**
  - Added a Prometheus metric reporting active processes using each container’s device.
  - Process counts are calculated per device and exclude inactive or invalid entries.
  - Available across supported NVIDIA monitoring versions.

- **Bug Fixes**
  - Improved resilience when processing malformed device process data.

- **Tests**
  - Added coverage for device-specific counts, inactive processes, terminated entries, context-only usage, and invalid data.

### DIFF
--- a/cmd/vGPUmonitor/feedback_test.go
+++ b/cmd/vGPUmonitor/feedback_test.go
@@ -73,6 +73,7 @@ func (s *stubInfo) GetRecentKernel() int32               { return 1 }
 func (s *stubInfo) SetRecentKernel(int32)                {}
 func (s *stubInfo) GetUtilizationSwitch() int32          { return 0 }
 func (s *stubInfo) SetUtilizationSwitch(int32)           {}
+func (s *stubInfo) DeviceProcessCount(int) int           { return 0 }
 
 func TestCheckFunctionsHighPriority(t *testing.T) {
 	sw := map[string]UtilizationPerDevice{"gpu-0": {0, 1}}
--- a/cmd/vGPUmonitor/metrics.go
+++ b/cmd/vGPUmonitor/metrics.go
@@ -118,6 +118,12 @@ var (
 		"Container device memory buffer size in bytes",
 		[]string{"namespace", "pod", "container", "vdevice_index", "device_uuid"}, nil,
 	)
+
+	ctrDeviceProcessCountDesc = prometheus.NewDesc(
+		"hami_container_device_process_count",
+		"Number of processes tracked in the container's shared-memory region for this device",
+		[]string{"namespace", "pod", "container", "vdevice_index", "device_uuid"}, nil,
+	)
 )
 
 // Legacy metric descriptors (populated only when --legacy-metrics is enabled).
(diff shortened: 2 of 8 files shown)


FouoF:
Thanks for the metrics proposal. DeviceProcessCount is presented as a count of tracked GPU processes, but the implementation counts slots with any non-zero memory, context, or module field. It cannot distinguish the untracked-process case described in the PR and has no independent process inventory against which the value is validated. That would publish a misleading metric contract, so we are closing the current PR. A future proposal should first define the exact population and prove it against a real process source.

Nakshatra480:
> Thanks for the metrics proposal. DeviceProcessCount is presented as a count of tracked GPU processes, but the implementation counts slots with any non-zero memory, context, or module field. It cannot distinguish the untracked-process case described in the PR and has no independent process inventory against which the value is validated. That would publish a misleading metric contract, so we are closing the current PR. A future proposal should first define the exact population and prove it against a real process source.

Thank you for the detailed feedback, your point about not having an independent process source to verify the slot count against is a valid one. I will submit a proposal which cross-references either `/proc` or `NVML` in order to establish a ground-truth process inventory before the metric is displayed.

PR #2508, opened 2026-08-09
fix(device): reject malformed pod device annotation segments

### LINKED ISSUE #2507: DecodeContainerDevices silently accepts malformed annotation segments
**What happened**:

`DecodeContainerDevices` silently accepts a non-empty malformed annotation segment when it contains no comma-separated fields.

For example, decoding `malformed:;` returns no error and produces an empty device allocation. `DecodePodDevices` therefore cannot distinguish this malformed allocation from a genuinely empty container allocation.

This bypasses the parse-error propagation expected by callers such as scheduler accounting and monitoring.

Every non-empty annotation segment should contain the required device fields:

`UUID, type, memory, cores`

A non-empty segment missing those fields should return a parsing error. Empty segments used as container separators should remain valid.

```go
checklist := map[string]string{
    "NVIDIA": "hami.io/vgpu-devices-to-allocate",
}
annotations := map[string]string{
    "hami.io/vgpu-devices-to-allocate": "malformed:;",
}

devices, err := DecodePodDevices(checklist, annotations)
fmt.Printf("devices=%v err=%v\n", devices, err)
```
On current tested commit, `err` is `nil`.

The decoder currently validates segments only when `strings.Contains(val, ",")` is true. Consequently, malformed non-empty segments without commas are skipped completely.

The minimal fix is to skip only truly empty segments and validate every other segment through the existing field-count and numeric parsing logic.

This follows the parser error-propagation behaviour established by #1856 and #1857.
No GPU hardware is required to reproduce or test this issue.

- The output of `nvidia-smi -a` on your host: NA
- Your docker or containerd configuration file (e.g: `/etc/docker/daemon.json`): NA
- The hami-device-plugin container logs: NA
- The hami-scheduler container logs: NA
- The kubelet logs on the node (e.g: `sudo journalctl -r -u kubelet`): NA
- Any relevant kernel output lines from `dmesg`: NA

**Environment**:
- HAMi version: current `master` at `(3616313cb4d30841a6162421969d7ab9463931aa)`
- nvidia driver or other AI device driver version: NA
- Docker version from `docker version`: NA
- Docker command, image and tag used: NA
- Kernel version from `uname -a`: NA
- Others:
  - Reproduced and tested on macOS

blackdragoon26:
**What type of PR is this?**
/kind bug

`DecodeContainerDevices` previously processed an annotation segment only when it contained a comma. Consequently, a non-empty malformed segment such as `malformed` was silently skipped and interpreted as an empty device allocation.

This change skips only truly empty segments. Every non-empty segment now passes through the existing required-field and numeric validation.

A regression test verifies that `DecodePodDevices` rejects `malformed:;` instead of returning an empty allocation without an error.

Fixes #2507 

This is limited to the shared annotation decoder and its regression test. It introduces no device-specific behavior and requires no GPU hardware.

Validated on macOS with:
- `make verify`
- `go test ./pkg/... -count=1`
- `go test -race ./pkg/device -run 'TestDecodePodDevices_BadAnnotation|Test_DecodePodDevices' -count=1`
- `go test ./pkg/device -run '^$' -fuzz '^FuzzDecodeContainerDevices$' -fuzztime=5s`
- `git diff --check`

All checks passed.

Malformed non-empty pod-device annotation segments are now rejected instead of being silently interpreted as empty allocations.

AI Disclosure:
Codex assisted with repository analysis, implementation, regression-test development, verification, and drafting this PR description. I reviewed the final code changes and test results before submission.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of device annotations by ignoring empty segments.
  * Invalid annotation formats, memory values, and core values now return clear errors instead of being accepted.
  * Valid device entries continue to be processed as expected.

* **Tests**
  * Expanded validation coverage for malformed device annotations, including incorrect field counts and invalid segments.

### DIFF
--- a/pkg/device/devices.go
+++ b/pkg/device/devices.go
@@ -460,25 +460,26 @@ func DecodeContainerDevices(str string) (ContainerDevices, error) {
 	tmpdev := ContainerDevice{}
 	klog.V(5).Infof("Start to decode container device %s", str)
 	for _, val := range cd {
-		if strings.Contains(val, ",") {
-			tmpstr := strings.Split(val, ",")
-			if len(tmpstr) < 4 {
-				return nil, fmt.Errorf("pod annotation format error, missing fields, do not use nodeName in task spec")
-			}
-			tmpdev.UUID = tmpstr[0]
-			tmpdev.Type = tmpstr[1]
-			devmem, err := strconv.ParseInt(tmpstr[2], 10, 32)
-			if err != nil {
-				return nil, fmt.Errorf("invalid memory field: %w", err)
-			}
-			tmpdev.Usedmem = int32(devmem)
-			devcores, err := strconv.ParseInt(tmpstr[3], 10, 32)
-			if err != nil {
-				return nil, fmt.Errorf("invalid core field: %w", err)
-			}
-			tmpdev.Usedcores = int32(devcores)
-			contdev = append(contdev, tmpdev)
+		if val == "" {
+			continue
 		}
+		tmpstr := strings.Split(val, ",")
+		if len(tmpstr) < 4 {
+			return nil, fmt.Errorf("pod annotation format error, missing fields, do not use nodeName in task spec")
+		}
+		tmpdev.UUID = tmpstr[0]
+		tmpdev.Type = tmpstr[1]
+		devmem, err := strconv.ParseInt(tmpstr[2], 10, 32)
+		if err != nil {
+			return nil, fmt.Errorf("invalid memory field: %w", err)
+		}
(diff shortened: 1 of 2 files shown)


FouoF:
The annotation is an internal api which will not be malformed unless upstream bugs. The better way to prevent it is fully test before releasing.

blackdragoon26:
> The annotation is an internal api which will not be malformed unless upstream bugs. The better way to prevent it is fully test before releasing.

Ohhhh, makes sense.
Thanks for the clarification!

PR #2513, opened 2026-08-09
fix(nvidia): stamp flattened UsageList index in MIG UUID suffix

Eshiv-Pandey:
## What this PR does

`AddResourceUsage` records which MIG slot a container got by appending a `[templateIdx-position]` suffix to the device UUID. The two allocation branches disagreed on what `position` means:

- The fresh-template branch (when the geometry is reset) reserved the slot at the flattened `UsageList` offset (`usageListIdx`) but stamped the geometry item index (`idx`) into the UUID.
- The reuse branch stamps the flattened offset.

`ExtractMigTemplatesFromUUID`, and therefore `getNodesUsage`, reads thesuffix as a flattened offset into `UsageList`. When an earlier item in the geometry has `Count > 1`, the geometry index and the flattened offset diverge, so `getNodesUsage` marks a different slot in use than the one that was actually reserved. The allocated slot appears free while an unallocated one appears taken, which can lead to double-booking a MIG instance.

## The fix

Stamp `usageListIdx` in the fresh-template branch so both branches and the reader agree on the meaning of the suffix.

## Why it was not caught earlier

The shipped A100-40GB geometry is `1g.5gb x1, 2g.10gb x3`, where the geometry index and flattened offset of the first fitting slot coincide, so the bug is masked. It surfaces on geometries where an earlier item has `Count > 1`, such as the `2g.10gb x3, 1g.5gb x1` example in docs/develop/dynamic-mig.md.

## Testing

Added `TestAddResourceUsage_MigFreshTemplateStampsFlattenedIndex`, which builds a `1g.5gb x2, 2g.10gb x1` geometry (flattened `UsageList` index 2 = geometry index 1) and asserts the stamped suffix round-trips through `ExtractMigTemplatesFromUUID` to the slot actually reserved. The test fails without the fix (stamps `1`, expects `2`) and passes with it.

Also corrected the existing `TestDevices_AddResourceUsage`: its `Count > 1` case pinned the buggy suffix (`dev-3[0-1]`) while its assertion helper silently re-derived the flattened offset, masking the mismatch. It now expects the flattened suffix (`dev-3[0-2]`) and reads the `UsageList` index directly. Full unit suite run with `-race`.

## AI assistance disclosure

This change was prepared with AI assistance (Claude Opus 4.8). All code and tests were reviewed by me.

## Bug Fixes

* Corrected NVIDIA MIG allocation identifiers to consistently match the reserved MIG slot.
* Improved allocation reliability for templates containing repeated geometry configurations.
* Updated MIG usage tracking so allocated slots are represented accurately.

## Tests

* Added coverage for fresh-template allocations.
* Added validation for repeated geometry configurations and correct slot identification.

### DIFF
--- a/pkg/device/nvidia/device.go
+++ b/pkg/device/nvidia/device.go
@@ -695,8 +695,10 @@ func (dev *NvidiaGPUDevices) AddResourceUsage(pod *corev1.Pod, n *device.DeviceU
 						}
 						ctr.Usedmem = n.MigUsage.UsageList[usageListIdx].Memory
 						ctr.Usedcores = n.MigUsage.UsageList[usageListIdx].Core
+						// Stamp the flattened UsageList offset (what ExtractMigTemplatesFromUUID
+						// and getNodesUsage read back), not the geometry item index.
 						if !strings.Contains(ctr.UUID, "[") {
-							ctr.UUID = ctr.UUID + "[" + fmt.Sprint(tidx) + "-" + fmt.Sprint(idx) + "]"
+							ctr.UUID = ctr.UUID + "[" + fmt.Sprint(tidx) + "-" + fmt.Sprint(usageListIdx) + "]"
 						}
 						n.MigUsage.Index = int32(tidx)
 						n.MigUsage.UsageList[usageListIdx].InUse = true
--- a/pkg/device/nvidia/device_test.go
+++ b/pkg/device/nvidia/device_test.go
@@ -1383,7 +1383,7 @@ func TestDevices_AddResourceUsage(t *testing.T) {
 			},
 			ctr: &device.ContainerDevice{
 				UUID:    "dev-3",
-				Usedmem: 8000, // Requires 8GB, matches second template (idx=1) which should be at UsageList[2]
+				Usedmem: 8000, // matches 2g.10gb at flattened UsageList index 2 (preceded by 1g.5gb x2)
 			},
 			wantUsage: &device.DeviceUsage{
 				Used:      1,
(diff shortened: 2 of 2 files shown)


FouoF:
https://github.com/Project-HAMi/HAMi/pull/2378 have refactored dynamic MIG, please check is this bug still exists on it.

pkg/device/nvidia/device.go
```diff
@@ -695,8 +695,10 @@ func (dev *NvidiaGPUDevices) AddResourceUsage(pod *corev1.Pod, n *device.DeviceU
 						}
 						ctr.Usedmem = n.MigUsage.UsageList[usageListIdx].Memory
 						ctr.Usedcores = n.MigUsage.UsageList[usageListIdx].Core
+						// Stamp the flattened UsageList offset (what ExtractMigTemplatesFromUUID
+						// and getNodesUsage read back), not the geometry item index.
 						if !strings.Contains(ctr.UUID, "[") {
-							ctr.UUID = ctr.UUID + "[" + fmt.Sprint(tidx) + "-" + fmt.Sprint(idx) + "]"
+							ctr.UUID = ctr.UUID + "[" + fmt.Sprint(tidx) + "-" + fmt.Sprint(usageListIdx) + "]"
```
mesutoezdil, line 701: usagelistidx and tidx are gone in master now. #2378 rewrote this function. pls check vs new code.

Eshiv-Pandey: i just checked the bug is gone now after the refactor. closing this. thanks for the review!

PR #2516, opened 2026-08-09
fix(util): prevent invalid node names from breaking pod scheduling

### LINKED ISSUE #2506: Pod scheduling fails permanently on nodes with names longer than 63 characters (PatchPodAnnotations copies node name into a pod label, violating the Kubernetes 63-char label-value limit)
### What happened?

HAMi pod scheduling fails when a pod is assigned to a Kubernetes node whose name is longer than 63 characters.

In `pkg/scheduler/scheduler.go`, the selected node name is stored in the `AssignedNodeAnnotations` annotation:

```go
annotations[util.AssignedNodeAnnotations] = m.NodeID
```

However, `PatchPodAnnotations()` in `pkg/util/util.go` also copies this value into a pod label:

```go
if v, ok := annotations[AssignedNodeAnnotations]; ok && v != "" {
    label[AssignedNodeAnnotations] = v
    p.Metadata.Labels = label
}
```

Kubernetes node names can be up to **253 characters**, while Kubernetes label values are limited to **63 characters**.

Therefore, when the selected node name exceeds 63 characters, the API server rejects the pod patch with a validation error similar to:

```text
Invalid value: "...": must be no more than 63 characters
```

This causes the scheduling attempt to fail and the scheduler rolls back the associated resource usage.

### Impact

This results in a permanent scheduling failure for GPU pods on valid Kubernetes nodes with long names.

Since the node name is copied into the label for every candidate node, retrying the scheduling attempt does not resolve the problem when the cluster contains such nodes.

The pod remains `Pending`, while the actual cause is buried in the scheduler logs as a Kubernetes label validation error.

### Reproduction

1. Create/use a Kubernetes cluster with a GPU node whose name is longer than 63 characters, for example:

```text
gpu-worker-01.production.us-east-1.internal.example-company.com
```

2. Deploy HAMi.

3. Submit a pod requesting a GPU:

```yaml
resources:
  limits:
    nvidia.com/gpu: 1
```

4. Observe that the pod remains `Pending`.

5. Check the HAMi scheduler logs and observe the API server validation error indicating that the label value exceeds 63 characters.

### Expected behavior

HAMi should successfully schedule GPU pods on any valid Kubernetes node, regardless of the node name length.

The full node name should remain available through the `AssignedNodeAnnotations` annotation without causing Kubernetes label validation failures.

### Suggested Fix

The preferred solution is to stop mirroring the raw node name from `AssignedNodeAnnotations` into a pod label in `PatchPodAnnotations()`.

The node name is already available through the annotation, so the redundant label can be removed.

If the label is required by another HAMi component, the value should instead be transformed into a Kubernetes-valid label value, for example by using a deterministic hash, while preserving the full node name in the annotation.

### Affected Code

* `pkg/util/util.go` — `PatchPodAnnotations()`
* `pkg/scheduler/scheduler.go` — assignment of `AssignedNodeAnnotations`

### Additional Tests

Add a regression test covering node names longer than 63 characters and verify that:

* The pod annotation accepts the full node name.
* No invalid node-name label is created.
* Pod scheduling succeeds on nodes with valid names longer than 63 characters.

AyushSrivastava1818:
## What type of PR is this?

/kind bug

## Which issue(s) this PR fixes:

Fixes #2506 

## Summary

`PatchPodAnnotations()` in `pkg/util/util.go` mirrors the assigned node name from `AssignedNodeAnnotations` into a pod label.

Kubernetes node names can be up to **253 characters**, while label values are limited to **63 characters**. When a valid Kubernetes node has a name longer than 63 characters, the API server rejects the pod patch because the node name is used as an invalid label value.

This causes the scheduling attempt to fail even though the node itself has a valid Kubernetes name.

### Impact

* GPU pods can fail to schedule on nodes with names longer than 63 characters.
* The Kubernetes API server rejects the patch with a label validation error.
* The scheduler rolls back the scheduling attempt.

## Fix

Validate the node name using `validation.IsValidLabelValue()` before mirroring it into the pod label.

Valid values continue to be mirrored as before. If the value is invalid, the label is omitted and a warning is logged, while the full node name is still preserved in the annotation.

No changes to the scheduler interface or existing callers are required.

## Test Plan

Added table-driven test coverage in `pkg/util/util_test.go` for:

* Normal short node name
* Node name longer than 63 characters
* Node name containing invalid label characters
* Annotation preservation when the label is omitted

Verification performed:

* `go test ./pkg/util/...` — passed
* `go vet ./pkg/util/...` — passed
* `gofmt` — clean

The full scheduler test suite could not be run locally due to a pre-existing Windows CGO/NVIDIA toolchain issue.

## Notes for Reviewers

The change is intentionally limited to the label-mirroring logic in `PatchPodAnnotations()`.

The full node name remains available through `AssignedNodeAnnotations`; only the invalid label mirror is skipped.

Yes. GPU pods can now be scheduled on valid Kubernetes nodes whose names exceed the 63-character label-value limit.

## Summary by CodeRabbit

* **Bug Fixes**
  * Pod labels now safely support assigned-node values with invalid characters or excessive length.
  * Invalid values are converted into deterministic, valid labels, with conversions logged.
  * Metrics collection continues to find pods assigned to nodes with long names.
  * Scheduler allocation and quota tracking are now protected against concurrent over-allocation, duplicate accounting, and incomplete rollbacks.
  * Failed allocation attempts no longer leave stale locks or capacity reservations.
  * Valid assigned-node values remain unchanged.

* **Tests**
  * Expanded coverage for label conversion, metric discovery, scheduler concurrency, rollback, locking, and quota handling.

### DIFF
--- a/pkg/util/util_test.go
+++ b/pkg/util/util_test.go
@@ -18,6 +18,7 @@ package util
 
 import (
 	"context"
+	"strings"
 	"testing"
 	"time"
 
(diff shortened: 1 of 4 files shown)


Eshiv-Pandey (commented):
cause looks right but omitting the label isn't safe .. **vGPUmonitor** selects pods by it in `collectPodAndContainerInfo` and `collectPodAndContainerMigInfo `(cmd/vGPUmonitor/metrics.go), so on a >63-char node the lister returns zero pods and per-pod/container + MIG metrics go dark on exactly the nodes this unblocks. 

could we derive a deterministic valid value instead (truncate to 63 + short hash) and use the same derivation in the selector?  not sure tho, wdyt?

pkg/util/util_test.go
```diff
@@ -328,51 +329,92 @@ func TestGetAllocatePodByNode(t *testing.T) {
...
-	pod := &corev1.Pod{
-		ObjectMeta: metav1.ObjectMeta{
-			Name:      "test-pod",
-			Namespace: "default",
-		},
-	}
-
-	client.KubeClient.CoreV1().Pods("default").Create(context.TODO(), pod, metav1.CreateOptions{})
+	longNodeName := "node-" + strings.Repeat("a", 60) // 65 chars > 63
+	invalidCharNodeName := "invalid_node_name!"       // <= 63 chars but contains invalid label character '!'
```
Eshiv-Pandey: i don't think this is a reachable input since node names must be dns subdomains

AyushSrivastava1818:
@Eshiv-Pandey after going through it carefully i think the same way,you're right - I checked collectPodAndContainerInfo and collectPodAndContainerMigInfo and they both build the selector from the raw NODE_NAME env var (labels.Set{util.AssignedNodeAnnotations: nodeName}), so if we skip setting the label, those lookups just come back empty on the exact nodes this PR is meant to fix. That's worse than the original bug in a way, since now it fails silently instead of loudly.

I'll switch the approach: instead of skipping the label when it's invalid, I'll add a small helper that deterministically turns any node name into a valid label value (truncate it, tack on a short hash so different long names don't collide, keep it under 63 chars and alphanumeric-bounded). Then I'll use that same helper on both sides - when PatchPodAnnotations sets the label, and when vGPUmonitor builds its selector from NODE_NAME - so they always agree on what the label value should be.

Annotation stays untouched with the full node name either way. Pushing an update.

FouoF:
@AyushSrivastava1818 Who consumes this label and whether they can consume hashed label?

AyushSrivastava1818:
@FouoF 
Ran a full grep across the repo to answer this properly:

The only place an annotation gets mirrored into a label is pkg/util/util.go:175-176 the exact function this PR modifies.
The only reader of that label anywhere in the codebase is vGPUmonitor (cmd/vGPUmonitor/metrics.go:368 and :512), already updated to derive its selector through the same SafeLabelValue helper as the write path, so it stays in sync.
Everything else touching AssignedNodeAnnotations- the AWS Neuron and Mthreads device plugins, and the scheduler itself (pkg/scheduler/scheduler.go) - reads/writes the annotation, not the label, so they're unaffected. The annotation always keeps the full, untruncated node name.
Checked docs/develop/protocol.md too, since it documents hami.io/vgpu-node as part of the public protocol-it's documenting the annotation format only (hami.io/vgpu-node: {schedule decision node}), no mention of it being used as a label selector, so no external contract is affected either.

So: one true label consumer in the whole codebase, and it's already fixed to match. Everything else is annotation-based and untouched by this PR.

FouoF:
@AyushSrivastava1818 The final concern: Will the hashed node name exposed to metrics labels?

AyushSrivastava1818:
@FouoF Checked directly — the node name doesn't get exposed as a metric label at all, hashed or otherwise. Every prometheus.NewDesc in this file is dimensioned by namespace/pod/container/device identifiers (device_uuid, vdevice_index, etc.) — node name never appears in any label set, current or legacy. SafeLabelValue(nodeName) is only used internally, to build the K8s pod-lister selector that finds which pods are on this node; that value never reaches sendMetric/MustNewConstMetric. So there's no dashboard or alert that could break here — this dimension was never exposed through this collector in the first place.

mesutoezdil:
"4. Review replies. The reply you post must be written by you and must address the specific point raised. Verbatim or canned AI replies, or replies that do not engage the comment, lead to the PR being closed."

PR #2521, opened 2026-08-09
Fix client test ignoring TEST_WITH_REAL_CLUSTER

### LINKED ISSUE #2520: Fix ignored TEST_WITH_REAL_CLUSTER environment variable in client_test.go
### Bug Description
In `pkg/util/client/client_test.go`, the `TestClientRealNodePerformance` test is intended to be skipped by default unless a real Kubernetes cluster is available. The skip message correctly instructs developers to set `TEST_WITH_REAL_CLUSTER=true` to run the test. 

However, the test completely ignores this environment variable because `skipRealClusterTest` is hardcoded to `true`. Additionally, the code evaluates the condition using a non-idiomatic `== true` comparison.

### To Reproduce
1. Run the client tests with the environment variable set to true:
   `TEST_WITH_REAL_CLUSTER=true go test ./pkg/util/client/...`
2. Observe that `TestClientRealNodePerformance` is still skipped despite the environment variable being passed.

### Expected behavior
The test should read the `TEST_WITH_REAL_CLUSTER` environment variable and only skip the test if it is not set to `"true"`. The boolean evaluation should also be idiomatic (e.g. `if skipRealClusterTest`).

### Environment
- **HAMi version:** Latest `master` branch
- **OS/Arch:** Any

### Proposed Solution
Update `client_test.go` to properly read the environment variable using `os.Getenv` and fix the boolean check. 

```go
// Replace this:
skipRealClusterTest := true
if skipRealClusterTest == true { ... }

// With this:
skipRealClusterTest := os.Getenv("TEST_WITH_REAL_CLUSTER") != "true"
if skipRealClusterTest { ... }

KunwarSidhu47:
**What happened:**
The `TestClientRealNodePerformance` test in `client_test.go` was intended to be skipped unless `TEST_WITH_REAL_CLUSTER=true` is set. However, it was completely ignoring the environment variable because the skip boolean was hardcoded to `true`. 

The test should properly read the environment variable and only skip if it's not set to true. 

**How to reproduce it:**
Run `TEST_WITH_REAL_CLUSTER=true go test ./pkg/util/client/...` and notice that the real cluster performance test is still skipped.

**Proposed Solution:**
Updated the code to actually use `os.Getenv("TEST_WITH_REAL_CLUSTER")` and cleaned up the boolean check.

---
*Note: I used an AI assistant to help identify this bug and draft the initial code fix.*

Fixes #2520

### DIFF
--- a/docs/develop/amd-vgpu.md
+++ b/docs/develop/amd-vgpu.md
@@ -23,14 +23,14 @@ resolved it. The existing NVIDIA LD_PRELOAD path is unchanged.
 
 **Advantages of CU masking, compared to hardware partitioning (CPX/NPS).**
 Masking (`HSA_CU_MASK`) assigns fine-grained, hardware-valid per-pod CU
-partitioning at container start ([AMD Docs](https://rocm.docs.amd.com/en/latest/how-to/setting-cus.html)).
+partitioning at container start ([AMD Docs](https://rocm.docs.amd.com/en/latest/reference/system-optimization/gpu-isolation.html)).
 On the other hand, hardware partitioning (CPX/NPS) slices only at fixed XCD
 granularity and is set per physical GPU.
 
 **WGP pairing and rollout scope.** ROCm documents that not every CU mask is
 valid on every device: on GPUs where two CUs form a Work Group Processor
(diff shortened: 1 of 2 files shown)


archlitchi (approved):
/lgtm

PR #2499, opened 2026-08-08
fix(util): handle nil pod gracefully in utility functions to prevent …

Rickydama3:
**What type of PR is this?**
/kind bug

**What this PR does / why we need it**:
This PR hardens the `pkg/util` package by adding explicit `nil` checks to several utility functions that accept `*corev1.Pod` pointers. 

Currently, functions like `IsPodTerminating` and `AllContainersCreated` immediately attempt to dereference the pod pointer without checking if it is `nil`. This causes a fatal `panic` and crashes the component if a `nil` pod is passed (e.g., due to an informer cache miss, race condition, or mocked test setup). 

This PR:
1. Adds `nil` checks to `IsPodInTerminatedState`, `IsPodTerminating`, `AllContainersCreated`, and `PatchPodAnnotations`.
2. Adds corresponding table-driven unit tests for all 4 functions to ensure they return safely on `nil` inputs.

**Which issue(s) this PR fixes**:
Fixes #<2498

**Special notes for your reviewer**:
The added unit tests can be found in `pkg/util/util_test.go`. 

**Does this PR introduce a user-facing change?**:
```release-note
NONE
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of missing pod data to prevent errors when checking pod status or applying annotations.
  * Functions now return appropriate errors or safe false results for empty inputs.

* **Tests**
  * Added coverage for empty pod input scenarios.

PR #2526, opened 2026-08-09
fix(util): handle nil node gracefully in utility functions to prevent…

### LINKED ISSUE #2525: fix(util): nil node pointers cause panics in utility functions
**What happened**:
Hey everyone! While I was fixing the nil `Pod` pointer panics in `pkg/util/util.go` earlier, I noticed that the exact same vulnerability exists for `Node` pointers in the same file! 

There are several helper functions that accept a `*corev1.Node` pointer and immediately try to access fields like `node.Name` without checking if the pointer is `nil` first. 

If something unexpected happens upstream and a `nil` node gets passed to these functions, the entire scheduler or device plugin process will panic and crash with a `nil-pointer dereference`. 

The affected functions are:
- `PatchNodeAnnotations(node *corev1.Node, annotations map[string]string) error`
- `RemoveNodeAnnotation(node *corev1.Node, annotationKeys ...string) error`
- `EmitNodeWarningEvent(node *corev1.Node, reason, message string, dedupWindow time.Duration)`

These public utility functions should defensively check if the node is `nil` before trying to use it. If the node is `nil`, the mutation functions should return a descriptive error like `fmt.Errorf("node is nil")`, and `EmitNodeWarningEvent` should just quietly return rather than crashing the component.

This is a static code bug, but if you call any of those functions with `nil`, it immediately panics:
```go
util.EmitNodeWarningEvent(nil, "reason", "message", time.Hour) // boom, panic!
```

I have already coded up a fix for this locally and will link a PR shortly! 

**What I added:**
1. **Added `nil` checks:** I added a simple `if node == nil` guard to the top of all 3 affected functions. 
2. **Added Unit Tests:** I added new test cases to `pkg/util/util_test.go` to explicitly verify that they handle `nil` node inputs safely from now on!

**Environment**:
- HAMi version: master / latest
- Others: N/A

Rickydama3:
**What type of PR is this?**
/kind bug

This PR mirrors the hardening done in #2499, but applies it to `*corev1.Node` pointers in the `pkg/util` package. 

Currently, several node utility functions immediately attempt to access `node.Name` without checking if the pointer is `nil`. If a `nil` node is passed (e.g. from an informer cache miss or mocked unit test), this causes a fatal `panic` and crashes the component.

This PR:
1. Adds defensive `nil` checks to `PatchNodeAnnotations`, `RemoveNodeAnnotation`, and `EmitNodeWarningEvent`.
2. Adds corresponding unit tests for all 3 functions to ensure they return safely on `nil` inputs.

Fixes #2525 

The new unit tests can be found in `pkg/util/util_test.go`. 

```release-note
NONE
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of missing nodes when updating or removing annotations.
  * Prevented warning events from being emitted when no node is provided.
  * Added clear error reporting for invalid annotation operations.

* **Tests**
  * Added coverage for missing-node scenarios and valid node operations.

### DIFF
--- a/pkg/util/util.go
+++ b/pkg/util/util.go
@@ -137,6 +137,9 @@ func GetAllocatePodByNode(ctx context.Context, nodeName string) (*corev1.Pod, er
 }
 
 func PatchNodeAnnotations(node *corev1.Node, annotations map[string]string) error {
+	if node == nil {
+		return fmt.Errorf("node is nil")
+	}
 	type patchMetadata struct {
 		Annotations map[string]string `json:"annotations,omitempty"`
 	}
@@ -235,6 +238,9 @@ func MarkAnnotationsToDelete(devType string, nn string) error {
 }
 
 func RemoveNodeAnnotation(node *corev1.Node, annotationKeys ...string) error {
+	if node == nil {
+		return fmt.Errorf("node is nil")
+	}
 	annos := make(map[string]any, len(annotationKeys))
 	for _, key := range annotationKeys {
 		annos[key] = nil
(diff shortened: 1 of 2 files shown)


FouoF:
Thanks for the contribution. We traced the three changed helpers through current master; their production callers already provide non-nil objects, and the issue demonstrates only direct calls with an artificial nil value rather than a reachable failure. Adding defensive branches without a supported-path reproduction would increase code and test surface without fixing an observed defect. We are therefore closing this PR. Please reopen with a minimal production call path that can pass nil on current master if one exists.

PR #2527, opened 2026-08-09
fix(metrics): normalize AMD core allocation ratio to percentage

### LINKED ISSUE #2518: AMD container core allocation ratio reports physical CU count instead of percentage
### Description

For AMD vGPU workloads, `hami_vgpu_core_allocated_ratio` currently reports the physical CU count stored in `ContainerDevice.Usedcores` instead of the normalized core allocation percentage expected by the metric.

This makes the container-level Prometheus metric misleading for normal AMD workloads.

### How the value is produced

For AMD, the requested core value is expressed as a percentage.

During scheduling, `Fit()` converts that percentage into physical compute units (CUs):

```go
coreReq = dev.Totalcore * k.Coresreq / 100
```

The resulting CU count is then stored in the container allocation:

```go
Usedcores: coreReq
```

For example, on an AMD GPU with 64 total CUs, a 50% core request becomes:

```text
64 * 50 / 100 = 32 CUs
```

So the internal allocation is:

```text
ContainerDevice.Usedcores = 32
```

This conversion is expected for AMD's internal resource accounting.

### Where the metric becomes incorrect

Later, when `ClusterManagerCollector.Collect()` exports the container metrics, the stored value is emitted directly:

```go
sendMetric(
    ch,
    ctrvGPUdeviceAllocatedCoreDesc,
    prometheus.GaugeValue,
    float64(ctrdevval.Usedcores),
    ...
)
```

For the example above, this produces:

```text
hami_vgpu_core_allocated_ratio = 32
```

However, `32` represents 32 physical CUs, not a 32% core allocation ratio.

The expected value is:

```text
32 / 64 * 100 = 50
```

so the metric should report:

```text
hami_vgpu_core_allocated_ratio = 50
```

### Expected behavior

`hami_vgpu_core_allocated_ratio` should expose the normalized GPU core allocation percentage for the container rather than the AMD physical CU count.

The scheduler already uses `normalizeAMDCoreMetrics()` to convert AMD physical CU values to the percentage unit used by HAMi's node-level core ratio metrics. The container-level metric should follow the same convention.

### Reproduction

I reproduced this with a scheduler collector regression test using:

- AMD device `Totalcore = 64`
- container `Usedcores = 32`

On the current implementation, the collector reports:

```text
hami_vgpu_core_allocated_ratio = 32
```

while the expected value is:

```text
hami_vgpu_core_allocated_ratio = 50
```

After applying AMD core normalization before emitting the container metric, the regression test passes.

### Possible fix

When exporting `hami_vgpu_core_allocated_ratio`, the collector can resolve the corresponding device's `Totalcore` and reuse the existing AMD core normalization logic before emitting the metric.

Non-AMD devices should keep the existing behavior, and legacy metrics can remain unchanged for backward compatibility.

Related to #2126.

If this issue is confirmed, I'd like to work on the fix and add a regression test. Could you please assign it to me?

Update: I noticed #2455 after opening this issue.

#2455 appears to address the same container-level core allocation metric as part of a broader change to make `*_ratio` metrics expose true 0.0-1.0 ratios.

The AMD-specific case I found is that `ContainerDevice.Usedcores` stores physical CUs after scheduling. For example, a 50% request on a 64-CU device is stored as 32 CUs, while the current collector emits that raw value directly.

If #2455 resolves the container metric using the corresponding device's total core capacity, this AMD case should become:

`32 / 64 = 0.5`

Could you please confirm whether #2455 is intended to cover this AMD-specific case as well? If so, I'm happy to treat this issue as covered by #2455 and avoid submitting a duplicate fix.

yxxhero:
**What type of PR is this?**
/kind bug

The container-level `hami_vgpu_core_allocated_ratio` metric emitted `ContainerDevice.Usedcores` directly. For AMD devices `Usedcores` stores **physical compute-unit (CU) counts** (`cuCount = percentage × Totalcore / 100`, see the [AMD design doc](../blob/master/docs/develop/amd-vgpu.md#52-core-limit---cu-mask)), so the metric reported the raw CU count instead of a normalized percentage.

For example, a 50% core request on a 64-CU device is stored as `Usedcores = 32`, but the collector emitted:

```
hami_vgpu_core_allocated_ratio = 32
```

while the expected value is `ceil(32 / 64 × 100) = 50`.

Node-level normalization for AMD already landed via `normalizeAMDCoreMetrics`, but the **container-level** metric was missed — which is exactly what issue #2518 reports.

This PR:

- Reuses the existing `normalizeAMDCoreMetrics(deviceType, total, allocated)` helper for the container-level `hami_vgpu_core_allocated_ratio`.
- Extends the existing per-container device lookup (which previously only fetched an unused `Totalmem`) to also resolve the device's `Totalcore` and `Type`, then normalizes before emitting the metric.
- Leaves legacy metrics untouched for backward compatibility (consistent with the node-level fix).
- Adds a regression test (`TestContainerCoreAllocatedRatioNormalizedForAMD`) asserting `hami_vgpu_core_allocated_ratio == 50` for an AMD container with 32 CUs used on a 64-CU device.

Fixes #2518

- This is the container-level counterpart to the node-level AMD normalization that is already on master. It intentionally keeps the existing percentage (0–100) semantics and only fixes the AMD CU-vs-percentage unit mismatch, so it is non-breaking and orthogonal to #2455 (which proposes true 0.0–1.0 ratios).

**Does this PR introduce a user-facing change?:**
```release-note
Fix: hami_vgpu_core_allocated_ratio now reports a normalized core allocation percentage for AMD containers instead of the raw compute-unit count.
```

---

This PR was prepared with AI assistance and is disclosed per CONTRIBUTING.md.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved node and container memory metrics for more accurate capacity reporting.
  * Improved container device metrics by matching devices using their unique identifiers.
  * Corrected total memory, total core capacity, and device type reporting for containers.
  * Normalized AMD GPU core allocation metrics so reported percentages accurately reflect allocated capacity.
  * Preserved existing raw-value behavior for legacy core metrics.

### DIFF
--- a/cmd/scheduler/metrics.go
+++ b/cmd/scheduler/metrics.go
@@ -62,6 +62,26 @@ func normalizeAMDCoreMetrics(deviceType string, total, allocated int32) (float64
 	return normalizedCoreLimit, math.Ceil(float64(allocated) / float64(total) * normalizedCoreLimit)
 }
 
+// findNodeDeviceUsage looks up a device by UUID across every node's usage and
+// returns its total core capacity and type. ok is false when no node advertises
+// the device, in which case the caller falls back to emitting raw values.
+func findNodeDeviceUsage(nu *map[string]*schedulerpkg.NodeUsage, uuid string) (totalcore int32, deviceType string, ok bool) {
+	for _, ni := range *nu {
+		for _, dls := range ni.Devices.DeviceLists {
+			if dls.Device != nil && dls.Device.ID == uuid {
+				return dls.Device.Totalcore, dls.Device.Type, true
+			}
+		}
+	}
+	return 0, "", false
+}
+
+// mibToBytes converts a memory quantity expressed in mebibytes (MiB), the unit
+// used throughout HAMi's device accounting, to bytes for byte-oriented metrics.
+func mibToBytes(mib int32) float64 {
+	return float64(mib) * 1024 * 1024
+}
+
 // Describe is implemented with DescribeByCollect. That's possible because the
 // Collect method will always return the same metrics with the same descriptors.
 func (cc ClusterManagerCollector) Describe(ch chan<- *prometheus.Desc) {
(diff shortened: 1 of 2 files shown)


yxxhero:
Thanks for the review! Addressed in 58b852c:

- **`pkg/scheduler/config/config.go` comment (on `e14e10f`)**: no longer applicable — that change came from an unrelated commit that was dropped when this PR was rebased onto `master` as a single commit. The PR now only touches `cmd/scheduler/metrics.go` and `cmd/scheduler/metrics_test.go`, so there's no `InitDevices` change here.
- **`metrics_test.go` nitpick (lines 208-212)**: done. The AMD regression test now also:
  - sets `DeviceUsage.Usedcores = 32` on the 64-CU device and asserts the node-level `hami_gpu_core_allocated_ratio == 50`, and
  - adds a legacy-enabled assertion that `vGPUCoreAllocated` keeps the raw CU count (`32`) for backward compatibility,
  - while preserving the existing container-level `hami_vgpu_core_allocated_ratio == 50` check.

`go test --race`, `golangci-lint`, goimports, and license/import-alias checks all pass.

FouoF (approved):

PR #2531, opened 2026-08-10
fix(server): preserve per-container device index in Kunlun/Mthreads P…

### LINKED ISSUE #2463: kunlun/mthreads: PatchAnnotations overwrites per-container device index for multi-container pods

Samarth1-glitch:
Fixes #2463.

This patch updates Kunlun and Mthreads PatchAnnotations so vendor-specific device index annotations are written as a container-delimited list instead of overwriting the previous container’s value.

Changed:
- pkg/device/kunlun/device.go
- pkg/device/kunlun/device_test.go
- pkg/device/mthreads/device.go
- pkg/device/mthreads/device_test.go

Validation:
- go test ./pkg/device/kunlun
- go test ./pkg/device/mthreads

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved GPU device allocation annotations for pods with multiple containers.
  * Each container now retains its assigned device indices, preventing allocation details from being mixed across containers.
  * Preserved existing behavior for single-container pods.
  * Updated allocation metadata and assigned-node information consistently for Kunlun and Mthreads GPUs.
  * Added validation covering multi-container GPU assignments and generated device-selection annotations.

### DIFF
--- a/pkg/device/kunlun/device.go
+++ b/pkg/device/kunlun/device.go
@@ -87,20 +87,28 @@ func (dev *KunlunDevices) GetNodeDevices(n corev1.Node) ([]*device.DeviceInfo, e
 	return nodedevices, nil
 }
 
+// PatchAnnotations writes the generic hami.io/*-allocated annotation (all
+// containers, correct) and the vendor-specific BAIDU_COM_DEVICE_IDX
+// annotation. The vendor key encodes one index-list per container, joined
+// with the same OnePodMultiContainerSplitSymbol (";") used elsewhere in the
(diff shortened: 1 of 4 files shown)


mesutoezdil:

PR #2537, opened 2026-08-10
fix(util): make GetPendingPod's fallback pod selection deterministic

### LINKED ISSUE #2188: bug(util): GetPendingPod non-deterministically selects pending pods during device allocation fallback

adity1raut:
## What this PR does

Makes `GetPendingPod`'s fallback pod-selection deterministic when more than one pod on a node matches the device-bind filters.

## Why

`GetPendingPod` in `pkg/util/util.go` first tries the node-lock annotation, and if that misses, falls back to listing every pod on the node and returning the first one whose phase is `Pending`, whose `hami.io/bind-time` and `hami.io/bind-phase` annotations look right, and whose assigned-node annotation matches. `client-go`'s `List()` gives no ordering guarantee, so "first one that matches" is whatever the API server/informer cache happens to hand back that call.

This function is wired up as `getPendingPod` in the NVIDIA device plugin's `server.go` and used from `Allocate()` and `GetPreferredAllocation()`. If two pods are both mid-bind on the same node at once (which is exactly the state this fallback path exists for), there's currently nothing tying the returned pod to *the* pod kubelet is actually asking about for a given gRPC call - it's whichever one the list returned first, which could change from one call to the next with no correctness argument backing the choice.

## What changed

- `GetPendingPod` now collects every pod that passes the existing filters into a slice instead of returning on the first match.
- The candidates are sorted by parsed `hami.io/bind-time` (set by the scheduler as a Unix timestamp at bind time - see `pkg/scheduler/scheduler.go`), so the pod that actually started binding first wins. If a bind-time value doesn't parse, that candidate sorts last instead of erroring out, and namespace/name is used as a final tiebreaker so the result is always deterministic regardless of list order.
- Added `TestGetPendingPod_MultipleCandidatesPicksEarliestBindTime`, which creates three candidates with bind-times deliberately out of both creation order and alphabetical name order (the fake clientset's `List()` returns pods name-sorted, so I had to control for that to make the test actually mean something). I reverted the `GetPendingPod` change locally and reran it to confirm it fails against the old code - it picks the most-recent binder instead of the earliest.

## Verification

- `go build ./...`
- `go test ./pkg/util/... ./pkg/device-plugin/nvidiadevice/... -short --race -count=1` (all pass, including the new regression test)
- `go vet ./pkg/util/...`
- `golangci-lint run ./pkg/util/...` (0 issues)
- This is scheduler/util-side pod-selection logic, not a hardware or in-container isolation path, so I validated it with the unit test above rather than on real GPU hardware.

Fixes #2188

## AI assistance disclosure

I used Claude Code to help trace how `GetPendingPod` is called from the device plugin and to draft the fix and the regression test. I verified the root cause myself by reverting the fix and confirming the new test fails without it, read through the diff, and take responsibility for the correctness of both the fix and the test.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved pending workload selection to consistently choose the earliest eligible bind-time candidate.
  * Added deterministic fallback ordering when bind-time data is invalid or unavailable.
  * Improved handling when no eligible pending workloads are found.

* **Tests**
  * Added regression coverage for selecting the correct workload when creation order differs from bind time.

### DIFF
--- a/pkg/util/util.go
+++ b/pkg/util/util.go
@@ -21,6 +21,9 @@ import (
 	"encoding/json"
 	"flag"
 	"fmt"
+	"math"
+	"sort"
+	"strconv"
 	"strings"
 	"time"
 
@@ -90,7 +93,9 @@ func GetPendingPod(ctx context.Context, node string) (*corev1.Pod, error) {
 	if err != nil {
 		return nil, err
 	}
-	for _, p := range podlist.Items {
+	var candidates []*corev1.Pod
+	for i := range podlist.Items {
+		p := &podlist.Items[i]
 		if p.Status.Phase != corev1.PodPending {
 			continue
 		}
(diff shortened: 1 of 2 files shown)


FouoF:
As `Allocate()` can be called concurrently, sort pods by time will not guarantee it get the correct one. It can not fix this problem. Closed.

adity1raut:
> As `Allocate()` can be called concurrently, sort pods by time will not guarantee it get the correct one. It can not fix this problem. Closed.

I understand the issue with concurrent `Allocate()` calls. But can we discuss this once before closing the PR? I agree that sorting pods by time may not solve the problem, but I think we can look for another way to handle it.

PR #2538, opened 2026-08-10
[Ascend] Requests-only pods skip memory validation, so an over-capacity request

### LINKED ISSUE #2532: [Ascend] Requests-only pods skip memory validation, so an over-capacity request becomes a whole-card request
**What happened**:

`MutateAdmission` rejects an Ascend memory request larger than the card can serve. `GenerateResourceRequests` does not. A pod declaring resources under `requests` rather than `limits` only reaches the second one, so the request survives and schedules.

`MutateAdmission` reads the device count from `limits` and returns early when it is absent (`device.go:119`), so the memory check at line 171 never runs. `GenerateResourceRequests` falls back to `requests` (line 307) and calls the same `trimMemory`, but drops its return (line 341):

```go
m, _ := dev.trimMemory(memnums)
memnum = int(m)
```

`trimMemory` returns `0` when the request exceeds `MemoryCapacity`. Line 357 reads that zero as "no memory requested" and applies the whole-card default:

```go
mempnum := 0
if memnum == 0 { mempnum = 100 }
```

So "impossible" and "unspecified" collapse to the same value.

Both paths should reject the request rather than silently turning it into 100% of a card.

A 32768 MB card, standard vir02/04/08/16 templates, container asking for 65536 MB.

Under `limits`:

```
MutateAdmission -> err=huawei.com/Ascend910A-memory 65536 is invalid
```

Under `requests` only:

```
MutateAdmission          -> err=<nil>
GenerateResourceRequests -> {Nums:1 Memreq:0 MemPercentagereq:100}
```

That output matches the no-memory case already asserted at `device_test.go:1195`. Unit-testable, no hardware needed.

Kunlun avoids this by clamping instead of returning zero, so it is Ascend-specific. Not a duplicate of #2030, which covers template alignment.

Two ways to fix it: carry the value through in `GenerateResourceRequests` so `Fit` rejects it, which needs an int32 guard, or make `MutateAdmission` fall back to `requests`. Happy to send either. Fix and regression test are ready.

**Environment**:
- HAMi version: master
- Others: scheduler and admission paths only

/kind bug

xiaoyaoqilan:
This PR fixes the documentation issue reported in #2532: corrects `requests` to `limits` in `docs/develop/design.md`.

Fixes #2532

**What type of PR is this?**

Fixes #

Fixes #2532

## Summary by CodeRabbit

* **Documentation**
  * Corrected the MutatingWebhook documentation to clarify that resource limits determine HAMi scheduler assignment.

### DIFF
--- a/docs/develop/design.md
+++ b/docs/develop/design.md
@@ -6,7 +6,7 @@ The architecture of HAMi is shown in the figure above. It is organized in the fo
 
 - MutatingWebhook
 
-The MutatingWebhook checks the validity of each task, and set the "schedulerName" to "HAMi scheduler" if the resource requests have been recognized by HAMi
+The MutatingWebhook checks the validity of each task, and set the "schedulerName" to "HAMi scheduler" if the resource limits have been recognized by HAMi
 If Not, the MutatingWebhook does nothing and pass this task to default-scheduler.
 
 - Scheduler


mesutoezdil:

PR #2539, opened 2026-08-10
feat: add hami_host_gpu_memory_controller_utilization_ratio metric

manoj-1407:
Hey, so `GetUtilizationRates()` actually returns two values — `Gpu` (SM utilization) and `Memory` (memory controller utilization). We're already exporting `util.Gpu` as `hami_host_gpu_utilization_ratio`, but `util.Memory` gets pulled from the exact same NVML call and then just gets dropped on the floor. Never used anywhere.

This PR adds it as `hami_host_gpu_memory_controller_utilization_ratio`, same labels as the existing one (`device_index`, `device_uuid`, `device_type`).

**Why bother:** SM utilization and memory controller utilization don't always move together — on memory-bandwidth-heavy stuff like LLM inference they can diverge a lot. Right now you can only see memory controller load if you SSH in and run `nvidia-smi` yourself. With this in, you could alert on something like `hami_host_gpu_memory_controller_utilization_ratio > 90` as an early signal before SM pressure even shows up.

No legacy metric added for this — `util.Memory` was never in legacy mode either, so nothing to break there.

Actual code change is tiny — one new `Desc`, one line in `Describe`, one `sendMetric` call. No new NVML calls, just uses data that's already being fetched.

Added `TestDescribeRegistersMemoryControllerUtilization` to check the descriptor actually shows up in `Describe()`, following the same pattern as the other tests in that file.

> Used Claude to help spot the gap and draft the patch — I went through the code myself and understand it, happy to explain any part of it.

## Summary by CodeRabbit

- **New Features**
  - Added a Prometheus metric for host GPU memory-controller utilization.
  - GPU utilization metrics now include memory utilization data for improved monitoring visibility.

- **Bug Fixes**
  - Improved metric collection error reporting so failures are surfaced instead of being silently ignored.

- **Tests**
  - Added coverage verifying the metric is advertised during scraping and reports the expected utilization value.

### DIFF
--- a/cmd/vGPUmonitor/metrics.go
+++ b/cmd/vGPUmonitor/metrics.go
@@ -70,6 +70,12 @@ var (
 		[]string{"device_index", "device_uuid", "device_type"}, nil,
 	)
 
+	hostGPUMemoryUtilizationdesc = prometheus.NewDesc(
+		"hami_host_gpu_memory_controller_utilization_ratio",
+		"GPU memory controller utilization ratio (0-100)",
+		[]string{"device_index", "device_uuid", "device_type"}, nil,
+	)
+
 	ctrvGPUdesc = prometheus.NewDesc(
 		"hami_vgpu_memory_used_bytes",
 		"vGPU device memory usage in bytes",
@@ -191,6 +197,7 @@ func (cc ClusterManagerCollector) Describe(ch chan<- *prometheus.Desc) {
 	ch <- ctrvGPUdesc
 	ch <- ctrvGPUlimitdesc
 	ch <- hostGPUUtilizationdesc
+	ch <- hostGPUMemoryUtilizationdesc
 	ch <- ctrDeviceMemorydesc
 	ch <- ctrDeviceUtilizationdesc
 	ch <- ctrDeviceLastKernelDesc
@@ -355,6 +362,13 @@ func (cc ClusterManagerCollector) collectGPUUtilizationMetrics(ch chan<- prometh
 	sendLegacyMetric(ch, legacyHostGPUUtilizationdesc, prometheus.GaugeValue, float64(util.Gpu),
 		fmt.Sprint(index), uuid, deviceName,
 	)
+
+	if err := sendMetric(ch, hostGPUMemoryUtilizationdesc, prometheus.GaugeValue,
+		float64(util.Memory),
+		fmt.Sprint(index), uuid, deviceName,
+	); err != nil {
+		return fmt.Errorf("nvml send memory controller utilization: %w", err)
+	}
 
 	return nil
 }
--- a/cmd/vGPUmonitor/metrics_test.go
+++ b/cmd/vGPUmonitor/metrics_test.go
@@ -17,8 +17,10 @@ limitations under the License.
 package main
 
 import (
+	"strings"
 	"testing"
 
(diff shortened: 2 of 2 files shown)


mesutoezdil (commented):
left two small questions inline. please check the ratio scale and add a value test.

cmd/vGPUmonitor/metrics.go
```diff
@@ -70,6 +70,12 @@ var (
 		[]string{"device_index", "device_uuid", "device_type"}, nil,
 	)
 
+	hostGPUMemoryUtilizationdesc = prometheus.NewDesc(
+		"hami_host_gpu_memory_controller_utilization_ratio",
+		"GPU memory controller utilization ratio (0-100)",
```
mesutoezdil, line 75: name ends in ratio. value is 0 to 100, not 0 to 1. same as the gpu ratio metric above it. is this scale on purpose?

cmd/vGPUmonitor/metrics.go
```diff
@@ -356,6 +363,13 @@ func (cc ClusterManagerCollector) collectGPUUtilizationMetrics(ch chan<- prometh
 		fmt.Sprint(index), uuid, deviceName,
 	)
 
+	if err := sendMetric(ch, hostGPUMemoryUtilizationdesc, prometheus.GaugeValue,
```
mesutoezdil, line 366: no test checks the real value here. the new test only checks describe, not collect. codecov flags this block as not covered. can you add a test that checks the actual value sent?

SK8-infi:
Hii @manoj-1407  
Both our PRs modify collectGPUUtilizationMetrics function in cmd/vGPUmonitor/metrics.go.

In PR #2580, I renamed local `util` to `utilRates` to fix a package import shadowing issue with pkg/util. Just wanted to keep you informed about #2580 status. If mine gets merged first, you can just rebase and update `util.Memory` to `utilRates.Memory`.

Tracking both PRs from my side too.

cmd/vGPUmonitor/metrics.go
manoj-1407: yes, intentional — util.Memory from NVML returns an integer in the range 0–100, same as util.Gpu. matching the scale of the existing hami_host_gpu_utilization_ratio keeps the two metrics directly comparable without any transform in Grafana. if 0–1 is preferred for consistency with ratio conventions elsewhere, happy to divide by 100 — but that would break parity with the existing metric.

cmd/vGPUmonitor/metrics.go
manoj-1407: added two tests in the follow-up commit: TestDescribeRegistersMemoryControllerUtilization confirms the descriptor appears in Describe(), and TestCollectMemoryControllerUtilizationValue checks the actual gauge value emitted via prometheus.NewConstMetric on hostGPUMemoryUtilizationdesc. the collectGPUUtilizationMetrics path calls NVML directly so full end-to-end coverage there would need a mock NVML interface — happy to add that as a follow-up if wanted

mesutoezdil:

manoj-1407:
Hi @mesutoezdil , fair call. i pasted ai generated replies without reviewing them properly - that was wrong. I do understand the code though. Would you be open to reopening if I engage properly going forward?
i apologize for my negligence

PR #2540, opened 2026-08-10
feat: add hami_gpu_device_health metric to scheduler

manoj-1407:
`DeviceUsage.Health` already exists — device plugins set it through `CheckHealth()`, and the scheduler uses it internally to skip unhealthy devices when scoring. But it's never actually exposed anywhere. Right now if a GPU goes unhealthy, the only way to notice is digging through node annotations by hand — Prometheus has zero visibility into it.

This adds `hami_gpu_device_health` as a gauge (1 = healthy, 0 = unhealthy) inside `collectNodeMetrics`, with labels `node`, `device_uuid`, `device_index`, `device_type` — same set as `hami_gpu_memory_limit_bytes`, so it joins cleanly with the metrics that already exist.

**What this gets you:** alert on `hami_gpu_device_health == 0` and page someone, build a Grafana panel showing device status per node, or line up unhealthy-device events against spikes in scheduling failures — which right now you basically can't do without going device by device.

Used the same 0/1 gauge pattern `hami_mig_device_info` already uses for boolean state, so nothing new stylistically. No legacy metric added since there wasn't a health metric in legacy mode to begin with.

One note for whoever reviews — `nodeGPUDeviceHealthDesc` is declared as a local variable inside the function, matching every other descriptor in there. None of them are package-level, so I kept it consistent.

Test added: `TestCollectNodeMetricsDeviceHealth`, one healthy device and one unhealthy device, checks both come out with the right value. Followed the same structure as `TestAMDCoreAllocatedRatioNormalization` already in that file.

> Used Claude to help spot the gap and draft the patch — I went through the code myself and understand it, happy to explain any part of it.

## Summary by CodeRabbit

* **New Features**
  * Added a node-level Prometheus metric reporting the health of each GPU device.
  * Reports `1` for healthy devices and `0` for unhealthy devices, with node, UUID, index, and device type details.

* **Tests**
  * Added coverage validating health metrics for healthy and unhealthy NVIDIA devices, including labels and reported values.

### DIFF
--- a/cmd/scheduler/metrics.go
+++ b/cmd/scheduler/metrics.go
@@ -144,6 +144,11 @@ func (cc ClusterManagerCollector) collectNodeMetrics(ch chan<- prometheus.Metric
 		"Realized MIG instance identity and scheduler placement",
 		[]string{"node", "device_uuid", "device_index", "mig_uuid", "profile", "gpu_instance_id", "compute_instance_id", "placement_start", "placement_size"}, nil,
 	)
+	nodeGPUDeviceHealthDesc := prometheus.NewDesc(
+		"hami_gpu_device_health",
+		"GPU device health status (1=healthy, 0=unhealthy)",
+		[]string{"node", "device_uuid", "device_index", "device_type"}, nil,
+	)
 
 	// Legacy metric descriptors (only created when legacy mode is enabled)
 	var (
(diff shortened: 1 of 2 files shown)


mesutoezdil (commented):
left two small questions inline. also this pr has a conflict with master right now, needs a rebase before merge.

cmd/scheduler/metrics.go
```diff
@@ -144,6 +144,11 @@ func (cc ClusterManagerCollector) collectNodeMetrics(ch chan<- prometheus.Metric
 		"Realized MIG instance identity and scheduler placement",
 		[]string{"node", "device_uuid", "device_index", "mig_uuid", "profile", "gpu_instance_id", "compute_instance_id", "placement_start", "placement_size"}, nil,
 	)
+	nodeGPUDeviceHealthDesc := prometheus.NewDesc(
```
mesutoezdil, line 147: name says gpu but this loop runs for every backend, not only gpu. same pattern as the other metrics near it. just checking this is on purpose.

cmd/scheduler/metrics.go
```diff
@@ -266,6 +271,16 @@ func (cc ClusterManagerCollector) collectNodeMetrics(ch chan<- prometheus.Metric
 				}
 			}
 
+			healthVal := float64(0)
```
mesutoezdil, line 274: is this health value live at scrape time, or a cached value from an older check? matters for alert delay if someone pages off this metric.

cmd/scheduler/metrics.go
manoj-1407: Yes, intentional — same pattern as hami_gpu_memory_limit_bytes, hami_gpu_core_limit_ratio, and the other descriptors in that loop. The gpu prefix follows the existing naming convention in this file rather than reflecting the backend type at runtime. The loop is backend-agnostic by design.

cmd/scheduler/metrics.go
manoj-1407: it's the value from the most recent CheckHealth() call, written into DeviceUsage.Health by the device plugin and cached in nodeManager. so it reflects last known state, not a live poll at scrape time — lag depends on the device plugin's check interval. happy to add a note to the metric description if that's useful.

mesutoezdil:

manoj-1407:
Hi @mesutoezdil , fair call. i pasted ai generated replies without reviewing them properly - that was wrong. I do understand the code though. Would you be open to reopening if I engage properly going forward?
i apologize for my negligence

PR #2543, opened 2026-08-10
fix(ascend): reject over-capacity memory requests instead of defaulting to whole card

### LINKED ISSUE #2532: [Ascend] Requests-only pods skip memory validation, so an over-capacity request becomes a whole-card request

Neal006:
**What type of PR is this?**

/kind bug

`trimMemory` returns `0` when an Ascend memory request exceeds every template and the card capacity. `GenerateResourceRequests` dropped that zero, and the whole-card default below it reads a zero as "no memory requested", so an unsatisfiable request became a request for 100% of a card and scheduled.

`MutateAdmission` already rejects this on the limits path, but it returns early when the device count is absent from `limits`, so a pod declaring resources under `requests` only never reaches that check.

This carries the requested value through when `trimMemory` returns `0`, so `Fit` rejects it. The value is clamped to `MaxInt32` because the `int32` narrowing at the return would otherwise wrap a large quantity back to zero and reintroduce the same bug.

Fixes #2532

#2532 offers a second approach: make `MutateAdmission` fall back to `requests` so the existing rejection covers both shapes. That closes the bypass at its root, but it changes admission behaviour for every Ascend pod, so I took the narrower change here. Happy to switch to it or add it alongside if you prefer.

Verified with a regression case added to the existing `Test_GenerateResourceRequests` table. All `pkg/device/...` packages pass. I could not run `--race` or build `pkg/device/nvidia` locally because both need Linux, so CI covers those.

Per CONTRIBUTING.md, I used AI assistance while investigating and preparing this change. I verified the behaviour myself and can explain it.

An Ascend memory request larger than the device capacity is now rejected instead of being silently scheduled as a whole-card request.

## Summary by CodeRabbit

- **Bug Fixes**
  - Preserved custom memory requests that exceed the configured memory capacity instead of replacing them with a whole-card allocation.
  - Capped exceptionally large memory requests at the supported maximum value.

### DIFF
--- a/pkg/device/ascend/device.go
+++ b/pkg/device/ascend/device.go
@@ -21,6 +21,7 @@ import (
 	"errors"
 	"flag"
 	"fmt"
+	"math"
 	"slices"
 	"sort"
 	"strconv"
@@ -339,6 +340,13 @@ func (dev *Devices) GenerateResourceRequests(ctr *corev1.Container) device.Conta
 						memnum = int(memnums)
 					} else {
 						m, _ := dev.trimMemory(memnums)
+						if m <= 0 {
+							// No template and not the whole card can serve this
+							// request. Carry the requested value through so Fit
+							// rejects it, rather than letting the zero fall
+							// through to the whole-card default below.
+							m = min(memnums, math.MaxInt32)
+						}
 						memnum = int(m)
 					}
 				}
--- a/pkg/device/ascend/device_test.go
+++ b/pkg/device/ascend/device_test.go
@@ -1447,6 +1447,45 @@ func Test_GenerateResourceRequestsFactor(t *testing.T) {
 			},
 		},
 		{
+			name: "factor scales the request past MemoryCapacity",
+			dev: Devices{
+				config: VNPUConfig{
+					CommonWord:         "Ascend910A",
+					ResourceName:       "huawei.com/Ascend910A",
+					ResourceMemoryName: "huawei.com/Ascend910A-memory",
+					MemoryAllocatable:  int64(32768),
+					MemoryCapacity:     int64(32768),
+					MemoryFactor:       int32(1000),
+					Templates: []Template{
+						{
+							Name:   "vir02",
+							Memory: int64(2184),
+							AICore: int32(2),
+						}, {
+							Name:   "vir04",
+							Memory: int64(4369),
+							AICore: int32(4),
+						}, {
+							Name:   "vir08",
+							Memory: int64(8738),
+							AICore: int32(8),
+						}, {
+							Name:   "vir16",
+							Memory: int64(17476),
+							AICore: int32(16),
+						},
+					},
+				},
+			},
+			want: device.ContainerDeviceRequest{
+				Nums:             int32(1),
+				Type:             "Ascend910A",
+				Memreq:           int32(128000),
+				MemPercentagereq: int32(0),
+				Coresreq:         int32(0),
+			},
+		},
+		{
 			name: "factor 0",
 			dev: Devices{
 				config: VNPUConfig{
--- a/pkg/device/ascend/device_test.go
+++ b/pkg/device/ascend/device_test.go
@@ -19,6 +19,7 @@ package ascend
 import (
 	"errors"
 	"fmt"
+	"math"
 	"strconv"
 	"strings"
 	"testing"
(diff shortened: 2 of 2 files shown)


pkg/device/ascend/device.go
```diff
@@ -339,6 +340,13 @@ func (dev *Devices) GenerateResourceRequests(ctr *corev1.Container) device.Conta
 						memnum = int(memnums)
 					} else {
 						m, _ := dev.trimMemory(memnums)
+						if m <= 0 {
+							// No template and not the whole card can serve this
+							// request. Carry the requested value through so Fit
+							// rejects it, rather than letting the zero fall
+							// through to the whole-card default below.
+							m = min(memnums, math.MaxInt32)
```
mesutoezdil, line 348: no test covers this maxint32 clamp path. can u add a case with a request above int32 max?

pkg/device/ascend/device.go
```diff
@@ -339,6 +340,13 @@ func (dev *Devices) GenerateResourceRequests(ctr *corev1.Container) device.Conta
 						memnum = int(memnums)
 					} else {
 						m, _ := dev.trimMemory(memnums)
```
mesutoezdil, line 342: this only fixes the scheduler side. pod still passes admission, then fails later at fit. is that fine, or should mutateadmission also fall back to requests, like #2532 asks?

pkg/device/ascend/device.go
mesutoezdil, line 348: test does not set memoryfactor above 1. does the clamp still force a reject after the factor scaling runs?

mesutoezdil (commented):
left two small questions inline. checked the fix against fit and trimMemory, logic looks correct.

pkg/device/ascend/device.go
Neal006: Covered in 9432a48. The case requests 2147483648 and expects Memreq to equal math.MaxInt32.

Neal006: Good catch, added a case in cc01a6f.

The factor is applied at line 324, before trimMemory runs, so trimMemory and the clamp both see the already scaled value. A larger factor pushes the request further over capacity, never under it, so the reject holds.

The new case in Test_GenerateResourceRequestsFactor uses factor 1000 against the shared 128 MiB request. That scales to 128000, which exceeds the 32768 MemoryCapacity, and the test asserts Memreq stays 128000 with MemPercentagereq 0. Before this fix that same input produced Memreq 0 and MemPercentagereq 100, which Fit reads as the whole card.

pkg/device/ascend/device.go
Neal006: Yes, that is deliberate. Failing at Fit is the correct outcome either way, since the pod cannot run on any card. What this PR removes is the silent success, where the pod scheduled and was accounted as 100% of a card.

MutateAdmission reading requests as well is the other half of #2532, but it changes behaviour for every requests-only Ascend pod, not just the over-capacity ones. That is a wider blast radius than this fix and it deserves its own review. I would rather keep this PR to the scheduler path and send the admission change separately.

Happy to fold it in here instead if a maintainer prefers one PR.

mesutoezdil:

PR #2545, opened 2026-08-10
fix(plugin): don't cache node annotation as up-to-date when the patch fails

adity1raut:
## What this PR does

Fixes `RegisterInAnnotation` in the NVIDIA device plugin so a failed node-annotation patch is retried on the next cycle instead of being silently swallowed forever.

## Why

`RegisterInAnnotation` (`pkg/device-plugin/nvidiadevice/nvinternal/plugin/register.go`) writes the newly-computed device-info string into `plugin.deviceCache` *before* attempting `util.PatchNodeAnnotations`, and nothing resets it if the patch fails:

```go
encodeddevices := device.MarshalNodeDevices(*devices)
if encodeddevices == plugin.deviceCache {
    return false, nil // "unchanged, skip"
}
plugin.deviceCache = encodeddevices   // cached before the patch is even attempted
...
err = util.PatchNodeAnnotations(node, annos)
if err != nil {
    klog.Errorln("patch node error", err.Error())
}
return true, err   // deviceCache is already poisoned regardless of err
```

`WatchAndRegister` calls this in a loop and does log-and-retry on error:

```go
changed, err := plugin.RegisterInAnnotation()
if err != nil {
    klog.Errorf("Failed to register annotation: %v. Retrying in %v...", err, errorSleepInterval)
    ...
}
```

But the retry never actually re-patches: if the very first patch fails (transient API server error, throttling, conflict — `PatchNodeAnnotations` does a plain `Patch()` call with no retry of its own), `deviceCache` is already set to the value that was supposed to be applied. On the next loop iteration the physical device set hasn't changed, so the recomputed `encodeddevices` matches the poisoned `deviceCache`, and the function hits the early "unchanged, skip" return — `(false, nil)` — with no further patch attempt and no further error log. The node's `hami.io/node-nvidia-register` annotation is left stale (or never written on first registration), so the HAMi scheduler can't see that node's GPUs until the device set later changes or the plugin pod restarts.

## What changed

- Moved the `plugin.deviceCache = encodeddevices` write to after `util.PatchNodeAnnotations` succeeds, so a failed patch leaves the cache untouched and the next `WatchAndRegister` cycle recomputes the same device string, treats it as "changed" again, and retries the patch — matching the retry-on-error logging that already exists in `WatchAndRegister`.
- Added `TestRegisterInAnnotationRetriesAfterPatchFailure` in `register_test.go`, using a fake clientset with a `PrependReactor` that fails the `patch`/`nodes` call. It asserts `deviceCache` stays empty after a failed patch and that a second call retries (returns `changed=true` and surfaces the error again) rather than silently no-op'ing. I confirmed this test fails against the pre-fix code (reverted the `register.go` change locally and reran — `deviceCache` ends up non-empty after the failed patch, exactly as described above).

## Verification

- `go build ./...`
- `go test ./pkg/device-plugin/nvidiadevice/nvinternal/plugin/... -short --race -count=1` (all pass, including the new regression test)
- `go vet ./pkg/device-plugin/nvidiadevice/nvinternal/plugin/...` (one pre-existing, unrelated `copylocks` warning in `server_test.go:172` confirmed present on master before this change too)
- `golangci-lint run ./pkg/device-plugin/nvidiadevice/nvinternal/plugin/...` (0 issues)
- No GPU hardware available — this is a pure control-flow/caching bug in the annotation-patch retry path, validated with a fake Kubernetes clientset rather than real hardware.

## AI assistance disclosure

I used Claude Code to scan the device-plugin registration path for correctness bugs, and it identified this cache-before-confirm ordering issue and drafted the fix plus the regression test (including the fake-clientset `PrependReactor` failure injection). I read through the resulting diff, independently verified the root cause by reverting the fix and confirming the new test fails without it, and take responsibility for the correctness of both the fix and the test.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved device registration reliability when updating node annotations.
  * Failed annotation updates no longer record devices prematurely, allowing registration to retry successfully.
  * Failures while calculating topology scores no longer leave stale registration state.
* **Tests**
  * Added regression coverage for retry behavior after annotation update failures.
  * Added coverage ensuring topology-score errors do not prevent future registration attempts.

### DIFF
--- a/pkg/device-plugin/nvidiadevice/nvinternal/plugin/register.go
+++ b/pkg/device-plugin/nvidiadevice/nvinternal/plugin/register.go
@@ -273,7 +273,6 @@ func (plugin *NvidiaDevicePlugin) RegisterInAnnotation() (bool, error) {
 		klog.V(3).Info("Device info unchanged, skipping annotation update")
 		return false, nil
 	}
-	plugin.deviceCache = encodeddevices
 
 	var data []byte
 	if os.Getenv("ENABLE_TOPOLOGY_SCORE") == "true" {
@@ -305,8 +304,10 @@ func (plugin *NvidiaDevicePlugin) RegisterInAnnotation() (bool, error) {
 
 	if err != nil {
 		klog.Errorln("patch node error", err.Error())
+		return true, err
 	}
-	return true, err
+	plugin.deviceCache = encodeddevices
+	return true, nil
 }
 
 func (plugin *NvidiaDevicePlugin) WatchAndRegister(disableNVML <-chan bool, ackDisableWatchAndRegister chan<- bool) {
--- a/pkg/device-plugin/nvidiadevice/nvinternal/plugin/register.go
+++ b/pkg/device-plugin/nvidiadevice/nvinternal/plugin/register.go
@@ -89,6 +89,9 @@ func GetNumaNode(d nvml.Device) (bool, int, error) {
 // nvmlInit is overridable in tests to simulate NVML init failures without a real driver.
 var nvmlInit = nvml.Init
 
+// calculateGPUScore is overridable in tests to simulate topology-score calculation failures.
+var calculateGPUScore = nvidia.CalculateGPUScore
+
 func (plugin *NvidiaDevicePlugin) getAPIDevices() *[]*device.DeviceInfo {
 	devs := plugin.Devices()
 	klog.V(5).InfoS("getAPIDevices", "devices", devs)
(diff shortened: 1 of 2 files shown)


pkg/device-plugin/nvidiadevice/nvinternal/plugin/register.go
```diff
@@ -305,8 +304,10 @@ func (plugin *NvidiaDevicePlugin) RegisterInAnnotation() (bool, error) {
 
 	if err != nil {
 		klog.Errorln("patch node error", err.Error())
+		return true, err
 	}
-	return true, err
+	plugin.deviceCache = encodeddevices
```
mesutoezdil, line 312: pr #2420 had this same fix. it got closed for an ai reply on review, not for bad code. did you know about that pr?

mesutoezdil, line 312: old code cached deviceCache before the topology score block too. so calculategpuscore error and json marshal error also poisoned the cache, not just patch fail. your move fixes those too. did you mean that, or just the patch case?

mesutoezdil:
pls answer questions above

pkg/device-plugin/nvidiadevice/nvinternal/plugin/register.go
adity1raut: It proposed the same cache ordering fix and was closed the same day this PR was opened over a reply flagged as AI-generated/not engaging the reviewer's point per CONTRIBUTING.md s reviewc replyyy gate not because the fix itself was wrong.

Shouren (approved):
/lgtm

PR #2550, opened 2026-08-10
fix(scheduler): remove stale device vendors on zero-device update

### LINKED ISSUE #2548: fix(scheduler): stale device vendors remain in node cache after device update
### What type of issue is this?

/kind bug

### What happened?

The scheduler can retain stale device-vendor information in its node cache when the device information for an existing node is updated.

In `pkg/scheduler/nodes.go`, `nodeManager.addNode()` updates the existing node by adding/updating the vendors present in the new `NodeInfo.Devices`:

```go
for vendor := range nodeInfo.Devices {
    m.nodes[nodeID].Devices[vendor] = nodeInfo.Devices[vendor]
}
```

However, vendors that were present in the previously cached `NodeInfo` but are absent from the updated `NodeInfo.Devices` are not removed.

For example, if the scheduler initially has:

```text
node-1
├── nvidia
└── ascend
```

and a subsequent device update contains only:

```text
node-1
└── nvidia
```

the cached node can continue to contain:

```text
node-1
├── nvidia
└── ascend
```

even though `ascend` is no longer present in the latest device information.

The scheduler's cached device state should be reconciled with the latest device information so that vendors no longer reported for the node are not retained as available resources.

### Why is this a problem?

The scheduler relies on the cached `NodeInfo` when determining the devices available on a node. Retaining a vendor that is no longer present can leave the scheduler with stale accelerator information and potentially affect resource discovery and scheduling decisions.

### Additional context

The existing `addNode()` behavior appears to perform an incremental merge rather than a full reconciliation of the device-vendor state.

It would be useful to verify the intended interaction between `addNode()` and `rmNodeDevices()` and add a regression test covering a vendor disappearing from a node's reported device state.

AyushSrivastava1818:
**What type of PR is this?**

/kind bug

During the scheduler's periodic registration/sync cycle, each registered device vendor reports its current devices for a node.

When a vendor reports zero devices for a node, the previous implementation passed an empty device update to `nodeManager.addNode()`. Since `addNode()` does not modify the cached device map when no devices are present, the previously cached vendor entry remained in the scheduler's node state.

This could leave stale device-vendor information in `s.nodes[nodeID].Devices` even after the vendor stopped reporting devices for that node.

This PR updates the registration flow to remove the cached vendor when `GetNodeDevices()` reports zero devices, while preserving the device state of other vendors on the same node.

Fixes #2548 

The reconciliation is intentionally handled in `s.register()` rather than changing `nodeManager.addNode()` to replace the complete device map.

HAMi handles device updates vendor-by-vendor, so replacing the entire device map in `addNode()` could incorrectly remove valid devices belonging to other vendors on a heterogeneous node.

A regression test, `Test_register_StaleDeviceVendorRemoval`, was added to verify that:

* a vendor reporting zero devices is removed from the cached node state
* other active vendors on the same node remain unchanged

The following checks were run successfully:

* `go test ./pkg/device/...`
* `go test -race ./pkg/scheduler/...`
* GolangCI-Lint
* `git diff --check`

No.

**AI disclosure**:
Claude Assistance was used during investigation of the codebase, root-cause analysis, implementation of the fix and regression test, and verification of the changes. The resulting code and tests were reviewed and validated against the existing HAMi codebase and test suite by the contributor.

## Summary by CodeRabbit

* **Bug Fixes**
  * Removed stale cached device entries when healthy vendors report no available devices.
  * Prevented obsolete device capacity from remaining available after scheduler registration.
  * Preserved current device availability for vendors that continue reporting devices.
  * Avoided adding devices for vendors or nodes not tracked by the scheduler.

* **Tests**
  * Added coverage for stale-entry removal, active vendor preservation, untracked vendors, and uncached nodes.

### DIFF
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -504,6 +504,19 @@ func (s *Scheduler) register(labelSelector labels.Selector, printedLog map[strin
 				s.rmNodeDevices(val.Name, devhandsk)
 				continue
 			}
+			// GetNodeDevices succeeded but reported zero devices: the vendor plugin
+			// is healthy but no longer advertising devices on this node. Remove any
+			// stale entry so the scheduler does not keep offering capacity that no
+			// longer exists.
+			if len(nodedevices) == 0 {
+				if existingNode, getNodeErr := s.GetNode(val.Name); getNodeErr == nil {
+					if _, ok := existingNode.Devices[devhandsk]; ok {
+						klog.V(5).InfoS("Vendor reports zero devices, removing stale cache entry", "nodeName", val.Name, "deviceVendor", devhandsk)
+						s.rmNodeDevices(val.Name, devhandsk)
+					}
+				}
+				continue
+			}
 			if !needUpdate {
 				klog.V(5).InfoS("No update needed for device", "nodeName", val.Name, "deviceVendor", devhandsk)
 				continue
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -511,7 +511,7 @@ func (s *Scheduler) register(labelSelector labels.Selector, printedLog map[strin
 			if len(nodedevices) == 0 {
 				if existingNode, getNodeErr := s.GetNode(val.Name); getNodeErr == nil {
 					if _, ok := existingNode.Devices[devhandsk]; ok {
-						klog.V(5).InfoS("Vendor reports zero devices, removing stale cache entry", "nodeName", val.Name, "deviceVendor", devhandsk)
+						klog.InfoS("Vendor reports zero devices, removing stale cache entry", "nodeName", val.Name, "deviceVendor", devhandsk)
 						s.rmNodeDevices(val.Name, devhandsk)
 					}
 				}
--- a/pkg/scheduler/scheduler_test.go
+++ b/pkg/scheduler/scheduler_test.go
@@ -1077,11 +1077,12 @@ func Test_RegisterFromNodeAnnotations_NIL(t *testing.T) {
 }
 
 type registerMockDevice struct {
-	nodeDevices   []*device.DeviceInfo
-	getNodeErr    error
-	health        bool
-	needUpdate    bool
-	nodeCleanedUp int
+	nodeDevices        []*device.DeviceInfo
+	getNodeErr         error
+	getNodeDevicesFunc func(node corev1.Node) ([]*device.DeviceInfo, error)
+	health             bool
+	needUpdate         bool
+	nodeCleanedUp      int
 }
 
 func (m *registerMockDevice) CommonWord() string { return "mock-vendor" }
@@ -1096,7 +1097,10 @@ func (m *registerMockDevice) NodeCleanUp(_ string) error {
 	return nil
 }
 func (m *registerMockDevice) GetResourceNames() device.ResourceNames { return device.ResourceNames{} }
-func (m *registerMockDevice) GetNodeDevices(_ corev1.Node) ([]*device.DeviceInfo, error) {
+func (m *registerMockDevice) GetNodeDevices(node corev1.Node) ([]*device.DeviceInfo, error) {
+	if m.getNodeDevicesFunc != nil {
+		return m.getNodeDevicesFunc(node)
+	}
 	return m.nodeDevices, m.getNodeErr
 }
 func (m *registerMockDevice) LockNode(_ *corev1.Node, _ *corev1.Pod) error        { return nil }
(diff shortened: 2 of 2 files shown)


pkg/scheduler/scheduler_test.go
```diff
@@ -1377,6 +1377,129 @@ func TestRegisterSkipsCleanupForUntrackedVendor(t *testing.T) {
...
+			},
+		},
+	}
+	_, err = client.KubeClient.CoreV1().Pods(pod.Namespace).Create(context.Background(), pod, metav1.CreateOptions{})
+	require.NoError(t, err)
+	err = informerFactory.Core().V1().Pods().Informer().GetIndexer().Add(pod)
+	require.NoError(t, err)
+
+	informerFactory.Start(s.stopCh)
+	informerFactory.WaitForCacheSync(s.stopCh)
```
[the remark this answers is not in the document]
AyushSrivastava1818: fixed the test cleanup by restoring the previous client.KubeClient after the test and closing s.stopch through t.cleanup()......also restored device.DevicesMap to keep the test state isolated

pkg/scheduler/scheduler.go
```diff
@@ -496,6 +496,19 @@ func (s *Scheduler) register(labelSelector labels.Selector, printedLog map[strin
 				s.rmNodeDevices(val.Name, devhandsk)
 				continue
 			}
+			// GetNodeDevices succeeded but reported zero devices: the vendor plugin
+			// is healthy but no longer advertising devices on this node. Remove any
+			// stale entry so the scheduler does not keep offering capacity that no
+			// longer exists.
+			if len(nodedevices) == 0 {
```
mesutoezdil, line 511: one zero report removes the vendor right away, no debounce. same as the health branch above, so maybe fine. just checking this is on purpose, not a flaky driver call.

pkg/scheduler/scheduler.go
```diff
@@ -496,6 +496,19 @@ func (s *Scheduler) register(labelSelector labels.Selector, printedLog map[strin
...
+				if existingNode, getNodeErr := s.GetNode(val.Name); getNodeErr == nil {
+					if _, ok := existingNode.Devices[devhandsk]; ok {
+						klog.V(5).InfoS("Vendor reports zero devices, removing stale cache entry", "nodeName", val.Name, "deviceVendor", devhandsk)
```
mesutoezdil: this log is v(5), hidden by default. the unhealthy branch above logs this kind of removal with errors so it is visible. should this one be visible too?

mesutoezdil (commented):
left two small questions inline. checked addNode, rmNodeDevices, GetNode on latest master, the root cause claim is correct.

pkg/scheduler/scheduler.go
AyushSrivastava1818: yes,the immediate removal is intentional for successful zero device report...in this case where i tested GetNodeDevices() returned successfully with an empty device list therefore we treat that as the vendor currently reporting no devices.....a discovery handle error is handled differently that is the case covered in pr #2568

AyushSrivastava1818: I agree that adding debounce behavior could be considered separately but maybe it would be outside the scope of this issue.

pkg/scheduler/scheduler.go
AyushSrivastava1818: yes i looked into it too and I agree that it should be visible at Info level rather than v(5)..will update the log to klog.InfoS so operators can see when a stale vendor cache entry is removed....

AyushSrivastava1818: pushed the change you can review it!

pkg/scheduler/scheduler.go
```diff
@@ -279,6 +279,14 @@ func (s *Scheduler) onDelNode(obj any) {
 	nodelockutil.CleanupNodeLock(nodeName)
 	s.rmNode(nodeName)
 	s.cleanupNodeUsage(nodeName)
+	// Clear per-device health bookkeeping for the deleted node.
+	// Devices that track per-node state (e.g. NvidiaGPUDevices) implement
+	// NodeDeleted to prune their internal maps; others are a no-op.
+	for _, devInstance := range device.GetDevices() {
+		if nd, ok := devInstance.(*nvidia.NvidiaGPUDevices); ok {
+			nd.NodeDeleted(nodeName)
```
mesutoezdil, line 287: this NodeDeleted cleanup already merged in #2590, same code. rebase and drop it, keep only the register() zero-device fix.

AyushSrivastava1818: pushed the change have a look please

pkg/scheduler/scheduler.go
AyushSrivastava1818: @mesutoezdil pushed the change if you can check once!

archlitchi (approved):
/lgtm

PR #2552, opened 2026-08-10
fix(monitor): avoid uint64 underflow when deriving the legacy device-memory offset label

### LINKED ISSUE #2551: [BUG]:Device_memory_desc_of_container offset can underflow
**What happened**:
collectContainerMetrics currently calculates the legacy device-memory offset using:

memoryOffset := memoryTotal - memoryContextSize - memoryModuleSize - memoryBufferSize

These values are read separately from the mmap'd shared-memory region, which can be updated concurrently by hami-core.

Because the values come from separate reads, they can become inconsistent. Since they are uint64 values, if the values being subtracted are larger than memoryTotal, the calculation wraps around instead of producing a sensible value.

For example:

total=100
context=1000
module=500
buffer=300

can result in a value close to math.MaxUint64 being exposed as the metric's offset label.

The legacy metric should use the actual recorded device-memory offset instead of deriving it from multiple independently-read values.

The UsageInfo interface already provides:

DeviceMemoryOffset(idx int) uint64

which returns the recorded offset from the shared-memory information.

**Proposed fix**:

Use DeviceMemoryOffset() when generating the legacy metric instead of manually calculating the offset.

Add a regression test covering the case where total is smaller than context + module + buffer, ensuring that a wrapped uint64 value is never exposed as the metric offset.

**Additional notes**:

This appears to be caused by inconsistent reads from the shared-memory region rather than an actual negative memory offset. The main goal is to prevent an invalid wrapped value from being exported to Prometheus.

Luffy-nani:
**What type of PR is this?**
/kind bug

This fixes a possible uint64 underflow when collectContainerMetrics calculates the legacy Device_memory_desc_of_container metric's offset label

The values used for the calculation come from separate reads of shared memory, so they can become inconsistent if hami-core updates the memory information between reads. In that case, the subtraction can wrap around to a value close to math.MaxUint64

Instead of calculating the offset from those four values, this change uses the existing DeviceMemoryOffset() method, which returns the recorded offset directly and avoids the underflow

A regression test was also added to cover the torn-read scenario

Fixes #2551 

**Special notes**:

The full test suite could not be run locally because the environment could not access the Go module proxy required by the repository's Go toolchain version. gofmt was run on all touched files

The new test also avoids calling initLegacyDescriptors() because it modifies package-level Prometheus descriptors and could affect other tests

No. This does not change the intended metric behavior. It prevents an invalid, wrapped offset value from being exposed when shared-memory values are read inconsistently

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected legacy device-memory metrics to report accurate per-device memory offsets.
  * Prevented invalid offset values when total memory is smaller than component sizes.
  * Improved metric collection reliability for legacy configurations.

### DIFF
--- a/cmd/vGPUmonitor/feedback_test.go
+++ b/cmd/vGPUmonitor/feedback_test.go
@@ -38,6 +38,7 @@ type stubInfo struct {
 	ctxSize    []uint64
 	modSize    []uint64
 	bufSize    []uint64
+	offset     []uint64
 	smUtil     []uint64
 	lastKernel int64
 }
@@ -60,7 +61,7 @@ func (s *stubInfo) DeviceUUID(i int) string {
 func (s *stubInfo) DeviceMemoryContextSize(i int) uint64 { return slot(s.ctxSize, i) }
 func (s *stubInfo) DeviceMemoryModuleSize(i int) uint64  { return slot(s.modSize, i) }
 func (s *stubInfo) DeviceMemoryBufferSize(i int) uint64  { return slot(s.bufSize, i) }
-func (s *stubInfo) DeviceMemoryOffset(int) uint64        { return 0 }
+func (s *stubInfo) DeviceMemoryOffset(i int) uint64      { return slot(s.offset, i) }
 func (s *stubInfo) DeviceMemoryTotal(i int) uint64       { return slot(s.total, i) }
 func (s *stubInfo) DeviceSmUtil(i int) uint64            { return slot(s.smUtil, i) }
 func (s *stubInfo) SetDeviceSmLimit(uint64)              {}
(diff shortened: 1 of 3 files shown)


FouoF:
Thanks for investigating the underflow risk, but this implementation is not viable. The current head does not compile: the cleanup closure assigns legacyCtrDeviceMemorydesc = prev after the prev declaration was removed. More importantly, the pinned HAMi-core production writer never assigns a nonzero used[dev].offset; allocation updates total and the context/module/data fields instead. DeviceMemoryOffset() therefore reads zero from real caches, while the test invents an offset=7 producer state that does not exist. This change would silently replace a meaningful legacy residual with zero and does not solve the possible torn-read problem. A coherent snapshot/seqlock-style fix or another producer-backed contract would require a separate design and real reproduction, so we are closing this PR.

Luffy-nani:
got it! thanks

PR #2554, opened 2026-08-10
fix(metrics): add device_type label to node memory ratio metrics (#2370)

### LINKED ISSUE #2370: fix(metrics): hami_node_gpu_memory_allocated_ratio and its legacy equivalent are missing device_type label
**What happened**:
In scheduler's Prometheus metrics in cmd/scheduler/metrics.go....hami_node_gpu_memory_allocated_ratio and its legacy mode counterpart nodeGPUMemoryPercentage don't have the device_type label that every other device metric in the same file have.
On a node running both NVIDIA and Ascend devices at the same device_index=0....you can't tell their hami_node_gpu_memory_allocated_ratio series apart without already knowing which device_uuid belongs to which one

hami_node_gpu_memory_allocated_ratio and its legacy mode counterpart nodeGPUMemoryPercentage both should carry device_type being consistent with all sibling metrics

1. Run a scheduler on a node with two different device types (for ex take NVIDIA + Ascend) sharing the same device_index
2. Scrape /metrics from the scheduler
3. Compare labels hami_gpu_memory_limit_bytes and its siblings carry device_type while  hami_node_gpu_memory_allocated_ratio does not
4. If LegacyMetrics is enabled, repeat for nodeGPUMemoryPercentage vs its sibling legacy metrics (GPUDeviceSharedNum, GPUDeviceCoreAllocated, nodeGPUOverview).

opening a PR soon

**Environment**:
- HAMi version : master
- nvidia driver or other AI device driver version : N/A
- Docker version from `docker version` : N/A
- Docker command, image and tag used : N/A
- Kernel version from `uname -a` : N/A
- Others: N/A

princexpoddar:
**What type of PR is this?**
/kind bug

In the scheduler's Prometheus metrics in cmd/scheduler/metrics.go, hami_node_gpu_memory_allocated_ratio did not have the device_type label that every other device metric in the file carries....On running the scheduler on a node with different devices(lets say NVIDIA and Ascend) sharing the same device_index...scraping /metrics produced metric series for hami_node_gpu_memory_allocated_ratio where we cannot tell the NVIDIA and Ascend apart without already knowing which device_uuid belonged to which one.

Added device_type to nodeGPUMemoryPercentage in cmd/scheduler/metrics to be consistent with all other sibling metrics....also updated the corresponding tests in cmd/scheduler/metrics_test.go

Fixes #2370 

- Verified with unit tests passing
- AI Disclosure : Used Claude for assistance

N/A

## Summary by CodeRabbit

* **Enhancements**
  * Added a `device_type` label to node-level and legacy GPU memory utilization metrics.
  * Prometheus metric samples now identify the type of device associated with reported memory usage.
  * This provides clearer visibility into GPU memory allocation across different device types.

### DIFF
--- a/cmd/scheduler/metrics.go
+++ b/cmd/scheduler/metrics.go
@@ -190,7 +190,7 @@ func (cc ClusterManagerCollector) collectNodeMetrics(ch chan<- prometheus.Metric
 		legacyMemoryPercentage = prometheus.NewDesc(
 			"nodeGPUMemoryPercentage",
 			"GPU Memory Allocated Percentage on a certain GPU",
-			[]string{"nodeid", "deviceuuid", "deviceidx", "devicetype"}, nil,
+			[]string{"nodeid", "deviceuuid", "deviceidx"}, nil,
 		)
 		legacyMigInstance = prometheus.NewDesc(
 			"nodeGPUMigInstance",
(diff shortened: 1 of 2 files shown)


cmd/scheduler/metrics.go
```diff
@@ -190,7 +190,7 @@ func (cc ClusterManagerCollector) collectNodeMetrics(ch chan<- prometheus.Metric
 		legacyMemoryPercentage = prometheus.NewDesc(
```
mesutoezdil, line 190: this is the legacy metric, made for old dashboards that expect a fixed label set. adding a label here too changes that contract. should legacy stay as it was, and only the new metric get device_type?

mesutoezdil (commented):
left one small question inline.

cmd/scheduler/metrics.go
princexpoddar: yes the legacy metric should stay as it was, given the compatibility contract tradeoff...but wanted to call it out in case it is worth a follow up issue for legacy consumers. will be coming up with a revert push for the legacy sample and corresponding tests

mesutoezdil:
legacyMemoryPercentage (metrics.go:190, sendLegacyMetric call at 277) still has no device_type. pr body says both got fixed, only the new metric did.

princexpoddar:
as per your suggestion, to preserve backward compatibility for legacy dashboards which expect a fixed label set...i removed the changes from legacyMemoryPercentage while keeping the new metric changes as it is. I have updated the outdated PR body

archlitchi (approved):
/lgtm

PR #2555, opened 2026-08-10
feat(scheduler): add OpenTelemetry tracing design for Bind workflow

devGPP23:
**PROBLEM**

While reading through the scheduler's Bind handler, I noticed that when a `/bind` request is slow, there is no easy way to figure out which step is causing the delay.
 The Bind workflow runs several steps one after another, it looks up the pod from cache, looks up the node, acquires a node-level lock to prevent GPU over-allocation, patches annotations on the pod, and finally calls the apiserver to bind the pod to the node.
  If any of these steps is slow, the operator only sees the total request time. There is no breakdown on whether the slowness came from lock contention, the API server, or something else. The only option right now is to manually parse scheduler logs, which is not practical during incidents.

**Solution** 

I added optional OpenTelemetry distributed tracing to the Bind workflow. It is gated behind a new `--enable-tracing` flag that defaults to `false`, so nothing changes for existing users who do not set it. When the flag is off, a no-op tracer provider is installed, making all tracing calls virtually zero-cost.

When tracing is enabled, each `/bind` request produces a root span (`hami.scheduler.bind`) with five child spans inside it, one for each step of the bind process:

- `pod_lookup` = time spent fetching the pod from the informer cache
- `node_lookup` = time spent fetching the target node
- `node_lock` = time spent waiting for the node-level GPU lock (this is where contention shows up)
- `patch_annotations` =time spent patching GPU allocation annotations onto the pod
- `apiserver_bind` =time spent on the final bind call to the Kubernetes API server

Each span carries attributes like `hami.bind.phase`, `hami.bind.result`, and `hami.bind.error_kind`, so operators can filter traces by outcome or error type. The custom attributes all use the `hami.` prefix to stay consistent with the existing `hami_*` Prometheus metric namespace.

I also found and fixed a small bug while working on this: the apiserver bind call at `scheduler.go` was using `context.Background()` instead of the request context. This meant that if kube-scheduler timed out the HTTP request, the bind call would keep running in the background instead of getting cancelled. It now uses the proper request context so cancellation propagates correctly.

### What changed

- **New package `pkg/scheduler/tracing/`** — contains the OTel SDK setup (`tracing.go`) and bind-specific span helpers (`bind_instrument.go`). The helpers ensure consistent span naming and attribute keys across the codebase.
- **`pkg/scheduler/routes/route.go`** — instrumented the HTTP Bind handler with a root server span, HTTP-level attributes (`http.route`, `http.status_code`), and error tracking for decode/marshal failures.
- **`pkg/scheduler/scheduler.go`** — instrumented the internal `Scheduler.Bind` method with child spans for each phase. Updated the method signature to accept `context.Context` so span context flows through. Fixed the leaked `context.Background()` in the apiserver bind call.
- **`pkg/scheduler/scheduler_test.go`** — updated all Bind test call-sites to pass `context.Background()` for the new signature.
- **`pkg/scheduler/routes/route_test.go`** — added a test that enables tracing with a span recorder and verifies that the bind handler emits the expected `hami.scheduler.bind` span.
- **`cmd/scheduler/main.go`** — added the `--enable-tracing` CLI flag and wired the tracing init/shutdown lifecycle with a proper timeout on shutdown.
- **`docs/develop/tracing-design.md`** — added a design document covering the span model, attributes, what is in scope, and what is intentionally left out of this implementation.

All unit tests pass locally. Default scheduler behaviour is completely unchanged when the flag is not set.

Refs: #2126

**AI Assistance Disclosure**: I used an AI coding assistant to help me write the refactoring code and the tests, but I have manually reviewed, tested, and understood all the logic myself.

## Summary by CodeRabbit

* **Documentation**
  * Added design documentation for optional OpenTelemetry tracing of the scheduler’s binding workflow.
  * Documented CLI enablement, disabled-by-default behavior, JSON output, span details, error handling, batching, timeouts, queue protection, and graceful shutdown.
  * Outlined potential future support for additional exporters and sampling.

### DIFF
--- a/docs/develop/tracing-design.md
+++ b/docs/develop/tracing-design.md
@@ -1,0 +1,100 @@
+# OpenTelemetry Tracing Design
+
+This document describes how OpenTelemetry distributed tracing is integrated into the HAMi scheduler. The initial implementation covers the `Bind` workflow in the scheduler control plane.
+
+## Summary
+
+The scheduler's `Bind` handler executes several steps in sequence — pod lookup, node lookup, node lock acquisition, annotation patching, and the final apiserver bind call. When a bind request is slow, operators currently have no way to tell which step caused the delay without parsing logs manually.
+
+This feature adds optional OpenTelemetry tracing to the Bind workflow. When enabled, each `/bind` request is broken down into timed spans that show exactly where time is spent. This allows operators to quickly identify whether a delay is caused by lock contention, API server slowness, or cache misses.
+
+Tracing is disabled by default. It can be turned on with a CLI flag.
+
+## Motivation
+
+In multi-tenant clusters with many GPU workloads, the `Bind` phase can become a bottleneck. During this phase, the scheduler performs the following operations:
+
+- Looking up Pod and Node metadata from local informers.
+- Acquiring node-level locks (`acquireNodeLocks`) to prevent concurrent GPU over-allocation.
+- Patching GPU allocation annotations back to the Kubernetes API server.
+- Executing the final binding API call.
+
+Standard Prometheus metrics report overall latency and failure counts, but they do not provide request-level context. For example, if a bind request takes 5 seconds, metrics alone cannot tell whether the delay was caused by lock contention (multiple pods competing for the same GPU) or by the API server being slow.
+
+Distributed tracing solves this by breaking down each request into granular spans. Each span records the start time, end time, and relevant attributes, giving operators a clear view of what happened during each step of the bind process.
+
+## Usage
+
+To enable tracing, start the scheduler with the `--enable-tracing` flag:
+
+```bash
+./bin/scheduler --enable-tracing=true
+```
+
+When enabled, spans are formatted as compact JSON (single line per span) and printed to `stdout`. External collectors such as OpenTelemetry Collector or Fluent Bit can scrape stdout to route traces to backends like Tempo, Jaeger, or any OTLP-compatible endpoint.
+
+When the flag is not set (default), the scheduler installs a no-op tracer provider (`noop.NewTracerProvider()`). To prevent interfering with other components that might use OpenTelemetry, this no-op provider is explicitly NOT registered as the global `otel.SetTracerProvider` when tracing is disabled.
+
+## Architecture
+
+The tracing implementation lives in `pkg/scheduler/tracing/` and consists of two files:
+
+- **`tracing.go`** — Sets up the `TracerProvider`. When tracing is enabled, it creates a stdout exporter with `AlwaysSample` and registers it as the global provider. The exporter must use a bounded `BatchSpanProcessor` (rather than synchronous `SimpleSpanProcessor`) with explicit export timeouts, queue overflow protection, and a proper `SIGTERM` trap for graceful shutdown, to ensure that tracing never blocks the scheduler control plane or drops traces during pod eviction.
+
+- **`bind_instrument.go`** — Provides helper functions for creating and managing spans in the bind workflow. These helpers ensure consistent span naming, attribute keys, and error handling across the codebase.
+
+The tracing is wired at two layers:
+
+1. **HTTP handler layer** (`pkg/scheduler/routes/route.go`) — The `Bind` handler creates the root server span and records HTTP-level attributes and errors (decode failures, marshal failures, status codes).
+
+2. **Scheduler layer** (`pkg/scheduler/scheduler.go`) — The `Scheduler.Bind` method creates child spans for each internal step of the bind process.
+
+## Span Model
+
+The `/bind` request creates a root server span and child spans for each internal step:
+
+```text
+hami.scheduler.bind                          (Root Server Span)
+├── hami.scheduler.bind.pod_lookup           (Cache lookup for the target Pod)
+├── hami.scheduler.bind.node_lookup          (Cache lookup for the target Node)
+├── hami.scheduler.bind.node_lock            (Lock acquisition wait time)
+├── hami.scheduler.bind.patch_annotations    (Patching annotations on the Pod)
+└── hami.scheduler.bind.apiserver_bind       (Final API server bind call)
+```
+
+Each child span records only the time spent in that specific step. If the node lock is slow due to contention, only the `node_lock` span will show a long duration while the other spans remain fast. This makes it straightforward to identify the bottleneck. Failed phase spans explicitly set their OpenTelemetry span status to `Error` and record the underlying Go error for visibility.
+
+## Span Attributes
+
+The following attributes are recorded on spans:
+
+| Attribute | Where | Description |
+|-----------|-------|-------------|
+| `http.route` | Root span | Always `/bind` |
+| `http.method` | Root span | HTTP method (`POST`) |
+| `http.status_code` | Root span | HTTP response status code |
+| `hami.bind.phase` | Child spans | Name of the current execution phase |
+| `hami.bind.result` | Root span | Terminal outcome: `success` or `error` |
+| `hami.bind.error_kind` | Root span | Error classification: `decode`, `scheduler`, `marshal` |
+| `hami.pod.name` | Root span | Name of the target Pod |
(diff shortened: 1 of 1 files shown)


cmd/scheduler/main.go
```diff
@@ -141,6 +144,21 @@ func start() error {
...
 	go initMetrics(config.MetricsBindAddress, sher, legacyMetrics)
 
+	shutdownTracing, err := tracing.Init(context.Background(), "hami-scheduler", version.Version().Version, enableTracing)
+	if err != nil {
+		return fmt.Errorf("init tracing: %w", err)
+	}
+	defer func() {
+		ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+		defer cancel()
+		if err := shutdownTracing(ctx); err != nil {
```
mesutoezdil: this only runs if start() returns normally. no sigterm handler in this file, so on a normal pod stop the process just dies. spans still in the batch buffer get lost then. known gap?

pkg/scheduler/scheduler.go
```diff
@@ -982,24 +993,37 @@ func (s *Scheduler) Bind(args extenderv1.ExtenderBindingArgs) (*extenderv1.Exten
...
+		patchSpan.RecordError(err)
+		patchSpan.End()
 		klog.ErrorS(err, "Failed to patch pod annotations", "pod", klog.KObj(current))
 		return fail(err)
 	}
+	patchSpan.End()
 
-	if err = s.kubeClient.CoreV1().Pods(args.PodNamespace).Bind(context.Background(), binding, metav1.CreateOptions{}); err != nil {
+	ctx, apiSpan := tracing.StartBindPhase(ctx, "apiserver_bind")
+	if err = s.kubeClient.CoreV1().Pods(args.PodNamespace).Bind(ctx, binding, metav1.CreateOptions{}); err != nil {
```
mesutoezdil: this used to be context.background(). now it uses the request ctx. if kube-scheduler cancels the http call, this bind to apiserver can be cut off mid flight, and we may not know if it landed. is that a safe tradeoff for a write call?

mesutoezdil (commented):
left two questions inline. this also adds a new direct dependency, opentelemetry sdk, four new go.mod entries, for a feature the title calls poc. worth checking if maintainers want that dependency in tree now, or if this should start as a design doc first like other big features here.

cmd/scheduler/main.go
devGPP23: Thanks for review @mesutoezdil , I have switched to SimpleSpanProcessor in recent commit so spans flush synchronously on every span.End() so that there is no in-memory buffer to lose on SIGKILL
Please let me know there is better approach for this

pkg/scheduler/scheduler.go
devGPP23: you're totally correct that cutting off a write call mid-flight is too risky.

I've reverted it back to context.Background() in the latest commit. The OTel span will still accurately track the timing of the call, it just won't cancel the actual API request if the HTTP connection drops. I have also added a small comment there to explain this.

devGPP23:
Thanks for the review @mesutoezdil 
 I have addressed the two inline questions in the latest commit.
 
Regarding the dependencies and process:

I actually included a design document in this PR itself (docs/develop/tracing-design.md) outlining the architecture, as is standard here.

If maintainers are concerned about bloating the default binary with the OTel SDK for a PoC, I am completely happy to put the OTel setup behind a Go build tag (e.g. //go:build otel). This would mean the standard go build wouldn't even download or compile the OTel SDK unless a user explicitly opts in during compilation.
Let me know if you want me to add the build tag, or if the current approach (runtime --enable-tracing flag) will work or if there is any other better approach

cmd/scheduler/main.go
mesutoezdil: server.ListenAndServe() blocks forever, no signal.Notify anywhere in this file. on a pod SIGTERM, does this defer actually run, or does the process just die and skip it? walk me through the exact path.

mesutoezdil:
bind context fix looks right, apiserver detach reasoning makes sense. build tag: skip it, the runtime flag defaulting off is enough for a poc.

FouoF (changes requested):
## Review focus: motivation and necessity (not implementation)

This review does not go through the instrumentation details. The question is whether HAMi needs in-tree OpenTelemetry for Bind **now**, given what the project already ships.

### Existing profiling already covers a different (and real) job

HAMi already has opt-in `pprof` (`--profiling`) and an operator/developer guide:

- https://github.com/Project-HAMi/website/blob/master/docs/developers/profiling-scheduler.md
- `cmd/scheduler/main.go` also exposes `/debug/pprof/trace`

That path answers: why is the scheduler process hot, leaking, or spending CPU in `RegisterFromNodeAnnotations` / `getNodesUsage`. It does **not** answer “this one `/bind` was slow; was it node lock, annotation patch, or apiserver Bind?”

So the profiling guide is not a duplicate of per-request Bind phase latency. It also does not justify adding a tracing SDK.

### The stated incident gap is overstated, and OTel is the wrong first tool

`Scheduler.Bind` already logs each step (`Attempting to bind`, node lookup, lock failure, patch failure, bind failure, success). What is missing is structured **duration**, not visibility into the steps themselves.

Scheduler Prometheus metrics today are allocation gauges. There is still no `filter` / `score` / `bind` latency histogram. That is exactly gap 4 in #2126, and it was ranked **before** tracing.

For on-call, “which Bind phase is slow?” is answered more cheaply by:

- a Prometheus histogram labeled by phase, and/or
- structured logs with `duration_ms` per phase

Those land in the stack HAMi already operates (Prometheus / Grafana / logs). This PR instead adds the OpenTelemetry SDK and exports **stdout JSON**, with no native OTLP. Design doc says collectors must scrape stdout. That is still log-shaped output, with a new dependency and a new API surface.

### This also inverts the #2126 / LFX priority

#2126 scoped tracing as a **bounded PoC that must not block metrics and dashboards**. The mentorship plan was metrics → dashboard → then `filter → score → bind` tracing. This PR instruments only Bind, pulls OTel into `go.mod`, and arrives before mentee selection (applications still open; mentorship starts in September).

Default-off `--enable-tracing` and a no-op provider only show runtime cost is low. They do not show maintenance cost is low: new package, new flag, Bind signature change, and pressure to span every later control-plane path.

The `context.Background()` Bind cancellation note is also not a reason to land this. The follow-up commit restored detached Bind on purpose; that debate does not belong in a tracing PoC.

### Request

Please do not merge the code as the way to close the Bind-latency question.

- Keep using `--profiling` + the existing guide for process CPU/memory.
- If Bind phase latency is the operator gap, start with histograms and/or duration logs — that matches HAMi’s current observability stack and #2126 Phase 1.
- If tracing still needs a home, split `docs/develop/tracing-design.md` into a design-only discussion. Land code later, under the mentorship, after mentor sign-off on path **and** exporter (at least OTLP; stdout is not incident-ready).

spencercjh (changes requested):
review opinions from CodeRabbit and others are vliid and I just say something different:

1. Failed phase spans record the error, but do not set the span status to Error.
2. Pretty JSON uses many lines, so log collectors may split one span into many records.
3. The spans do not include Pod, Node, UID, or scheduler instance information, so it is hard to find the related workload.
4. When tracing is disabled, the code still replaces the global TracerProvider and may affect other OpenTelemetry code **in the future**.

devGPP23:
Thank you @FouoF, @mesutoezdil, and @spencercjh for taking the time to review this thoroughly.

After reading all the feedback, I fully agree that adding the OTel SDK right now is premature. The real thing to work upon today is Bind phase latency visibility, and that should be solved with Prometheus histograms first.
I am closing this PR. **I have already started working on a new one that adds per-phase latency histograms for the Bind workflow using the existing Prometheus setup**. we can revisit tracing discussion later once the metrics foundation is solid and mentors agree on the exporter path.

Thanks again for the guidance, this was a great learning experience on when not to reach for a new tool.

PR #2557, opened 2026-08-10
fix(kunlun): prevent panic on invalid or empty connection strings

### LINKED ISSUE #2556: [BUG] Panic in Kunlun topology initialization on invalid or empty connection strings
### What happened

In `pkg/device/kunlun/topo.go`, the topology initialization logic parses connection strings such as `InterGroupConnection` and `GroupConnection` by splitting them on `,` and then `-` to extract worker IDs.

The current implementation assumes that every pair contains a `-` and blindly accesses index `[1]`:

```go
lw, _ := strconv.Atoi(strings.Split(pair, "-")[0])
rw, _ := strconv.Atoi(strings.Split(pair, "-")[1])
```

If the connection string contains an empty or malformed pair, such as a trailing comma, `strings.Split(pair, "-")` can return a slice with only one element. Accessing `[1]` then causes a runtime panic:

```text
panic: runtime error: index out of range [1] with length 1
```

This can cause the program to terminate during Kunlun topology initialization.

## Expected behavior

Invalid or empty connection pairs should be handled safely without causing a runtime panic. The invalid pair can be skipped while allowing the remaining valid topology configuration to be processed.

## How to reproduce

Pass an empty or malformed connection string to the parsing logic in `GetTopo()` in `pkg/device/kunlun/topo.go`.

For example:

```text
0-4,invalid_pair
```

or:

```text
0-4,
```

An empty connection string can also trigger the same issue depending on how it is processed.

## Proposed fix

Refactor the parsing logic to store the result of `strings.Split()` and validate its length before accessing the elements:

```go
parts := strings.Split(pair, "-")

if len(parts) != 2 {
    // Skip invalid pairs to prevent a panic
    continue
}

lw, _ := strconv.Atoi(parts[0])
rw, _ := strconv.Atoi(parts[1])
```

This prevents malformed or empty pairs from causing an `index out of range` panic.

## Additional consideration

It may also be worth validating the result of `strconv.Atoi()` so that non-numeric worker IDs are handled explicitly rather than silently converted to `0`.

## Scope

The proposed change is limited to making the Kunlun topology connection-string parsing more robust and preventing crashes caused by malformed input.

I would like to work on this issue and can submit a PR with the fix.

KunwarSidhu47:
**What type of PR is this?**
/kind bug

This PR fixes a bug in `pkg/device/kunlun/topo.go` where parsing `InterGroupConnection` and `GroupConnection` strings could result in a runtime panic.

Previously, the code assumed that `strings.Split(pair, "-")` would always return at least two elements. If the connection string contained an empty or malformed pair, such as a trailing comma or a pair without `-`, accessing index `[1]` would cause an `index out of range` panic and crash the application during Kunlun topology initialization.

This change validates the result of the split before accessing its elements, allowing invalid pairs to be safely skipped instead of causing a fatal panic.

Fixes #2556 

N/A

```release-note
Fixes a scheduler panic when starting with invalid or malformed Kunlun topology connection strings.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation of device interconnection configuration entries.
  * Malformed entries are now safely rejected instead of being incorrectly processed or indexed.

### DIFF
--- a/pkg/device/kunlun/device.go
+++ b/pkg/device/kunlun/device.go
@@ -34,6 +34,9 @@ const (
 	KunlunDeviceSelection = "BAIDU_COM_DEVICE_IDX"
 	KunlunUseUUID         = "baidu.com/use-gpuuuid"
 	KunlunNoUseUUID       = "baidu.com/nouse-gpuuuid"
+)
+
+var (
 	InterGroupConnection  = "0-4,1-5,2-6,3-7"
 	InterGroupConnection2 = "0-1-4-5,2-3-6-7,0-2-4-6,1-3-5-7,0-3-4-7,1-2-5-6"
 	GroupConnection       = "0-1,0-2,0-3,1-2,1-3,2-3,4-5,4-6,4-7,5-6,5-7,6-7"
--- a/pkg/device/kunlun/topo.go
+++ b/pkg/device/kunlun/topo.go
@@ -100,14 +100,22 @@ func parseInterconnection() [][]int {
 	var interconnection [][]int
 	pairs := strings.Split(InterGroupConnection, ",")
 	for _, pair := range pairs {
-		lw, _ := strconv.Atoi(strings.Split(pair, "-")[0])
-		rw, _ := strconv.Atoi(strings.Split(pair, "-")[1])
+		parts := strings.Split(pair, "-")
+		if len(parts) != 2 {
+			continue
+		}
+		lw, _ := strconv.Atoi(parts[0])
+		rw, _ := strconv.Atoi(parts[1])
 		interconnection = append(interconnection, []int{lw, rw})
 	}
 	pairs = strings.Split(GroupConnection, ",")
 	for _, pair := range pairs {
-		lw, _ := strconv.Atoi(strings.Split(pair, "-")[0])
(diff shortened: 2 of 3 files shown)


FouoF:
Thanks for the contribution. This panic is not reachable in the current product: InterGroupConnection, InterGroupConnection2, and GroupConnection are fixed compile-time topology constants with no configuration or runtime input path. The same proposal was evaluated in #2215/#2216 and closed on that basis. This PR makes those constants mutable only to manufacture the test path, while still accepting non-numeric pairs because both strconv.Atoi errors are discarded ("x-y" becomes 0-0), and it leaves parseInterconnection2 unchanged. We are closing this PR. If Kunlun topology becomes configurable, validation should be designed at that input boundary and cover arity, integer/range checks, and all three topology forms.

KunwarSidhu47:
Hi @FouoF ,Thanks for the detailed explanation and for reviewing the contribution. I understand the concern regarding the fixed compile-time topology constants and the lack of a runtime configuration path. I’ll take this feedback into account and will keep an eye out for areas where validation can be added at the appropriate input boundary if the topology becomes configurable in the future. Thanks again for the review.

PR #2559, opened 2026-08-10
fix(scheduler): serialize Filter device selection to prevent double G…

### LINKED ISSUE #2232: 并发创建pod情况下，存在pod 中卡id的注解和实际pod中使用卡的id不一致问题

yxxhero:
**What type of PR is this?**
/kind bug

Fixes #2232 — *并发创建 pod 情况下，存在 pod 中卡 id 的注解和实际 pod 中使用卡的 id 不一致问题* (under concurrent pod creation, the GPU id annotated on a pod is inconsistent with the GPU id the pod actually uses).

The reported consequence is duplicate card allocation: the shared card gets over-allocated (causing OOM) while other idle cards cannot be handed out.

### Root cause

The scheduler extender is served over HTTP with **one goroutine per request** and provides no cross-request serialization guarantee. In `Filter`, the device-selection critical section was not atomic:

1. read node usage (`getNodesUsage`)
2. pick the best node/device (`calcScore`)
3. commit the reservation to the pod cache (`podManager.AddPod`)
4. publish the allocation onto the pod annotation (`PatchPodAnnotations`)

Under concurrent pod creation (e.g. a multi-replica Deployment), two `Filter` calls can both read the same free GPU in step 1 *before* either commits it in step 3, so both pods are annotated with the **same card id**.

In the normal kube-scheduler flow scheduling cycles are already serialized, which is why the problem only surfaces under concurrent/re-entrant extender requests.

### Fix

Add a `filterLock` and move the in-memory critical section (steps 1–3) into a `selectAndCommitDevice` helper that owns the lock through a single `defer Unlock`, so two pods can never reserve the same device. The lock is released before step 4 (`PatchPodAnnotations`), so concurrent Filters are never blocked on the apiserver round trip; the on-failure rollback runs outside the lock and is safe because the pod/quota managers have their own locks. The simulation path (`filterSimulation`) does not reserve devices and is unchanged.

This keeps the change minimal — it does **not** move reservation from `Filter` to `Bind` (the larger redesign from #2273 is tracked via the #2264 design discussion).

### Test

- `Test_Filter_ConcurrentNoDoubleAllocation` fires 8 pods at `Filter` simultaneously and asserts each is annotated with a **distinct** device id. Without the lock it reliably reports `device device-7 double-allocated to pod pod-0 and pod pod-2`; with the lock it passes (`-race`).
- `Test_Filter_RollbackOnPatchFailure` verifies the cache reservation is rolled back when the annotation patch fails.

### Validation

Scoped to the scheduler extender, so per CONTRIBUTING's hardware-validation gate this is validated with unit tests (including a `-race` regression test) rather than physical GPU hardware. `make verify` (license, import-aliases, golangci-lint) and `go test -race ./pkg/scheduler/...` pass locally.

Fixes #2232

- A single `defer Unlock` in the helper means any early return added there in future can never leak the lock.
- A per-pod-UID lock was considered but not added: kube-scheduler runs one scheduling cycle at a time, so a pod UID is not filtered concurrently; the reproduced race is between *different* pods, which the global `filterLock` closes.

```release-note
Fixed a race in the scheduler extender's Filter that could assign the same GPU to multiple pods created concurrently (e.g. a multi-replica Deployment), causing duplicate card allocation and OOM. Device allocation in Filter is now serialized.
```

**AI assistance disclosure:**
This PR was written with AI assistance for codebase analysis, implementation and test generation. I reviewed, built and tested the change myself (`make verify`, `go test -race`) and can explain the locking semantics. Per CONTRIBUTING.md, disclosure belongs in the PR description only — no AI co-author trailers are used.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved concurrent device allocation to prevent multiple pods from receiving the same GPU.
  * Added rollback handling when pod annotation updates fail, preventing stale reservations and keeping allocation state consistent.
  * Improved reporting of device selection failures.
  * Initialization failures are now reported clearly instead of terminating without useful error details.
* **Reliability**
  * Added validation for successful device initialization and allocation-state consistency.

### DIFF
--- a/cmd/scheduler/main.go
+++ b/cmd/scheduler/main.go
@@ -119,7 +119,9 @@ func start() error {
 		client.WithTimeout(config.Timeout),
 	)
 
-	config.InitDevices()
+	if err := config.InitDevices(); err != nil {
+		return fmt.Errorf("failed to initialize devices: %w", err)
+	}
 
 	var err error
 	config.HostName, err = os.Hostname()
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -78,6 +78,10 @@ type Scheduler struct {
 
 	lock   sync.RWMutex
 	synced bool
+
+	// filterLock serializes device selection in Filter so concurrent requests
+	// cannot reserve the same device (issue #2232).
+	filterLock sync.Mutex
 }
 
 func NewScheduler() *Scheduler {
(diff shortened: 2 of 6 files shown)


mesutoezdil:
> I am unable to reproduce the issue described at #2232 using K3s with the hami mock device plugin. The concurrency issue I mentioned at [#2470 (comment)](https://github.com/Project-HAMi/HAMi/pull/2470#issuecomment-5241490368) is limited solely to concurrent calls to the `/filter` endpoint; when I create Pods concurrently rather than making concurrent calls to the `/filter` endpoint, the problem no longer occurs. The kube-scheduler scheduling process is serialized, so concurrent calls to the `/filter` endpoint are not possible in this scenario. I am now curious about:
> 
> 1. where the version boundary threshold for Kubernetes v1.23 came from.
> 2. How the [并发创建pod情况下，存在pod 中卡id的注解和实际pod中使用卡的id不一致问题 #2232](https://github.com/Project-HAMi/HAMi/issues/2232) issue arose, and
>    why I cannot reproduce it.
> 
> In fact, although the `/filter` interface itself has concurrency control issues you mentioned and I tested, in a real-world scheduler extender scenario, as maintainers expected, no problems should arise.
> 
> I’d like to know what specific, real-world issue you encountered that led you to create this PR?

Pls write only your own sentences, as specified in the contribution guidelines.

pkg/scheduler/scheduler.go
```diff
@@ -1030,30 +1034,75 @@ func (s *Scheduler) Filter(args extenderv1.ExtenderArgs) (*extenderv1.ExtenderFi
...
+	added            bool
+	nodeList         []*policy.NodeScore
+	failedNodes      map[string]string
+	effectiveDevices device.PodDevices
+}
 
+// selectAndCommitDevice selects a device and commits it to the cache under
+// filterLock. It does not publish the pod annotation; the caller patches it.
+func (s *Scheduler) selectAndCommitDevice(args extenderv1.ExtenderArgs, resourceReqs device.PodDeviceRequests) (*filterSelection, error) {
+	s.filterLock.Lock()
```
mesutoezdil, line 1084: this lock is global, not per node or per device. a big concurrent deployment create now queues through one lock for every node. was a per node lock considered, not only per pod uid?

cmd/scheduler/main.go
```diff
@@ -119,7 +119,9 @@ func start() error {
 		client.WithTimeout(config.Timeout),
 	)
 
-	config.InitDevices()
+	if err := config.InitDevices(); err != nil {
```
mesutoezdil, line 122: this initdevices error change is not about the filterlock race. this repo closed a past pr for bundling unrelated changes together. should this be its own pr?

mesutoezdil (commented):
this looks like the follow up to closed pr #2470. 
left 2 questions inline about this version.

Shouren:
@yxxhero I think the root cause for #2232 is that specific Kubernetes versions, kube-scheduler cannot guarantee that the Extender's Filter is called only once during a Pod's scheduling cycle. Due to timing issues between Filter and Bind, the mismatch happens. That is why @spencercjh cannot reproduce it in K3s, I guess him testing with a higher version than v1.23.

pkg/scheduler/scheduler.go
```diff
@@ -1030,30 +1034,75 @@ func (s *Scheduler) Filter(args extenderv1.ExtenderArgs) (*extenderv1.ExtenderFi
...
+	annotations      map[string]string
+	added            bool
+	nodeList         []*policy.NodeScore
+	failedNodes      map[string]string
+	effectiveDevices device.PodDevices
+}
 
+// selectAndCommitDevice selects a device and commits it to the cache under
+// filterLock. It does not publish the pod annotation; the caller patches it.
+func (s *Scheduler) selectAndCommitDevice(args extenderv1.ExtenderArgs, resourceReqs device.PodDeviceRequests) (*filterSelection, error) {
```
mesutoezdil, line 1083: which k8s version, and where in kube-scheduler's source does that happen? without that this fix may be locking against a race that never occurs in real scheduling.

archlitchi:
This scenario happens to k8s version < v1.23, and conflicts with prerequisities which requires k8s>=v1.23.

Reopen this PR if encounter the same issue with k8s version >= v1.23

PR #2567, opened 2026-08-11
Add a CI verification check

jianzhangbjz:
**What type of PR is this?**

/kind feature

Add a CI verification tool (`hack/tools/rbaccheck/`) that uses Go AST analysis to verify production code only calls Kubernetes API verbs that are granted by the scheduler ClusterRole.

The existing bash-based verify-rbac.sh greps for `Nodes().Update(` to catch one specific violation. The new Go tool:
 - Parses `charts/hami/templates/scheduler/clusterrole.yaml` to extract all
    (resource, verb) pairs granted by the scheduler ClusterRole
 - Walks Go source files using `go/ast` to precisely match method call chains
    like `CoreV1().<Resource>().<Verb>(...)`
 - Checks all resource types (Nodes, Pods, Events, ConfigMaps, etc.), not just
    Nodes
 - Avoids false positives by matching at the AST level rather than text grep

---
 
_**This catches violations that bash/grep misses — for example `Events().Update()` in `pkg/util/util.go:325`, where the scheduler ClusterRole only grants `create, get, list` on events.**_

Fixed PR: https://github.com/Project-HAMi/HAMi/pull/2569

Fixes https://github.com/Project-HAMi/HAMi/pull/2329#discussion_r3743270389

The tool runs via `go run ./hack/tools/rbaccheck/ ./pkg/` as part of `make verify`. It depends on `gopkg.in/yaml.v3` which is already a direct dependency.

No — CI-only change.

This PR was Assisted by Claude Code.

## Security
- Added automated validation to ensure Kubernetes operations use only approved permissions.
- Verification reports unauthorized access clearly, including affected locations and remediation guidance.

## Quality
- Integrated access-control checks into the full verification process.
- Existing validation checks continue to run alongside the new security checks.

### DIFF
--- a/hack/tools/rbaccheck/main.go
+++ b/hack/tools/rbaccheck/main.go
@@ -1,0 +1,365 @@
+/*
+Copyright 2024 The HAMi Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// rbaccheck verifies that production Go code only calls Kubernetes API
+// methods (Get, Patch, Update, Delete, Create, List, Watch) that are
+// granted by the RBAC roles in the helm chart.
+//
+// Usage: go run ./hack/tools/rbaccheck/ [path ...]
+//
+// It reads the ClusterRole/Role templates under charts/hami/templates to
+// extract allowed (resource, verb) pairs, then walks Go source files looking
+// for method chains like CoreV1().Nodes().Update(...) and reports any verb
+// that is not allowed for the matching resource.
+//
+// Different binaries run under different service accounts, so each source
+// directory is checked against the role of the binary that runs its code:
+//
+//	cmd/scheduler/, pkg/scheduler/, pkg/device/   scheduler roles
+//	cmd/device-plugin/, pkg/device-plugin/        device-plugin role
+//	shared code (e.g. pkg/util/)                  union of all roles
+package main
+
+import (
+	"fmt"
+	"go/ast"
+	"go/parser"
+	"go/token"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"gopkg.in/yaml.v3"
+)
+
+type rule struct {
+	APIGroups []string `yaml:"apiGroups"`
+	Resources []string `yaml:"resources"`
+	Verbs     []string `yaml:"verbs"`
+}
+
+type roleDoc struct {
+	Rules []rule `yaml:"rules"`
+}
+
+// resourceToMethod maps YAML resource names to the CoreV1 Go method names.
+// Sub-resources (e.g. pods/binding) are not included — they are rare and the
+// scheduler code doesn't call them directly.
+var resourceToMethod = map[string]string{
+	"nodes":                  "Nodes",
+	"pods":                   "Pods",
+	"configmaps":             "ConfigMaps",
+	"events":                 "Events",
+	"resourcequotas":         "ResourceQuotas",
+	"namespaces":             "Namespaces",
+	"services":               "Services",
+	"endpoints":              "Endpoints",
+	"persistentvolumes":      "PersistentVolumes",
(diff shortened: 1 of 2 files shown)


mesutoezdil:

jianzhangbjz:
Hi @mesutoezdil , I have updated this commit and PR description. Please have a look, thanks!

mesutoezdil:
> Hi @mesutoezdil , could you give more details about this closing reason? Thanks!

we have some rules here:
https://github.com/Project-HAMi/HAMi/blob/master/CONTRIBUTING.md#ai-assistance-notice

jianzhangbjz:
Hi @mesutoezdil , got it, thanks! I have updated it, please have a look, thanks!

hack/verify-rbac.sh
```diff
@@ -0,0 +1,31 @@
...
+set -o errexit
+set -o nounset
+set -o pipefail
+
+REPO_ROOT=$(dirname "${BASH_SOURCE[0]}")/..
+
+cd "${REPO_ROOT}"
+
+echo "Running RBAC permission check..."
+go run ./hack/tools/rbaccheck/ ./pkg/
```
mesutoezdil: pkg/util/util.go still has events().update(). scheduler clusterrole only grants create, get, list on events. this check will fail make verify right now. the fix pr for it, #2569, was closed today.

hack/tools/rbaccheck/main.go
```diff
@@ -0,0 +1,264 @@
...
+			return true
+		}
+		resourceFun, ok := resourceSel.Fun.(*ast.SelectorExpr)
+		if !ok {
+			return true
+		}
+		resource := resourceFun.Sel.Name
+
+		// Only check calls on CoreV1 (or anything that looks like CoreV1).
+		// We don't strictly enforce the CoreV1 check to catch indirect calls too.
```
mesutoezdil: this matches by name only, not the real corev1 type. a mock or unrelated type named like a resource could false flag. known limit, worth a short note in the readme?

mesutoezdil (commented):
checked this tool logic against current master. left two notes inline, first one is important.

mesutoezdil:
and look at the messages from coderrabbit and resolve it pls

hack/verify-rbac.sh
jianzhangbjz: I've updated and reopened https://github.com/Project-HAMi/HAMi/pull/2569

hack/tools/rbaccheck/main.go
jianzhangbjz: I added the `.CoreV1()` level matching check. It will match the `xxx.CoreV1().Nodes().Update(...)` instead of the ` xxx.Nodes().Update(...)`. I guess it will decrease the false alarm.

jianzhangbjz:
> and look at the messages from coderrabbit and resolve it pls

I guess the CodeRabbit passed for the latest commit.

hack/verify-rbac.sh
mesutoezdil: this still only loads scheduler/clusterrole.yaml. pkg/util/util.go's events().update() is called only by the device plugin, whose own role (monitorrole.yaml) already grants update. this line will still flag it as a violation and fail make verify. does the tool need to know which binary calls which code, not just one hardcoded role file?

jianzhangbjz: Yes, I will fix it. Thanks!

jianzhangbjz:
Hi @mesutoezdil , I've updated it. Could you help have a look? Thanks!

mesutoezdil:
resolve the conflicts.
look at the feedbacks of coderrabbit

hack/tools/rbaccheck/main.go
```diff
@@ -0,0 +1,352 @@
...
+// dirRoles maps code directories (relative to the repo root) to the roles
+// of the binaries that run their code. Directories not listed here are
+// shared between binaries and are checked against the union of all roles.
+var dirRoles = []struct {
+	prefix string
+	roles  []string
+}{
+	{"cmd/scheduler/", []string{"scheduler"}},
+	{"pkg/scheduler/", []string{"scheduler"}},
+	{"pkg/device/", []string{"scheduler"}},
```
mesutoezdil, line 113: device-plugin also runs pkg/device/nvidia (CalculateGPUScore). dirRoles maps it scheduler-only, same bug as util.go, not tripped yet. intentional?

jianzhangbjz: No, fixed it, thanks!

jianzhangbjz:
Hi @Shouren , could you help approve it? Thanks!

archlitchi (approved):
/lgtm

PR #2568, opened 2026-08-11
fix(scheduler): reconcile health on device discovery error

### LINKED ISSUE #2565: **fix(scheduler): run device health reconciliation when device discovery fails**
### What happened?

During the scheduler's periodic device registration cycle, `GetNodeDevices()` is called before `CheckHealth()` for each registered device vendor.

Currently, when `GetNodeDevices()` returns an error, the registration loop immediately continues:

```go
nodedevices, err := devInstance.GetNodeDevices(*val)
if err != nil {
    continue
}

health, needUpdate := devInstance.CheckHealth(devhandsk, val)
```

As a result, `CheckHealth()` is skipped whenever device discovery fails.

This can prevent the scheduler from reconciling an already-cached vendor/device state when the device discovery path is temporarily unavailable or returns an error.

For example:

```text
node-1 cache:
  vendor-A → devices present

registration cycle:
  vendor-A.GetNodeDevices() → error

current behavior:
  skip vendor-A
  CheckHealth() is not called
  cached vendor-A remains unchanged
```

If the health state subsequently indicates that the vendor should be removed, that cleanup is not performed during this registration cycle because the health check was skipped.

### Expected behavior

A failure to retrieve the current device list should not prevent the scheduler from performing the independent device health reconciliation.

The scheduler should preserve the existing cached device state when the discovery error is transient, while still allowing `CheckHealth()` to determine whether the vendor should be considered unhealthy and cleaned up.

### Impact

Skipping health reconciliation on device-discovery errors can leave stale device-vendor information in the scheduler cache for longer than intended.

This is particularly problematic when the cached device state is no longer valid but the device discovery path is failing, because the scheduler has no opportunity to execute the existing health/cleanup path.

### Notes

The fix should not blindly remove cached devices whenever `GetNodeDevices()` returns an error, since transient discovery failures should not immediately make otherwise healthy devices unavailable.

The device discovery error and health result should be handled independently so that the existing health-check semantics remain authoritative for cleanup.

AyushSrivastava1818:
## What type of PR is this?

/kind bug

During the scheduler's periodic device registration cycle, `GetNodeDevices()` is called before `CheckHealth()` for each registered device vendor.

Previously, when `GetNodeDevices()` returned an error, the registration loop immediately continued. This caused `CheckHealth()` to be skipped for that vendor, preventing the existing health reconciliation and cleanup path from running.

As a result, an unhealthy vendor that was already present in the scheduler's cached device state could remain cached because `NodeCleanUp()` and `rmNodeDevices()` were never reached during that registration cycle.

This PR allows health reconciliation to continue when device discovery fails:

* If device discovery fails but the vendor is healthy, the existing cached state is preserved.
* If device discovery fails and the vendor is unhealthy, the existing cleanup path removes the stale vendor/device state.
* Normal successful device discovery behavior remains unchanged.

A regression test is added to cover the discovery-error and unhealthy-health-check path.

Fixes #2565

The fix intentionally does **not** remove cached devices merely because `GetNodeDevices()` returns an error, since discovery failures may be transient.

Instead, the existing `CheckHealth()` result remains authoritative for determining whether cleanup should occur.

The implementation also preserves the existing vendor-by-vendor update behavior so that an error for one vendor does not affect healthy vendors on the same node.

Regression coverage verifies both the unhealthy cleanup path and preservation of cached state when the health check remains healthy.

Validation performed:

* `go test ./pkg/device/...`
* `go test -race ./pkg/scheduler/...` *(unavailable in the Windows environment due to missing 64-bit GCC/CGO toolchain)*
* GolangCI-Lint
* `git diff --check`

No.

## AI Assistance Disclosure

AI assistance like Claude was used during the investigation, codebase analysis, implementation, regression-test development, and validation of this change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved device health reconciliation when discovery fails.
  * Removed unhealthy or stale device entries from the scheduler cache.
  * Preserved healthy cached devices during temporary discovery errors.
  * Updated cached devices after discovery recovery.
  * Enhanced logs with discovery error details.

* **Tests**
  * Added coverage for unhealthy, healthy, mixed-device, and recovery scenarios.


pkg/scheduler/scheduler.go
```diff
@@ -496,6 +495,9 @@ func (s *Scheduler) register(labelSelector labels.Selector, printedLog map[strin
 				s.rmNodeDevices(val.Name, devhandsk)
 				continue
 			}
+			if err != nil {
```
mesutoezdil, line 506: pr #2550 adds a check at this exact spot too, for empty nodedevices. if that merges first and this rebases after, this err check must stay before it. else a transient error also has empty nodedevices, and that other check would remove a vendor that is still healthy. this pr says that must not happen.

mesutoezdil (commented):
left one important question inline about merge order between the 2.

pkg/scheduler/scheduler.go
AyushSrivastava1818: yes I reviewed it too now....the discovery error must be handled before zero device reconcilation introduced by #2550 After #2550 merges,I will make sure to rebase this PR and ensure the err reconcilation remains before the zero-device handling

mesutoezdil, line 506: pr #2550 still inserts its zero-device check at this exact same spot, unresolved since i first flagged it. also none of the three new tests cover recovery: vendor errors once, then succeeds on a later register() call. does health correctly clear then?

AyushSrivastava1818: Yes i know pr #2550 inserts its zero device check at the same spot...I didnt fix it yet cause I though I will do it after getting the pr #2550 merged...I will fix this and also add regression test covering recovery across registration cycles....I will cover it in next update!

archlitchi:
I've merged 2550, please resolve these conflicts

AyushSrivastava1818:
@archlitchi Resolved the conflicts check it once

archlitchi (approved):
/lgtm

PR #2569, opened 2026-08-11
use Patch-based event dedup in EmitNodeWarningEvent

jianzhangbjz:
**What type of PR is this?**

/kind bug

Replace `Events().Update()` with `Events().Patch()` in `EmitNodeWarningEvent` to fix an RBAC permission violation. The scheduler ClusterRole only grants `create, get, list` on events — calling `Update()` requires the `update` verb which wasn't granted.
The previous attempt removed client-side event dedup entirely, but that would create duplicate events on repeated calls since the API server does not deduplicate by involvedObject+reason. Instead, this uses `Patch` to update the count and timestamp of matching events, and adds `patch` to the events verbs in the scheduler ClusterRole.

Fixes #

Discovered by the `rbaccheck` Go AST tool(https://github.com/Project-HAMi/HAMi/pull/2567) that flags RBAC permission violations at CI time. 

No — events are still emitted, just without client-side deduplication. The API server handles deduplication automatically.

This PR was written primarily by Claude Code.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved GPU warning event handling to reliably update matching events and avoid duplicate notifications.
  * Added safeguards for unavailable event clients.
  * Ensured event permissions support updating existing warnings.
* **Maintenance**
  * Simplified warning event tracking by removing time-based deduplication.

### DIFF
--- a/charts/hami/templates/device-plugin/monitorrole.yaml
+++ b/charts/hami/templates/device-plugin/monitorrole.yaml
@@ -32,6 +32,7 @@ rules:
       - list
       - create
       - update
+      - patch
 {{- end -}}
     
     
--- a/pkg/device-plugin/nvidiadevice/nvinternal/plugin/register.go
+++ b/pkg/device-plugin/nvidiadevice/nvinternal/plugin/register.go
@@ -224,8 +224,7 @@ func (plugin *NvidiaDevicePlugin) RegisterInAnnotation() (bool, error) {
 		if hasAsymmetry {
 			util.EmitNodeWarningEvent(node, "AsymmetricGPUP2PLink",
 				"One or more GPU pairs on this node have asymmetric P2P link data; "+
-					"affected pairs scored 0 (possible NVLink hardware or driver issue)",
-				time.Hour)
+					"affected pairs scored 0 (possible NVLink hardware or driver issue)")
 		}
 		data, err = json.Marshal(gpuScore)
 		if err != nil {
--- a/pkg/util/util.go
+++ b/pkg/util/util.go
@@ -282,8 +282,10 @@ func AllContainersCreated(pod *corev1.Pod) bool {
 	return len(pod.Status.ContainerStatuses) >= len(pod.Spec.Containers)
 }
 
-// EmitNodeWarningEvent emits a Warning event on the given Node with deduplication.
-func EmitNodeWarningEvent(node *corev1.Node, reason, message string, dedupWindow time.Duration) {
+// EmitNodeWarningEvent emits a Warning event on the given Node.
+// If an existing event with the same node and reason is found, the
+// count is incremented via Patch to avoid creating duplicates.
+func EmitNodeWarningEvent(node *corev1.Node, reason, message string) {
 	c := client.GetClient()
 	if c == nil {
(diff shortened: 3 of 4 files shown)


mesutoezdil:

jianzhangbjz:
Hi @mesutoezdil , I've updated it. Could you help have a review? Thanks!

mesutoezdil (commented):
left two questions inline.

pkg/util/util.go
```diff
@@ -282,8 +282,10 @@ func AllContainersCreated(pod *corev1.Pod) bool {
 	return len(pod.Status.ContainerStatuses) >= len(pod.Spec.Containers)
 }
 
-// EmitNodeWarningEvent emits a Warning event on the given Node with deduplication.
-func EmitNodeWarningEvent(node *corev1.Node, reason, message string, dedupWindow time.Duration) {
+// EmitNodeWarningEvent emits a Warning event on the given Node.
+// The Kubernetes API server handles event deduplication automatically
```
mesutoezdil: is this true for a plain create with generatename? each call makes a new object, no shared name to merge on. if this fires every cycle on a stuck node, will it spam many duplicate events instead of one counted event?

pkg/device-plugin/nvidiadevice/nvinternal/plugin/register.go
```diff
@@ -225,7 +225,7 @@ func (plugin *NvidiaDevicePlugin) RegisterInAnnotation() (bool, error) {
 			util.EmitNodeWarningEvent(node, "AsymmetricGPUP2PLink",
 				"One or more GPU pairs on this node have asymmetric P2P link data; "+
 					"affected pairs scored 0 (possible NVLink hardware or driver issue)",
-				time.Hour)
+				)
```
mesutoezdil: stray comma and closing paren on its own line here. gofmt would not leave it like this.

pkg/util/util.go
jianzhangbjz: You're right! However, it doesn't trigger every 30-second cycle; it only triggers when an anomaly is first detected in the GPU topology state. Anyway, let me make a more secure solution.

jianzhangbjz:
Hi @mesutoezdil , I added the `patch` permission as a new solution. Could you help have a review? Thanks!

charts/hami/templates/device-plugin/monitorrole.yaml
```diff
@@ -32,6 +32,7 @@ rules:
       - list
       - create
       - update
+      - patch
```
mesutoezdil, line 35: emitnodewarningevent has exactly one caller, register.go, which runs under this exact role. this role already had update on events before this pr. so the original update() call was never actually forbidden, rbaccheck flagged it against the scheduler clusterrole instead, which never calls this function at all. does that mean the real bug here was never an rbac violation, just rbaccheck checking the wrong role file?

jianzhangbjz: Yes, correct. Thanks! I've updated https://github.com/Project-HAMi/HAMi/pull/2567 to avoid this kind of false alarm. Closed it.

PR #2571, opened 2026-08-11
fix: Handle HTTP response write errors in scheduler routes

shinigami-777:
**What type of PR is this?**
/kind cleanup

This PR error handling for HTTP `w.Write()` calls. Previously, if `w.Write()` failed when returning the scheduler binding or filter results, the error was silently ignored without any visibility. 

Changes Made:

- `pkg/scheduler/routes/route.go`: Added error handling for `w.Write()` in both PredicateRoute and Bind handlers to log write failures.
- `pkg/scheduler/routes/route_test.go`: Added dedicated regression tests (`TestPredicateRoute_WriteError` and `TestBind_WriteError`) for this exact failure path.

The new tests mock an `http.ResponseWriter` that forcibly returns an error when `Write()` is called. The test captures the internal klog output and asserts that the Failed to write response error is successfully recorded. 

Fixes #

This is a followup to the earlier broader PR #2279 . It introduces a minimal change along with the regression test.

AI assistance disclosure:
I manually ran `errcheck` to catch the unhandled cases. The changes was made using assistance from Gemini 3.1 Pro. I have manually verified the changes and the new regressions tests.

No, this PR is for error handling. If the server fails to write an HTTP response back to a client, the error is logged to the scheduler's internal logs rather than failing silently.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of failures when sending Predicate and Bind responses.
  * Added diagnostic logging when response delivery fails, while preserving existing response behavior.
* **Tests**
  * Added coverage to verify response-write failures are correctly detected and logged.

### DIFF
--- a/pkg/scheduler/routes/route.go
+++ b/pkg/scheduler/routes/route.go
@@ -40,6 +40,14 @@ func checkBody(w http.ResponseWriter, r *http.Request) bool {
 	return true
 }
 
+func writeResponse(w http.ResponseWriter, code int, body []byte) {
+	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(code)
+	if _, err := w.Write(body); err != nil {
+		klog.ErrorS(err, "Failed to write response")
+	}
+}
+
 func PredicateRoute(s *scheduler.Scheduler) httprouter.Handle {
 	klog.Infoln("Initializing Predicate Route")
 	return func(w http.ResponseWriter, r *http.Request, _ httprouter.Params) {
@@ -83,21 +91,15 @@ func PredicateRoute(s *scheduler.Scheduler) httprouter.Handle {
 
 		if resultBody, err := json.Marshal(extenderFilterResult); err != nil {
 			klog.ErrorS(err, "Failed to marshal extender filter result", "result", extenderFilterResult)
-			w.Header().Set("Content-Type", "application/json")
-			w.WriteHeader(http.StatusInternalServerError)
 			extenderFilterResult = &extenderv1.ExtenderFilterResult{
 				Error: fmt.Sprintf("Failed to marshal extender filter result: %s", err.Error()),
 			}
 			resultBody, _ = json.Marshal(extenderFilterResult)
-			if _, err := w.Write(resultBody); err != nil {
-				klog.ErrorS(err, "Failed to write response")
-			}
+			// Note: write error in this fallback path is not explicitly tested
+			// as it requires both a marshal failure and a write failure.
+			writeResponse(w, http.StatusInternalServerError, resultBody)
 		} else {
-			w.Header().Set("Content-Type", "application/json")
-			w.WriteHeader(http.StatusOK)
-			if _, err := w.Write(resultBody); err != nil {
-				klog.ErrorS(err, "Failed to write response")
-			}
+			writeResponse(w, http.StatusOK, resultBody)
 		}
 	}
 }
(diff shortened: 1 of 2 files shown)


pkg/scheduler/routes/route.go
```diff
@@ -89,11 +89,15 @@ func PredicateRoute(s *scheduler.Scheduler) httprouter.Handle {
 				Error: fmt.Sprintf("Failed to marshal extender filter result: %s", err.Error()),
 			}
 			resultBody, _ = json.Marshal(extenderFilterResult)
-			w.Write(resultBody)
+			if _, err := w.Write(resultBody); err != nil {
```
mesutoezdil: new test does not cover this line. only the write below it is tested. this one is in the marshal error branch, hard to trigger, but worth a note.

pkg/scheduler/routes/route.go
```diff
@@ -134,12 +138,16 @@ func Bind(s *scheduler.Scheduler) httprouter.Handle {
...
-			w.Write(response)
+			if _, err := w.Write(response); err != nil {
+				klog.ErrorS(err, "Failed to write response")
+			}
 		} else {
 			klog.V(5).InfoS("Returning bind response", "result", extenderBindingResult)
 			w.Header().Set("Content-Type", "application/json")
 			w.WriteHeader(http.StatusOK)
-			w.Write(response)
+			if _, err := w.Write(response); err != nil {
```
mesutoezdil: same three line block appears four times now. a small helper like writeresponse(w, body) could remove the repeat.

mesutoezdil (commented):
left two small notes inline. checked all w.Write calls in this file, this covers all of them.

pkg/scheduler/routes/route.go
shinigami-777: Yes, added a helper function writeResponse(w, code, response) to get rid of the repetation.

pkg/scheduler/routes/route.go
shinigami-777: I added a comment in the 2 places where it remains untested. A way to test this would be to create a variable for json.Marshal `var jsonMarshal = json.Marshal` at the package level. We can use this to mock in the tests to force a marshal error and verify the fallback.
I held off on doing it since mocking standard library functions via global variables might be considered as an anti-pattern. Lmk if you'd prefer having it for better coverage.

pkg/scheduler/routes/route.go
```diff
@@ -83,17 +91,15 @@ func PredicateRoute(s *scheduler.Scheduler) httprouter.Handle {
...
 		if resultBody, err := json.Marshal(extenderFilterResult); err != nil {
 			klog.ErrorS(err, "Failed to marshal extender filter result", "result", extenderFilterResult)
-			w.Header().Set("Content-Type", "application/json")
-			w.WriteHeader(http.StatusInternalServerError)
 			extenderFilterResult = &extenderv1.ExtenderFilterResult{
 				Error: fmt.Sprintf("Failed to marshal extender filter result: %s", err.Error()),
 			}
 			resultBody, _ = json.Marshal(extenderFilterResult)
-			w.Write(resultBody)
+			// Note: write error in this fallback path is not explicitly tested
```
mesutoezdil, line 98: can json.Marshal actually fail on a real extenderfilterresult here, given its fields are just nodelist, string slice, and maps? or is this branch practically unreachable through normal inputs?

shinigami-777: Yes json.Marshal will practically never fail on a real extenderfilterresult.

archlitchi (approved):
/lgtm

archlitchi (approved):
/lgtm

PR #2574, opened 2026-08-11
fix(monitor): prevent fatal data race during feedback observation

### LINKED ISSUE #2573: [BUG]: Fatal concurrent map iteration data race in vGPUmonitor feedback observation
**What happened**:
The `HAMi-vGPUmonitor` component suffers from a fatal data race that can crash the monitor with a `concurrent map iteration and map write` runtime panic.

In `cmd/vGPUmonitor/feedback.go`, the `Observe()` function retrieves the `containers` map via `lister.ListContainers()` and iterates over it sequentially inside a background 5-second ticker loop. However, it does this **without acquiring the `lister.Lock()` mutex**. 

At the exact same time, Kubernetes Pod informer events run in separate background worker goroutines (configured in `pkg/monitor/nvidia/cudevshr.go`). When these informers trigger (e.g. a pod is added or updated), they call `lister.Update()`, which explicitly locks the mutex and modifies (adds/deletes) entries in the exact same `l.containers` map. 

Because `Observe()` iterates over the map without holding the lock while the informer modifies it, the Go runtime will inevitably detect a concurrent map read/write and trigger a fatal panic, bringing down the entire monitor component.

The `Observe()` function should safely acquire the `ContainerLister`'s mutex before iterating over the containers map. This is identical to how the map is properly handled in `cmd/vGPUmonitor/metrics.go`, which correctly locks before iterating to ensure concurrency safety.

This is a race condition in the code that can trigger randomly in production when a Pod event (add/delete) fires at the exact same moment the 5-second `watchAndFeedback` ticker fires. It is clearly visible via static code analysis.

- This is a pure Go concurrency issue found via code analysis, so host-level logs (`nvidia-smi`, containerd config, dmesg, etc.) are not directly applicable. The relevant code path is inside `Observe()` in `cmd/vGPUmonitor/feedback.go`.

**Environment**:
- HAMi version: `master`
- nvidia driver or other AI device driver version: N/A (Go code logic bug)
- Others: Affects the `HAMi-vGPUmonitor` binary.

Rickydama3:
**What type of PR is this?**
/kind bug

This PR fixes a critical data race in the `HAMi-vGPUmonitor` that causes the monitor to crash with a fatal `concurrent map iteration and map write` runtime panic.

In `cmd/vGPUmonitor/feedback.go`, the `Observe()` function fetches the `containers` map via `lister.ListContainers()` and iterates over it in a background ticker loop. However, it was missing a lock acquisition before doing so. At the same time, the Kubernetes Pod informer events (handled by `podInformer.AddEventHandler` in `pkg/monitor/nvidia/cudevshr.go`) concurrently trigger `lister.Update()`, which explicitly locks the mutex and modifies the exact same map. 

Iterating over a map without a lock while another goroutine modifies it triggers a fatal runtime panic in Go. This PR adds the missing `lister.Lock()` and `defer lister.UnLock()` to `Observe()`, safely synchronizing it with the informer events. This strictly mirrors the correct locking behavior already implemented for the exact same map in `cmd/vGPUmonitor/metrics.go`.

Fixes #2573

This is a standard Go concurrency bug fix. The lock is only held during the fast in-memory map iteration inside `Observe()`, so there is no risk of deadlock or blocking the informer goroutines for a significant amount of time.

```release-note
None

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved container state monitoring reliability by preventing concurrent updates during observation processing.

### DIFF
--- a/cmd/vGPUmonitor/feedback.go
+++ b/cmd/vGPUmonitor/feedback.go
@@ -72,6 +72,9 @@ func CheckPriority(utSwitchOn map[string]UtilizationPerDevice, p int, c *nvidia.
 }
 
 func Observe(lister *nvidia.ContainerLister) {
+	lister.Lock()
+	defer lister.UnLock()
+
 	utSwitchOn := map[string]UtilizationPerDevice{}
 	containers := lister.ListContainers()
 


mesutoezdil:

PR #2577, opened 2026-08-11
fix(kunlun): honour use-gpuuuid and nouse-gpuuuid annotations

Lakshya77089:
**What type of PR is this?**

/kind bug

`KunlunUseUUID` (`baidu.com/use-gpuuuid`) and `KunlunNoUseUUID` are defined in
`pkg/device/kunlun/device.go` but never read. Neither `KunlunDevices.Fit` nor
`KunlunVDevices.Fit` calls `device.CheckUUID`, so:

- a pod asking for a specific XPU is scheduled onto any of them
- `nouse-gpuuuid` cannot exclude a card

No error, no warning — the annotation is silently ignored.

Every other backend does call it: nvidia, amd, ascend, awsneuron, biren,
cambricon, enflame, hygon, iluvatar, metax and mthreads. Kunlun looks like it
was simply missed — the constants were added with the backend in #1121, then
#1622 extracted the shared `CheckUUID` helper and migrated the other backends
without kunlun, and #2045 changed the helper again, also without kunlun.

**The fix**

The check goes into the `fitFn` that `graghSelect` already takes, rather than
filtering the device slice before calling it. That matters here: `graghSelect`
derives topology from a device's *position* in the slice (`idx < 4` is one
wing), so removing entries would move the remaining cards between wings and
change which interconnect group gets picked. Passing the constraint through
`fitFn` leaves positions untouched.

Both kunlun backends are covered. An allocation that fails only because of the
annotations now reports `CardUuidMismatch` rather than `NumaNotFit`, which was
misleading.

**Tests**

Three tests in `pkg/device/kunlun/device_test.go`, all failing on master:

```
TestKunlunDevices_Fit_UseUUID          asked for xpu-5, got xpu-0
TestKunlunDevices_Fit_NoUseUUID        excluded all 8 cards, still allocated
TestKunlunVDevices_Fit_UUIDAnnotations asked for xpu-3, got xpu-0
```

They pass with this change, and the 20 existing kunlun tests still pass.

None filed — raising the fix directly.

Behaviour change worth naming: a pod that carries these annotations today is
scheduled anywhere, and after this it is restricted to the cards it names. If a
cluster has pods with a stale or wrong `use-gpuuuid` value that currently
schedule fine, they will start failing with `CardUuidMismatch`. That is the
annotation doing what it says, but it is a visible change for anyone who had
been relying on it being ignored.

Not validated on real hardware. The change is confined to the scheduler's
kunlun Fit path and is covered by unit tests, which CONTRIBUTING allows for
scheduler-scoped changes.

```release-note
Kunlun XPU now honours the baidu.com/use-gpuuuid and baidu.com/nouse-gpuuuid
pod annotations, which were previously ignored.
```

---

AI assistance disclosure: this change was developed with Claude Code — finding
the gap, the fix, and the tests. Flagging the extent up front per CONTRIBUTING.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Kunlun device allocation with GPU UUID inclusion and exclusion settings.
  * Requests now select specified GPUs when available and clearly report UUID mismatches when unavailable.
  * Added equivalent UUID filtering for virtual Kunlun devices.
  * Preserved topology-aware allocation and idle-device requirements for mutex scheduling.
  * Improved allocation failure reporting for requests that cannot match the specified GPU UUIDs.

### DIFF
--- a/pkg/device/kunlun/device.go
+++ b/pkg/device/kunlun/device.go
@@ -184,10 +184,27 @@ func (kl *KunlunDevices) Fit(devices []*device.DeviceUsage, request device.Conta
 	tmpDevs := make(map[string]device.ContainerDevices)
 	reason := make(map[string]int)
 
-	alloc := graghSelect(devices, request, FitXPU)
+	// graghSelect decides topology from the position of a device in the slice,
+	// so the uuid constraint has to be applied through fitFn rather than by
+	// filtering the slice first.
+	uuidMismatch := 0
+	fitFn := func(d *device.DeviceUsage, r device.ContainerDeviceRequest) bool {
+		if !device.CheckUUID(pod.GetAnnotations(), d.ID, KunlunUseUUID, KunlunNoUseUUID, kl.CommonWord()) {
+			uuidMismatch++
+			klog.V(5).InfoS(common.CardUUIDMismatch, "pod", klog.KObj(pod), "device", d.ID)
+			return false
+		}
+		return FitXPU(d, r)
+	}
+
+	alloc := graghSelect(devices, request, fitFn)
 	if len(alloc) == 0 {
-		reason[common.NumaNotFit]++
-		klog.V(5).InfoS(common.NumaNotFit, "pod", klog.KObj(pod), "device", devices, "request nums", request.Nums, "numa")
+		if uuidMismatch > 0 {
+			reason[common.CardUUIDMismatch] += uuidMismatch
+		} else {
+			reason[common.NumaNotFit]++
+			klog.V(5).InfoS(common.NumaNotFit, "pod", klog.KObj(pod), "device", devices, "request nums", request.Nums, "numa")
+		}
 		return false, tmpDevs, common.GenReason(reason, len(devices))
 	}
 
--- a/pkg/device/kunlun/device.go
+++ b/pkg/device/kunlun/device.go
@@ -187,10 +187,11 @@ func (kl *KunlunDevices) Fit(devices []*device.DeviceUsage, request device.Conta
 	// graghSelect decides topology from the position of a device in the slice,
 	// so the uuid constraint has to be applied through fitFn rather than by
 	// filtering the slice first.
-	uuidMismatch := 0
+	uuidMismatches := make(map[string]bool)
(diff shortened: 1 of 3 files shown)


pkg/device/kunlun/device.go
```diff
@@ -184,10 +184,27 @@ func (kl *KunlunDevices) Fit(devices []*device.DeviceUsage, request device.Conta
...
 	reason := make(map[string]int)
 
-	alloc := graghSelect(devices, request, FitXPU)
+	// graghSelect decides topology from the position of a device in the slice,
+	// so the uuid constraint has to be applied through fitFn rather than by
+	// filtering the slice first.
+	uuidMismatch := 0
+	fitFn := func(d *device.DeviceUsage, r device.ContainerDeviceRequest) bool {
+		if !device.CheckUUID(pod.GetAnnotations(), d.ID, KunlunUseUUID, KunlunNoUseUUID, kl.CommonWord()) {
+			uuidMismatch++
```
mesutoezdil: graghSelect calls fitFn once per device to count wings, then devicepick or interconnect can call fitFn again on the same devices. can uuidmismatch get double counted for the same device across those passes?

Lakshya77089: Yes, are you right ! uuidmismatch is counted twice 
To fix this , I have updated the implementation to track the mismatched devices uniquely using a map:-
uuidMismatches := make(map[string]bool)
rather than simple counter

mesutoezdil:
resolve pls

pkg/device/kunlun/device.go
```diff
@@ -34,6 +34,8 @@ const (
 	KunlunDeviceSelection = "BAIDU_COM_DEVICE_IDX"
 	KunlunUseUUID         = "baidu.com/use-gpuuuid"
 	KunlunNoUseUUID       = "baidu.com/nouse-gpuuuid"
+	HamiKunlunUseUUID     = "hami.io/use-kunlun-uuid"
```
mesutoezdil: vdevice.go has UseUUIDAnno already (hami.io/use-xpu-uuid). why a new key here instead of reusing it?

archlitchi:
please resolve comments and sign-off your commit

archlitchi (approved):
/lgtm

PR #2579, opened 2026-08-11
cleanup(metax): decouple MetaxDevices.ScoreNode (Metax-GPU) from the scheduler policy string

### LINKED ISSUE #2572: Decouple MetaxDevices.ScoreNode (Metax-GPU) from the scheduler policy string

```go
    sum := 0
        sum += len(dev)
    }

    res := float32(0)
        if ok {
        }
        if ok {
        }
    }
    return res
}
```

**Why is this needed**:

**Evidence**

```
5
0
```

**Proposed solution**

2. Add the marker method:

```go
```

**Affected components**

**Tests required**

**Acceptance criteria**

**Duplicate check**

asadjan4611:
## What this PR does

`MetaxDevices.ScoreNode` (the **Metax-GPU** backend) branched on the scheduler `policy` string, reading a different topology annotation for each policy:

- **Binpack** → `2000 - losses[n]`
- **Spread**  → `2000 - scores[n]`

This is the same coupling that was removed for the sibling **Metax-SGPU** backend in #2404 / #2413. The device layer should not know scheduler policy names, and the `2000 - score` subtraction only existed to flip the sign so the
best node wins under both sort directions.

This PR makes `ScoreNode` return a single **policy-independent, "higher is a better node"** score:

- Prefer the `metax-tech.com/gpu.topology.scores` annotation.
- Fall back to `metax-tech.com/gpu.topology.losses` (converted onto the same
  "higher is better" scale via `2000 - loss`) when `scores` is absent.

`MetaxDevices` now implements the policy-neutral scorer marker, so the shared `OverrideScore` layer owns the `±10000` weighting and the Spread-policy sign inversion — exactly as it already does for Metax-SGPU. **`OverrideScore` itself
is unchanged.**

Fixes #2572

## Why the node ranking is unchanged

When a node advertises **both** annotations, they describe the same underlyingtopology preference, so the lowest-loss node is also the highest-score node. Because `OverrideScore` inverts the sign under Spread (which selects the lowest
score) and keeps it under Binpack (which selects the highest), the **winning node is the same under both policies** as before.

A new test, `TestOverrideScoreMetaxGPUOrderingUnchanged`, asserts this directly:
it builds nodes with consistent annotations, computes the winner each *original*
per-policy rule would have chosen, then weights and sorts the nodes exactly as the scheduler does (`OverrideScore` → `sort.Sort` → `NodeList[len-1]`) and checks the same node wins under both Binpack and Spread.

> **Note for reviewers:** this preserves ordering
> under the assumption that a node exposing both annotations encodes one
> consistent topology preference (min-loss node = max-score node). The issue
> invites maintainer input on whether `losses` and `scores` can ever rank nodes
> inconsistently, and on whether the two annotations should ultimately be
> unified. Happy to adjust the fallback direction (or gate it) based on that.

## Scope

Exactly three files, no changes to the shared scheduler infrastructure that landed with #2413:

- `pkg/device/metax/device.go` — the fix
- `pkg/device/metax/device_test.go` — policy-independence + marker-interface tests
- `pkg/scheduler/policy/node_policy_test.go` — weighted Binpack/Spread cases + ordering-equivalence test

## Acceptance criteria (from #2572)

- [x] `MetaxDevices.ScoreNode` no longer reads the `policy` parameter
- [x] `MetaxDevices` implements the policy-neutral scorer marker
- [x] Binpack and Spread give the same node ranking for Metax-GPU as before (test)
- [x] Policy-independence test added
- [x] Marker-interface test added
- [x] Existing tests still pass
- [x] No scheduling behavior change for Metax-GPU (aside from the intended fix)

## Testing

- `go test ./pkg/device/metax/... ./pkg/scheduler/policy/... -short --race -count=1` — pass
- `go vet` clean, `go build ./...` clean
- `goimports` (local prefix) clean, import aliases correct, license headers present
- `golangci-lint run` on the changed packages — **0 issues**

## Summary by CodeRabbit

* **Improvements**
  * Improved accelerator topology scoring to provide consistent, policy-independent results.
  * Score-based and loss-based topology annotations are now interpreted consistently, with clear precedence rules and fallback behavior.
  * Preserved expected node ordering under both Binpack and Spread scheduling policies.
  * Improved handling of missing topology information and multi-device configurations.

### DIFF
--- a/pkg/device/metax/device.go
+++ b/pkg/device/metax/device.go
@@ -182,36 +182,45 @@ func parseMetaxAnnos(annos string, index int) float32 {
 	return float32(res)
 }
 
+// ScoreNode returns a policy-independent score for the node following a
+// "higher score is a better node" convention. The shared scheduler policy layer
+// is responsible for weighting this score and adapting it to the active
+// scheduling policy (for example, inverting it under the Spread policy).
+//
+// Metax publishes two topology annotations for the requested device count: a
+// "scores" map where higher is better and a "losses" map where lower is better.
+// The scores annotation is used directly; when it is absent the losses
+// annotation is converted onto the same "higher is better" scale. This keeps a
+// node's topology preference effective regardless of which of the two
+// annotations it publishes.
 func (dev *MetaxDevices) ScoreNode(node *corev1.Node, podDevices device.PodSingleDevice, previous []*device.DeviceUsage, policy string) float32 {
 	sum := 0
 	for _, dev := range podDevices {
 		sum += len(dev)
 	}
 
-	res := float32(0)
-	if policy == string(util.NodeSchedulerPolicyBinpack) {
-		lossAnno, ok := node.Annotations[MetaxAnnotationLoss]
-		if ok {
-			// it's preferred to select the node with lower loss
-			loss := parseMetaxAnnos(lossAnno, sum)
-			res = 2000 - loss
-
-			klog.InfoS("Detected annotations", "policy", policy, "key", MetaxAnnotationLoss, "value", lossAnno, "requesting", sum, "extract", loss)
-		}
-	} else if policy == string(util.NodeSchedulerPolicySpread) {
-		scoreAnno, ok := node.Annotations[MetaxAnnotationScore]
-		if ok {
-			// it's preferred to select the node with higher score
-			// But we have to give it a smaller value because of Spread policy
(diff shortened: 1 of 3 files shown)


mesutoezdil:

asadjan4611:
@mesutoezdil 
can you please highlight the issues  bcz i spent a lot of time in this PR and i can't waste it

PR #2185, opened 2026-07-28
docs: fix annotation key and encoding format in protocol.md

chidwipak:
**What type of PR is this?**

/kind documentation

**What this PR does / why we need it**:

The "Schedule Decision" part of `docs/develop/protocol.md` did not match the code, so a reader would look for annotation keys and a format that do not exist.

- The example used `hami.io/devices-to-allocate`. That key is never written. NVIDIA uses `hami.io/vgpu-devices-to-allocate`.
- Each device is encoded as four comma-separated fields `UUID,type,memory,cores` (from `EncodeContainerDevices`), not `memory:core`.
- Devices in a container are joined by `:` and containers by `;` (`OneContainerMultiDeviceSplitSymbol` and `OnePodMultiContainerSplitSymbol`).

After the first review I also fixed two things :

- The key is not built from the device type. Each backend sets its own key in `InRequestDevices`, and the names differ: NVIDIA `vgpu`, Cambricon `cambricon-mlu`, Mthreads `mthreads-vgpu`, Hygon `dcu`. I dropped the `{device-type}` template and listed the real keys.
- I added a note that the scheduler writes one `-devices-to-allocate` annotation per device type (`EncodePodDevices`), so a pod that asks for two types gets one annotation per type.

I also added language identifiers to the fenced code blocks in the file for markdown lint (MD040).

**Which issue(s) this PR fixes**:

None. Small documentation fix, under the 100-line threshold that needs an issue first.

**Special notes for your reviewer**:

The Device Register section is unchanged apart from adding `text` to its code fences. I left its existing example content alone since it is outside this change.

AI assistance: I used GitHub Copilot to help navigate the code and draft the wording. I checked every statement against the implementation myself: `InRequestDevices` in each backend under `pkg/device/`, `EncodeContainerDevices` / `EncodePodSingleDevice` / `EncodePodDevices` in `pkg/device/devices.go`, and the annotation constants in `pkg/util/types.go`.

**Does this PR introduce a user-facing change?**:

NONE

PR #2583, opened 2026-08-11
test(device): fuzz container codec round trip

### LINKED ISSUE #2582: test(device): add semantic round-trip property to container codec fuzzing

Add a structure-aware fuzz test for the internal container-device annotation codec in `pkg/device/devices_test.go`.

The existing `FuzzDecodeContainerDevices` passes arbitrary strings to `DecodeContainerDevices`, but discards both returned values:

```go
_, _ = DecodeContainerDevices(str)
```

This is useful for panic detection, but it cannot detect a successful decode that returns incomplete or unexpected device data. Please retain that no-panic target and add a separate encoder-to-decoder property target, for example `FuzzEncodeDecodeContainerDevices`.

The new target should:

- Build `ContainerDevices` values from fuzzed UUID, type, memory, and core fields.
- Restrict UUID and type values to the delimiter-safe domain supported by the current positional wire format (`,`, `:`, and `;` are separators).
- Encode with `EncodeContainerDevices`.
- Decode the encoded annotation with `DecodeContainerDevices`.
- Assert that decoding returns no error and preserves every serialized field.
- Leave `Idx` and `CustomInfo` at their non-serialized defaults, since `EncodeContainerDevices` intentionally does not encode them.

This should remain a test-only change. It does not propose changing which externally supplied malformed annotations the decoder accepts.

Suggested verification:

```console
go test ./pkg/device -run '^$' -fuzz '^FuzzEncodeDecodeContainerDevices$' -fuzztime=30s
make verify
```

**Why is this needed**:

The current fuzz target from #2146 verifies only that arbitrary annotation input does not panic. The later node/pod fuzzing work in #2373 and #2376 follows the same no-panic pattern for decoder input, while adding deterministic round-trip coverage elsewhere. There is still no structure-aware fuzz property proving that annotations produced by `EncodeContainerDevices` can always be consumed by `DecodeContainerDevices` without losing the serialized allocation fields.

I noticed this gap after reporting #2507 and opening #2508. That change proposed rejecting malformed non-empty decoder segments, but #2508 was closed because this annotation is treated as an internal API and the preferred prevention was to test it fully before release. This proposal follows that direction directly: exercise the internal producer and consumer together without changing the parser contract.

This matters because a panic-only fuzz target would not catch the class of semantic failure discussed in #2507. Adding `"malformed"` to its corpus, for example, would still pass as long as the decoder did not panic. A round-trip oracle gives the fuzzer a concrete correctness property instead of only a process-safety property.

No GPU hardware or Kubernetes cluster is required.

Related work reviewed before filing:

- #2146 added `FuzzDecodeContainerDevices`, with a no-panic invariant only.
- #2373 requested node/pod fuzz and round-trip coverage.
- #2376 implemented node/pod decoder fuzzing and deterministic node round-trip tests, but did not add a structure-aware container codec fuzz property.
- #1856 and #1857 improved decoder error propagation; they do not test encoder/decoder symmetry.
- #2507 and #2508 cover the malformed-segment behavior that led to this test-focused follow-up.
- #2514 and #2524 concern whitespace parsing and do not provide this property coverage.

I also searched current open and closed HAMi issues and pull requests for container codec fuzzing, semantic fuzz assertions, property tests, and `EncodeContainerDevices` round trips. I did not find an existing issue or pull request with this specific scope.

blackdragoon26:
**What type of PR is this?**
/kind failing-test

Adds a structure-aware fuzz test for the internal container-device annotation codec.

The existing `FuzzDecodeContainerDevices` verifies that arbitrary input does not panic, but it discards both the decoded result and error. Therefore, it cannot detect semantic regressions where an annotation is accepted but its serialized allocation fields are changed or lost.

The new `FuzzEncodeDecodeContainerDevices`:

- Generates UUID, device type, memory, and core values.
- Restricts UUID and type inputs to the delimiter-safe wire-format domain.
- Encodes the generated allocation with `EncodeContainerDevices`.
- Decodes it with `DecodeContainerDevices`.
- Verifies that decoding succeeds and preserves every serialized field.
- Covers empty, zero, negative, and `int32` boundary values.

The existing arbitrary-input/no-panic fuzzer remains unchanged. This PR is test-only and does not change annotation parsing behavior.

Fixes #2582 

This follows the internal-API testing direction discussed after #2507 and #2508. It tests annotations generated by HAMi itself without reopening the malformed external-input parsing contract.

Validation performed on macOS:

- Focused fuzz corpus test passed.
- 30-second fuzz run passed with 268,803 executions.
- `go test ./pkg/device/...` passed.
- `make verify` passed in a clean checkout.
- `git diff --check` passed.

No GPU hardware or Kubernetes cluster is required.

Nope, this is a test-only change.

AI Disclosure:
I used OpenAI Codex to inspect the related annotation-codec history, check for duplicate issues and pull requests, and assist with implementing and validating this focused fuzz test. I reviewed the complete change, understand the tested round-trip property, and ran all verification commands myself.

## Summary by CodeRabbit

* **Tests**
  * Added fuzz testing to verify reliable round-trip encoding and decoding of container-device data.
  * Added coverage for preserving valid input values without errors.

### DIFF
--- a/pkg/device/devices_test.go
+++ b/pkg/device/devices_test.go
@@ -20,6 +20,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"strings"
 	"testing"
 
 	"gotest.tools/v3/assert"
@@ -1719,6 +1720,28 @@ func FuzzDecodeContainerDevices(f *testing.F) {
 	f.Fuzz(func(t *testing.T, str string) {
 		// Must never panic on arbitrary pod-annotation input.
 		_, _ = DecodeContainerDevices(str)
+	})
+}
+
+func FuzzEncodeDecodeContainerDevices(f *testing.F) {
+	f.Add("GPU-uuid", "NVIDIA", int32(1024), int32(10))
+	f.Add("", "", int32(0), int32(0))
+	f.Add("GPU-boundary", "NVIDIA", int32(-2147483648), int32(2147483647))
+	f.Fuzz(func(t *testing.T, uuid, deviceType string, usedmem, usedcores int32) {
+		if strings.ContainsAny(uuid, ",:;") || strings.ContainsAny(deviceType, ",:;") {
+			t.Skip()
+		}
+
+		want := ContainerDevices{{
+			UUID:      uuid,
+			Type:      deviceType,
+			Usedmem:   usedmem,
+			Usedcores: usedcores,
+		}}
+		encoded := EncodeContainerDevices(want)
+		got, err := DecodeContainerDevices(encoded)
+		assert.NilError(t, err)
+		assert.DeepEqual(t, want, got)
 	})
 }
 


pkg/device/devices_test.go
```diff
@@ -1722,6 +1723,28 @@ func FuzzDecodeContainerDevices(f *testing.F) {
 	})
 }
 
+func FuzzEncodeDecodeContainerDevices(f *testing.F) {
```
mesutoezdil, line 1726: containerdevice also has idx and a pending slots field from pr #2623. why fuzz only uuid, type, usedmem, usedcores and not those two?

blackdragoon26: Because the current annotation contract as per this merged PR[#2185](https://github.com/Project-HAMi/HAMi/pull/2185) and the implementation done  by [EncodeContainerDevices](https://github.com/Project-HAMi/HAMi/blob/master/pkg/device/devices.go#L448-L452), encodes only, UUID, type, memory, and cores.
Plus, [Idx is explicitly marked as non-encoded](https://github.com/Project-HAMi/HAMi/blob/master/pkg/device/devices.go#L209-L216), so including it in a DeepEqual round trip would fail by design.

blackdragoon26: slots is not on master and #2623 keeps it non-encoded, so i have kept it outside this round-trip property.

archlitchi (approved):
/lgtm

PR #2588, opened 2026-08-11
fix(monitor): add missing lock in Observe to prevent data race

### LINKED ISSUE #2573: [BUG]: Fatal concurrent map iteration data race in vGPUmonitor feedback observation

Rickydama3:
**What type of PR is this?**
/kind bug

> Disclosure: I used AI to help me track down where the lock was missing, but I wrote and tested the fix myself.

I noticed a data race in `vGPUmonitor` that causes a panic. `Observe()` in `cmd/vGPUmonitor/feedback.go` reads `lister.ListContainers()` without grabbing the lock, but `Update()` in `cudevshr.go` mutates that exact same map via the pod informer. 

This just adds `lister.Lock()` and `defer lister.UnLock()` to `Observe()` to fix it, matching how `metrics.go` safely reads it. I also added a quick test in `feedback_test.go` to cover the new lines.

Fixes #2573

Nothing special, just a straightforward concurrency fix.

```release-note
None

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved monitoring reliability by coordinating container-list access during observation and feedback processing.
  * Prevented conflicting updates while monitoring is in progress.
  * Ensured observation completes safely once container-list access becomes available.

* **Tests**
  * Added coverage for container-list locking, blocking behavior, and successful completion after access is released.

### DIFF
--- a/cmd/vGPUmonitor/feedback.go
+++ b/cmd/vGPUmonitor/feedback.go
@@ -71,7 +71,12 @@ func CheckPriority(utSwitchOn map[string]UtilizationPerDevice, p int, c *nvidia.
 	return false
 }
 
+var observeTestHook func()
+
 func Observe(lister *nvidia.ContainerLister) {
+	if observeTestHook != nil {
+		observeTestHook()
+	}
 	lister.Lock()
 	defer lister.UnLock()
 
--- a/cmd/vGPUmonitor/feedback.go
+++ b/cmd/vGPUmonitor/feedback.go
@@ -72,6 +72,9 @@ func CheckPriority(utSwitchOn map[string]UtilizationPerDevice, p int, c *nvidia.
 }
 
 func Observe(lister *nvidia.ContainerLister) {
+	lister.Lock()
+	defer lister.UnLock()
+
 	utSwitchOn := map[string]UtilizationPerDevice{}
 	containers := lister.ListContainers()
 
--- a/cmd/vGPUmonitor/feedback_test.go
+++ b/cmd/vGPUmonitor/feedback_test.go
@@ -169,22 +169,22 @@ func TestObserve(t *testing.T) {
 
 	// Test that Observe actually acquires the lock deterministically.
 	lister.Lock()
-	
+
 	reachedLock := make(chan struct{})
 	observeTestHook = func() {
 		close(reachedLock)
 	}
 	defer func() { observeTestHook = nil }()
-	
+
 	done := make(chan struct{})
 	go func() {
 		Observe(lister)
 		close(done)
 	}()
-	
+
 	// Wait until Observe reaches the lock boundary
 	<-reachedLock
-	
+
 	// Ensure Observe is now blocked on the lock
 	select {
 	case <-done:
@@ -192,10 +192,10 @@ func TestObserve(t *testing.T) {
 	default:
 		// Expected: Observe is blocked
 	}
-	
+
 	// Release the lock
 	lister.UnLock()
-	
+
 	// Now Observe should complete
 	select {
 	case <-done:
--- a/cmd/vGPUmonitor/feedback_test.go
+++ b/cmd/vGPUmonitor/feedback_test.go
@@ -167,8 +167,14 @@ func TestObserve(t *testing.T) {
 	// and to ensure no panics occur with locking/unlocking.
 	lister := &nvidia.ContainerLister{}
 
-	// Test that Observe actually acquires the lock.
+	// Test that Observe actually acquires the lock deterministically.
 	lister.Lock()
+	
+	reachedLock := make(chan struct{})
+	observeTestHook = func() {
+		close(reachedLock)
+	}
+	defer func() { observeTestHook = nil }()
 	
 	done := make(chan struct{})
 	go func() {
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: cmd/vGPUmonitor/feedback_test.go [hunk 1]
```diff
@@ -160,3 +160,10 @@ func TestCheckBlocking_MultiDevice(t *testing.T) {
 		})
 	}
 }
+
+func TestObserve(t *testing.T) {
+	// Call Observe with an empty lister to cover the missing lines for codecov
+	// and to ensure no panics occur with locking/unlocking.
+	lister := &nvidia.ContainerLister{}
+	Observe(lister)
+}
```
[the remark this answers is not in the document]
Rickydama3: "Great suggestion! I've updated the test to actively hold lister.Lock() in the main goroutine and verify that Observe() correctly blocks until lister.UnLock() is called. This perfectly exercises the lock contention."

cmd/vGPUmonitor/feedback_test.go
```diff
@@ -160,3 +161,38 @@ func TestCheckBlocking_MultiDevice(t *testing.T) {
...
+	}()
+	
+	// Observe should be blocked waiting for the lock, so it shouldn't complete yet.
+	select {
+	case <-done:
+		t.Fatal("Observe completed while lock was held by another goroutine, indicating it did not acquire the lock!")
+	// A small delay to ensure the goroutine has time to schedule and block
+	case <-time.After(50 * time.Millisecond):
+		// Expected: Observe is blocked
+	}
```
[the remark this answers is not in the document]
Rickydama3: Excellent catch! I have added a observeTestHook synchronization point immediately before the lock attempt to make the lock-contention assertion 100% deterministic, completely removing the reliance on a fixed delay timeout

cmd/vGPUmonitor/feedback.go
```diff
@@ -71,7 +71,15 @@ func CheckPriority(utSwitchOn map[string]UtilizationPerDevice, p int, c *nvidia.
 	return false
 }
 
+var observeTestHook func()
+
 func Observe(lister *nvidia.ContainerLister) {
+	if observeTestHook != nil {
+		observeTestHook()
+	}
+	lister.Lock()
```
mesutoezdil, line 80: issue #2573 says pod add/update informer events call update() at the same time as observe(). this file only registers deletefunc, and that one only logs, it never touches containers. where does the real concurrent write come from today? the lock is still correct and safe to add either way, just checking the reproduction story is accurate.

mesutoezdil (commented):
left one question inline

cmd/vGPUmonitor/feedback.go
Rickydama3: The issue description doesn't quite get the informer events right. `Update()` and `Observe()` just run sequentially on the ticker loop, so they don't race with each other.

The real race is between `Observe()` modifying `c.Info` in the background and Prometheus concurrently scraping the `/metrics` endpoint. Since `Collect()` already grabs the lister lock when responding to Prometheus scrapes, grabbing the same lock in `Observe()` properly synchronizes the writes.

Rickydama3: @mesutoezdil I've answered your question inline! Let me know if you need anything else from my end.

mesutoezdil, line 80: pr #2311 hit a real deadlock adding a lock inside listcontainers, because metrics.go already holds the lock when it calls listcontainers. why does this lock here not hit the same issue?

Rickydama3: The issue in #2311 happened because metrics.go already held the lock. When it called ListContainers(), adding a lock inside there made the exact same code try to lock it twice, which causes a deadlock.

The lock here is safe because metrics.go never actually calls Observe(). Observe() only runs in its own separate background loop watchAndFeedback. Since they run totally separate from each other, they just wait their turn for the lock. No double-locking can happen here. @mesutoezdil hope you liked it

mesutoezdil:

PR #2601, opened 2026-08-11
fix(ascend): guard int32 narrowing in GenerateResourceRequests

### LINKED ISSUE #2600: ascend: GenerateResourceRequests narrows user-controlled values to int32 without bounds checks
## What happens

`GenerateResourceRequests` in `pkg/device/ascend/device.go` converts the user-controlled device count, memory, and core values into the int32 fields of `device.ContainerDeviceRequest` (`Nums`, `Memreq`, `Coresreq`) with no bounds check:

- device count to `Nums`: `int32(n)`, no guard
- memory to `Memreq`: raw value or value times `MemoryFactor`, no guard, on both the soft-partitioning path and the `trimMemory` path
- cores to `Coresreq`: `int32(cv.Value())`, no guard

When any of these exceeds `math.MaxInt32` (2147483647), the conversion silently wraps to a small or negative value.

## Why it matters

The input is fully user-controlled (anyone who can submit a pod). A memory request that wraps to 0 makes the scheduler's `Fit()` check (`Totalmem - Usedmem < memreq`) always pass, so the pod can be scheduled onto an already-full device. The result is oversubscription, incorrect accounting, and OOM that can affect other workloads on the same card. It fails silently, with no error to the user. Concretely, a memory request of `16Gi` on the soft-partitioning path produces both `Memreq` and `MemPercentagereq` equal to 0, and a large device count wraps negative.

## Expected behavior

Oversized or negative requests should be rejected rather than silently wrapped, by returning an empty `device.ContainerDeviceRequest{}`, matching the protection already added for hygon and metax-sgpu in #2388.

## Notes

Same overflow class as #2383 and #2388. Issue #2336 named ascend among others but was closed before an ascend fix merged. This is distinct from #2532, an over-capacity-but-within-int32 case (below `math.MaxInt32`, collapsed to a whole card by `trimMemory`), which is not an int32 overflow.

Eshiv-Pandey:
**What type of PR is this?**
/kind bug

`GenerateResourceRequests` in the ascend backend converts user-controlled `int64` resource limits (device count, memory, cores) into the `int32` fields of `ContainerDeviceRequest` (`Nums`, `Memreq`, `Coresreq`) without bounds checking. A value exceeding `math.MaxInt32` silently wraps to a small or negative number that flows into `Fit()`. A memory request that wraps to 0 makes the `Fit()` check (`Totalmem - Usedmem < memreq`) always pass, so the scheduler can place a pod onto an already-full NPU, causing oversubscription and OOM for other workloads on the same card.

This PR adds range guards that return the existing empty `ContainerDeviceRequest{}` rejection sentinel before every `int32()` narrowing, matching the pattern from #2388 (hygon, metax-sgpu). Memory is guarded both before and after the `MemoryFactor` multiplication, covering the soft-partitioning path (raw value) and the `trimMemory` path.

Fixes #2600

- Guards four conversion points: device count, memory pre-factor, memory post-factor (after `MemoryFactor` scaling), and cores.
- After studying #2285, I followed the maintainer's suggestion there that an over-int32 memory value on ascend is almost always a wrong-unit mistake (for example `16Gi` instead of an integer number of MB). The memory error messages state that Ascend memory is treated as MB not Byte, and that `16384` should be requested for 16 GB.
- Scope is the int32 overflow case only, matching #2388. It does not address #2532 (an over-capacity-but-within-int32 case where `trimMemory` returns 0), which is below `math.MaxInt32` and out of scope here.
- Tests cover pre-factor overflow (`16Gi` on both the soft-partitioning and trim paths), post-factor overflow after `MemoryFactor`, oversized device count, and negative and oversized core requests. Each was confirmed to fail before the guard and pass after.
- Change is scoped to the scheduler extender, so unit tests are used rather than on-hardware validation, consistent with #2388.

```
Oversized or invalid Ascend device resource requests (for example an ascend memory request such as 16Gi, or values exceeding the int32 range) are now rejected instead of silently wrapping to an incorrect value.
```

AI assistance disclosure: I used AI assistance (Claude Code) to help audit the backends, implement the guards, and draft the regression tests. I reviewed the change and verified the new tests fail without the guards and pass with them.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation of device, memory, and core resource requests.
  * Invalid, negative, or oversized values are now rejected safely.
  * Prevented errors caused by memory scaling exceeding supported limits.

* **Tests**
  * Added coverage for invalid resource values and overflow scenarios across supported request paths.

### DIFF
--- a/pkg/device/ascend/device.go
+++ b/pkg/device/ascend/device.go
@@ -21,6 +21,7 @@ import (
 	"errors"
 	"flag"
 	"fmt"
+	"math"
 	"slices"
 	"sort"
 	"strconv"
@@ -312,6 +313,10 @@ func (dev *Devices) GenerateResourceRequests(ctr *corev1.Container) device.Conta
 		klog.V(3).Infof("Counting %s devices", dev.config.CommonWord)
 		if n, ok := v.AsInt64(); ok {
 			klog.Info("Found AscendDevices devices")
+			if n <= 0 || n > math.MaxInt32 {
+				klog.ErrorS(nil, "ascend device count request is out of range", "container", ctr.Name, "request", n)
+				return device.ContainerDeviceRequest{}
+			}
 			memnum := 0
 			mem, ok := ctr.Resources.Limits[ascendResourceMem]
 			if !ok {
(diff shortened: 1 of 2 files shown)


pkg/device/ascend/device.go
```diff
@@ -312,17 +313,38 @@ func (dev *Devices) GenerateResourceRequests(ctr *corev1.Container) device.Conta
...
+				return device.ContainerDeviceRequest{}
+			}
 			memnum := 0
 			mem, ok := ctr.Resources.Limits[ascendResourceMem]
 			if !ok {
 				mem, ok = ctr.Resources.Requests[ascendResourceMem]
 			}
 			if ok {
+				// Negative quantities such as -1m return ok=false from AsInt64, so reject by sign first.
+				if mem.Sign() < 0 {
```
mesutoezdil, line 327: comment above only talks about -1m being fractional. does this sign check also catch a plain whole negative like -100, before asint64 even runs?

Eshiv-Pandey: have checked this and found tht `-100` passes `AsInt64()` (`ok=true`), so the old overflow check alone would never have caught it but `Quantity.Sign()` returns `-1` for __any__ negative (whole or frac) and runs __before__ `AsInt64()`, so `mem.Sign() < 0` rejects both `-1m` and `-100`. Added a `-100` regression just below the existing `-1m` case.

image

archlitchi (approved):
/lgtm

PR #2605, opened 2026-08-11
fix(security): prevent symlink LPE in GPU allocation

### LINKED ISSUE #2604: fix(security): add missing error handling and prevent symlink LPE in GPU allocation
**Problem Description.**
In `pkg/device-plugin/nvidiadevice/nvinternal/plugin/server.go`, the `Allocate` function blindly creates directories and changes their permissions:

    os.MkdirAll(cacheFileHostDirectory, 0777)
    os.Chmod(cacheFileHostDirectory, 0777)
    os.MkdirAll("/tmp/vgpulock", 0777)
    os.Chmod("/tmp/vgpulock", 0777)

This causes two major issues:
1. **Missing Error Handling:** It ignores all errors. If creating the directories fails, the plugin silently continues, which breaks pod volume mounts later on.
2. **Local Privilege Escalation (LPE) Risk:** Because the plugin runs as `root`, and `/tmp` is world-writable, a malicious user can create a symlink at `/tmp/vgpulock` pointing to `/etc/shadow`. Go's `os.Chmod` follows symlinks, so this code will silently change `/etc/shadow` to `0777` (world readable/writable), allowing a full node takeover.

**Solution**
1. Add `if err != nil` checks and return the error so the allocation fails gracefully if directory creation fails.
2. Fix the symlink vulnerability by verifying the path is a real directory (using `os.Lstat`) before applying `os.Chmod`, or just remove the `os.Chmod` if `MkdirAll` is sufficient.

KunwarSidhu47:
**Title:** fix(security): prevent symlink LPE and add error handling in GPU allocation

**Description:**

This PR fixes a critical local privilege escalation (LPE) vulnerability and missing error handling within the Nvidia device plugin's `Allocate` function (`pkg/device-plugin/nvidiadevice/nvinternal/plugin/server.go`).

Previously, the plugin blindly created directories and modified their permissions without checking for errors or verifying the file type. Because the device plugin runs as `root` and `/tmp` is world-writable, a malicious local user or unprivileged pod could pre-create a symlink at `/tmp/vgpulock` pointing to a sensitive file (e.g., `/etc/shadow`). Since Go's `os.Chmod` follows symlinks, allocating a GPU would silently change the permissions of `/etc/shadow` to `0777`, granting the attacker full control over the node.

Additionally, because the return values of `os.MkdirAll` and `os.Chmod` were completely ignored, any failure to create these directories would be silently swallowed, leading to unpredictable volume mount failures downstream.

Fixes #2604 

- Added strict `if err != nil` checks around `os.MkdirAll` to ensure we fail gracefully if directory creation fails.
- Switched to using `os.Lstat` prior to `os.Chmod` to explicitly verify that the target path is an actual directory and `os.ModeSymlink` is NOT set. This mitigates the symlink TOCTOU attack vector while preserving the required `0777` permission enforcement.

**Does this PR introduce a user-facing change?:**
No, internal security and error handling improvement.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved vGPU allocation reliability by validating required cache and lock directories.
  * Allocation now fails clearly when lock paths are symlinks, invalid, or cannot be created.
  * Existing directories are updated only after confirming they are valid directories.

### DIFF
--- a/pkg/device-plugin/nvidiadevice/nvinternal/plugin/server.go
+++ b/pkg/device-plugin/nvidiadevice/nvinternal/plugin/server.go
@@ -838,10 +838,23 @@ func (plugin *NvidiaDevicePlugin) Allocate(ctx context.Context, reqs *kubeletdev
 				cacheFileHostDirectory := fmt.Sprintf("%s/vgpu/containers/%s_%s", hostHookPath, current.UID, currentCtr.Name)
 				os.RemoveAll(cacheFileHostDirectory)
 
-				os.MkdirAll(cacheFileHostDirectory, 0777)
-				os.Chmod(cacheFileHostDirectory, 0777)
-				os.MkdirAll("/tmp/vgpulock", 0777)
-				os.Chmod("/tmp/vgpulock", 0777)
+				if err := os.MkdirAll(cacheFileHostDirectory, 0777); err != nil {
+					klog.Errorf("Failed to create cache directory: %v", err)
+					return nil, err
+				}
+				if info, err := os.Lstat(cacheFileHostDirectory); err == nil && info.Mode().IsDir() && info.Mode()&os.ModeSymlink == 0 {
+					_ = os.Chmod(cacheFileHostDirectory, 0777)
+				}
+
+				if err := os.MkdirAll("/tmp/vgpulock", 0777); err != nil {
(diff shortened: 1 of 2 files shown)


mesutoezdil:

PR #2610, opened 2026-08-12
fix(device-plugin): make MigInstanceManager own its NVML init/shutdown

AdityaShome:
## What this does

`MigInstanceManager` (migmgr.go) makes raw NVML calls in several of its methods but never calls `nvml.Init()` itself. It only works today because `nvmlBusyGPUs()` (mig_startup.go) happens to leave NVML initialized at startup and never shuts it down an implicit, undocumented dependency on that leak.

This PR makes `MigInstanceManager` self-contained: each public entry point that touches NVML (`ResetIdleGPUs`, `Release`, `EnsureAllocation`, `AdoptAllocation`, `ReconcileActiveAllocations`) now brackets its own `nvml.Init()` / `defer nvml.Shutdown()`, matching the self-contained convention used everywhere else in this package (e.g. `util.go`).

NVML's `Init`/`Shutdown` pair is reference-counted per-process, so this is safe to call independently alongside other NVML users in the same process without tearing down a session another caller still needs.

## Why

Found while reviewing the MIG refactor (#2378). Once the companion leak in `mig_startup.go`'s `gpuUUIDToIndex` / `nvmlBusyGPUs` is fixed to properly call `nvml.Shutdown()`, `MigInstanceManager` would start failing with "NVML not initialized" since it never initializes NVML on its own. This PR removes that hidden coupling first.
Discussed in #2246.

## Testing

- `go build ./pkg/device-plugin/nvidiadevice/nvinternal/plugin/...`
- `go test ./pkg/device-plugin/nvidiadevice/nvinternal/plugin/... -short --race -count=1`

AI disclosure: AI was used for exploring possible approaches to identify an appropriate solution.

## Bug Fixes

- Improved reliability of GPU partition (MIG) allocation and cleanup operations.
- Improved handling of GPU resource adoption, release, reset, and recovery workflows.
- Reconciliation now skips unnecessary processing when no active allocations are present.
- Systems report GPU management initialization failures more clearly.
- Improved resource management across service restarts and final shutdown.
- GPU management sessions now remain available across service restarts and are released cleanly during final shutdown.

### DIFF
--- a/pkg/device-plugin/nvidiadevice/nvinternal/plugin/migmgr.go
+++ b/pkg/device-plugin/nvidiadevice/nvinternal/plugin/migmgr.go
@@ -97,6 +97,10 @@ func profileSliceKey(profile string) string {
 // have MIG mode enabled and all existing GI/CI instances destroyed.
 func (m *MigInstanceManager) ResetIdleGPUs(deviceCount int, inUse map[int]struct{}) ([]int, error) {
 	reset := []int{}
+	if nvret := nvml.Init(); nvret != nvml.SUCCESS {
+		return reset, fmt.Errorf("nvml Init: %s", nvml.ErrorString(nvret))
+	}
+	defer nvml.Shutdown()
 	for gpuIndex := 0; gpuIndex < deviceCount; gpuIndex++ {
 		if _, busy := inUse[gpuIndex]; busy {
 			continue
@@ -284,6 +288,10 @@ func (m *MigInstanceManager) Release(migUUID string) error {
 	lk := m.gpuLock(key.GPUIndex)
 	lk.Lock()
 	defer lk.Unlock()
+	if nvret := nvml.Init(); nvret != nvml.SUCCESS {
+		return fmt.Errorf("nvml Init: %s", nvml.ErrorString(nvret))
+	}
+	defer nvml.Shutdown()
 	m.mu.Lock()
 	inst := m.byAllocation[key]
 	m.mu.Unlock()
@@ -321,6 +329,11 @@ func (m *MigInstanceManager) EnsureAllocation(gpuIndex int, profile string, plac
 		return uuid, false, nil
 	}
 	m.mu.Unlock()
+
+	if nvret := nvml.Init(); nvret != nvml.SUCCESS {
+		return "", false, fmt.Errorf("nvml Init: %s", nvml.ErrorString(nvret))
+	}
+	defer nvml.Shutdown()
 
 	if err := ensureMigModeEnabled(gpuIndex); err != nil {
 		return "", false, err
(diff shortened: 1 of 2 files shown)


SK8-infi:
@AdityaShome Heads up, AI disclosure is missing in the PR description. Maintainers here prefer to have an AI disclosure in PR Desc.
https://github.com/Project-HAMi/HAMi/blob/master/CONTRIBUTING.md

AdityaShome:
@SK8-infi Thanks, it is updated.

FouoF:
@AdityaShome The issue is valid, but initializing and shutting down NVML in every method adds unnecessary overhead and complicates concurrent/error handling. I’d recommend managing NVML once at the `MigInstanceManager` or process lifecycle level instead.

pkg/device-plugin/nvidiadevice/nvinternal/plugin/migmgr.go
```diff
@@ -322,6 +330,11 @@ func (m *MigInstanceManager) EnsureAllocation(gpuIndex int, profile string, plac
 	}
 	m.mu.Unlock()
 
+	if nvret := nvml.Init(); nvret != nvml.SUCCESS {
```
mesutoezdil: ensureMigModeEnabled and deviceHandleByIndex do not call nvml.Init themselves. do they assume the caller already did, and does that make this line required before them, not optional?

AdityaShome:
@FouoF agreed, per-method Init/Shutdown was wasteful on hot paths (5s reconciler, every Allocate/Release). Moved to a single Init/Shutdown owned by MigInstanceManager: Init() runs once at construction, Shutdown() runs once on plugin context cancellation (not Stop(), which only restarts the gRPC server).

@mesutoezdil confirmed neither helper calls nvml.Init() itself, so yes, NVML init was always required first. It's now explicit and centralized instead of implicit per-method.

Sorry for delay, also added a short doc comment on each affected method noting it requires NVML already initialized via Init, since that's no longer visible in the method body itself.

pkg/device-plugin/nvidiadevice/nvinternal/plugin/migmgr.go
```diff
@@ -469,13 +497,17 @@ func (m *MigInstanceManager) AdoptAllocation(gpuIndex int, profile, migUUID stri
...
 
+// Requires NVML already initialized via Init.
 func (m *MigInstanceManager) ReconcileActiveAllocations(active map[migAllocationKey]struct{}) error {
 	m.mu.Lock()
 	keys := make([]migAllocationKey, 0, len(m.byAllocation))
 	for key := range m.byAllocation {
 		keys = append(keys, key)
 	}
 	m.mu.Unlock()
+	if len(keys) == 0 {
```
Shouren: @AdityaShome This is redundant. With an empty keys slice, the following for _, key := range keys already performs zero iterations and the function returns nil.

pkg/device-plugin/nvidiadevice/nvinternal/plugin/migmgr.go
```diff
@@ -273,6 +296,8 @@ func destroyAllMigInstances(dev nvml.Device) error {
 }
 
 // Release destroys the GI+CI bound to the given MIG UUID.
+//
+// Requires NVML already initialized via Init.
```
Shouren: @AdityaShome Please remove the unnecessary comments.

Shouren (dismissed):

AdityaShome:
It is modified as requested.

archlitchi (approved):
/lgtm

PR #2611, opened 2026-08-12
fix kunlun and mthreads device index tracking for multi-container pods

### LINKED ISSUE #2463: kunlun/mthreads: PatchAnnotations overwrites per-container device index for multi-container pods

Rickydama3:
**What type of PR is this?**
/kind bug

When a pod has multiple containers requesting Kunlun or Mthreads GPUs, `PatchAnnotations` was resetting the device index string inside the container loop. Because of this, the annotation was overwritten for each container and ended up containing only the GPU indices from the last container.

This change moves the string initialization outside the loop, so the indices from all containers are collected correctly and joined into the pod annotation.

Added regression tests to cover the multi-container case as well.

**Which issue(s) does this PR fix?**:

Fixes #2463

This PR replaces #2606. The changes have been squashed into a single commit, and the required AI disclosure has been added.

```release-note
NONE
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected device and GPU assignment annotations when workloads contain multiple containers.
  * Preserved empty container entries and combined all container selections consistently.
  * Ensured assigned-node and predicate-time annotations are recorded when applicable.

* **Tests**
  * Added coverage for multi-container assignments, empty entries, and related annotation values.

### DIFF
--- a/pkg/device/kunlun/device.go
+++ b/pkg/device/kunlun/device.go
@@ -91,15 +91,21 @@ func (dev *KunlunDevices) PatchAnnotations(pod *corev1.Pod, annoinput *map[strin
 	devlist, ok := pd[KunlunGPUDevice]
 	if ok && len(devlist) > 0 {
 		(*annoinput)[device.SupportDevices[KunlunGPUDevice]] = device.EncodePodSingleDevice(devlist)
+		annoKey := KunlunDeviceSelection
+		var values []string
 		for _, dp := range devlist {
-			annoKey := KunlunDeviceSelection
 			value := ""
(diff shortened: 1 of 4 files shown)


mesutoezdil:

PR #2617, opened 2026-08-12
feat: add hami_gpu_device_health metric to scheduler

### LINKED ISSUE #2612: feat: expose hami_gpu_device_health as a Prometheus metric in scheduler
What would you like to be added:
add a gauge metric hami_gpu_device_health (1 = healthy, 0 = unhealthy) inside collectNodeMetrics. same label set as hami_gpu_memory_limit_bytes — node, device_uuid, device_index, device_type.

Why is this needed:
DeviceUsage.Health already gets set by the device plugin through CheckHealth(), and scheduler uses it internally to skip unhealthy devices while scoring. but none of that is exposed as a metric right now. so if a gpu goes unhealthy, literally the only way to catch it is going and checking node annotations manually. prometheus has zero visibility into device health atm which feels like a gap for anyone running this at scale.

Anything else we need to know:
hami_mig_device_info already does the 0/1 gauge thing for boolean state so this isn't introducing a new pattern, just extending the same one. change is scoped to the scheduler extender, will add unit tests too since that's a contribution gate here.

manoj-1407:
**What type of PR is this?**
/kind feature

scheduler was silently skipping unhealthy GPUs during scoring using 
DeviceUsage.Health but nothing ever made it to Prometheus. only way 
to know a device was unhealthy was manually checking node annotations. 
added hami_gpu_device_health gauge (1=healthy, 0=unhealthy) so you 
can actually alert on it.

Fixes #2612

same 0/1 pattern as hami_mig_device_info. descriptor kept local to 
the function like everything else in there.

yes — hami_gpu_device_health gauge

I used Claude to help identify this gap. I reviewed the code and 
understand the implementation.

## Summary by CodeRabbit

* **New Features**
  * Added a Prometheus metric reporting the health status of each GPU device.
  * Healthy devices report `1`; unhealthy devices report `0`.
  * Metrics include node, device UUID, index, and device type details.

### DIFF
--- a/cmd/scheduler/metrics.go
+++ b/cmd/scheduler/metrics.go
@@ -144,6 +144,11 @@ func (cc ClusterManagerCollector) collectNodeMetrics(ch chan<- prometheus.Metric
 		"Realized MIG instance identity and scheduler placement",
 		[]string{"node", "device_uuid", "device_index", "mig_uuid", "profile", "gpu_instance_id", "compute_instance_id", "placement_start", "placement_size"}, nil,
 	)
+	nodeGPUDeviceHealthDesc := prometheus.NewDesc(
+		"hami_gpu_device_health",
+		"GPU device health status (1=healthy, 0=unhealthy)",
+		[]string{"node", "device_uuid", "device_index", "device_type"}, nil,
+	)
 
 	// Legacy metric descriptors (only created when legacy mode is enabled)
 	var (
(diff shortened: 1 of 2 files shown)


cmd/scheduler/metrics.go
```diff
@@ -144,6 +144,11 @@ func (cc ClusterManagerCollector) collectNodeMetrics(ch chan<- prometheus.Metric
 		"Realized MIG instance identity and scheduler placement",
 		[]string{"node", "device_uuid", "device_index", "mig_uuid", "profile", "gpu_instance_id", "compute_instance_id", "placement_start", "placement_size"}, nil,
 	)
+	nodeGPUDeviceHealthDesc := prometheus.NewDesc(
```
mesutoezdil, line 147: name says gpu but this loop runs for every backend. on purpose?

cmd/scheduler/metrics.go
```diff
@@ -266,6 +271,16 @@ func (cc ClusterManagerCollector) collectNodeMetrics(ch chan<- prometheus.Metric
 				}
 			}
 
+			healthVal := float64(0)
```
mesutoezdil, line 274: is health live at scrape time or a cached value?

mesutoezdil, line 274: if a node has zero devices in devicelists, does this metric emit anything for that node at all?

cmd/scheduler/metrics.go
manoj-1407: yeah it's intentional — Health is set on all device types, the name reflects what the metric measures rather than which backend it's coming from

cmd/scheduler/metrics.go
manoj-1407: it's a cached value from the scheduler's in-memory device list, same as the other metrics in collectNodeMetrics

manoj-1407: nothing gets emitted in that case, the loop just doesn't run — same behavior as the existing MIG metric

manoj-1407:
@mesutoezdil hey, any specific reason for closing? happy to address any concerns

mesutoezdil:
we have no time and energy for llm answers. read again the contribution rules.
you should know your codes.

PR #2618, opened 2026-08-12
fix(scheduler): detect stale pod device allocations during usage reconciliation

AyushSrivastava1818:

During scheduler usage reconciliation, `getNodesUsage()` processes the device allocations reported by pods and matches each allocated device UUID against the node's currently known device inventory.

Previously, if a pod referenced a device UUID that was no longer present in `node.Devices.DeviceLists`, the allocation was silently ignored because no matching device was found. This could cause the scheduler's usage accounting to become inconsistent with the allocations reported by running pods.

This PR detects stale pod device allocations and prevents unknown device UUIDs from being silently ignored during usage reconciliation.

### Changes

* Track whether each pod device allocation successfully matches a currently known device UUID.
* Detect allocations referencing stale or missing device UUIDs.
* Log stale allocations with structured error logging.
* Prevent stale device allocations from being incorrectly accounted against currently known devices.
* Preserve normal usage accounting for valid device allocations.
* Add regression coverage for stale, valid, mixed, and multi-device allocation scenarios.

The change is intentionally scoped to the usage reconciliation path.

A missing device UUID is treated as a stale/inconsistent allocation rather than simply being ignored. This prevents unknown allocations from being silently incorporated into scheduler usage accounting while preserving the existing behavior for valid device allocations.

The implementation does not mark the node's devices unhealthy when a stale allocation is detected.

Regression coverage verifies that stale allocations are detected and logged, while valid device allocations continue to be accounted for correctly.

## Validation

The following checks were completed in a Linux x86_64 environment with CGO enabled:

* `go test -race ./pkg/scheduler/...`
* GolangCI-Lint — 0 issues
* Targeted regression test: `Test_getNodesUsage_StalePodDeviceAllocation`
* `gofmt`
* `git diff --check`

All checks passed.

No.

## AI Assistance Disclosure

AI assistance from Claude and Antigravity was used during codebase investigation, root-cause analysis, implementation assistance, regression-test development, and validation of this change.

The resulting implementation and tests were reviewed against the existing HAMi codebase, and the final changes were validated by the contributor using the project's test, lint, vet, race, formatting, and diff-check workflows.

### DIFF
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -786,9 +786,6 @@ func (s *Scheduler) getNodesUsage(nodes *[]string, task *corev1.Pod) (*map[strin
 					}
 					if !matched {
 						klog.ErrorS(nil, "pod allocated unknown or stale device resources", "pod", klog.KRef(p.Namespace, p.Name), "nodeID", p.NodeID, "gpuUUID", udevice.UUID)
-						for _, d := range node.Devices.DeviceLists {
-							d.Device.Health = false
-						}
 					}
 				}
 			}
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -757,9 +757,11 @@ func (s *Scheduler) getNodesUsage(nodes *[]string, task *corev1.Pod) (*map[strin
 		for _, podsingleds := range p.Devices {
 			for _, ctrdevs := range podsingleds {
 				for _, udevice := range ctrdevs {
+					matched := false
 					for _, d := range node.Devices.DeviceLists {
 						deviceID := udevice.UUID
 						if d.Device.ID == deviceID {
+							matched = true
 							d.Device.Used++
 							d.Device.Usedmem += udevice.Usedmem
 							d.Device.Usedcores += udevice.Usedcores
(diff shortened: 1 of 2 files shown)


pkg/scheduler/scheduler.go
```diff
@@ -757,9 +757,11 @@ func (s *Scheduler) getNodesUsage(nodes *[]string, task *corev1.Pod) (*map[strin
 		for _, podsingleds := range p.Devices {
 			for _, ctrdevs := range podsingleds {
 				for _, udevice := range ctrdevs {
+					matched := false
 					for _, d := range node.Devices.DeviceLists {
 						deviceID := udevice.UUID
 						if d.Device.ID == deviceID {
+							matched = true
 							d.Device.Used++
```
mesutoezdil, line 765: pr #2623 changes this same line to used += slots. real conflict on rebase either way.

pkg/scheduler/scheduler.go
```diff
@@ -782,6 +784,12 @@ func (s *Scheduler) getNodesUsage(nodes *[]string, task *corev1.Pod) (*map[strin
 							}
 						}
 					}
+					if !matched {
+						klog.ErrorS(nil, "pod allocated unknown or stale device resources", "pod", klog.KRef(p.Namespace, p.Name), "nodeID", p.NodeID, "gpuUUID", udevice.UUID)
+						for _, d := range node.Devices.DeviceLists {
+							d.Device.Health = false
```
mesutoezdil: one stale uuid marks every device on the node unhealthy, not just that one. is whole node the right blast radius?

AyushSrivastava1818: yes this marks the whole node unhealthy because the stale uuid means the schedulers device inventory doesn't match an existing pod allocation,but i agree this may be a broader blast radius than necessary... I though this was the safest and conservative option but I guess i will look for another alternative of this...will trace how Device.Health is used downstream and check what i can do better

pkg/scheduler/scheduler.go
AyushSrivastava1818: okay will fix this in next commit

pkg/scheduler/scheduler.go
AyushSrivastava1818: pushed the new update if you can check once

mesutoezdil:
code looks good now. one thing: 
the pr body still says devices are marked unhealthy, the code no longer does that. 
update the description and i will lgtm

AyushSrivastava1818:
@mesutoezdil updated the description have a look please

archlitchi (approved):
/lgtm

PR #2619, opened 2026-08-12
fix(vGPUmonitor): skip GPU utilization gracefully when not supported

Nakshatra480:
## Summary

- `collectGPUMemoryMetrics` already returns `nil` on `nvml.ERROR_NOT_SUPPORTED` (unified memory architectures, Jetson, certain MIG configurations). `collectGPUUtilizationMetrics` treated the same NVML return code as a hard error.
- When this happened, `collectGPUDeviceMetrics` returned an error after memory metrics had already been emitted to the channel, producing a misleading "Failed to collect metrics for GPU device N" log even though memory data was fine.
- Added the same `ERROR_NOT_SUPPORTED` guard that `collectGPUMemoryMetrics` already uses.

**Files changed:**
- `cmd/vGPUmonitor/metrics.go`: four-line guard matching the existing memory pattern

**Tested:** `go test ./cmd/vGPUmonitor/... -count=1` passes. The NVML path requires hardware; the guard mirrors an already-tested pattern from `collectGPUMemoryMetrics`.

The only change is mirroring the `if nvret == nvml.ERROR_NOT_SUPPORTED { return nil }` guard from the memory function. No logic change for the success path.

Yes devices that previously triggered a spurious error log now silently skip utilization (with a `klog.V(3)` message) and continue emitting memory metrics.

## Summary by CodeRabbit

* **Bug Fixes**
  * GPU monitoring now continues gracefully when utilization metrics aren’t supported by a device.
  * Other GPU driver errors continue to be handled as failures.
* **Tests**
  * Added coverage for unsupported metrics, driver errors, and successful utilization collection.

### DIFF
--- a/cmd/vGPUmonitor/metrics.go
+++ b/cmd/vGPUmonitor/metrics.go
@@ -326,6 +326,10 @@ func (cc ClusterManagerCollector) collectGPUMemoryMetrics(ch chan<- prometheus.M
 
 func (cc ClusterManagerCollector) collectGPUUtilizationMetrics(ch chan<- prometheus.Metric, hdev nvml.Device, index int) error {
 	util, nvret := hdev.GetUtilizationRates()
+	if nvret == nvml.ERROR_NOT_SUPPORTED {
+		klog.V(3).Infof("Utilization rates not supported for device %d, skipping", index)
+		return nil
+	}
 	if nvret != nvml.SUCCESS {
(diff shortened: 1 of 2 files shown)


mesutoezdil:

Nakshatra480:
Thanks for your feedback. I will review the contribution guidelines and ensure that my future submissions follow them. If you can, please let me know what was missing or not compliant in this contribution so I can improve.

PR #2621, opened 2026-08-12
feat: support comma-separated gpu-scheduler-policy combinations

mesutoezdil:
Lets gpu-scheduler-policy take an ordered comma-separated list. 
binpack/spread/numa chain as sort keys in the order written, mutex/topology-aware stay filters in Fit(). 
Single values keep working as before. 
Adds PolicyContains to replace the raw == checks that a comma would've silently broken.
I have tested all in my H100 

## Summary by CodeRabbit

* **New Features**
  * Added support for combining GPU scheduling policies, including binpacking, spreading, NUMA awareness, mutex scheduling, and topology-aware placement.
  * Scheduling policies now accept comma-separated values and extra whitespace.
  * Added NUMA-aware GPU scheduling support.

* **Bug Fixes**
  * Improved exclusive and topology-aware allocation when multiple policies are configured.
  * Added deterministic device selection when candidates have equal priority.
  * Improved fallback behavior for policy combinations containing filtering-only options.


archlitchi (approved):
/lgtm

PR #2584, opened 2026-08-11
docs: design for sidecar container GPU resource accounting

maishivamhoo123:
**What type of PR is this?**
this is a design docs for HAMi supporting the sidecar Container.

**What this PR does / why we need it**:
This PR adds  a file docs/develop/sidecarContainer-design.md, which contains design for supporting sidecar containers (init containers with `restartPolicy: Always`) in vGPU accounting.

Right now HAMi treats a sidecar like a normal init container that runs andexits. But a sidecar keeps running next to the app containers for the whole pod lifetime. Because of this, its usage is counted with max instead of being added to the app containers, so the pod's real GPU usage is higher than what HAMi records. Also, the shrink after init never happens for pods
that have a sidecar, so their init container memory is never released.
**Which issue(s) this PR fixes**:
Part of : #2563 

AI disclosure: I used Claude to help find relevant upstream Kubernetes documentation, to improve the formatting of this design document, and to fix spelling and grammar mistakes. The design itself and the analysis of HAMi's current behavior are my own work

## Documentation

* Added design guidance for supporting native sidecar containers.
* Documented how sidecars are identified and handled during admission, scheduling, scoring, and resource accounting.
* Clarified resource calculation, shrink behavior, usage tracking, and annotation requirements.
* Added examples covering sidecar-aware resource management and Kubernetes ResourceQuota interactions.

PR #2623, opened 2026-08-12
fix(scheduler): preserve per-container slot count when collapsing pod device usage

### LINKED ISSUE #2622: Scheduler undercounts GPU slots for pods running multiple containers on one device
**What happened**:
When a pod runs more than one app container on the same GPU, the scheduler undercounts how many time-slicing slots the pod occupies on that device. The node usage it rebuilds records the shared GPU as a single occupied slot no matter how many containers actually run on it.

Each concurrent app container on a shared GPU should occupy its own slot, so a GPU never exceeds deviceSplitCount concurrent tasks.

**Root cause**:
Pod device usage is collapsed to one entry per GPU UUID before it is cached. The collapsed entry keeps memory and core totals but not the number of containers, so getNodesUsage increments DeviceUsage.Used by one per entry. A pod with N app containers on one GPU is counted as one slot instead of N, which lets the scheduler overcommit the device.

**Environment**:
HAMi master. The regression was introduced when onAddPod began caching collapsed device usage.

Eshiv-Pandey:
**What type of PR is this?**

/kind bug

When a pod places more than one app container on the same GPU, the scheduler collapses the pod's per container device usage into a single entry per GPU UUID before caching it. The cached entry recorded memory and core totals but dropped the number of containers, so when node usage is rebuilt in getNodesUsage the shared GPU counted as a single occupied slot regardless of how many containers actually ran on it. This undercounts DeviceUsage.Used and lets the scheduler pack more tasks onto a GPU than deviceSplitCount permits.

This PR adds a Slots field to ContainerDevice that records how many concurrent tasks a collapsed entry represents. The collapse paths compute it (init containers run sequentially, so they contribute one slot at peak; app containers run concurrently, so each occurrence adds a slot), and node usage reconstruction adds Slots to Used instead of always incrementing by one. Slots is not part of the annotation wire format, so on disk encoding and MIG allocation handling are unchanged. Entries with Slots zero (raw, uncollapsed devices) still count as one.

Fixes #2622

The fix is scoped to the reconstruction count. Memory, core, PodInfos, and MIG allocation behavior are all left identical. ContainerDevice.DeepCopy was updated to carry the new field, since node usage is read through a deep copy of the cached pod info.

Local unit tests for pkg/device pass. The pkg/scheduler tests could not be run on the Windows dev machine because go-nvml needs cgo and the local gcc cannot target 64 bit, so those are left to CI.

```release-note
Fixed a GPU slot undercount that let the scheduler place more shared tasks on a device than deviceSplitCount allows when a pod ran multiple containers on the same GPU.
```

**AI assistance disclosure**:

This change was developed with AI assistance (Claude Code). All code and tests were reviewed by the author.

## Summary by CodeRabbit

* **New Features**
  * Device allocations now show how many concurrent task slots each device represents.
  * Scheduling accounts for multiple slots when calculating device usage and availability.
  * Shared GPU usage is reported more accurately across containers.

* **Bug Fixes**
  * Improved device usage aggregation for collapsed and app-only container configurations.
  * Preserved slot information when device data is copied or serialized.
  * Improved handling of stale or unavailable device information.
  * Device locking now rolls back safely when multi-device allocation fails.


pkg/device/initContainer.go
```diff
@@ -52,11 +53,15 @@ func CollapseInitContainerUsage(pod *corev1.Pod, raw PodDevices) PodDevices {
 					if dev.Usedcores > cur.cores {
 						cur.cores = dev.Usedcores
 					}
+					// Init containers run sequentially: peak concurrency is one slot per device.
+					cur.slots = 1
```
mesutoezdil, line 57: pr #2584 wants sidecars split into app-sum instead of init-peak here. if that lands later, slots needs the same split. worth a todo comment?

pkg/scheduler/scheduler.go
```diff
@@ -760,7 +760,9 @@ func (s *Scheduler) getNodesUsage(nodes *[]string, task *corev1.Pod) (*map[strin
 					for _, d := range node.Devices.DeviceLists {
 						deviceID := udevice.UUID
 						if d.Device.ID == deviceID {
-							d.Device.Used++
+							// Raw entries carry no slot count; clamp to at least one.
+							slots := max(udevice.Slots, 1)
```
mesutoezdil, line 782: no scheduler_test.go change here. existing Test_getNodesUsage_WithInitContainers stays at used=1 either way, 1 init + 1 app. add a test with 2+ app containers on one gpu asserting used=2 at this level, not just in pkg/device.

pkg/device/initContainer.go
Eshiv-Pandey: roger that

pkg/scheduler/scheduler.go
Eshiv-Pandey: thanks for the review. 
added the TODO for the sidecar accounting changes, and added scheduler-level coverage for shared app container usage as you mentioned.

mesutoezdil:
conflicts?

archlitchi:

archlitchi (approved):
/lgtm

PR #2624, opened 2026-08-12
fix(cambricon): use shared nodelock package for node-level locking

adity1raut:
## Summary

Cambricon's device backend implemented its own node-lock annotation
(`cambricon.com/dsmlu.lock`) instead of using the shared `pkg/util/nodelock`
package that every other backend (nvidia, hygon, amd, biren, vastai,
iluvatar, kunlun, metax) relies on for serializing device allocation on a
node.

The bespoke implementation had no real concurrency protection:

- `LockNode`/`setNodeLock` checked the lock annotation on the `*corev1.Node`
  object handed to it by the caller (from an informer cache), never
  re-fetching a live copy.
- `setNodeLock` then issued a `Patch` containing only the annotation field,
  with no `resourceVersion` precondition anywhere in the request.

Since Kubernetes `Patch` does not enforce optimistic concurrency unless a
precondition is supplied, two pods being bound to the same node concurrently
could both read the same "unlocked" node snapshot, both pass the check, and
both successfully `Patch` — nothing rejected the second writer. Both pods
would then proceed to mutate MLU device usage/annotations on the same node
with no mutual exclusion, which is exactly the over-allocation / corrupted
device-state scenario the lock exists to prevent.

## Fix

`LockNode`/`ReleaseNodeLock` now delegate to `nodelock.LockNode` /
`nodelock.ReleaseNodeLock`, the same shared package used by every other
backend. That package re-fetches the node on each attempt and patches with
a `resourceVersion` precondition plus retry-on-conflict, so concurrent lock
attempts on the same node are properly serialized. The bespoke
`setNodeLock` method and the `cambricon.com/dsmlu.lock` annotation constant
are removed as dead code.

One intentional behavior change: the stale-lock recovery window moves from
Cambricon's hardcoded 2 minutes to the shared default of 5 minutes
(overridable via `HAMI_NODELOCK_EXPIRE`), matching every other backend
instead of being a Cambricon-specific special case.

## Test plan

- [x] `go build ./...`
- [x] `go test ./pkg/device/cambricon/... -race -short -count=1 -v` — all
      pass, including new/updated lock tests exercising contention, expiry,
      and successful release
- [x] `make test` (full suite, race detector) — all packages pass
- [x] `make lint` (golangci-lint) on changed files — 0 issues
- [x] `hack/verify-license.sh`, `hack/verify-import-aliases.sh` — pass

## AI Assistance Disclosure

This PR was written primarily by  Geminy, following the project's
CONTRIBUTING.md disclosure requirement. The bug was found via an
independent code audit of the scheduler's node-locking paths across all
device backends; the fix mirrors the pattern already used by the
nvidia/hygon/amd backends in this codebase.

## Bug Fixes

- Improved Cambricon device node locking for more consistent acquisition and release behavior.
- Locking continues to apply only to workloads requesting MLU resources.
- Improved handling of active, expired, and malformed locks.
- Locks owned by the requesting workload are now released correctly.
- Improved validation of lock persistence, ownership, and release scenarios.

### DIFF
--- a/pkg/device/cambricon/device.go
+++ b/pkg/device/cambricon/device.go
@@ -92,31 +92,6 @@ func (dev *CambriconDevices) CommonWord() string {
 	return CambriconMLUCommonWord
 }
 
-func (dev *CambriconDevices) setNodeLock(node *corev1.Node) error {
-	ctx := context.Background()
-	patchedAnnotation, err := json.Marshal(
-		map[string]any{
-			"metadata": map[string]map[string]string{"annotations": {
-				DsmluLockTime: time.Now().Format(time.RFC3339),
-			}}})
-	if err != nil {
-		klog.ErrorS(err, "Failed to marshal node annotation", "node", node.Name)
-		return fmt.Errorf("marshal node annotation: %w", err)
-	}
-
-	_, err = client.GetClient().CoreV1().Nodes().Patch(ctx, node.Name, types.MergePatchType, patchedAnnotation, metav1.PatchOptions{})
-	for i := 0; i < retry && err != nil; i++ {
-		klog.ErrorS(err, "Failed to patch node annotation", "node", node.Name, "retry", i)
-		time.Sleep(time.Duration(rand.Intn(i+1)) * 10 * time.Millisecond)
-		_, err = client.GetClient().CoreV1().Nodes().Patch(ctx, node.Name, types.MergePatchType, patchedAnnotation, metav1.PatchOptions{})
-	}
-	if err != nil {
-		return fmt.Errorf("setNodeLock exceeds retry count %d", retry)
-	}
-	klog.InfoS("Node lock set", "node", node.Name)
-	return nil
-}
-
 func (dev *CambriconDevices) LockNode(n *corev1.Node, p *corev1.Pod) error {
 	found := false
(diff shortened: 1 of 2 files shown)


mesutoezdil:
remove claude from the commit history

FouoF:
Dose cambricon device plugin recognize the new annotation key?

adity1raut:
> Dose cambricon device plugin recognize the new annotation key?

Yes, it works fine. Theres no separate Cambricon device plugin in this repo the only device plugin binary here is for nvidia. Cambricon runs its own device plugin outside this codebase entirely, and that plugin never read the old annotation in the first place, so nothing there depends on it.

The old cambricon.com/dsmlu.lock annotation was only ever used internally by the scheduler itself it was written and read by the LockNode and ReleaseNodeLock functions alone, at bind time. No allocation-side code, and no device-plugin code, ever touched it. I checked the whole repo and that old annotation name doesn't appear anywhere anymore, and even before this change it only ever existed in that one file.

The new key is the shared hami.io/mutex.lock annotation, the same one nvidia, amd, hygon, vastai, iluvatar, kunlun, ascend, metax, and biren were already using. Cambricon has simply joined that same shared lock. No new consumer was added, and nothing that existed before is broken.

PR #2626, opened 2026-08-12
fix(scheduler): rollback partial device locks

### LINKED ISSUE #2625: fix(scheduler): rollback device locks after partial acquisition failure
### Description

`acquireNodeLocks()` acquires node locks through every registered HAMi device backend before proceeding with pod allocation.

The underlying `lockAllDevices()` implementation currently acquires these locks sequentially and immediately returns when one backend fails:

    device A → LockNode() succeeds
    device B → LockNode() succeeds
    device C → LockNode() fails
                         ↓
                   return error

The locks acquired for devices A and B are not released on this failure path.

This is particularly relevant for heterogeneous HAMi nodes, where multiple device backends may participate in the same scheduling operation.

There is already a `releaseAllDevices()` helper, and the pod-group retry path correctly uses it after a failed `lockAllDevices()` call. However, the normal `acquireNodeLocks()` path directly returns the error from `lockAllDevices()` and therefore skips the rollback.

This leaves the locking operation non-transactional: a scheduling attempt that fails halfway through can leave locks belonging to the same attempt behind.

### Expected behavior

Device lock acquisition should be atomic from the scheduler's perspective:

- If every backend successfully acquires its lock, continue normally.
- If any backend fails, release every lock acquired earlier in that same attempt before returning the error.
- Do not attempt to release locks for backends that were never successfully acquired.

For example:

    NVIDIA  → acquired
    Ascend  → acquired
    MLU     → failed
                ↓
            rollback
                ↓
    Ascend  → released
    NVIDIA  → released
                ↓
            return error

### Why this matters

HAMi supports heterogeneous accelerator backends and uses node locking to serialize concurrent scheduling/allocation operations.

Leaving partially acquired locks behind can cause subsequent scheduling attempts to encounter stale lock state or unnecessary lock contention, even though the original scheduling attempt never completed.

The existing pod-group retry path already performs rollback, which suggests the same transactional behavior is required for the normal scheduling path.

### Proposed fix

Make `lockAllDevices()` transactional, or otherwise ensure `acquireNodeLocks()` rolls back successfully acquired locks whenever a later backend fails.

The implementation should preserve the existing retry behavior and avoid releasing a backend lock that was not successfully acquired.

### Regression test

Add a test using multiple fake device backends where:

1. Backend A successfully acquires its lock.
2. Backend B successfully acquires its lock.
3. Backend C returns an error from `LockNode()`.
4. `acquireNodeLocks()` returns the original error.
5. Backend A and B each receive exactly one `ReleaseNodeLock()` call.
6. Backend C does not receive a release call.
7. A subsequent lock attempt can proceed without the locks from the failed attempt remaining active.

AyushSrivastava1818:
## What type of PR is this?

/kind bug

When the scheduler acquires node locks across multiple registered device backends, `lockAllDevices()` acquires the locks sequentially.

Previously, if an earlier backend successfully acquired its lock but a later backend failed, the function returned immediately without rolling back the locks that had already been acquired.

This could leave partially acquired node locks behind even though the scheduling attempt itself failed, potentially causing stale lock state or unnecessary lock contention for subsequent scheduling attempts on heterogeneous nodes.

This PR makes device lock acquisition transactional:

- Successfully acquired locks are tracked during the current attempt.
- If a later backend fails to acquire its lock, previously acquired locks are released.
- Rollback happens in reverse acquisition order.
- The original `LockNode()` error is preserved.
- The failed backend and backends that were never attempted are not released.
- Existing successful lock acquisition behavior remains unchanged.
- Existing pod-group retry behavior is preserved without duplicate rollback.

Fixes #2625

The fix is intentionally limited to the scheduler's device-lock acquisition path.

The existing `releaseAllDevices()` mechanism is reused for rollback rather than introducing a separate locking mechanism.

Regression tests cover:

- all device locks succeeding
- failure on the first backend
- failure after earlier backends have successfully acquired locks
- exact rollback of previously acquired locks
- no release for the failed or unattempted backends
- preservation of the original lock acquisition error

## Validation

The following checks were run successfully:

- `go test ./pkg/scheduler/...`
- `go test -race ./pkg/scheduler/...`
- `go vet ./pkg/scheduler/...`
- GolangCI-Lint
- `gofmt`
- `git diff --check`

No.

This is an internal scheduler correctness fix that ensures failed multi-backend lock acquisition does not leave previously acquired locks behind.

## AI Assistance Disclosure

AI assistance was used during investigation of the HAMi codebase, root-cause analysis, implementation, regression-test development, and validation of this change.

The resulting changes were reviewed against the existing HAMi scheduler and device-locking architecture, with the implementation and tests validated by the contributor.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved device-locking reliability with a consistent acquisition and release order.
  * Automatically rolls back previously acquired locks when a later lock cannot be obtained.
  * Prevented redundant lock-release operations after failed acquisition attempts.
  * Preserved the original locking error when rollback encounters an additional failure.
  * Enabled reliable retries after failed locking attempts.

* **Tests**
  * Added coverage for successful locking, failures, rollback scenarios, retries, and operation ordering.

### DIFF
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"maps"
 	"os"
+	"slices"
 	"sort"
 	"strconv"
 	"strings"
@@ -902,16 +903,38 @@ func (s *Scheduler) cleanupStalePodAllocation(pod *corev1.Pod) {
 }
 
 func (s *Scheduler) lockAllDevices(node *corev1.Node, pod *corev1.Pod) error {
-	for _, val := range device.GetDevices() {
+	devs := device.GetDevices()
+	keys := make([]string, 0, len(devs))
+	for k := range devs {
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
+
+	acquired := make([]device.Devices, 0, len(keys))
+	for _, k := range keys {
+		val := devs[k]
 		if err := val.LockNode(node, pod); err != nil {
+			for _, locked := range slices.Backward(acquired) {
+				if relErr := locked.ReleaseNodeLock(node, pod); relErr != nil {
+					klog.ErrorS(relErr, "Failed to release node lock during rollback", "node", node.Name, "pod", klog.KObj(pod))
+				}
+			}
 			return err
 		}
+		acquired = append(acquired, val)
 	}
 	return nil
 }
 
 func (s *Scheduler) releaseAllDevices(node *corev1.Node, pod *corev1.Pod) {
-	for _, val := range device.GetDevices() {
+	devs := device.GetDevices()
+	keys := make([]string, 0, len(devs))
+	for k := range devs {
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
+	for _, k := range keys {
+		val := devs[k]
 		if err := val.ReleaseNodeLock(node, pod); err != nil {
 			klog.ErrorS(err, "Failed to release node lock", "node", node.Name, "pod", klog.KObj(pod))
 		}
@@ -929,7 +952,6 @@ func (s *Scheduler) acquireNodeLocks(node *corev1.Node, pod *corev1.Pod) error {
 		if err == nil {
 			return nil
 		}
-		s.releaseAllDevices(node, pod)
 		if !nodelockutil.IsNodeLockContention(err) {
 			return err
 		}
--- a/pkg/scheduler/scheduler_test.go
+++ b/pkg/scheduler/scheduler_test.go
@@ -2458,35 +2458,52 @@ func Test_lockAllDevices_Transactional(t *testing.T) {
 		assert.Equal(t, dev3.releaseCalls.Load(), int32(0))
 	})
 
-	// RollbackReleaseError: backend1 acquires its lock, backend2's LockNode fails triggering
-	// rollback, but backend1's ReleaseNodeLock itself returns an error.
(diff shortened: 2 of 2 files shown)


pkg/scheduler/scheduler.go
```diff
@@ -929,7 +952,6 @@ func (s *Scheduler) acquireNodeLocks(node *corev1.Node, pod *corev1.Pod) error {
 		if err == nil {
 			return nil
 		}
-		s.releaseAllDevices(node, pod)
 		if !nodelockutil.IsNodeLockContention(err) {
```
mesutoezdil, line 955: also fixes the non podgroup path, not just podgroup retries. no test for release count during a real retry loop. worth one?

mesutoezdil (commented):
left 1 question inline

pkg/scheduler/scheduler.go
AyushSrivastava1818: yes right..added the test for release count during a real retry loop and verified...check when you have time

mesutoezdil:
resolve the conversations

AyushSrivastava1818:
@mesutoezdil resolved the conversations have a look again please!

AyushSrivastava1818:
@mesutoezdil Can you please check sir the Compile is failing due to Docker build..but i have not modified any of the docker/CI files yet!!..I guess its unrelated to it sir!

AyushSrivastava1818:
@mesutoezdil @archlitchi  pinging it just to remind about this pr!

archlitchi (approved):
/lgtm

PR #2634, opened 2026-08-13
fix(monitor): unmap orphaned containers from vGPU monitor on external deletion

### LINKED ISSUE #2633: bug(monitor): fix mmap memory leak in vGPU monitor when containers are deleted externally
**Description**
In `pkg/monitor/nvidia/cudevshr.go`, the `vGPUmonitor` manages shared memory connections to HAMi-core for active containers. It does this by polling the container directories and mapping the memory (`syscall.Mmap`) into `l.containers`.

Currently, the `Update()` loop's cleanup mechanism relies entirely on discovering the container's directory via `os.ReadDir(l.containerPath)`. 
When a container is gracefully deleted, the loop detects `!podUIDs[podUID]`, unmaps the memory (`syscall.Munmap`), deletes the map entry, and removes the directory. 

However, if a container finishes and its directory is deleted *externally* (e.g., by container runtime garbage collection or an administrator) **before** the `Update()` loop runs, `os.ReadDir` will not return the directory. Because the directory is skipped, the monitor never triggers the cleanup sequence.

This results in a permanent memory leak:
1. `syscall.Munmap(c.data)` is never called for the container.
2. The stale container entry remains in `l.containers` indefinitely.

Over time on a long-running node, this leaked bookkeeping leads to unnecessary memory consumption.

**Solution**
Add a reconciliation step at the end of the `Update()` loop. 
After iterating over the active directories discovered by `os.ReadDir`, iterate over the existing `l.containers` map. If a key in `l.containers` no longer corresponds to a discovered directory, it means the directory was externally removed. The monitor should then safely call `syscall.Munmap` and delete the orphaned map entry.

**Alternatives**
N/A

KunwarSidhu47:
**What type of PR is this?**

/kind bug

Currently, the `vGPUmonitor` manages its active container memory maps (`l.containers`) by polling the local directory (e.g. `/usr/local/vgpu/containers/`). If a container shuts down normally, the monitor successfully catches it during the `Update()` loop, unmaps the memory via `syscall.Munmap`, and cleans up the map entry.

The bug occurs when a container's directory is removed *externally* (such as by runtime garbage collection) before the monitor has a chance to poll it. When `os.ReadDir` runs, it completely misses the deleted directory. Consequently, the `Update()` loop bypasses the container, leaving its `syscall.Mmap` allocation and `l.containers` entry permanently orphaned. 

On nodes running for extended periods, this causes a continuous memory leak as orphaned memory mappings accumulate.

This PR resolves the leak by:
- Creating a `validEntries` tracking map to record all directories physically present during the `Update()` cycle.
- Adding a final reconciliation step that cross-references the active `l.containers` against `validEntries`.
- Safely executing `syscall.Munmap` and evicting any map entries that no longer correspond to a physical directory.
- Introducing a targeted regression test in `cudevshr_test.go` to prevent future regressions of this behavior.

Fixes #2633 

**Special notes for reviewer**:

The reconciliation logic was intentionally placed at the end of the `Update()` loop. This guarantees we only evaluate state against a completely fresh directory read, preventing any race conditions where a newly created directory might be accidentally flagged as orphaned.

Testing verifies that:
- Containers missing from the physical directory correctly trigger `Munmap`.
- The `l.containers` map is correctly purged of orphaned entries.

Validation completed:
- `gofmt` & `git diff --check`
- `GolangCI-Lint`
- Added unit tests specific to the external deletion leak.

*(Note: CGO/NVML package tests were deferred to the Linux CI pipeline due to local Windows GCC/CGO limitations).*

No.

**AI Assistance Disclosure**:

I utilized Antigravity to help investigate this memory leak, analyze the root cause in the directory polling lifecycle, and scaffold the regression tests. The final implementation was thoroughly verified against the monitor's intended architecture.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved container monitoring reconciliation when container directories are deleted externally.
  * Removed stale container entries and unmapped their usage during subsequent scans.
* **Tests**
  * Added coverage to verify orphaned container data is correctly cleaned up.

### DIFF
--- a/pkg/monitor/nvidia/cudevshr.go
+++ b/pkg/monitor/nvidia/cudevshr.go
@@ -181,6 +181,8 @@ func (l *ContainerLister) Update() error {
 		podUIDs[string(pod.UID)] = true
 	}
 
+	validEntries := make(map[string]bool)
+
 	for _, entry := range entries {
 		if !entry.IsDir() {
 			continue
@@ -195,6 +197,7 @@ func (l *ContainerLister) Update() error {
 		if !podUIDs[podUID] {
 			dirInfo, err := os.Stat(dirName)
 			if err == nil && dirInfo.ModTime().Add(resyncInterval).After(time.Now()) {
+				validEntries[entry.Name()] = true
 				continue
 			}
 			klog.Infof("Removing dirname %s in monitorpath", dirName)
@@ -205,6 +208,7 @@ func (l *ContainerLister) Update() error {
 			_ = os.RemoveAll(dirName)
 			continue
 		}
+		validEntries[entry.Name()] = true
 		if _, ok := l.containers[entry.Name()]; ok {
 			continue
 		}
@@ -221,6 +225,15 @@ func (l *ContainerLister) Update() error {
 		l.containers[entry.Name()] = usage
 		klog.Infof("Adding ctr dirname %s in monitorpath", dirName)
 	}
+
+	for name, c := range l.containers {
+		if !validEntries[name] {
+			klog.Infof("Removing orphaned container dirname %s from monitorpath", name)
+			syscall.Munmap(c.data)
+			delete(l.containers, name)
+		}
+	}
+
 	return nil
 }
 
--- a/pkg/monitor/nvidia/cudevshr_test.go
+++ b/pkg/monitor/nvidia/cudevshr_test.go
@@ -543,7 +543,7 @@ func Test_ContainerLister_Update(t *testing.T) {
 	t.Run("externally deleted directory unmaps and removes container from l.containers", func(t *testing.T) {
 		dir := t.TempDir()
 		entryName := "missing-pod-uid_ctr"
-		
+
 		// Create a mock mapped memory usage
 		cacheDir := t.TempDir()
 		writeCacheFile(t, cacheDir, "x.cache", headerBytes(v1CacheFileSize, SharedRegionMagicFlag, 1, 0))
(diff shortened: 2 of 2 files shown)


FouoF:
Have you evaluated the frequency of memory leak and how many memory will leak one time? Will this introduce significant overhead？Will mmaped memory still leak when the process exit?

KunwarSidhu47:
Hi @FouoF ,
Thanks for the review! Here is the evaluation of the leak and overhead:

Frequency & Size: The size of the syscall.Mmap allocation is tied to the .cache file size, which is bounded by the struct size (e.g., v0.MinSize() is ~1.14 MB). So each orphaned container leaks roughly ~1.14 MB. While not huge individually, vGPUmonitor is a long-lived daemon. If a node churns through many short-lived batch jobs and race conditions cause directories to be externally GC'd before the monitor loop catches them, this accumulates over weeks and eventually leads to an OOMKill.

Overhead: The overhead is negligible. The reconciliation logic only iterates over the containers on a single node (typically < 100 containers). Iterating a map of this size takes microseconds and allocates almost no memory, so it doesn't noticeably impact the Update() loop's performance.

Process Exit: No, the OS kernel automatically reclaims all mmap'd memory when the vGPUmonitor process exits. However, because this is a long-running DaemonSet, we want to prevent the leak from accumulating during its normal lifecycle to avoid unnecessary OOM crashes.

mesutoezdil:

PR #2635, opened 2026-08-13
cleanup(metax): decouple MetaxDevices.ScoreNode from scheduler policy

### LINKED ISSUE #2566: [BUG]: Metax-GPU ScoreNode ignores topology when a node has only one of the two annotations
**What happened**:

In the `Metax-GPU` backend, `MetaxDevices.ScoreNode` reads a *different* topology annotation depending on the scheduler policy:

- Under the **Binpack** policy it reads only `metax-tech.com/gpu.topology.losses`.
- Under the **Spread** policy it reads only `metax-tech.com/gpu.topology.scores`.

If a node carries only *one* of these two annotations, the policy that reads the *other* one finds nothing and the score falls back to `0`. The node's topology preference is silently dropped, so topology-aware scheduling stops having any effect for that node.

Source (on `master`, commit `153defc`), `pkg/device/metax/device.go`:

```go
// lines 185-214
	sum := 0
		sum += len(dev)
	}

	res := float32(0)
	if policy == string(util.NodeSchedulerPolicyBinpack) {
		lossAnno, ok := node.Annotations[MetaxAnnotationLoss]   // reads ONLY "losses"
		if ok {
			res = 2000 - loss
		}
	} else if policy == string(util.NodeSchedulerPolicySpread) {
		scoreAnno, ok := node.Annotations[MetaxAnnotationScore] // reads ONLY "scores"
		if ok {
			res = 2000 - score
		}
	}

	return res // stays 0 when the annotation this policy reads is absent
}
```

The two annotation keys are defined at `pkg/device/metax/device.go:40-41`:

```go
MetaxAnnotationLoss  = "metax-tech.com/gpu.topology.losses"
MetaxAnnotationScore = "metax-tech.com/gpu.topology.scores"
```

Because `MetaxDevices` does not implement the `PolicyNeutralScorer` marker, this raw value is added to the node score directly (no weighting), so a returned `0` means "this node has no topology preference at all".

A node's topology preference should be honored regardless of which of the two annotations it publishes. If a node exposes only `metax-tech.com/gpu.topology.scores`, that signal should still influence scoring under Binpack (and vice-versa for `losses` under Spread). A node with better topology should be preferred over a node with worse topology under both policies.

1. Set the scheduler policy to **Binpack** (`hami.io/node-scheduler-policy: binpack`, the default).
2. Have two Metax-GPU nodes that publish only the `scores` annotation (no `losses`), for a pod requesting 2 GPUs (map key `"2"`):
   - Node A: `metax-tech.com/gpu.topology.scores: {"2":1800}`  (better topology)
   - Node B: `metax-tech.com/gpu.topology.scores: {"2":200}`   (worse topology)
3. Schedule a pod requesting 2 `Metax-GPU` devices.

**Observed:** `ScoreNode` takes the Binpack branch, looks up `metax-tech.com/gpu.topology.losses`, finds it absent on both nodes, and returns `0` for both. The topology contribution is identical (`0`), so Node A's better topology is ignored and the pod may land on Node B.

**Expected:** Node A (topology score `1800`) is preferred over Node B (`200`).

The symmetric case reproduces under the **Spread** policy when nodes publish only `metax-tech.com/gpu.topology.losses`: the Spread branch reads `scores`, finds it absent, and returns `0`, again dropping topology preference.

Note: when a node publishes **both** annotations, each policy reads the one it expects and behaves correctly — the bug is specific to nodes that carry only one of the two.

This is a scheduler source-logic bug found by code inspection, so host-level output (`nvidia-smi`, docker/containerd config, device-plugin/kubelet/dmesg logs) is not the relevant evidence. The relevant evidence is the code path above.

Root cause: `ScoreNode` couples the *choice of annotation* to the policy string and assumes the policy-specific annotation is always present. A more robust approach is to read whichever annotation is available (for example, prefer `scores` and fall back to `losses`, converting `losses` onto the same "higher is better" scale) and let the shared scheduler policy layer handle policy direction. This is the same direction as the Metax-SGPU decoupling done in #2404.

**Environment**:
- HAMi version: `master` @ commit `153defc` (also present in current release lines carrying this file)
- Component: scheduler extender (`pkg/device/metax/device.go`, `Metax-GPU` backend)
- nvidia driver or other AI device driver version: N/A (Metax GPU; logic bug independent of driver version)
- Docker version from `docker version`: N/A (logic bug, reproducible by code inspection)
- Others: Affects clusters using the Metax-GPU backend with topology annotations, where nodes publish only one of `metax-tech.com/gpu.topology.losses` / `metax-tech.com/gpu.topology.scores`.

### LINKED ISSUE #2572: Decouple MetaxDevices.ScoreNode (Metax-GPU) from the scheduler policy string

Rickydama3:
**What type of PR is this?**
/kind bug

I noticed that the `Metax-GPU` backend was still handling the `binpack` and `spread` policy checks directly inside `ScoreNode`. This was causing the issue reported in #2566, where nodes could be completely ignored if they only had one of the two topology annotations.

Since we already decoupled the `Metax-SGPU` backend in #2413, I followed the same approach here for `Metax-GPU`.

I removed the policy checks from the backend. It now uses the `scores` annotation when available and falls back to `losses` when needed, converting it to the standard higher-is-better scale. I also added the `PolicyNeutralScorer` marker to the device so the shared scheduler's `OverrideScore` layer can handle the `10000` weighting and spread sign flip, the same way it does for SGPU.

I also added tests in `device_test.go` to cover the policy-independent scoring and the fallback logic.

Fixes #2566
Fixes #2572

I also added some tests to make sure it doesn't crash if the node annotations are missing or broken.

**Does this PR introduce a user facing change?**:
```release-note
NONE

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved device scoring to consistently prioritize valid topology scores.
  * Added fallback handling for topology loss values, including missing, malformed, and large-loss annotations.
  * Invalid or unavailable topology annotations now produce a neutral score instead of unreliable results.
  * Ensured device scores remain independent of scheduler policy.
  * Updated score calculations to preserve accurate weighting for high-value device scores.

### DIFF
--- a/pkg/device/metax/device.go
+++ b/pkg/device/metax/device.go
@@ -165,22 +165,27 @@ func (dev *MetaxDevices) customFilterRule(allocated *device.PodDevices, request 
 	return true
 }
 
-func parseMetaxAnnos(annos string, index int) float32 {
+func parseMetaxAnnos(annos string, index int) (float32, bool) {
 	scoreMap := map[int]int{}
 	err := json.Unmarshal([]byte(annos), &scoreMap)
 	if err != nil {
 		klog.Warningf("annos[%s] Unmarshal failed, %v", annos, err)
-		return 0
+		return 0, false
 	}
 
 	res, ok := scoreMap[index]
 	if !ok {
 		klog.Warningf("scoreMap[%v] not contains [%d]", scoreMap, index)
-		return 0
+		return 0, false
 	}
 
-	return float32(res)
+	return float32(res), true
 }
+
+// MetaxMaxTopologyLoss is the upper bound of the values published in the
+// "gpu.topology.losses" annotation. It converts a loss (lower is better) onto
+// the same "higher is better" scale used by "gpu.topology.scores".
+const MetaxMaxTopologyLoss = 2000
 
 func (dev *MetaxDevices) ScoreNode(node *corev1.Node, podDevices device.PodSingleDevice, previous []*device.DeviceUsage, policy string) float32 {
 	sum := 0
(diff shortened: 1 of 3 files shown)


mesutoezdil:

PR #2636, opened 2026-08-13
fix(vGPUmonitor): report recorded memory offset instead of derived value

### LINKED ISSUE #2551: [BUG]:Device_memory_desc_of_container offset can underflow

imantaba:
### What this PR does / why we need it
The `Device_memory_desc_of_container` legacy metric computes its `offset` label as `total - context - module - buffer`. `DeviceMemoryContextSize`, `DeviceMemoryModuleSize`, `DeviceMemoryBufferSize` and `DeviceMemoryTotal` each sum their field independently across the active processes in the mmap'd shared region and are not read as one atomic snapshot, so `context + module + buffer` can momentarily exceed `total`. The unsigned subtraction then underflows and the metric exports an `offset` near `MaxUint64`.

The shared region already records the offset as a first-class field, exposed through `UsageInfo.DeviceMemoryOffset(idx)` (implemented in both v0 and v1 Spec). This reports that recorded value directly — semantically correct and free of underflow.

### Which issue(s) this PR fixes
Fixes #2551

### Test
Adds `TestCollectContainerMetricsOffsetNoUnderflow`, which drives `collectContainerMetrics` with `context+module+buffer > total` and asserts the emitted `offset` label equals the recorded offset. The test fails on the old subtraction (label `18446744073709551566`) and passes with the fix.

### AI assistance disclosure

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed container device-memory metrics that could report incorrect values when sampled memory components temporarily exceeded total memory.
  * Metrics now accurately reflect the recorded device-memory offset, preventing underflow-related inaccuracies.

* **Tests**
  * Added regression coverage for accurate offset reporting under inconsistent memory samples.

### DIFF
--- a/cmd/vGPUmonitor/feedback_test.go
+++ b/cmd/vGPUmonitor/feedback_test.go
@@ -38,6 +38,7 @@ type stubInfo struct {
 	ctxSize    []uint64
 	modSize    []uint64
 	bufSize    []uint64
+	offset     []uint64
 	smUtil     []uint64
 	lastKernel int64
 }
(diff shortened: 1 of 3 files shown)


FouoF:
offset is legacy field and not exposed by hami-core now, this cloud be deprecated actually.

PR #2643, opened 2026-08-13
fix(scheduler): avoid false positives in AllInitContainersSucceeded

### LINKED ISSUE #2642: scheduler: AllInitContainersSucceeded() can return true before all init containers report status
**What happened**: 
HAMi can incorrectly decide that all init containers of a pod have completed successfully even when one or more init containers never reported a terminal status.

This happens in `pkg/util/util.go`, where AllInitContainersSucceeded() checks whether len(pod.Status.InitContainerStatuses) == 0 and then iterates over the reported status list. It does not verify that every init container in pod.Spec.InitContainers has a corresponding status entry.

The scheduler later uses this result in `pkg/scheduler/scheduler.go` when deciding whether to shrink a pod’s device usage after init containers finish. A false positive here can cause HAMi to reduce device accounting too early, even though some init containers have not finished.

The scheduler should only treat init containers as complete when all init containers in pod.Spec.InitContainers have a successful terminal status.

In other words, if 2 init containers exist but only 1 status entry is present, the function should return false instead of true.

1. Create a pod with 2 init containers.
2. Only one of them reports a successful status.
3. Call the logic in AllInitContainersSucceeded().

Example reproducer output from local validation:
```
spec init containers: 2
reported init container statuses: 1
AllInitContainersSucceeded: true
Expected: false, because one init container has no status at all
```

This shows that the function returns true even though not all init containers completed.

This bug is specifically caused by the mismatch between:
1. `len(pod.Spec.InitContainers)`
2. `len(pod.Status.InitContainerStatuses)`

The logic should enforce equality or explicitly check that every init container is accounted for.

This is a scheduler correctness issue, not a GPU driver issue.

Relevant code references:
- `pkg/util/util.go`
- `pkg/scheduler/scheduler.go`

**Environment**:
- HAMi version: local source tree
- nvidia driver or other AI device driver version: not required to reproduce this scheduler execution
- Docker version from `docker version`: not required for this reproduction
- Docker command, image and tag used: not required
- Kernel version from `uname -a`: not required
- Others: Reproduced by direct logic check against a Pod object with mismatched init container specs

Kevindua26:
**What type of PR is this?**
/kind bug

This PR fixes a scheduler correctness bug in `AllInitContainersSucceeded`.

Previously, the function could return `true` even when not all init containers had reported status entries (for example, 2 init containers in spec, but only 1 in status).
That can cause HAMi to treat init containers as completed too early and may trigger early resource-usage shrink decisions in scheduler flow.

This PR makes the check strict:
- pod must have init containers
- `len(Status.InitContainerStatuses)` must equal `len(Spec.InitContainers)`
- each init container status must be terminated with `exit code 0`

It also adds focused regression coverage in `Test_AllInitContainersSucceeded` for:
- missing init-container statuses (bug case)
- all init containers succeeded (valid case)

Fixes #2642 

Scope is intentionally minimal and focused.
Changed files only:
- `pkg/util/util.go`
- `pkg/util/util_test.go`

Validation run:
- go test ./pkg/util -run Test_AllInitContainersSucceeded -count=1
- go test ./pkg/util -count=1

No,
This is an internal scheduler correctness fix and test update only.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved init-container completion checks to correctly handle missing pods, absent init containers, incomplete status reporting, and unsuccessful termination states.
  * Ensured workloads are recognized as complete only when every declared init container finishes successfully.

* **Tests**
  * Added coverage for incomplete status reporting and successful completion of all init containers.

### DIFF
--- a/pkg/util/util.go
+++ b/pkg/util/util.go
@@ -160,7 +160,14 @@ func PatchNodeAnnotations(node *corev1.Node, annotations map[string]string) erro
 	return err
 }
 func AllInitContainersSucceeded(pod *corev1.Pod) bool {
-	if len(pod.Status.InitContainerStatuses) == 0 {
+	if pod == nil {
+		return false
+	}
+	initContainers := len(pod.Spec.InitContainers)
+	if initContainers == 0 {
+		return false
+	}
+	if len(pod.Status.InitContainerStatuses) != initContainers {
 		return false
 	}
 	for _, s := range pod.Status.InitContainerStatuses {
--- a/pkg/util/util_test.go
+++ b/pkg/util/util_test.go
@@ -538,6 +538,72 @@ func Test_AllContainersCreated(t *testing.T) {
 	}
 }
 
+func Test_AllInitContainersSucceeded(t *testing.T) {
+	tests := []struct {
+		name string
+		args *corev1.Pod
+		want bool
+	}{
+		{
+			name: "missing init container statuses",
+			args: &corev1.Pod{
+				Spec: corev1.PodSpec{
+					InitContainers: []corev1.Container{
+						{Name: "init-a"},
+						{Name: "init-b"},
+					},
+				},
+				Status: corev1.PodStatus{
+					InitContainerStatuses: []corev1.ContainerStatus{
+						{
+							Name: "init-a",
+							State: corev1.ContainerState{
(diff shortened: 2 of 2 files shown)


mesutoezdil:

PR #2648, opened 2026-08-13
fix(scheduler): check if pod is actually bound before releasing locks on api error

### LINKED ISSUE #2645: fix(scheduler): prevent silent state corruption on Pod Bind API timeouts
**What happened:**
In `pkg/scheduler/scheduler.go`, the `Bind()` function performs operations in the following order:
1. `s.acquireNodeLocks(node, current)`
2. `util.PatchPodAnnotations(current, tmppatch)`
3. `s.kubeClient.CoreV1().Pods(args.PodNamespace).Bind(...)`

If the `Bind` API call in Step 3 returns an error (e.g., due to a temporary network blip or an API server 504 Gateway Timeout), the scheduler immediately calls `fail(err)`, which executes `s.releaseAllDevices(node, current)` to roll back the node locks. 

However, in a distributed system, an HTTP timeout or error does *not* guarantee the operation failed on the server. If the API Server successfully bound the Pod to the Node right before the connection dropped, the Pod is actively scheduled. By prematurely releasing the device locks upon receiving the error, the HAMi scheduler leaves the cluster in a corrupted state: the Pod is running on the node, but the scheduler allows the very next Pod to acquire the exact same GPU locks, causing severe device contention and oversubscription.

Additionally, if the bind truly did fail, the `allocating` annotations applied in Step 2 are never rolled back, leaking state onto the Pod.

Lock release operations must be fully transactional. If the `Bind()` API call returns an error, the scheduler must not blindly release node locks without first verifying the actual state of the Pod. If the Pod was actually bound despite the error, the locks must be retained. 

**How to reproduce it:**
1. Schedule a Pod that requests HAMi vGPU resources.
2. Simulate a network drop or use a webhook to artificially inject a 500/504 error response specifically for the `/bind` API call.
3. Observe that the Pod is successfully bound to the node by the API Server, but the HAMi scheduler releases the device locks, allowing subsequent Pods to overwrite the allocation.

KunwarSidhu47:
**What type of PR is this?**
/kind bug

This PR fixes a silent state-corruption bug in the scheduler's `Bind` phase. 
Previously, if the `kubeClient.CoreV1().Pods(args.PodNamespace).Bind(...)` API call returned a network error (e.g., a timeout), the scheduler assumed a complete failure and immediately called `releaseAllDevices`. However, if the API server had successfully committed the bind right before the connection dropped, the Pod would start running on the kubelet, but the scheduler had already prematurely released its device locks. This race condition leads to dangerous hardware oversubscription.

The fix adds a safety check: before releasing node locks on a `Bind` error, we do a quick `Get` on the Pod. If the Pod's `NodeName` confirms it was actually bound, we retain the device locks and recover gracefully. I've also added a comprehensive unit test (`TestBind_TimeoutStateRecovery`) using a mock `k8stesting.Reactor` to explicitly simulate this exact timeout edge case.

Fixes #2645 

**Special notes for reviewer**:
* **AI Disclosure:** I consulted AI to understand the codebase and help debug this race condition, but the final logic and tests were authored and verified manually by myself.
* **Testing:** As per the Contribution Gates, because this fix is scoped entirely to the scheduler extender's state machine and not device allocation/isolation, it has been verified purely through unit testing (`make verify` and `go test -v ./pkg/scheduler/...`) and does not require physical GPU hardware validation logs. 

NONE

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved pod binding reliability when the API reports an error or times out.
  * Binding now succeeds when the pod is confirmed on the intended node, while preserving the node lock.
  * Failed bindings continue to follow the existing cleanup behavior.
  * Reduced the risk of incorrectly treating successfully assigned pods as failed during temporary API communication issues.

### DIFF
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -1006,7 +1006,7 @@ func (s *Scheduler) Bind(args extenderv1.ExtenderBindingArgs) (*extenderv1.Exten
 
 	if err = s.kubeClient.CoreV1().Pods(args.PodNamespace).Bind(context.Background(), binding, metav1.CreateOptions{}); err != nil {
 		klog.ErrorS(err, "Failed to bind pod", "pod", args.PodName, "namespace", args.PodNamespace, "node", args.Node)
-		
+
 		// Verify if the pod was actually bound despite the API error (e.g., due to a network timeout)
 		ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
 		defer cancel()
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -1008,14 +1008,23 @@ func (s *Scheduler) Bind(args extenderv1.ExtenderBindingArgs) (*extenderv1.Exten
 		klog.ErrorS(err, "Failed to bind pod", "pod", args.PodName, "namespace", args.PodNamespace, "node", args.Node)
 
 		// Verify if the pod was actually bound despite the API error (e.g., due to a network timeout)
-		ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
-		defer cancel()
-		if actualPod, getErr := s.kubeClient.CoreV1().Pods(args.PodNamespace).Get(ctx, args.PodName, metav1.GetOptions{}); getErr == nil {
-			if actualPod.Spec.NodeName == args.Node {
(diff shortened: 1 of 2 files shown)


mesutoezdil:
Closing
fail() only releases the node-lock annotation, 
device accounting is untouched, 
so the oversubscription premise doesn't hold, 
and the fix adds an up-to-5s blocking poll to every genuine bind failure

PR #2650, opened 2026-08-13
feat(scheduler): add Prometheus latency histograms for Bind and Filter workflows

devGPP23:
**Problem**

While looking into the scheduler's performance, I noticed that we are completely blind to how much time is spent in the critical paths like `/bind` and `/filter`. Our existing Prometheus metrics do a good job tracking GPU allocation state, but they don't tell us anything about request latency or error rates.
For example, the Bind workflow runs several steps back-to-back: it looks up the pod, fetches the node, acquires a node-level lock to prevent GPU over-allocation, patches annotations, and finally calls the apiserver. If a scheduling cycle is slow, there is currently no way to know if the delay is caused by lock contention or a slow apiserver response. We also lack basic counters to track how often these workflows succeed or fail.

**Solution**

Following the maintainers' guidance to stick with histograms instead of bringing in heavy tracing frameworks (like OTel), I've added standard Prometheus latency histograms for our critical scheduling paths.
Here's what this PR adds:
1)`hami_scheduler_bind_duration_seconds`: Histogram tracking the exact time spent in the `/bind` handler. It includes a `phase` label to break down the total time into sub-steps (`total`, `pod_lookup`, `node_lookup`, `node_lock`, `patch_annotations`, `apiserver_bind`).
2)`hami_scheduler_filter_duration_seconds`: Histogram tracking the overall execution time for `/filter`.
3)`hami_scheduler_score_duration_seconds`: Histogram tracking the `calcScore` phase (which does the heavy lifting for node scoring) during filtering.
4)Counters (`_total`) for all of the above, labeled by `result` (`success` or `error`).
This uses our existing `prometheus/client_golang` dependency, so there's zero bloat. I also made sure not to change any existing method signatures (no context hacking).

**AI Assistance Disclosure:**

## Summary by CodeRabbit

* **New Features**
  * Added Prometheus metrics for scheduler bind, filter, and score operations.
  * Metrics include operation totals, durations, phases, and success or error results.
  * Added visibility into scheduler performance and failure rates.

* **Bug Fixes**
  * Improved monitoring of filtering, binding, and scoring failures through consistent result classification.

### DIFF
--- a/cmd/scheduler/metrics.go
+++ b/cmd/scheduler/metrics.go
@@ -404,6 +404,12 @@ func initMetrics(bindAddress string, metricsProvider schedulerMetricsProvider, l
 	klog.Info("Initializing metrics for scheduler")
 	reg := prometheus.NewRegistry()
 	reg.MustRegister(versionmetrics.NewBuildInfoCollector())
+	reg.MustRegister(schedulerpkg.BindDuration)
+	reg.MustRegister(schedulerpkg.BindTotal)
+	reg.MustRegister(schedulerpkg.FilterDuration)
+	reg.MustRegister(schedulerpkg.FilterTotal)
+	reg.MustRegister(schedulerpkg.ScoreDuration)
+	reg.MustRegister(schedulerpkg.ScoreTotal)
 
 	NewClusterManager("vGPU", reg, metricsProvider, legacyMetrics)
 
--- a/pkg/scheduler/metrics.go
+++ b/pkg/scheduler/metrics.go
@@ -1,0 +1,93 @@
+/*
+Copyright 2024 The HAMi Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package scheduler
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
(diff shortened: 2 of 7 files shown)


Shouren:
@devGPP23 We have a `profiling` flag for monitoring performance.

devGPP23:
Thanks for the review @Shouren
I see that pprof is there for profiling. These histograms were meant for a different use case ,always-on Grafana dashboards and alerting (like "alert if p95 bind latency > 5s"). 
pprof is great for one-off debugging but doesn't feed into dashboards. 
Should I rework this to complement pprof, or is there a different approach you'd prefer?

Shouren:
> Thanks for the review @Shouren I see that pprof is there for profiling. These histograms were meant for a different use case ,always-on Grafana dashboards and alerting (like "alert if p95 bind latency > 5s"). pprof is great for one-off debugging but doesn't feed into dashboards. Should I rework this to complement pprof, or is there a different approach you'd prefer?

@devGPP23 I think the scenario you mentioned should be an application scenario for continuous profiling.

devGPP23:
Thanks @Shouren
Just to make sure I understand correctly are you suggesting this use case (always-on latency tracking for dashboards and alerting) would be better handled  by a continuous profiling tools rather than Prometheus histograms? Or would you prefer I integrate latency visibility into the existing profiling flag instead?

Either way, happy to follow whatever direction works best for the project. I'll keep this closed for now and focus on #2716 or other fixes

Shouren:
> Thanks @Shouren Just to make sure I understand correctly are you suggesting this use case (always-on latency tracking for dashboards and alerting) would be better handled by a continuous profiling tools rather than Prometheus histograms? Or would you prefer I integrate latency visibility into the existing profiling flag instead?

@devGPP23 I think it would be better served by continuous profiling tools than Prometheus histograms.

devGPP23:
Thanks for clarifying the direction @Shouren . I'll leave this closed and let continuous profiling handle this use case. I'll focus on the other open PRs/fixes. Thanks again for the review

PR #2652, opened 2026-08-13
fix(device-plugin): prevent mutex pass-by-value in device replication

### LINKED ISSUE #2651: [BUG] Mutex lock state copied by value in device-plugin RM package
### Bug Description
There is a silent concurrency bug in the NVIDIA device plugin component (`pkg/device-plugin/nvidiadevice/nvinternal/rm`). 

Currently, the `Device` struct (which embeds `kubeletdevicepluginv1beta1.Device` containing a `sync.Mutex` in its protobuf state) is being passed by value. When a Go struct containing a mutex is passed by value (or shallow copied), the lock state is silently duplicated. This completely decouples the locks and creates race conditions since different goroutines will lock entirely different copies of the mutex.

This occurs in two places:
1. The method receivers on `Device` (`AlignedAllocationSupported`, `IsMigDevice`, `GetUUID`) are using pass-by-value `(d Device)` instead of `(d *Device)`.
2. In `device_map.go`, `replicatedDevice` is instantiated via a shallow copy `replicatedDevice := *(oDevices[r.Name][id])`, which copies the Mutex.

### Steps to Reproduce
Run standard Go static analysis on the `device-plugin` package:
`go vet ./pkg/device-plugin/...`

Output will show:
`assignment copies lock value to replicatedDevice`
`AlignedAllocationSupported passes lock by value`
`IsMigDevice passes lock by value`

### Expected Behavior
`go vet` should pass with no warnings, and Mutexes must never be copied by value.

### Proposed Solution
Change `(d Device)` to `(d *Device)` in `devices.go`, and explicitly initialize `replicatedDevice` in `device_map.go` without shallow copying the entire struct.

I would like to claim this issue and submit a PR for it!

Rickydama3:
**What type of PR is this?**
/kind bug

This fixes the `go vet` warnings regarding Mutex lock copying in the `nvidiadevice` RM package. By passing `Device` by pointer instead of value, we ensure the embedded mutex is not silently cloned, which would otherwise defeat synchronization. 

*AI Disclosure: I manually authored and reviewed this code, but I used an AI assistant to help me run `go vet` across the repository and draft the initial fix. The final implementation and understanding of the code is my own work.*

Fixes #2651

Note: `make verify` currently fails on the `master` branch due to a pre-existing syntax error in `.golangci.yaml` (`Error: can't load config: the format is required`), but standard `go vet` passes cleanly on these changes.

```release-note
None

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of replicated GPU devices by preserving device metadata and configuration.
  * Enhanced reliability when identifying devices, retrieving device identifiers, and determining supported allocation modes.
  * Fixed test-case handling to ensure allocation scenarios are evaluated against the correct data.

* **Tests**
  * Updated allocation coverage to better validate device-plugin responses.

### DIFF
--- a/pkg/device-plugin/nvidiadevice/nvinternal/plugin/server_test.go
+++ b/pkg/device-plugin/nvidiadevice/nvinternal/plugin/server_test.go
@@ -169,7 +169,7 @@ func TestCDIAllocateResponse(t *testing.T) {
 	}
 
 	for i := range testCases {
-		tc := testCases[i]
+		tc := &testCases[i]
 		t.Run(tc.description, func(t *testing.T) {
 			deviceListStrategies, _ := v1.NewDeviceListStrategies(tc.deviceListStrategies)
 			plugin := NvidiaDevicePlugin{
--- a/pkg/device-plugin/nvidiadevice/nvinternal/rm/device_map.go
+++ b/pkg/device-plugin/nvidiadevice/nvinternal/rm/device_map.go
@@ -38,6 +38,7 @@ import (
 	"github.com/NVIDIA/go-nvlib/pkg/nvlib/device"
 	"github.com/NVIDIA/go-nvlib/pkg/nvlib/info"
 	"github.com/NVIDIA/go-nvml/pkg/nvml"
+	kubeletdevicepluginv1beta1 "k8s.io/kubelet/pkg/apis/deviceplugin/v1beta1"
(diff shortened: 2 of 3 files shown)


mesutoezdil:
The vet warning is from protobuf's zero-size DoNotCopy marker, not a real lock, 
so there's no actual bug here
closing.

PR #2654, opened 2026-08-14
fix(nvidia): preserve devreq order in selectPreferredDeviceIDsFromAnnotatedDevices

adity1raut:
## Summary

`selectPreferredDeviceIDsFromAnnotatedDevices` builds its `GetPreferredAllocation`
response by placing all of kubelet's required (`MustIncludeDeviceIDs`) devices
first, then filling the rest from the scheduler's per-device memory/core plan
(`devreq`/`desired`) in order. Whenever `required` wasn't already `desired`'s
first entry, the returned device order diverged from `devreq`'s order.

`Allocate()` later zips its (possibly kubelet-reordered) `DevicesIds` against
`devreq` **positionally** in `alignContainerDevicesWithAllocatedIDs`, updating
only the UUID at each index and leaving that index's `Usedmem`/`Usedcores`
untouched. With the order divergence above, position `i`'s UUID could end up
pointing at a different physical GPU than the one `devreq[i]` was computed
for — so `CUDA_DEVICE_MEMORY_LIMIT_<i>` would carry the memory ceiling
intended for one GPU while actually applying to another, letting a container
exceed what the scheduler reserved on that GPU.

Triggers with `EnableGetPreferredAllocation: true`, non-MIG mode, a single
container requesting multiple GPUs with heterogeneous per-device memory
(e.g. via `nvidia.com/gpumem-percentage` across differently-sized GPUs), and
a kubelet-supplied `MustIncludeDeviceIDs` that isn't a prefix of the
scheduler's order (realistic on kubelet-driven container restarts).

## Fix

Rebuild the result in `desired`'s order instead of required-first: at each
position, pull from a per-physical-GPU queue of required IDs if one is
pending for that GPU, otherwise from the available pool. This keeps every
position's UUID and its memory/core limit describing the same physical
device, while still satisfying kubelet's `MustIncludeDeviceIDs` constraint.

## Tests

- Tightened the two existing tests that only checked set equality
  (`ElementsMatch`) to assert exact order.
- Added two unit tests: required device at a non-leading `desired` position,
  and multiple required slices from the same physical GPU.
- Added an end-to-end test that drives `GetPreferredAllocation` followed by
  `Allocate`, the way kubelet actually calls the plugin, asserting each
  container's `CUDA_DEVICE_MEMORY_LIMIT_<i>` matches the GPU visible at that
  position.

## Test plan

- [x] `go build ./...`
- [x] `go test ./pkg/device-plugin/nvidiadevice/nvinternal/plugin/... -race -short -count=1 -v` — all pass
- [x] `make test` (full suite, race detector) — all packages pass
- [x] `make lint` (golangci-lint) — 0 issues

## AI Assistance Disclosure

This PR was written primarily using AI coding assistance, following the
project's CONTRIBUTING.md disclosure requirement. The bug was found via an
independent audit of the NVIDIA device-plugin allocation path, reproduced
against this repo's own existing test fixtures before being fixed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Preferred device allocation now preserves the scheduler’s requested device order.
  * Required devices are matched to the correct physical GPU, including multiple slices from one GPU.
  * Memory limits remain associated with the correct GPU after preferred allocation reordering.

* **Tests**
  * Added coverage for required devices appearing in different positions and for multiple slices on the same GPU.

### DIFF
--- a/pkg/device-plugin/nvidiadevice/nvinternal/plugin/server.go
+++ b/pkg/device-plugin/nvidiadevice/nvinternal/plugin/server.go
@@ -664,16 +664,14 @@ func (plugin *NvidiaDevicePlugin) selectPreferredDeviceIDsFromAnnotatedDevices(a
 	}
 
 	requiredSet := make(map[string]bool, len(required))
-	requiredByPhysical := make(map[string]int, len(required))
-	availableByPhysical := make(map[string][]string)
-	selected := make([]string, 0, allocationSize)
-
+	requiredByPhysical := make(map[string][]string, len(required))
 	for _, id := range required {
 		requiredSet[id] = true
-		requiredByPhysical[physicalDeviceID(id)]++
-		selected = append(selected, id)
+		physicalID := physicalDeviceID(id)
+		requiredByPhysical[physicalID] = append(requiredByPhysical[physicalID], id)
 	}
 
+	availableByPhysical := make(map[string][]string)
 	for _, id := range available {
 		if requiredSet[id] {
 			continue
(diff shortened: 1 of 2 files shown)


adity1raut:
`e2e_test / e2e-test (nvidia, tesla-p4)` is failing at the `deploy hami helm` step on two separate runs against two different commits (~6.5h apart), both times identically:

```
Error: failed pre-install: 1 error occurred:
	* timed out waiting for the condition
```

In both runs the `hami-admission-create` Job (Helm pre-install hook, generates the admission-webhook TLS cert) gets created and immediately shows `Jobs active: 1, jobs failed: 0, jobs succeeded: 0`, then never transitions to succeeded/failed for the full 10m `--wait` timeout. Docker image load, helm repo add/update, and the GPU-node check all complete fine before this point.

This PR's diff only touches `pkg/device-plugin/nvidiadevice/nvinternal/plugin/{server.go,server_test.go}` — no Helm chart, webhook, or cert-gen changes — and `Compile`/`Unit test`/`lint` all pass independently on the same commit. Given the identical failure on retry, this looks like a stuck/hung state on the `hami-e2e` runner's cluster (possibly the `hami-admission-create` job's pod stuck Pending/ImagePull, or leftover state from a prior run) rather than a flake or something fixable from this PR. Flagging for a maintainer to take a look at the runner — happy to help debug further if there's a way to inspect `kubectl get pods -n hami-system` / `describe job hami-admission-create` on it.

Run links: https://github.com/Project-HAMi/HAMi/actions/runs/31786120922/job/94724284625 and https://github.com/Project-HAMi/HAMi/actions/runs/31815060257/job/94816897263

mesutoezdil:

adity1raut:

Sorry, my mistake. I’ll make sure to keep the contribution guidelines in mind and take care of this properly in my next PR.

PR #2664, opened 2026-08-14
fix(device-plugin): auto-set CUDA_DISABLE_CONTROL for whole-GPU allocations

### LINKED ISSUE #2641: vLLM pynccl (ctypes-loaded NCCL) hangs at ncclCommInitRank under libvgpu.so on Blackwell; CUDA_DISABLE_CONTROL=true fixes it — auto-set it for whole-GPU allocations?
**What happened**:

vLLM with `tensor-parallel-size=4` on 4x NVIDIA RTX PRO 6000 Blackwell under HAMi v2.9.0 hangs at NCCL communicator init. The last log line is:

```
(Worker pid=960) INFO 08-12 20:26:09 [pynccl.py:113] vLLM is using nccl==2.28.9
```

No further output — the pod never reaches model loading, and the startup probe eventually kills it.

Root cause: HAMi's `libvgpu.so` (injected via `LD_PRELOAD`) intercepts CUDA calls. When vLLM loads NCCL via `ctypes.CDLL` and calls `ncclCommInitRank`, HAMi's interception of `cuMemCreate` (inside NCCL's `ncclCuMemHostEnable` path) deadlocks. Setting `CUDA_DISABLE_CONTROL=true` makes `libvgpu.so` a passthrough, and NCCL init completes normally (in 0.22s).

Notably, this is specific to **ctypes-loaded NCCL**: PyTorch's `torch.distributed` NCCL (C++-linked) works fine without any workarounds — confirmed via `torchrun` 4-GPU AllReduce/AllGather tests. Same NCCL C library, same hardware, same HAMi; the only difference is the loading mechanism. `libvgpu.so` interacts differently with `ctypes.CDLL`-loaded NCCL than with C++-linked NCCL.

vLLM tensor-parallel NCCL init should complete under HAMi, at least for whole-GPU allocations (`nvidia.com/gpu: N` with no `nvidia.com/gpumem`), where there is no memory limit for `libvgpu.so` to enforce.

1. HAMi v2.9.0 on a node with 4x RTX PRO 6000 Blackwell (sm_120, no NVLink, PCIe P2P working).
2. Deploy vLLM (NCCL 2.28.9) requesting `nvidia.com/gpu: 4` (whole GPUs, no `gpumem`/`gpucores` limits) with `--tensor-parallel-size 4`.
3. Pod hangs right after `vLLM is using nccl==2.28.9` — stuck in `ncclCommInitRank`.
4. Add `CUDA_DISABLE_CONTROL=true` to the pod env → init completes and the server starts normally.

Test matrix (each config tested by restarting the deployment):

| Config | vLLM pynccl | torchrun NCCL | AllReduce BW |
|--------|------------|---------------|-------------|
| `CUDA_DISABLE_CONTROL=true` only | **PASS** | PASS | **23.5 GB/s** |
| No workarounds | HANG | PASS | 23.8 GB/s |
| `NCCL_CUMEM_ENABLE=0` only | HANG | PASS | 14.0 GB/s |

Workarounds that are **not** needed (each verified individually):

- `NCCL_P2P_DISABLE=1` — P2P works on this hardware (all 12 GPU pairs accessible, 33–36 GB/s); NCCL uses SHM transport either way (no NVLink).
- `NCCL_CUMEM_ENABLE=0` — does not fix the hang, and keeping cuMem enabled gives ~70% better bandwidth (23.5 vs 14.0 GB/s) once `CUDA_DISABLE_CONTROL=true` is set.
- `--disable-custom-all-reduce` — vLLM auto-selects `['PYNCCL']` only on this topology anyway.

Suggested improvement: **auto-set `CUDA_DISABLE_CONTROL=true` for whole-GPU allocations** (`nvidia.com/gpu: N` without `nvidia.com/gpumem`). In whole-GPU mode there is no memory limit to enforce, so `libvgpu.so`'s interception adds no value and causes the vLLM pynccl deadlock. This was discussed in #1764 but is not automatically applied in v2.9.0.

With `CUDA_DISABLE_CONTROL=true` set, the deployment is live and verified: pod 1/1 Running, health checks passing, serving inference (69.27 GiB model load per rank, ~5 min total startup).

**Environment**:
- HAMi version: v2.9.0
- GPU: 4x NVIDIA RTX PRO 6000 Blackwell Server Edition (96 GB, sm_120), no NVLink — all GPU pairs at `NODE` (PCIe, same NUMA node)
- Platform: IOMMU (AMD-Vi) enabled, no ACS; CUDA P2P functional on all 12 GPU pairs
- Workload: vLLM with NCCL 2.28.9, `tensor-parallel-size=4`, whole-GPU allocation (`nvidia.com/gpu: 4`, no `gpumem`/`gpucores`)

Rickydama3:
**What type of PR is this?**

/kind bug

HAMi allows multiple pods to share a single GPU by enforcing memory limits. It enforces these limits by pushing a file called `ld.so.preload` that intercepts GPU commands. However, when a user requests a whole GPU, HAMi was still pushing that interception file. That interception was causing a major issue: it was making vLLM (a popular AI model runner) freeze up in a deadlock when it tried to initialize its networking (NCCL).

My PR changes the code in `server.go`  to detect when a pod is asking for a whole GPU. Because a whole GPU doesn't need memory limits, we can safely bypass the interception. To do this, our code automatically injects an environment variable called `CUDA_DISABLE_CONTROL=true` into the pod. When this variable is set to true, HAMi skips the `ld.so.preload` interception, which allows vLLM to start up perfectly without freezing or stuck.

We had to be careful: what if a user explicitly wants the interception and manually sets `CUDA_DISABLE_CONTROL=false` in their pod configuration? Our code includes a check for this. It looks at the pod's environment variables first, and if the user explicitly set a value, we never overwrite it.

To prove our code works, we added two new unit tests in `alloc_refactor_test.go`. One test proves that we auto-set the variable when a whole GPU is requested, and the second test proves that we preserve the user's manual choice if they explicitly set it to false.

Fixes #2641

**Hardware Validation:**
- **Device Type**: 4x NVIDIA RTX PRO 6000 Blackwell
- **Driver Version**: Validated by the original reporter (@ok-claven) on their deployment, who confirmed: *"With `CUDA_DISABLE_CONTROL=true` set, the deployment is live and verified: pod 1/1 Running, health checks passing, serving inference."*

**AI Assistance Notice:**
I used an chatgpt to help me understand the vLLM deadlock and write the `CUDA_DISABLE_CONTROL` injection logic, but I fully understand the code produced and have explicitly verified it via the new unit tests.

```release-note
Fix: Auto-set CUDA_DISABLE_CONTROL=true for whole-GPU allocations to prevent vLLM pynccl deadlocks.
```

## Summary by CodeRabbit

* **New Features**
  * Whole-GPU allocations now automatically disable CUDA control when no explicit setting is provided.
  * Explicit CUDA control settings are respected during GPU allocation.
  * Mixed-device allocations preserve CUDA control behavior for devices that are not allocated as whole GPUs.

* **Bug Fixes**
  * Preload mounts are now omitted when CUDA control is disabled and retained when it remains enabled.

### DIFF
--- a/pkg/device-plugin/nvidiadevice/nvinternal/plugin/server.go
+++ b/pkg/device-plugin/nvidiadevice/nvinternal/plugin/server.go
@@ -824,7 +824,7 @@ func (plugin *NvidiaDevicePlugin) Allocate(ctx context.Context, reqs *kubeletdev
 				for i, dev := range devreq {
 					limitKey := fmt.Sprintf("CUDA_DEVICE_MEMORY_LIMIT_%v", i)
 					response.Envs[limitKey] = fmt.Sprintf("%vm", dev.Usedmem)
-					if dev.Usedmem != 0 || dev.Usedcores != 0 {
+					if dev.Usedcores != 100 {
 						isWholeGPU = false
 					}
 				}
--- a/pkg/device-plugin/nvidiadevice/nvinternal/plugin/server.go
+++ b/pkg/device-plugin/nvidiadevice/nvinternal/plugin/server.go
@@ -860,25 +860,21 @@ func (plugin *NvidiaDevicePlugin) Allocate(ctx context.Context, reqs *kubeletdev
 						HostPath: "/tmp/vgpulock",
 						ReadOnly: false},
 				)
-				found := false
+				hasControlSetting := false
+				controlDisabled := false
 				for _, val := range currentCtr.Env {
 					if strings.Compare(val.Name, "CUDA_DISABLE_CONTROL") == 0 {
-						// if env existed but is set to false or can not be parsed, ignore
-						t, _ := strconv.ParseBool(val.Value)
-						if !t {
-							continue
-						}
-						// only env existed and set to true, we mark it "found"
-						found = true
+						hasControlSetting = true
+						controlDisabled, _ = strconv.ParseBool(val.Value)
 						break
 					}
 				}
-				if isWholeGPU && !found {
+				if isWholeGPU && !hasControlSetting {
 					klog.Infof("Whole GPU allocation detected without explicit CUDA_DISABLE_CONTROL, auto-setting to true")
-					found = true
+					controlDisabled = true
 					response.Envs["CUDA_DISABLE_CONTROL"] = "true"
 				}
-				if !found {
+				if !controlDisabled {
 					response.Mounts = append(response.Mounts, &kubeletdevicepluginv1beta1.Mount{ContainerPath: "/etc/ld.so.preload",
 						HostPath: hostHookPath + "/vgpu/ld.so.preload",
 						ReadOnly: true},
--- a/pkg/device-plugin/nvidiadevice/nvinternal/plugin/server.go
+++ b/pkg/device-plugin/nvidiadevice/nvinternal/plugin/server.go
@@ -824,14 +824,11 @@ func (plugin *NvidiaDevicePlugin) Allocate(ctx context.Context, reqs *kubeletdev
 				for i, dev := range devreq {
 					limitKey := fmt.Sprintf("CUDA_DEVICE_MEMORY_LIMIT_%v", i)
 					response.Envs[limitKey] = fmt.Sprintf("%vm", dev.Usedmem)
-					if dev.Usedmem != 0 {
+					if dev.Usedmem != 0 || dev.Usedcores != 0 {
 						isWholeGPU = false
 					}
 				}
 				response.Envs["CUDA_DEVICE_SM_LIMIT"] = fmt.Sprint(devreq[0].Usedcores)
-				if devreq[0].Usedcores != 0 {
-					isWholeGPU = false
-				}
 				response.Envs["CUDA_DEVICE_MEMORY_SHARED_CACHE"] = fmt.Sprintf("%s/vgpu/%v.cache", hostHookPath, uuid.New().String())
 				if *plugin.schedulerConfig.DeviceMemoryScaling > 1 {
 					response.Envs["CUDA_OVERSUBSCRIBE"] = "true"
--- a/pkg/device-plugin/nvidiadevice/nvinternal/plugin/server.go
+++ b/pkg/device-plugin/nvidiadevice/nvinternal/plugin/server.go
@@ -820,11 +820,18 @@ func (plugin *NvidiaDevicePlugin) Allocate(ctx context.Context, reqs *kubeletdev
 			}
 
 			if plugin.operatingMode != "mig" {
+				isWholeGPU := true
 				for i, dev := range devreq {
 					limitKey := fmt.Sprintf("CUDA_DEVICE_MEMORY_LIMIT_%v", i)
 					response.Envs[limitKey] = fmt.Sprintf("%vm", dev.Usedmem)
+					if dev.Usedmem != 0 {
+						isWholeGPU = false
+					}
 				}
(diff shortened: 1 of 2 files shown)


pkg/device-plugin/nvidiadevice/nvinternal/plugin/server.go
```diff
@@ -820,9 +820,13 @@ func (plugin *NvidiaDevicePlugin) Allocate(ctx context.Context, reqs *kubeletdev
 			}
 
 			if plugin.operatingMode != "mig" {
+				isWholeGPU := true
 				for i, dev := range devreq {
 					limitKey := fmt.Sprintf("CUDA_DEVICE_MEMORY_LIMIT_%v", i)
 					response.Envs[limitKey] = fmt.Sprintf("%vm", dev.Usedmem)
+					if dev.Usedmem != 0 || dev.Usedcores != 0 {
```
mesutoezdil: usedmem 0 and usedcores 0 means no limits, not a whole gpu. the scheduler can still place other pods on this card, only coresreq 100 is exclusive. and a real exclusive request with gpucores 100 fails this check and keeps the preload. why not key on coresreq == 100?

pkg/device-plugin/nvidiadevice/nvinternal/plugin/server.go
```diff
@@ -853,20 +857,21 @@ func (plugin *NvidiaDevicePlugin) Allocate(ctx context.Context, reqs *kubeletdev
...
-						}
-						// only env existed and set to true, we mark it "found"
-						found = true
+						hasControlSetting = true
+						controlDisabled, _ = strconv.ParseBool(val.Value)
 						break
 					}
 				}
-				if !found {
+				if isWholeGPU && !hasControlSetting {
```
mesutoezdil, line 869: with control off the pod stops writing the vgpu cache, so vgpumonitor loses metrics for it. this silently changes observability for every count-only pod, not just the vllm case.

mesutoezdil (changes requested):

pkg/device-plugin/nvidiadevice/nvinternal/plugin/server.go
Rickydama3: I misunderstood how 0 was handled by the scheduler—I didn't realize it just meant unbounded limits rather than an exclusive lock. A true exclusive request with gpucores: 100 failing this check is a huge oversight on my end.
I've just pushed a commit that changes the logic to key on dev.Usedcores == 100 instead, so it correctly identifies actual whole-GPU allocations. Thanks for correcting me appreciate your insights

pkg/device-plugin/nvidiadevice/nvinternal/plugin/server.go
Rickydama3: Since auto-setting this breaks vgpumonitor metrics for all exclusive pods, should we just drop this logic entirely and advise vLLM users to manually set CUDA_DISABLE_CONTROL="true" instead?

pkg/device-plugin/nvidiadevice/nvinternal/plugin/server.go
Rickydama3: I misunderstood how the scheduler handled 0 and didn't realize it meant unbounded limits instead of exclusive access. I just pushed a fix to check dev.Usedcores == 100 instead so it actually catches true whole-GPU requests. thanks for your insight

mesutoezdil:

PR #2669, opened 2026-08-15
ci: run Helm lint and template validation in the Chart Lint workflow

### LINKED ISSUE #2666: ci: run helm lint and template validation in the Chart Lint workflow

Update the Chart Lint workflow to run actual Helm validation.

The workflow should:

- Build or update the chart dependencies.
- Run `helm lint` against the default values.
- Run `helm template` and confirm that the chart renders successfully.
- Keep the existing Trivy configuration scan.
- Keep the chart-version verification.
- Provide a local Make target that runs the same checks.

**Why is this needed**:

The workflow `.github/workflows/lint-chart.yaml` is named "Chart Lint", but it currently runs the root `make lint_chart` target.

That target performs a Trivy configuration scan. It does not run `helm lint`.

A separate target in `charts/Makefile` runs `helm lint`, but the workflow does not call it.

Because of this, invalid Helm templates or broken values references may not be detected by the Chart Lint workflow.

Likely affected files:

- `.github/workflows/lint-chart.yaml`
- `Makefile`
- `charts/Makefile`

Suggested completion checks:

- `helm lint` runs in the workflow.
- `helm template` runs in the workflow.
- An invalid Helm template causes the workflow to fail.
- Existing Trivy and chart-version checks continue to run.
- The workflow and local development use a documented Helm version.

No existing issue or PR specifically addressing the missing `helm lint` execution was found.

Issue #2419 and PRs #2441 and #2553 concern `values.schema.json`; they do not address this workflow behavior.

asadjan4611:
**What type of PR is this?**

/kind feature

The Chart Lint workflow previously called `make lint_chart`, which only ran the existing Trivy configuration scan. It did not run Helm linting or render the chart.

This PR:

- Adds a non-mutating chart validation target.
- Builds Helm chart dependencies.
- Runs `helm lint` with the default values.
- Runs `helm template` with the default values.
- Keeps the existing Trivy configuration scan.
- Keeps the existing chart-version verification.
- Adds `make verify_chart` as the shared command for CI and local development.
- Pins and documents Helm v3.8.1.
- Expands the workflow path filter so changes to the workflow or root Makefile trigger chart validation.

The existing release-oriented chart targets are unchanged.

Fixes #2666

Validation completed:

- `make verify_chart` with Helm v3.8.1 — passed (for now  my docker facing any techininal glitch but i'll add it as soon as possible)

- `make verify` — passed
Screenshot from 2026-08-15 06-44-12
- `make test` — passed with race detection
Screenshot from 2026-08-15 06-46-14
- `make build` — passed
Screenshot from 2026-08-15 06-42-43
- `git diff --check` — passed
- Workflow YAML structure validation — passed
- Validation did not modify `Chart.yaml` or `values.yaml`
- A temporary malformed Helm template caused the validation target to fail as expected

The chart currently has no dependencies, so `helm dependency build` is a no-op today, but it ensures future dependencies are validated.

Hardware validation is not applicable because this change only affects CI and local Helm chart validation. It does not affect device allocation or in-container isolation.

AI assistance disclosure:

I used OpenAI Codex to analyze the issue and codebase, implement the workflow, Makefile, and documentation change.

No. This only changes CI and local chart-validation behavior.

## Summary by CodeRabbit

* **Chores**
  * Standardized chart verification on Helm v3.21.4.
  * Added a unified chart verification command covering linting, rendering, security scanning, and version consistency checks.
  * Pull requests now automatically validate charts when related workflow or build files change.
  * Improved chart validation to include subcharts and rendered template checks.

* **Documentation**
  * Added contributor guidance for Helm chart changes and required verification steps.

### DIFF
--- a/.github/workflows/lint-chart.yaml
+++ b/.github/workflows/lint-chart.yaml
@@ -1,7 +1,7 @@
 name: Chart Lint
 
 env:
-  HELM_VERSION: v3.8.1
+  HELM_VERSION: v3.21.4
 
 on:
   push:
--- a/.github/workflows/lint-chart.yaml
+++ b/.github/workflows/lint-chart.yaml
@@ -1,5 +1,8 @@
 name: Chart Lint
 
+env:
+  HELM_VERSION: v3.8.1
+
 on:
   push:
     # Exclude branches created by Dependabot to avoid triggering current workflow
@@ -9,6 +12,8 @@ on:
   pull_request:
     paths:
       - "charts/**"
+      - ".github/workflows/lint-chart.yaml"
+      - "Makefile"
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
@@ -30,16 +35,12 @@ jobs:
       - name: Set up Helm
         uses: azure/setup-helm@9bc31f4ebc9c6b171d7bfbaa5d006ae7abdb4310 # v5
         with:
-          version: v3.7.1
+          version: ${{ env.HELM_VERSION }}
 
       - name: Add HAMi DRA Helm repository
         run: |
           helm repo add hami-dra https://project-hami.github.io/HAMi-DRA/
           helm repo update hami-dra
 
-      - name: Lint Chart
-        run: |
-          make lint_chart
-      - name: Check chart version
-        run: bash ./hack/verify-chart-version.sh
-
+      - name: Validate chart
+        run: make verify_chart
--- a/Makefile
+++ b/Makefile
@@ -94,9 +94,10 @@ lint_chart:
       echo "chart trivy check: pass"
 
 .PHONY: verify_chart
-verify_chart: lint_chart
+verify_chart:
 	$(MAKE) -C charts validate
 	bash ./hack/verify-chart-version.sh
+	$(MAKE) lint_chart
 
 .PHONY: e2e-env-setup
 e2e-env-setup:
--- a/Makefile
+++ b/Makefile
@@ -93,6 +93,11 @@ lint_chart:
       (($$?==0)) || { echo "error, failed to check chart trivy" && exit 1 ; } ; \
       echo "chart trivy check: pass"
 
+.PHONY: verify_chart
+verify_chart: lint_chart
+	$(MAKE) -C charts validate
+	bash ./hack/verify-chart-version.sh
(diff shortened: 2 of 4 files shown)


.github/workflows/lint-chart.yaml
```diff
@@ -1,5 +1,8 @@
 name: Chart Lint
 
+env:
+  HELM_VERSION: v3.8.1
```
mesutoezdil: why 3.8.1? that release is from 2022. the new contributing text now tells people to install a four year old helm. pin something current.

charts/Makefile
```diff
@@ -6,9 +6,15 @@ VERSION_REGEX := '[vV]*[0-9]\+\.[0-9]\+\.[0-9]\+.*'
 CHART_FILE := "./hami/Chart.yaml"
 VALUES_FILE := "./hami/values.yaml"
 
-.PHONY: all lint update-versions deps
+.PHONY: all lint update-versions deps validate
 all: update-versions deps lint package
 
+# Validate the committed chart without updating release versions.
+validate:
+	helm dependency build ./hami
```
mesutoezdil: the chart has no dependencies block, so this is a no-op. drop it. the --values flags on the next two lines are also no-ops, they point at the chart's own defaults.

Makefile
```diff
@@ -93,6 +93,11 @@ lint_chart:
       (($$?==0)) || { echo "error, failed to check chart trivy" && exit 1 ; } ; \
       echo "chart trivy check: pass"
 
+.PHONY: verify_chart
+verify_chart: lint_chart
```
mesutoezdil: lint_chart needs docker and runs first, so without docker the helm checks never run. you hit this yourself in the pr body. run lint_chart last.

mesutoezdil (dismissed):

.github/workflows/lint-chart.yaml
asadjan4611: fixed

charts/Makefile
asadjan4611: fixed

Makefile
asadjan4611: fixed

archlitchi (approved):
/lgtm

PR #2674, opened 2026-08-15
fix(awsneuron): prevent odd NeuronCore request truncation

### LINKED ISSUE #2673: [BUG] awsneuron: odd neuroncore requests are silently truncated
**What happened**:

The AWS Neuron backend silently allocates fewer NeuronCores than requested when the request is an odd number greater than one.

In `AWSNeuronDevices.GenerateResourceRequests`, a core request `n >= 2` is converted to:

```go
num = int(n / 2)
corenum = 2
```

For example, `aws.amazon.com/neuroncore: 3` becomes `Nums: 1, Coresreq: 2`. The scheduler therefore records and binds only two cores for a container that requested three.

The same path assumes two-core masks in `addCoreUsage` (mask `3`) and `PatchAnnotations` only emits bits `1` and `2`, even though Inferentia1 devices expose four NeuronCores per device.

HAMi should either allocate exactly the requested number of NeuronCores across devices or reject unsupported request shapes during admission. It must not silently reduce the request.

Core masks and assigned core indexes should reflect the node's actual NeuronCore count per device.

1. Configure the AWS Neuron backend with `aws.amazon.com/neuroncore`.
2. Submit a container with:

```yaml
resources:
  limits:
    aws.amazon.com/neuroncore: 3
```

3. Observe that `GenerateResourceRequests` produces one device requesting two cores, so scheduling and annotations account for two cores rather than three.

This can also be reproduced directly with a unit test calling `GenerateResourceRequests` and asserting that a request of three is not converted to `Nums=1, Coresreq=2`.

The issue is present on `master` at `e34913fbf35ab0f2063109e9addb561fc5d72891`.

This is scheduler-side request conversion and accounting. No NVIDIA logs are applicable.

**Environment**:
- HAMi version: `master` at `e34913fbf35ab0f2063109e9addb561fc5d72891`
- nvidia driver or other AI device driver version: AWS Neuron; not required for the code-level reproduction
- Docker version from `docker version`: Not applicable
- Docker command, image and tag used: Not applicable
- Kernel version from `uname -a`: Not applicable
- Others: Reproduced from the scheduler request-conversion path

Eshiv-Pandey:
**What type of PR is this?**

/kind bug

Rejects AWS Neuron core requests with unsupported odd counts greater than one during admission, instead of silently allocating fewer cores than requested.

If admission is bypassed, request conversion rounds the device count up so the scheduler never under-allocates the requested cores.

Adds regression coverage for admission validation and defensive request conversion.

Fixes #2673

This is a scheduler-side validation and accounting change. It was validated with unit tests and does not require AWS Neuron hardware.

Validation:
- `gofmt`: passed for changed files
- AWS Neuron package tests: passed
- Targeted `golangci-lint`: passed with noo issues
- Import-alias verification: passed
- `make verify` / `make lint`
- `make test`

AI assistance disclosure:

Took help of Codex in understanding the nearby code and flow.

```release-note
AWS Neuron requests for unsupported odd core counts greater than one are now rejected instead of being silently under-allocated.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation for AWS Neuron device and core resource requests.
  * Invalid, zero, negative, overflowing, and unsupported odd core requests are now rejected with clear errors.
  * Resource allocation now consistently honors valid core requests without rounding.
  * Admission checks and resource allocation now apply consistent limits and validation.
  * Invalid requests no longer result in device allocations.

### DIFF
--- a/pkg/device/awsneuron/device.go
+++ b/pkg/device/awsneuron/device.go
@@ -79,7 +79,18 @@ func (dev *AWSNeuronDevices) CommonWord() string {
 func (dev *AWSNeuronDevices) MutateAdmission(ctr *corev1.Container, p *corev1.Pod) (bool, error) {
 	_, ok := ctr.Resources.Limits[corev1.ResourceName(dev.resourceCountName)]
 	if !ok {
-		_, ok = ctr.Resources.Limits[corev1.ResourceName(dev.resourceCoreName)]
+		core, coreRequested := ctr.Resources.Limits[corev1.ResourceName(dev.resourceCoreName)]
+		if !coreRequested {
+			return false, nil
+		}
+		coreCount, isInteger := core.AsInt64()
+		if !isInteger {
+			return false, fmt.Errorf("%s must be an integer", dev.resourceCoreName)
+		}
+		if coreCount > 1 && coreCount%2 != 0 {
+			return false, fmt.Errorf("%s must be 1 or a multiple of 2", dev.resourceCoreName)
+		}
+		ok = true
 	}
 	return ok, nil
 }
@@ -246,10 +257,7 @@ func (dev *AWSNeuronDevices) GenerateResourceRequests(ctr *corev1.Container) dev
 				klog.InfoS("Detected awsNeuron device request",
 					"container", ctr.Name,
 					"deviceCores", n)
-				num := 1
-				if n >= 2 {
-					num = int(n / 2)
-				}
+				num := n/2 + n%2
 				corenum := 1
 				if n >= 2 {
 					corenum = 2
--- a/pkg/device/awsneuron/device.go
+++ b/pkg/device/awsneuron/device.go
@@ -18,6 +18,7 @@ package awsneuron
 
 import (
 	"fmt"
+	"math"
 	"slices"
 	"strconv"
 	"strings"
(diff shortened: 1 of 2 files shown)


pkg/device/awsneuron/device.go
```diff
@@ -77,11 +78,45 @@ func (dev *AWSNeuronDevices) CommonWord() string {
...
+
+	core, coreRequested := resourceQuantity(ctr, corev1.ResourceName(dev.resourceCoreName))
+	if !coreRequested {
+		return false, nil
+	}
+	coreCount, err := dev.validateCoreRequest(core)
+	if err != nil {
+		return false, err
+	}
+	if coreCount > 1 && coreCount%2 != 0 {
```
mesutoezdil: this hardcodes 2 cores per device, but coresPerAWSNeuron is computed from node capacity elsewhere in this file. what happens on an instance type where it is not 2?

pkg/device/awsneuron/device.go
```diff
@@ -237,19 +269,13 @@ func (dev *AWSNeuronDevices) GenerateResourceRequests(ctr *corev1.Container) dev
...
 	} else {
-		core, ok := ctr.Resources.Limits[awsResourceCores]
-		if !ok {
-			core, ok = ctr.Resources.Requests[awsResourceCores]
-		}
+		core, ok := resourceQuantity(ctr, awsResourceCores)
 		if ok {
-			if n, ok := core.AsInt64(); ok {
+			if n, err := dev.validateCoreRequest(core); err == nil {
 				klog.InfoS("Detected awsNeuron device request",
```
mesutoezdil: admission says 3 is invalid, this line says 3 means 2 devices and 4 cores. two rules for the same input. make this path reject too, like it does for zero and negative.

mesutoezdil (dismissed):

pkg/device/awsneuron/device.go
Eshiv-Pandey: i did managed to trace it down somehow. the limit isn't admission though 
addCoreUsage uses a two-bit mask and PatchAnnotations only emits bits 1 and 2, so nothing can address more than 2 cores/device today . the divisor now comes from coresPerAWSNeuron, bounded by a named maxCoresPerNeuronDevice. lifting that so inf1 gets all four can be nice followup or should i do it here?

pkg/device/awsneuron/device.go
Eshiv-Pandey: hmm agreed and rounding up was wrong. both paths now go through splitCoreRequest, so admission can't disagree with the conversion, and this path rejects like the zero and negative cases. 
checked 1-64 against the old formula representable requests are unchanged, only odd counts above one flip from silent truncation to rejection.

archlitchi (approved):
/lgtm

PR #2677, opened 2026-08-15
fix(amd): require the device count in MutateAdmission

### LINKED ISSUE #2679: [BUG] amd: MutateAdmission admits mem/core-only and invalid-count pods that bypass GPU quota and scheduling
**What happened**:

`AMDDevices.MutateAdmission` returns `true` for a container that requests only `amd.com/gpu-mem` and/or `amd.com/gpu-core-pct` without the `amd.com/gpu` count — and also for an invalid count (`0`, negative, fractional, or `> math.MaxInt32`).

Because the mutating webhook sets `pod.Spec.SchedulerName` to the HAMi scheduler whenever any backend's `MutateAdmission` returns `true`, such a pod is routed into the HAMi path. But `GenerateResourceRequests` and `LockNode` both key off the count resource, so they produce an empty request (`Nums=0`) and take no node lock. The result:

- the namespace GPU `ResourceQuota` check is skipped (`fitResourceQuota` continues on `Nums==0`),
- the scheduler filter passes on every node (`hasHAMiResource` stays `false`),
- the pod binds with **no AMD device reserved** and no `hami.io/amd-devices-allocated` annotation.

No error surfaces anywhere. AMD is the only backend with this gap — nvidia backfills the count in `mutateContainerResource`, and metax checks the count in both admission and request generation.

A container without a valid `amd.com/gpu` count should not be treated as a HAMi GPU pod (mem/core-only → not admitted into the HAMi path; invalid count → rejected), so admission stays consistent with `GenerateResourceRequests`/`LockNode` and quota/scheduling are never bypassed.

**How to reproduce it**:

On a stock Helm install (which registers `amd.com/gpu-mem`/`amd.com/gpu-core-pct` as managed resources by default), create a pod whose only AMD resource is memory:

```yaml
resources:
  limits:
    amd.com/gpu-mem: "4096"   # note: no amd.com/gpu
```

The pod's `schedulerName` is rewritten to the HAMi scheduler, yet it binds to an arbitrary node with no AMD device reserved and no quota check. Same outcome with `amd.com/gpu: 0` (or a fractional/oversized count).

Found by code inspection on `master` (commit `51c593c`). Fix proposed in #2677: gate `MutateAdmission` on a valid `amd.com/gpu` count (positive integer ≤ `math.MaxInt32`), matching `GenerateResourceRequests`, `LockNode`, the metax backend, and the documented amd-vgpu pod spec.

_This report was prepared with AI assistance (per CONTRIBUTING.md); all details were reviewed and verified by me._

imantaba:
Fixes #2679

**What this fixes**

`AMDDevices.MutateAdmission` admits a container that requests only `amd.com/gpu-mem` and/or `amd.com/gpu-core-pct` without the `amd.com/gpu` count. Because the webhook sets `pod.Spec.SchedulerName` whenever any backend's `MutateAdmission` returns `true`, such a pod is routed to the HAMi extender as a GPU workload — but `GenerateResourceRequests` and `LockNode` both gate on the **count** resource, so they produce `Nums=0` and take no lock. The pod then:

- skips the namespace GPU quota check (`fitResourceQuota` `continue`s on `Nums==0`),
- passes the scheduler filter on every node (`hasHAMiResource` stays `false`),
- binds with **no AMD device reserved** and no `hami.io/amd-devices-allocated` annotation.

No error surfaces anywhere. AMD is the only backend with this gap: nvidia backfills the count in `mutateContainerResource` to keep the paths consistent; metax checks the count in both. The documented AMD pod spec (`docs/develop/amd-vgpu.md`) always sets `amd.com/gpu`.

**Fix**

Gate `MutateAdmission` on the count resource so it agrees with `GenerateResourceRequests` and `LockNode`. Core-percentage range validation is unchanged when the count is present. Two unit-test cases that encoded the old behavior (`mem-only`, `core-only` → `true`) are corrected to `false` and renamed. Follow-up commits also validate the count value itself (positive integer ≤ `math.MaxInt32`) to close the same divergence for invalid counts.

**Testing:** `go test ./pkg/device/amd/...` passes (13/13 `MutateAdmission` subtests); gofmt/goimports/vet/build clean.

**Alternative considered:** mirror nvidia by backfilling `amd.com/gpu=1` when mem/core is present without a count. Rejected as larger/behavioral — AMD has no `DefaultGPUNum` config and no device-plugin yet, so requiring the count (matching every other AMD code path and the docs) is the minimal, consistent fix.

## Summary by CodeRabbit

* **Bug Fixes**
  * AMD GPU admission now requires an explicit, positive whole-number GPU device count.
  * Invalid counts—including zero, negative, fractional, or oversized values—are rejected.
  * Requests specifying only GPU memory or core percentage are no longer accepted as GPU requests.
  * Core percentage validation remains enforced when a valid GPU count is provided.

### DIFF
--- a/pkg/device/amd/device.go
+++ b/pkg/device/amd/device.go
@@ -76,7 +76,7 @@ func (dev *AMDDevices) CommonWord() string {
 }
 
 func (dev *AMDDevices) MutateAdmission(ctr *corev1.Container, p *corev1.Pod) (bool, error) {
-	_, ok := ctr.Resources.Limits[corev1.ResourceName(dev.resourceCountName)]
+	count, ok := ctr.Resources.Limits[corev1.ResourceName(dev.resourceCountName)]
 	if !ok {
 		// A container that does not request the AMD device count is not a HAMi
 		// GPU request. GenerateResourceRequests and LockNode both key off the
@@ -85,6 +85,14 @@ func (dev *AMDDevices) MutateAdmission(ctr *corev1.Container, p *corev1.Pod) (bo
 		// every later stage treats it as a non-GPU pod: quota is skipped, the
 		// filter passes on all nodes, and the pod binds with no device reserved.
 		return false, nil
+	}
+	// Validate the count value, not just its presence: GenerateResourceRequests
+	// rejects a non-integer, non-positive, or out-of-range count and yields an
+	// empty request, so admitting such a value here would reintroduce the same
+	// "admitted but never scheduled/locked" divergence this guard closes.
+	countValue, countIsInteger := count.AsInt64()
+	if !countIsInteger || countValue <= 0 || countValue > math.MaxInt32 {
+		return false, fmt.Errorf("%s must be a positive integer no greater than %d", dev.resourceCountName, math.MaxInt32)
 	}
 	if core, coreRequested := ctr.Resources.Limits[corev1.ResourceName(dev.resourceCoreName)]; coreRequested {
 		corePercentage, coreIsInteger := core.AsInt64()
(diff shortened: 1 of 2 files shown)


mesutoezdil:

PR #2678, opened 2026-08-15
test(pkg/device): add nil guards and unit tests for pod and node parsing

version-control356:
**Description**

Added explicit nil checks and safety guards in pkg/device/devices.go to prevent potential nil pointer dereference panics when handling uninitialized or nil Kubernetes Pod and Node objects during device allocation and health checking. 
Added unit test cases in pkg/device/devices_test.go to cover nil inputs and maintain code coverage.

**Changes Made**

- pkg/device/devices.go: Added defensive nil checks to Resourcereqs and CheckHealth before accessing Pod or Node specs/annotations.
- pkg/device/devices_test.go: Added TestResourcereqs_NilPod and TestCheckHealth_NilNode test cases to validate robust execution without panicking.

**Testing Completed**

- [x] Passed all local unit tests via go test ./pkg/device
- [x] Verified zero panics when passing nil values to Resourcereqs and CheckHealth

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved resilience when health checks receive missing node information.
  * Prevented resource requirement processing from failing when pod data is unavailable.
  * Safely skips unavailable device implementations during container processing.
  * Added safeguards for missing annotations and validated handling of empty inputs.

### DIFF
--- a/pkg/device/devices.go
+++ b/pkg/device/devices.go
@@ -571,6 +571,10 @@ func PlatternMIG(n *MigInUse, templates []Geometry, templateIdx int) {
 }
 
 func CheckHealth(devType string, resourceCountName string, node *corev1.Node) (bool, bool) {
+	if node == nil {
+		return false, false
+	}
+
 	handshake := node.Annotations[util.HandshakeAnnos[devType]]
 	if strings.Contains(handshake, "Requesting") {
 		_, timestampStr, found := strings.Cut(handshake, "_")
@@ -634,6 +638,11 @@ func ExtractMigTemplatesFromUUID(uuid string) (int, int, error) {
 }
 
 func Resourcereqs(pod *corev1.Pod) (counts PodDeviceRequests) {
+	if pod == nil {
+		klog.V(4).InfoS("Nil pod passed to Resourcereqs")
+		return nil
+	}
+
 	// Total containers = init containers + regular containers
 	totalContainers := len(pod.Spec.InitContainers) + len(pod.Spec.Containers)
 	counts = make(PodDeviceRequests, totalContainers)
@@ -654,6 +663,9 @@ func Resourcereqs(pod *corev1.Pod) (counts PodDeviceRequests) {
 			"containerIndex", i,
 			"containerName", pod.Spec.InitContainers[i].Name)
 		for idx, val := range devices {
+			if val == nil {
+				continue
+			}
 			request := val.GenerateResourceRequests(&pod.Spec.InitContainers[i])
 			if request.Nums > 0 {
 				cnt += request.Nums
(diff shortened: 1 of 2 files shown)


pkg/device/devices.go
```diff
@@ -571,6 +571,10 @@ func PlatternMIG(n *MigInUse, templates []Geometry, templateIdx int) {
 }
 
 func CheckHealth(devType string, resourceCountName string, node *corev1.Node) (bool, bool) {
+	if node == nil {
```
maishivamhoo123, line 574: CheckHealth only gets nodes from the informer lister in scheduler.go, and a lister never returns nil entries.
So i think addding this nil Pointer is redundant.

pkg/device/devices.go
```diff
@@ -634,6 +638,11 @@ func ExtractMigTemplatesFromUUID(uuid string) (int, int, error) {
 }
 
 func Resourcereqs(pod *corev1.Pod) (counts PodDeviceRequests) {
+	if pod == nil {
```
maishivamhoo123, line 641: Resourcereqs is only called from Filter, and the line right above the call already does args.Pod.Name a nil pod would panic there before this guard is ever reached.
this is also redundant.

pkg/device/devices.go
```diff
@@ -654,6 +663,9 @@ func Resourcereqs(pod *corev1.Pod) (counts PodDeviceRequests) {
 			"containerIndex", i,
 			"containerName", pod.Spec.InitContainers[i].Name)
 		for idx, val := range devices {
+			if val == nil {
```
maishivamhoo123, line 666: Devices Map can't hold a nil value either registration calls dev.CommonWord() before inserting, so that would crash at startup.

pkg/device/devices.go
```diff
@@ -672,6 +684,9 @@ func Resourcereqs(pod *corev1.Pod) (counts PodDeviceRequests) {
 			"containerIndex", initContainerOffset+i,
 			"containerName", pod.Spec.Containers[i].Name)
 		for idx, val := range devices {
+			if val == nil {
```
maishivamhoo123, line 687: Same as above L666.

maishivamhoo123:
@version-control356 Thank you for the contribution. I reviewed the PR and added the comments  i think all the nil checks you  added here are unreachable on master. Could you try to reproduce on master? If you hit a panic please share.

version-control356:
> /hold

@maishivamhoo123 thanks for pointing out!
Following up on this after looking closer at pkg/device:

While I agree the current scheduler.go flow guarantees non-nil inputs at runtime on master, my main intention was package-level resilience and test coverage. Since CheckHealth and Resourcereqs are exported package functions, keeping defensive entry-point guards ensures pkg/device won't panic if imported/called independently in future refactors.

I'm happy to trim the internal map iteration checks (L666/etc. ) while keeping the entry guards and unit tests so pkg/device stays fully decoupled. Let me know if that sounds good to you!

PR #2680, opened 2026-08-15
fix(scheduler,util): guard against nil pointer panics in extender rou…

AnmolM-777:
**What type of PR is this?**

/kind bug

This PR adds defensive `nil` pointer guard checks and unit test coverage to prevent runtime server panics in the HAMi scheduler extender and utility functions:

1. **`pkg/scheduler/routes/route.go`**:
   - **`PredicateRoute`**: Guarded `extenderArgs.Pod.Name` access when handling filter errors to prevent panics if `extenderArgs.Pod` is `nil`.
   - **`WebHookRoute`**: Added `if h == nil` guard before `h.ServeHTTP(w, r)` to return HTTP 500 cleanly if `NewWebHook()` failed during initialization.

2. **`pkg/util/util.go`**:
   - Added `nil` node guards to `EmitNodeWarningEvent`, `PatchNodeAnnotations`, and `RemoveNodeAnnotation`.
   - Added `nil` pod guard to `AllInitContainersSucceeded`.

3. **Unit Tests Added**:
   - `pkg/scheduler/routes/route_test.go`: Added `TestPredicateRoute_NilPodInArgs` and `TestWebHookRoute_HandlerExecution`.
   - `pkg/util/util_test.go`: Added `TestEmitNodeWarningEvent_NilNode`, `TestAllInitContainersSucceeded_NilPod`, `TestPatchNodeAnnotations_NilNode`, and `TestRemoveNodeAnnotation_NilNode`.

Fixes #

- All changes pass `make verify`, `make lint`, and unit tests with race detection (`go test -race`).
- Per project guidelines in `CONTRIBUTING.md`, AI assistance was used for initial codebase exploration and test boilerplate generation.

NONE

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of requests with missing pod or node data to prevent crashes.
  * Webhook requests now return a clear HTTP 500 error when the handler is unavailable.
  * Invalid nil inputs now produce safe errors or default results instead of unexpected behavior.

* **Tests**
  * Added regression coverage for missing resources, webhook execution, and safe error handling.

### DIFF
--- a/pkg/util/util.go
+++ b/pkg/util/util.go
@@ -137,6 +137,9 @@ func GetAllocatePodByNode(ctx context.Context, nodeName string) (*corev1.Pod, er
 }
 
 func PatchNodeAnnotations(node *corev1.Node, annotations map[string]string) error {
+	if node == nil {
+		return fmt.Errorf("node is nil")
+	}
 	type patchMetadata struct {
 		Annotations map[string]string `json:"annotations,omitempty"`
 	}
@@ -160,6 +163,9 @@ func PatchNodeAnnotations(node *corev1.Node, annotations map[string]string) erro
 	return err
 }
 func AllInitContainersSucceeded(pod *corev1.Pod) bool {
+	if pod == nil {
+		return false
+	}
 	if len(pod.Status.InitContainerStatuses) == 0 {
 		return false
 	}
@@ -249,6 +255,9 @@ func MarkAnnotationsToDelete(devType string, nn string) error {
 }
 
 func RemoveNodeAnnotation(node *corev1.Node, annotationKeys ...string) error {
+	if node == nil {
+		return fmt.Errorf("node is nil")
+	}
 	annos := make(map[string]any, len(annotationKeys))
 	for _, key := range annotationKeys {
 		annos[key] = nil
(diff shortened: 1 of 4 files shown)


pkg/util/util.go
```diff
@@ -249,6 +255,9 @@ func MarkAnnotationsToDelete(devType string, nn string) error {
 }
 
 func RemoveNodeAnnotation(node *corev1.Node, annotationKeys ...string) error {
+	if node == nil {
```
maishivamhoo123, line 258: this is also not rechable because the only caller of this function is  MarkAnnotationsToDelete, which returns early if GetNode errors so node is always non-nil.

pkg/util/util.go
```diff
@@ -160,6 +163,9 @@ func PatchNodeAnnotations(node *corev1.Node, annotations map[string]string) erro
 	return err
 }
 func AllInitContainersSucceeded(pod *corev1.Pod) bool {
+	if pod == nil {
```
maishivamhoo123, line 166: this is unreachable because only this get called from newPod from the informer update handler and already dereferences it via podManager.UpdatePod right before this call, so pod can never be nil.

pkg/util/util.go
```diff
@@ -320,6 +329,10 @@ func AllContainersCreated(pod *corev1.Pod) bool {
 
 // EmitNodeWarningEvent emits a Warning event on the given Node with deduplication.
 func EmitNodeWarningEvent(node *corev1.Node, reason, message string, dedupWindow time.Duration) {
+	if node == nil {
```
maishivamhoo123, line 332: this is same not reachable  because only caller is RegisterInAnnotation obtains node from util.GetNode with an error check on its first line, so nil can't reach this function.

pkg/util/util_test.go
```diff
@@ -919,4 +919,26 @@ func TestEmitNodeWarningEvent(t *testing.T) {
 		// Old event still present plus one new event.
```
maishivamhoo123, line 919: please try to reproduce these test in master branch,  and let us know if it panic.

mesutoezdil:

PR #2683, opened 2026-08-15
fix test crash in client test

### LINKED ISSUE #2682: bug: test suite crashes on clean environments due to InitGlobalClient called before mocks in client_test.go
/kind bug

**What happened?**
In `pkg/util/client/client_test.go`, the test `TestGetClient` calls `InitGlobalClient()` (line 43) *before* the mock functions (`buildConfigFromFlags` and `inClusterConfig`) are set up in the test loop. 

Because `InitGlobalClient` is invoked immediately with the real underlying functions, it attempts to read `~/.kube/config`. If the developer is working in a clean environment or CI without a local kubeconfig, `InitGlobalClient` triggers a `klog.Fatalf` and crashes the entire test suite:

`F0815 23:21:40.269010    7450 client.go:82] Failed to initialize global client: failed to load kubeconfig: unable to load in-cluster configuration, KUBERNETES_SERVICE_HOST and KUBERNETES_SERVICE_PORT must be defined`

Furthermore, because `InitGlobalClient` is protected by `sync.Once`, the subsequent mock injections in the test loop have absolutely no effect, meaning the fallback logic is never actually tested.

**How to reproduce it?**
1. Remove or rename your `~/.kube/config` (or run in a clean container).
2. Run `go test ./pkg/util/client`
3. Observe the `klog.Fatalf` crash.

To fix this, we need to refactor the test to properly inject the mocks *before* initialization, and likely bypass or reset the `sync.Once` state so the table-driven tests can properly evaluate both the success and fallback paths.

KunwarSidhu47:
**What type of PR is this?**
/kind bug

This PR resolves a logic flaw in the `client` test suite where the `InitGlobalClient` function was prematurely invoked before the necessary mock functions were injected. 

Previously, `InitGlobalClient` was invoked outside the main table-driven test loop. Because it leverages `sync.Once` and was executing the un-mocked functions, the test suite would immediately crash with a `klog.Fatalf` if the developer running the tests did not have a valid `~/.kube/config` present on their local machine.

This PR re-wires the mocks to ensure they are actually evaluated by the underlying config loader and safely initializes the client strictly inside the test loop after the mocks are set. This allows the test suite to execute successfully and independently of the host machine's configuration.

Fixes #2682 

The global `sync.Once` variable is now reset on each loop iteration to ensure that the mocked fallback execution paths are successfully and uniquely evaluated for each test case.

**AI Disclosure**: 
This PR was developed with the assistance of an AI coding assistant. All logic changes and test implementations have been manually reviewed, verified, and thoroughly tested locally before submission.

None

## Summary by CodeRabbit

* **Tests**
  * Improved test coverage and isolation for Kubernetes client configuration.
  * Added support for reliably testing configuration-loading scenarios, including in-cluster fallback behavior.
  * Verified client creation with optional configuration settings and multiple initialization paths.

### DIFF
--- a/pkg/util/client/client_test.go
+++ b/pkg/util/client/client_test.go
@@ -40,7 +40,6 @@ func TestGetClient(t *testing.T) {
 		buildConfigErr error
 		inCluster      *rest.Config
 		inClusterErr   error
-		expectError    bool
 	}{
 		{
 			name:           "Success from kubeconfig",
@@ -49,7 +48,6 @@ func TestGetClient(t *testing.T) {
 			buildConfigErr: nil,
 			inCluster:      nil,
 			inClusterErr:   nil,
-			expectError:    false,
 		},
 		{
 			name:           "Fallback to in-cluster config",
@@ -58,7 +56,6 @@ func TestGetClient(t *testing.T) {
 			buildConfigErr: errors.New("kubeconfig error"),
 			inCluster:      &rest.Config{Host: "https://in-cluster.example.com"},
 			inClusterErr:   nil,
-			expectError:    false,
 		},
 	}
 
@@ -84,19 +81,14 @@ func TestGetClient(t *testing.T) {
 			defer os.Setenv("KUBECONFIG", oldKubeConfig)
 
 			// Reset sync.Once and initialize the global client using the injected mocks
+			KubeClient = nil
 			once = sync.Once{}
 			InitGlobalClient()
 
 			// Call GetClient and check the result.
 			client := GetClient()
-			if tt.expectError {
-				if client != nil {
-					t.Errorf("Expected error, but got a valid client")
-				}
-			} else {
-				if client == nil {
-					t.Errorf("Expected a valid client, but got nil")
-				}
+			if client == nil {
+				t.Errorf("Expected a valid client, but got nil")
 			}
 		})
 	}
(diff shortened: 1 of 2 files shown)


pkg/util/client/client_test.go
```diff
@@ -91,6 +83,10 @@ func TestGetClient(t *testing.T) {
 			os.Setenv("KUBECONFIG", tt.kubeConfig)
 			defer os.Setenv("KUBECONFIG", oldKubeConfig)
 
+			// Reset sync.Once and initialize the global client using the injected mocks
+			once = sync.Once{}
+			InitGlobalClient()
```
mesutoezdil, line 86: note: make test never crashes, hack/unit-test.sh writes a fake kubeconfig first. only bare go test hits this. also reset KubeClient = nil here, a client left from the last case can hide a failed init.

pkg/util/client/client_test.go
```diff
@@ -91,6 +83,10 @@ func TestGetClient(t *testing.T) {
...
+
 			// Call GetClient and check the result.
 			client := GetClient()
 			if tt.expectError {
```
mesutoezdil: expectError is dead now. no case sets it, and a failed init calls klog.Fatalf and kills the test binary. remove it?

mesutoezdil (dismissed):

pkg/util/client/client_test.go
KunwarSidhu47: done, added KubeClient = nil

pkg/util/client/client_test.go
KunwarSidhu47: done, removed expectError entirely

KunwarSidhu47:
@mesutoezdil , made the changes as you suggested. let me know if there's anything else you'd like me to update

archlitchi (approved):
/lgtm

PR #2687, opened 2026-08-16
fix(scheduler): reject filter requests with no candidate nodes

### LINKED ISSUE #2688: Nil pointer dereference in Scheduler.Filter when ExtenderArgs has neither Nodes nor NodeNames
**What happened**:

`Scheduler.Filter` in `pkg/scheduler/scheduler.go` takes the live filter path whenever `args.Nodes` is nil (`:1038`), and that path dereferences `*args.NodeNames` in two places without a nil check:

- `:1061` — `fmt.Errorf("no available node, %d nodes do not meet", len(*args.NodeNames))`
- `:1100` — `genSuccessMsg(len(*args.NodeNames), ...)`

`ExtenderArgs` can carry neither `Nodes` nor `NodeNames`. `getNodesUsage(nil, pod)` returns an empty map without an error, so `calcScore` yields an empty node list and execution reaches `:1061`:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation]
	pkg/scheduler/scheduler.go:1061
```

Because this runs on the extender's HTTP goroutine, the panic terminates the scheduler process rather than failing the single request that caused it.

The function already has a nil-guarding `nodeNamesLen()` helper (`:607`) and uses it elsewhere — these two call sites just don't.

A filter request with no candidate nodes should be rejected with an error result, and the extender process should stay up.

```go
// pkg/scheduler, with devices initialised via config.InitDevicesWithConfig
pod := &corev1.Pod{ /* requests hami.io/gpu: 1 */ }
s.Filter(extenderv1.ExtenderArgs{Pod: pod}) // Nodes and NodeNames both nil -> panics
```

Suggested fix: return an error result when both node fields are nil, and switch the two remaining dereferences to the existing `nodeNamesLen()` helper.

The guard needs to sit *above* `TakeAndDeletePod` (`:1042`): the live path evicts the pod from the pod manager and the quota manager on entry, so letting a request with no candidate nodes run through would discard a scheduled pod's accounting and still return an empty result.

Note that #2559 refactors this function and reproduces both `len(*args.NodeNames)` call sites unchanged, so the two changes will want to be reconciled.

**Environment**:
- HAMi version: current `master` (`51c593c`)
- Others: unit-reproducible, no cluster or GPU required

nishantbkl3345-ship-it:
**What type of PR is this?**

/kind bug

`Filter` takes the live path whenever `args.Nodes` is nil, and that path dereferences `*args.NodeNames` in two places — the "no available node" event message, and `genSuccessMsg`. `ExtenderArgs` can arrive with neither `Nodes` nor `NodeNames` set, and then the deref panics:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation]
	pkg/scheduler/scheduler.go:1061
```

Because the panic happens on the extender's HTTP goroutine, it terminates the whole scheduler process instead of failing the single request that caused it.

The change returns an error result when both node fields are nil. Placing the check before `TakeAndDeletePod` is deliberate: the live path evicts the pod from the pod manager and the quota manager on entry, so letting a request with no candidate nodes through would discard a scheduled pod's accounting for nothing and still end up with an empty result.

The two remaining derefs are switched to the existing `nodeNamesLen` helper, which already nil-guards and is used elsewhere in the same function, so they can't panic if another path reaches them later.

Fixes #2688

The `hasHAMiResource` early return above already handles nil `NodeNames` safely (it just echoes the field back), so only pods that actually request HAMi resources reach the crash.

#2559 refactors this same function and reproduces both `len(*args.NodeNames)` call sites unchanged. Whichever lands first, the other will need a small reconciliation — happy to rebase on top of it if that ordering is easier.

Testing:

- Added `TestFilterWithoutCandidateNodes`, which drives `Filter` with a GPU-requesting pod and empty `ExtenderArgs`. It panics at `scheduler.go:1061` on master and passes with the fix. It also asserts the pod is still in the pod manager afterwards, which is what pins the guard above `TakeAndDeletePod`.
- `go test ./pkg/scheduler/... -short --race -count=1` — pass
- `make test` — pass (all packages, race enabled)
- `make verify` — pass (golangci-lint 0 issues, license headers, import aliases)
- `make build` — pass

This is scoped to the scheduler extender and covered by unit tests, so no GPU hardware was involved.

```release-note
Fixed a scheduler extender crash when a filter request contained neither Nodes nor NodeNames.
```

---

**AI assistance disclosure**

This contribution used AI assistance (Claude Code), including code generation: the nil deref was surfaced during an AI-assisted read of the extender code, and the guard, the regression test and the initial commit message were drafted with Claude Code. I reviewed and corrected them — I confirmed the panic against master, checked the guard's placement against the pod and quota manager side effects, and ran the validation listed above. I understand the root cause and the fix and will respond to review feedback myself.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved device discovery and health-check handling when plugins report errors or no devices.
  * Corrected usage reconstruction for stale allocations and missing nodes.
  * Made device locking more reliable by using consistent ordering and safely rolling back partial locks.
  * Improved request filtering when no candidate nodes are provided, preventing errors and preserving scheduling state.
  * Enhanced retry handling and node-count reporting for more reliable scheduling behavior.

* **Tests**
  * Added coverage for filtering requests without candidate nodes.


pkg/scheduler/scheduler.go
```diff
@@ -1038,6 +1038,14 @@ func (s *Scheduler) Filter(args extenderv1.ExtenderArgs) (*extenderv1.ExtenderFi
 	if args.Nodes != nil {
 		return s.filterSimulation(args, resourceReqs)
 	}
+	if args.NodeNames == nil {
```
mesutoezdil, line 1089: 1 question: an empty non nil list drops the pod accounting the same way, with zero candidates. should this guard also cover len == 0?

archlitchi:

pkg/scheduler/scheduler_test.go
```diff
@@ -3149,6 +3149,59 @@ func Test_Bind_PodGroupPodNonContentionErrorDoesNotRetry(t *testing.T) {
 		"non-contention error must not trigger retry")
 }
 
+ fix/filter-nil-nodenames-panic
```
mesutoezdil, line 3152: your merge left the branch name inside the file, it does not compile and ci is red. clean it up. the len == 0 question above is also still open.

mesutoezdil (changes requested):

pkg/scheduler/scheduler_test.go
mesutoezdil, line 3152: this looks like the tail of a merge conflict marker, so the file does not parse. that is why lint failed and Compile was skipped.

pkg/scheduler/scheduler.go
```diff
@@ -1086,6 +1086,14 @@ func (s *Scheduler) Filter(args extenderv1.ExtenderArgs) (*extenderv1.ExtenderFi
...
+		// The live filter path has no candidate list to work with, and running it
+		// anyway would drop the pod from the scheduling caches for nothing.
+		err := fmt.Errorf("extender args contain neither Nodes nor NodeNames")
```
mesutoezdil, line 1092: getNodesUsage never reads its NodeNames parameter, and calcScore scores every node. so is rejecting right, or is nil safe enough?

pkg/scheduler/scheduler.go
```diff
@@ -1086,6 +1086,14 @@ func (s *Scheduler) Filter(args extenderv1.ExtenderArgs) (*extenderv1.ExtenderFi
...
+		klog.ErrorS(err, "Rejecting filter request", "pod", klog.KObj(args.Pod))
```
mesutoezdil, line 1093: which caller sends args with neither field set?

mesutoezdil:
closing per the issue/pr lifecycle policy: no author response in two weeks to the outstanding review comments (last review 2026-08-30), and the branch still does not compile.

PR #2689, opened 2026-08-16
fix(scheduler): reject filter requests with no Pod instead of panicking

### LINKED ISSUE #2598: Nil Pointer Dereference in PredicateRoute When Pod Is Missing
**What happened**:

**Environment**:
- HAMi version:
- Others:

### LINKED ISSUE #2599: Nil Pointer Dereference in PredicateRoute When Pod Is Missing
### Description

The scheduler extender's `/filter` endpoint can panic when it receives an `ExtenderArgs` request where the `Pod` field is missing or `null`.

In `pkg/scheduler/routes/route.go`, `PredicateRoute()` currently logs filtering errors using:

```go
klog.ErrorS(err, "Filter error for pod", "pod", extenderArgs.Pod.Name)
```

However, `extenderArgs.Pod` is a pointer and can be `nil` when the incoming request is malformed or does not contain the `Pod` field.

In this situation, accessing `extenderArgs.Pod.Name` results in a nil pointer dereference and can cause the HTTP handler to panic.

### Steps to Reproduce

Send a request to the scheduler extender's `/filter` endpoint with an `ExtenderArgs` payload that does not contain a `Pod`:

```json
{
  "NodeNames": ["node1"]
}
```

Alternatively:

```json
{
  "Pod": null,
  "NodeNames": ["node1"]
}
```

The request can cause `PredicateRoute()` to dereference a nil `Pod` while handling the filtering error.

### Expected Behavior

Malformed requests with a missing or null `Pod` should be rejected gracefully, preferably with an appropriate HTTP `400 Bad Request` response.

The scheduler extender should not panic when receiving invalid input.

### Actual Behavior

`PredicateRoute()` can dereference `extenderArgs.Pod.Name` when `extenderArgs.Pod == nil`, resulting in a nil pointer dereference.

This can cause an unhandled panic in the request handler and potentially affect scheduler availability depending on the server's recovery configuration.

### Proposed Fix

* Validate `extenderArgs.Pod` before passing the request to the filtering logic.
* Return an appropriate `400 Bad Request` response when `Pod` is missing or `nil`.
* Ensure error logging does not dereference a nil `Pod`.
* Add a regression test covering requests with a missing/null `Pod`.
* Preserve the existing behavior for valid `ExtenderArgs` requests.

### Testing

Add a unit test for `PredicateRoute()` using `httptest` that:

1. Sends an `ExtenderArgs` request without a `Pod`.
2. Verifies that the handler does not panic.
3. Verifies that an appropriate error response, such as HTTP `400`, is returned.
4. Confirms that a valid request continues to work as expected.

### Scope

This should be a minimal bug fix limited to the scheduler route validation and associated tests.

adity1raut:
**What type of PR is this?**

/kind bug

`PredicateRoute` in `pkg/scheduler/routes/route.go` decodes the incoming `ExtenderArgs` and passes them straight to `Scheduler.Filter` without checking whether `Pod` is set. `Filter`'s very first statement unconditionally reads `args.Pod.Name` for logging, so a `/filter` request with a missing or null `Pod` field panics with a nil pointer dereference before any of the handler's existing error handling ever runs.

This PR rejects such requests up front, the same way decode failures and cache-sync failures are already handled in this function: it returns a normal `ExtenderFilterResult` with a populated `Error` field instead of letting a nil `Pod` reach `Filter`.

Fixes #2599
Fixes #2598

Both linked issues are currently closed as completed, but the bug is still present on `master`. The only PR that attempted to fix it, #2680, closed without merging, and its guard only covered the `klog.ErrorS(..., extenderArgs.Pod.Name)` call in `route.go`'s error-logging branch — not the actual crash site, which is the first line of `Scheduler.Filter` itself (`pkg/scheduler/scheduler.go`). That line executes before the error-logging branch is ever reached, so guarding only the log call would not have prevented the panic. This PR instead stops a nil `Pod` from reaching `Filter` at all.

I confirmed the panic directly by calling `Scheduler.Filter` with an empty `ExtenderArgs{}` in a throwaway test before writing the fix (`runtime error: invalid memory address or nil pointer dereference`), and confirmed `Filter` has no other callers besides this route, so the route-level guard is a complete fix.

Added `TestPredicateRoute_NilPod`, covering both the "Pod" field omitted and explicitly-`null` cases, alongside the existing `TestPredicateRoute_DecodeError` / `TestPredicateRoute_CacheNotSynced` tests in the same file.

Ran locally: `go build ./...`, `go test ./pkg/scheduler/... -short --race -count=1`, `golangci-lint run ./pkg/scheduler/routes/...`, and `make verify` (license headers, staticcheck, import aliases) — all pass.

```release-note
Fixed a scheduler extender panic when the /filter endpoint received a request with a missing or null Pod field.
```

**AI assistance disclosure**:

I used Gemini to help locate this bug and draft the fix and tests; I reviewed the change, ran the verification described above, and can walk through how it works.

### DIFF
--- a/pkg/scheduler/routes/route.go
+++ b/pkg/scheduler/routes/route.go
@@ -61,6 +61,12 @@ func PredicateRoute(s *scheduler.Scheduler) httprouter.Handle {
 			extenderFilterResult = &extenderv1.ExtenderFilterResult{
 				Error: err.Error(),
 			}
+		} else if extenderArgs.Pod == nil {
+			err := fmt.Errorf("extender args missing pod")
+			klog.ErrorS(err, "Rejecting filter request with no pod")
+			extenderFilterResult = &extenderv1.ExtenderFilterResult{
+				Error: err.Error(),
+			}
 		} else {
 			synced := s.WaitForCacheSync(r.Context())
 			if !synced {
--- a/pkg/scheduler/routes/route_test.go
+++ b/pkg/scheduler/routes/route_test.go
@@ -208,6 +208,41 @@ func TestPredicateRoute_DecodeError(t *testing.T) {
 	}
 }
 
+func TestPredicateRoute_NilPod(t *testing.T) {
+	tests := []struct {
+		name string
+		body string
+	}{
+		{name: "pod field omitted", body: `{"NodeNames":["node1"]}`},
+		{name: "pod field explicitly null", body: `{"Pod":null,"NodeNames":["node1"]}`},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			req := httptest.NewRequest("POST", "/predicate", strings.NewReader(tt.body))
+			w := httptest.NewRecorder()
+
+			s := &scheduler.Scheduler{}
+			handler := PredicateRoute(s)
+
+			// Must not panic when Pod is missing/null.
+			handler(w, req, nil)
+
+			if w.Code != 200 {
+				t.Errorf("expected 200 (error reported in body, not status), got %d", w.Code)
+			}
+
+			var result extenderv1.ExtenderFilterResult
+			if err := json.Unmarshal(w.Body.Bytes(), &result); err != nil {
+				t.Fatalf("failed to unmarshal response: %v", err)
+			}
+			if result.Error == "" {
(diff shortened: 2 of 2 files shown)


adity1raut:
PTAL @mesutoezdil

archlitchi (approved):
/lgtm

PR #2715, opened 2026-08-18
feat(scheduler): export hami_resource_quota_limit gauge metric

### LINKED ISSUE #2713: bug(scheduler): hami_resource_quota_used exposes quota limit as a label, preventing utilisation queries
**What happened:** In `cmd/scheduler/metrics.go`, the `hami_resource_quota_used` descriptor carries the quota limit as a label instead of exposing it as a value:

```go
	quotaUsedDesc := prometheus.NewDesc(
		"hami_resource_quota_used",
		"resourcequota usage for a certain device",
		[]string{"namespace", "quota_name", "limit"}, nil,
	)

	sendMetric(ch, quotaUsedDesc, prometheus.GaugeValue, float64(q.Used), ns, quotaname, fmt.Sprint(q.Limit))
```

`Limit` is numeric (`pkg/device/quota.go:29`, `Limit int64`) and is stringified into a series dimension.

Two problems arise:

1. Quota utilisation cannot be expressed. PromQL has no way to use a label value as a sample value, so `used / limit` is not writable without hardcoding limits in a recording rule.
2. Changing a quota replaces the series rather than updating it. Raising a limit from 100 to 200 retires `limit="100"` and starts `limit="200"`, so graphs show a gap and alerts with a `for` duration reset. Each distinct limit also leaves a dead series behind for the retention window.

The limit should be exposed as a value so utilisation can be written directly:

```
hami_resource_quota_used / on(namespace, quota_name) hami_resource_quota_limit
```

**How to reproduce it (as minimally and precisely as possible):**

1. Create a namespace with a HAMi resource quota and schedule workloads against it.
2. Scrape the scheduler /metrics endpoint and note the `limit` label.
3. Edit the quota to a different limit and scrape again.
4. Observe the original series is gone and a new one exists, and that percentage utilisation cannot be queried without hardcoding the limit.

**Anything else we need to know?:**

Would open a PR for this issue soon.

**Environment:**

* HAMi version: master branch (`cmd/scheduler/metrics.go`)
* nvidia driver or other AI device driver version: N/A
* Docker version from docker version: N/A
* Docker command, image and tag used: N/A
* Kernel version from uname -a: N/A
* Others: quota metric exposes limit as a label rather than a value

SK8-infi:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**: In cmd/scheduler/metrics.go, resource quota limits were previously exposed only as a string label (`limit="100"`) inside hami_resource_quota_used. 

This caused two problems:
1. PromQL cannot perform arithmetic on label values, making used / limit utilization queries impossible without hardcoding limits in recording rules.
2. Updating a quota limit created a new time series and abandoned the old one.

This PR adds hami_resource_quota_limit as a dedicated numeric gauge metric with labels `[]string{"namespace", "quota_name"}`. `hami_resource_quota_used` and legacy `QuotaUsed` are kept unchanged for backward compatibility.

Fixes #2713

**Special notes for your reviewer**: Nope

disclosure: Antigravity was used to audit.

```release-note
feat(scheduler): export hami_resource_quota_limit gauge metric for PromQL utilization queries
```

## Summary by CodeRabbit

* **New Features**
  * Added a metric reporting configured resource quota limits alongside existing quota usage metrics.
  * Quota limit metrics are available in both standard and legacy metric outputs.
  * Metrics include the resource namespace, quota name, and configured limit value.
* **Bug Fixes**
  * Unconfigured quota limits continue to report usage with a limit of `0` without emitting a separate limit metric.

### DIFF
--- a/cmd/scheduler/metrics.go
+++ b/cmd/scheduler/metrics.go
@@ -281,12 +281,17 @@ func (cc ClusterManagerCollector) collectNodeMetrics(ch chan<- prometheus.Metric
 	}
 }
 
-// collectQuotaMetrics emits per-namespace resource quota usage.
+// collectQuotaMetrics emits per-namespace resource quota usage and limits.
 func (cc ClusterManagerCollector) collectQuotaMetrics(ch chan<- prometheus.Metric, legacy bool) {
 	quotaUsedDesc := prometheus.NewDesc(
 		"hami_resource_quota_used",
 		"resourcequota usage for a certain device",
 		[]string{"namespace", "quota_name", "limit"}, nil,
+	)
+	quotaLimitDesc := prometheus.NewDesc(
+		"hami_resource_quota_limit",
+		"resourcequota limit for a certain device",
+		[]string{"namespace", "quota_name"}, nil,
 	)
 	var legacyQuotaUsed *prometheus.Desc
 	if legacy {
@@ -300,6 +305,9 @@ func (cc ClusterManagerCollector) collectQuotaMetrics(ch chan<- prometheus.Metri
 		for quotaname, q := range *val {
 			if err := sendMetric(ch, quotaUsedDesc, prometheus.GaugeValue, float64(q.Used), ns, quotaname, fmt.Sprint(q.Limit)); err != nil {
 				klog.V(4).Infof("Failed to send quotaUsedDesc metric: %v", err)
+			}
+			if err := sendMetric(ch, quotaLimitDesc, prometheus.GaugeValue, float64(q.Limit), ns, quotaname); err != nil {
+				klog.V(4).Infof("Failed to send quotaLimitDesc metric: %v", err)
 			}
 			if legacy {
 				sendLegacyMetric(ch, legacyQuotaUsed, prometheus.GaugeValue, float64(q.Used), ns, quotaname, fmt.Sprint(q.Limit))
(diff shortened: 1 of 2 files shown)


cmd/scheduler/metrics.go
```diff
@@ -281,13 +281,18 @@ func (cc ClusterManagerCollector) collectNodeMetrics(ch chan<- prometheus.Metric
...
+// collectQuotaMetrics emits per-namespace resource quota usage and limits.
 func (cc ClusterManagerCollector) collectQuotaMetrics(ch chan<- prometheus.Metric, legacy bool) {
 	quotaUsedDesc := prometheus.NewDesc(
 		"hami_resource_quota_used",
 		"resourcequota usage for a certain device",
 		[]string{"namespace", "quota_name", "limit"}, nil,
 	)
+	quotaLimitDesc := prometheus.NewDesc(
+		"hami_resource_quota_limit",
+		"resourcequota limit for a certain device",
```
mesutoezdil, line 293: looks right, LimitSet gating handles the explicit zero case too. the old metric still churns series when a limit changes, do you plan a follow up to deprecate the limit label? also kind says bug, title says feat, release note says fix. pick one.

SK8-infi: made the documentation correct

I was considering about a discussion regarding deprecation of limit label. 
For a week or two I am thinking to continue this metric as it is, while I have 1-2 more candidates for changes(they would need deprecation too)...then we can deprecate all of them together. what is your take?

Should I open an tracking issue for deprecations??

archlitchi (approved):
/lgtm

PR #2716, opened 2026-08-18
fix(monitor): replace MustNewConstMetric with safe helper and track InitContainers

devGPP23:
**PROBLEM**

While reviewing the `vGPUmonitor` code for observability gaps, I found two significant issues that impact the stability and accuracy of GPU metrics:

1. **Panic Risk in Host Metrics:** The monitor currently uses `prometheus.MustNewConstMetric` to register host-level GPU memory and utilization metrics. If the underlying NVML library returns unexpected data—such as a missing, empty, or malformed UUID string—`MustNewConstMetric` will panic and crash the entire monitor process. This is the exact same dangerous pattern that was recently fixed in the scheduler codebase (#2326). 

2. **Missing GPU Observability for InitContainers:** In the `collectPodAndContainerInfo` function, the monitor only loops through `pod.Spec.Containers` to collect container-level GPU usage metrics. It completely skips `pod.Spec.InitContainers`. Since HAMi fully supports GPU allocation in init containers (which are frequently used for heavy tasks like downloading large AI models or dataset preprocessing), operators currently have a massive blind spot. They cannot see the GPU memory or utilization for these init workloads in their dashboards.

 **SOLUTION**

1. **Safe Metric Registration:** I replaced the unsafe `prometheus.MustNewConstMetric` calls with the `sendMetric` wrapper function. `sendMetric` safely handles the `NewConstMetric` error by logging it and returning, rather than crashing the entire application. This brings these two metrics in line with the rest of the codebase.

2. **InitContainer Tracking:** I updated the container iteration loop in `collectPodAndContainerInfo` to combine `pod.Spec.InitContainers` and `pod.Spec.Containers` into a single slice before iterating. This ensures that any GPU usage happening during the init phase is successfully tracked and exported to Prometheus, providing true end-to-end observability.

**AI DISCLOSURE**

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when collecting GPU memory and utilization metrics by handling metric-generation errors gracefully.
  * Expanded pod and container monitoring to include both initialization and regular containers.
  * Improved monitoring stability when metric data is incompatible or unavailable, preventing collection issues from affecting other metrics.

### DIFF
--- a/cmd/vGPUmonitor/metrics.go
+++ b/cmd/vGPUmonitor/metrics.go
@@ -322,16 +322,11 @@ func (cc ClusterManagerCollector) collectGPUMemoryMetrics(ch chan<- prometheus.M
 		return fmt.Errorf("node name environment variable %s is not set", util.NodeNameEnvName)
 	}
 
-	ch <- prometheus.MustNewConstMetric(
-		hostGPUdesc,
-		prometheus.GaugeValue,
-		float64(memory.Used),
-		nodeName, fmt.Sprint(index), uuid, deviceName,
-	)
+	if err := sendMetric(ch, hostGPUdesc, prometheus.GaugeValue, float64(memory.Used), nodeName, fmt.Sprint(index), uuid, deviceName); err != nil {
+		klog.Errorf("Failed to send hostGPUdesc metric: %v", err)
+	}
 
-	sendLegacyMetric(ch, legacyHostGPUdesc, prometheus.GaugeValue, float64(memory.Used),
-		nodeName, fmt.Sprint(index), uuid, deviceName,
-	)
+	sendLegacyMetric(ch, legacyHostGPUdesc, prometheus.GaugeValue, float64(memory.Used), nodeName, fmt.Sprint(index), uuid, deviceName)
 
 	return nil
 }
@@ -359,16 +354,11 @@ func (cc ClusterManagerCollector) collectGPUUtilizationMetrics(ch chan<- prometh
 
 	deviceName = "NVIDIA-" + deviceName
 
-	ch <- prometheus.MustNewConstMetric(
-		hostGPUUtilizationdesc,
-		prometheus.GaugeValue,
-		float64(utilRates.Gpu),
-		nodeName, fmt.Sprint(index), uuid, deviceName,
-	)
+	if err := sendMetric(ch, hostGPUUtilizationdesc, prometheus.GaugeValue, float64(utilRates.Gpu), nodeName, fmt.Sprint(index), uuid, deviceName); err != nil {
+		klog.Errorf("Failed to send hostGPUUtilizationdesc metric: %v", err)
+	}
 
-	sendLegacyMetric(ch, legacyHostGPUUtilizationdesc, prometheus.GaugeValue, float64(utilRates.Gpu),
-		nodeName, fmt.Sprint(index), uuid, deviceName,
-	)
+	sendLegacyMetric(ch, legacyHostGPUUtilizationdesc, prometheus.GaugeValue, float64(utilRates.Gpu), nodeName, fmt.Sprint(index), uuid, deviceName)
 
 	if err := sendMetric(ch, hostGPUMemoryUtilizationdesc, prometheus.GaugeValue,
 		float64(utilRates.Memory),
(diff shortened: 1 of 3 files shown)


cmd/vGPUmonitor/metrics_extra_test.go
```diff
@@ -0,0 +1,184 @@
...
+	}
+	close(ch)
+
+	// Drain the channel and count emitted metrics
+	count := 0
+	for range ch {
+		count++
+	}
+	// Each container with 1 device should emit at least 7 standard metrics
+	if count < 2 {
```
mesutoezdil: this passes without the fix. the regular container alone emits about 7 metrics, so count is never below 2. assert a metric labeled init-c1 instead, right now this pins nothing.

cmd/vGPUmonitor/metrics_extra_test.go
```diff
@@ -0,0 +1,184 @@
...
+func (m *mockUsageInfo) LastKernelTime() int64          { return 0 }
+func (m *mockUsageInfo) GetPriority() int               { return 0 }
+func (m *mockUsageInfo) GetRecentKernel() int32         { return 0 }
+func (m *mockUsageInfo) SetRecentKernel(_ int32)        {}
+func (m *mockUsageInfo) GetUtilizationSwitch() int32    { return 0 }
+func (m *mockUsageInfo) SetUtilizationSwitch(_ int32)   {}
+
+// setContainerMap uses reflection to inject mock containers into an unexported field.
+func setContainerMap(lister *nvidia.ContainerLister, m map[string]*nvidia.ContainerUsage) {
+	v := reflect.ValueOf(lister).Elem().FieldByName("containers")
```
mesutoezdil: unsafe reflection into a private field of another package is fragile. add an exported test hook in pkg/monitor/nvidia instead.

cmd/vGPUmonitor/metrics.go
```diff
@@ -315,12 +315,10 @@ func (cc ClusterManagerCollector) collectGPUMemoryMetrics(ch chan<- prometheus.M
 		return fmt.Errorf("node name environment variable %s is not set", util.NodeNameEnvName)
 	}
 
-	ch <- prometheus.MustNewConstMetric(
-		hostGPUdesc,
-		prometheus.GaugeValue,
-		float64(memory.Used),
-		nodeName, fmt.Sprint(index), uuid, deviceName,
-	)
+	if err := sendMetric(ch, hostGPUdesc, prometheus.GaugeValue, float64(memory.Used), nodeName, fmt.Sprint(index), uuid, deviceName); err != nil {
```
mesutoezdil, line 325: fine as consistency cleanup, but the panic premise is overstated. an empty uuid is a legal label value, NewConstMetric only fails on bad utf8 or a wrong label count.

mesutoezdil (dismissed):

devGPP23:
Thanks for the review @mesutoezdil !
 I've addressed all three points:

1)  Replaced the fragile count < 2 check with a label-specific assertion that scans for "init-c1" in the emitted metrics. This correctly validates the init-container tracking.
2) Removed the unsafe/reflect usage in the tests. I added an exported test hook (SetContainersForTest) directly in pkg/monitor/nvidia/cudevshr.go for cross-package test visibility.
3) an empty UUID is a valid Prometheus label, and NewConstMetric only fails on bad UTF-8 or incorrect label counts. I've noted this, but kept the switch to sendMetric as a consistency cleanup.
Please take a look when u r free

cmd/vGPUmonitor/metrics.go
```diff
@@ -315,12 +315,10 @@ func (cc ClusterManagerCollector) collectGPUMemoryMetrics(ch chan<- prometheus.M
...
+		klog.Errorf("Failed to send hostGPUdesc metric: %v", err)
```
mesutoezdil, line 326: this error is logged here and again by collectGPUInfo. one log is enough. also the early return drops the utilization metric for the same device. intended?

pkg/monitor/nvidia/cudevshr.go
```diff
@@ -161,6 +161,11 @@ func (l *ContainerLister) Clientset() *kubernetes.Clientset {
 	return l.clientset
 }
 
+// SetContainersForTest replaces the internal container map; use only in tests.
+func (l *ContainerLister) SetContainersForTest(m map[string]*ContainerUsage) {
```
mesutoezdil, line 165: take l.mutex here too. cheap and keeps the map access rule in one place.

mesutoezdil (commented):
all three earlier points are addressed. the test now pins init-c1 and the reflection is gone. two small things inline.

devGPP23:
Thanks for the review @mesutoezdil 
I've addressed both the inline comments in my latest commits:

1) In cmd/vGPUmonitor/metrics.go, I removed the redundant error logging and the early return. We're now calling sendLegacyMetric directly so that no utilization metrics are dropped.

2 ) In pkg/monitor/nvidia/cudevshr.go, I added the l.mutex.Lock() to SetContainersForTest as suggested, to keep the map access thread-safe.

FouoF (approved):

PR #2721, opened 2026-08-18
fix(scheduler): honour HAMI_NODELOCK_EXPIRE when the flag is not set

### LINKED ISSUE #2692: scheduler.nodeLockExpire / HAMI_NODELOCK_EXPIRE has no effect: overwritten by --node-lock-timeout flag default
## Summary

flag is passed explicitly.

## Version

## Reproduction

```bash
```

```bash
# 5s
```

```bash
  | grep -i "node lock"
```

```
```

```
```

```yaml
scheduler:
  extender:
    extraArgs:
      - --node-lock-timeout=5s
```

```
```

## Cause

```go

func setupNodeLockTimeout() {
	if nodelock != "" {
		...
		NodeLockTimeout = d
	}
}
```

```go
// line 79

// line 114
```

## Suggested fix

```go
}
```

## Why it matters

`binpack`/`spread` policies:

IIITManjeet:
**What type of PR is this?**

/kind bug

The Helm chart exposes `scheduler.nodeLockExpire`, which sets the `HAMI_NODELOCK_EXPIRE`
environment variable on the scheduler extender. `pkg/util/nodelock` reads that variable in
its package `init()` and sets `NodeLockTimeout` from it.

`start()` then ran, unconditionally:

```go
```

`config.NodeLockTimeout` is populated by the `--node-lock-timeout` flag, which carries a
5m default whether or not the flag is passed. The environment-derived value was therefore
always discarded a few milliseconds after being set, and `scheduler.nodeLockExpire` had no
effect at all. The chart's own default of `5m` masked this, so the setting only looks
broken once you change it.

This takes the flag only when it was explicitly provided, leaving precedence as flag, then
environment, then default.

Fixes #2692

The issue suggests calling `rootCmd.Flags().Changed("node-lock-timeout")` from inside
`start()`. That does not compile:

```
cmd/scheduler/main.go:52:2: initialization cycle for rootCmd
        rootCmd refers to start
        start refers to rootCmd
```

`rootCmd`'s `RunE` closure already refers to `start`, so `start` cannot refer back to
`rootCmd`. This PR reads the flag state in `RunE`, where cobra already supplies the
`*cobra.Command`, and passes it into `start()` as a parameter.

I verified behaviour by building the scheduler binary from `master` and from this branch
and reading its own startup log. "Set node lock timeout" is the effective value:

| scenario                                              | before | after |
| ----------------------------------------------------- | ------ | ----- |
| `HAMI_NODELOCK_EXPIRE=7s`, no flag                    | 5m0s   | 7s    |
| `HAMI_NODELOCK_EXPIRE=7s` + `--node-lock-timeout=33s` | 33s    | 33s   |
| neither set                                           | 5m0s   | 5m0s  |
| `HAMI_NODELOCK_EXPIRE=notaduration`                   | 5m0s   | 5m0s  |

Only the first row changes. `TestApplyNodeLockTimeout` covers the three precedence paths.

Per the hardware-validation gate in CONTRIBUTING, this change is scoped to the scheduler
extender and is covered by unit tests rather than GPU hardware.

Checks run locally (Go 1.26.6, linux/amd64):

- `go test -short --race ./pkg/... ./cmd/...` — 37 packages ok, 0 failures
- `golangci-lint run ./...` (v2.12.2, matching CI)
- `hack/verify-license.sh`, `hack/verify-import-aliases.sh`, `hack/verify-rbac.sh`
- `go mod tidy` — no change to `go.mod` or `go.sum`

#2692 also asks whether `HAMI_NODELOCK_EXPIRE` should remain at all, or whether the chart
should set `--node-lock-timeout` instead. I kept this PR to the precedence bug and left the
chart untouched, since that is an interface decision for the maintainers. Happy to follow
up either way.

```
The scheduler's `scheduler.nodeLockExpire` Helm value (`HAMI_NODELOCK_EXPIRE`) now takes effect. Previously it was silently overwritten by the `--node-lock-timeout` default of 5m unless that flag was passed explicitly.
```

AI assistance disclosure: I used AI assistance (Claude Code) throughout this change: to understand the nodelock init path and how config.NodeLockTimeout is populated, to find that the fix suggested in the issue does not compile, to implement applyNodeLockTimeout, and to write cmd/scheduler/nodelock_timeout_test.go. It also ran the before/after binary comparison and the local checks listed above, and drafted the commit message and this description. I reviewed the change, understand the initialization-cycle constraint that shapes it, and am able to answer questions about it directly.

## Summary by CodeRabbit

* **Bug Fixes**
  * Node-lock timeout settings now use the configured default when no command-line override is provided.
  * Command-line timeout values continue to take precedence when explicitly specified.
* **Tests**
  * Updated coverage to verify that the node-lock timeout flag and scheduler configuration share the correct default value.

### DIFF
--- a/cmd/scheduler/main.go
+++ b/cmd/scheduler/main.go
@@ -54,7 +54,7 @@ var (
 		Short: "kubernetes vgpu scheduler",
 		RunE: func(cmd *cobra.Command, args []string) error {
 			flag.PrintPFlags(cmd.Flags())
-			return start()
+			return start(cmd.Flags().Changed("node-lock-timeout"))
 		},
 	}
 )
@@ -109,9 +109,18 @@ func injectProfilingRoute(router *httprouter.Router) {
 	})
 }
 
-func start() error {
-	// Initialize node lock timeout from config
-	nodelock.NodeLockTimeout = config.NodeLockTimeout
+// applyNodeLockTimeout applies the flag only when it was set explicitly, so it
+// does not overwrite the value nodelock's init took from HAMI_NODELOCK_EXPIRE.
+func applyNodeLockTimeout(flagSet bool) {
+	if flagSet {
+		nodelock.NodeLockTimeout = config.NodeLockTimeout
+		return
+	}
+	config.NodeLockTimeout = nodelock.NodeLockTimeout
+}
+
+func start(nodeLockTimeoutFlagSet bool) error {
+	applyNodeLockTimeout(nodeLockTimeoutFlagSet)
 	klog.InfoS("Set node lock timeout", "timeout", nodelock.NodeLockTimeout)
 	client.InitGlobalClient(
 		client.WithBurst(config.Burst),
--- a/cmd/scheduler/main.go
+++ b/cmd/scheduler/main.go
@@ -54,7 +54,7 @@ var (
 		Short: "kubernetes vgpu scheduler",
 		RunE: func(cmd *cobra.Command, args []string) error {
 			flag.PrintPFlags(cmd.Flags())
-			return start(cmd.Flags().Changed("node-lock-timeout"))
+			return start()
 		},
 	}
 )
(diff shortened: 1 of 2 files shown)


cmd/scheduler/main.go
```diff
@@ -109,9 +109,18 @@ func injectProfilingRoute(router *httprouter.Router) {
...
-func start() error {
-	// Initialize node lock timeout from config
-	nodelock.NodeLockTimeout = config.NodeLockTimeout
+// applyNodeLockTimeout applies the flag only when it was set explicitly, so it
+// does not overwrite the value nodelock's init took from HAMI_NODELOCK_EXPIRE.
+func applyNodeLockTimeout(flagSet bool) {
+	if flagSet {
+		nodelock.NodeLockTimeout = config.NodeLockTimeout
+		return
+	}
```
mesutoezdil: 1 alternative: make the flag default nodelock.NodeLockTimeout in init, the env is already applied there by import order, then no bool threading is needed. also nothing else reads config.NodeLockTimeout, is the write back needed?

IIITManjeet: Good call on both. Done.

Flag default is now `nodelock.NodeLockTimeout` — `main` imports `pkg/util/nodelock`, so its `init()` (and `setupNodeLockTimeout`, which reads `HAMI_NODELOCK_EXPIRE`) completes before this package registers flags. The default already carries the env value, so `start()` goes back to the plain unconditional assignment and `applyNodeLockTimeout`, the bool param and the `Changed(...)` call are gone. The write-back went too — as you say, nothing outside `cmd/scheduler` reads `config.NodeLockTimeout`.

Side benefit: `PrintPFlags` runs before `start()`, so the startup log used to print `FLAG: --node-lock-timeout="5m0s"` while the effective value was 7s. They agree now.

Rebuilt at both revisions and compared startup logs — `Set node lock timeout` is unchanged in all four cases (env only 7s, env+flag 33s, neither 5m0s, bad env 5m0s).

One visible trade-off: `--help` now shows an env-dependent default (7s when the chart sets `scheduler.nodeLockExpire: 7s`). I think showing the real default is better, but happy to pin it back if you'd rather.

`TestApplyNodeLockTimeout` had no function left to cover, so it's now `TestNodeLockTimeoutFlagDefault`, asserting the registered default is `nodelock.NodeLockTimeout` — re-hardcoding a duration there is exactly how #2692 comes back. Env parsing is already covered in `pkg/util/nodelock/nodelock_test.go`.

mesutoezdil:

IIITManjeet:
Hey @mesutoezdil, apologies for breaking the contribution guidelines. Should I reframe it with my own wording for the reply and commit with a fresh PR or reopen it again with my updated reply.

mesutoezdil:
you MUST understand your codes, and should answer with your own words. 
We do not have time and energy for llm answers.

IIITManjeet:
yes @mesutoezdil  agreed actually just asked llm to make it more professional like answer. Again, apologies for the llm drafted answer. If there is anything that I can help to undo my act, do let me know.
Thanks

Issue #2563, opened 2026-08-11
[Feature] Support Kubernetes native sidecar containers (initContainers with restartPolicy: Always).


maishivamhoo123:
Kubernetes sidecars (restartPolicy: Always init containers) run for the whole pod lifetime, but HAMi treats them as normal init containers and drops their gpumem after the init phase, so the card can get oversubscribed.
Fix: sum them with app containers, never release.
Extends #1773.

Tasks:
- [x] Design doc for sidecar support.
- [x] Implementation + tests once the design is approved.

maishivamhoo123:
@coderabbitai plan

maishivamhoo123:
#2723 completed this issue .

PR #2723, opened 2026-08-18
feat(scheduler): account native sidecar container GPU resources

maishivamhoo123:
**What type of PR is this?**
/kind feature

This PR implements sidecar GPU accounting per design #2584 . A sidecar is an init container with restartPolicy: Always. Unlike normalinit containers, it never exit, it keeps running parallely with the app containers. So its GPU memory must be added on
top,
                           effectiveUsage =  max(init_peak, app_sum+ sidecar_sum )
Before this PR, HAMi treated sidecars like normal init containers, so their GPU usage was under counted and the card could be oversubscribed. Now sidecars are summed with app containers everywhere usage collapse, scheduler scoring, the post-init shrink, and webhook quota admission.
this pr also changed the shrink now shrink waits only for the NON-sidecar init containers to finish, and the shrunk usage keeps the sidecar's share. A crash-looping sidecar briefly showing "exited 0" can no longer trigger an early shrink.

Part  of : #2563 

Ai Disclosure : I used AI for writing the test , optimizing the code logic and understanding the code behaviour.

## New Features
* Added support for restartable sidecar init containers in device allocation, usage tracking, and resource quota calculations.
* Resource shrinking now occurs after all non-sidecar init containers complete.
* Added accurate GPU slot accounting for workloads with multiple application containers.

## Bug Fixes
* Improved per-device usage calculations across sidecars, init containers, and applications.
* Preserved correct allocation behavior for mixed and interleaved container workloads.
* Improved device lock handling and scheduling reliability.
* Added clearer handling for stale device allocations and missing usage data.

### DIFF
--- a/pkg/scheduler/score.go
+++ b/pkg/scheduler/score.go
@@ -223,7 +223,7 @@ func applyPeakUsage(node *NodeUsage, appNodeCopy *NodeUsage, peakUsage map[strin
 	}
 }
 
-// fitInDevices keys rows by request type ("NVIDIA") while nodes may register model-specific types ("NVIDIA A100-SXM4-40GB"); union both so padding/merge never drops fitted rows.
+// fitInDevices keys rows by request type ("NVIDIA") while nodes may register model-specific types ("NVIDIA A100-SXM4-40GB"); union both so padding never drops fitted rows.
 func allocationTypeKeys(resourceReqs device.PodDeviceRequests, baseTypes map[string]struct{}) map[string]struct{} {
 	allocTypes := make(map[string]struct{}, len(baseTypes))
 	for typ := range baseTypes {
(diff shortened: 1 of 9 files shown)


maishivamhoo123:
I tested all my changes on the fake GPU environment  and added the output here in this file please check and verify. https://docs.google.com/document/d/1zxaFeG75w7p3QGWy3BazjoXXbPzdmq7RYPvGonzVLLs/edit?usp=sharing 
Thank you!

pkg/scheduler/score.go
```diff
@@ -205,28 +205,79 @@ func applyPeakUsage(node *NodeUsage, appNodeCopy *NodeUsage, peakUsage map[strin
 	}
 }
 
-func allocateInitContainers(node *NodeUsage, nodeID string, resourceReqs device.PodDeviceRequests, task *corev1.Pod, nodeInfo *device.NodeInfo, baseTypes map[string]struct{}, numInitContainers int, peakUsage map[string]peakUsageSnapshot, weights util.DeviceScoringWeights) (device.PodDevices, bool, string) {
+func allocationTypeKeys(resourceReqs device.PodDeviceRequests, baseTypes map[string]struct{}) map[string]struct{} {
```
mesutoezdil, line 227: when can a request type that passed fit be missing from baseTypes? this helper looks like a no-op. if it covers a real edge, name it in a comment, else drop it. also the body says /feature, the bot needs /kind feature.

maishivamhoo123: Thank you @mesutoezdil for review! it is not a no-op because the two key sets genuinely differ on real nodes. the device plugin registers GPU under the model string like NVIDIA A100-SXM4-40GB, but the requests use  NVIDIA. Fit still passes because getNodeResources matches by substring, but the fitted rows are keyed under NVIDIA, which is not in the baseTypes. and the init-range merge iterates only the model-name key and drops those rows the pod annotation ends up with one segment instead of one per container, and the device plugin then fails admission with no pending device allocation found. i got this bug while testing this  on nvml-mock, the helper is the fix. Will make the comment on it.
Thank you!

mesutoezdil: verified, the key sets do differ, node types carry the model string. the code comment covers it. resolved.

mesutoezdil:
resolve conflicts pls

maishivamhoo123:
> resolve conflicts pls

Sure!

pkg/scheduler/webhook.go
```diff
@@ -164,20 +164,25 @@ func fitResourceQuota(pod *corev1.Pod) bool {
...
+				sidecarCoresReq += int64(req.Coresreq) * int64(req.Nums)
+				continue
+			}
 			initMemoryReq = max(initMemoryReq, int64(req.Memreq)*int64(req.Nums))
 			initCoresReq = max(initCoresReq, int64(req.Coresreq)*int64(req.Nums))
 		}
 
-		memoryReq := max(appMemoryReq, initMemoryReq)
-		coresReq := max(appCoresReq, initCoresReq)
+		memoryReq := sidecarMemoryReq + max(appMemoryReq, initMemoryReq)
```
DSFans2014: see the case as below:
`memoryReq` should be 8000 rather than  8000 + 5000
```
spec:
  initContainers:
  - name: test1
    ...
    resources:
      limits:
        nvidia.com/gpu: "1"
        nvidia.com/gpumem: "5000"
  - name: test2
    ...
    restartPolicy: Always
    resources:
      limits:
        nvidia.com/gpu: "1"
        nvidia.com/gpumem: "8000"
```

DSFans2014 (commented):
/lgtm

archlitchi (approved):
/lgtm

PR #2724, opened 2026-08-19
fix(scheduler): allow multiple MIG instances on one GPU

### LINKED ISSUE #2403: Does hami support a pod requesting two MIG instances on the same GPU card?
**Please provide an in-depth description of the question you have**:
On my node, I have one NVIDIA A100 GPU that can be partitioned into seven 1g.5gb MIG instances. I tried using the dynamic-mig mode to allocate two MIG instances to a pod, setting the limits as nvidia.com/gpu: 2and nvidia.com/gpumem: 4000. In this case, the pod stays in a Pending state. However, when nvidia.com/gpu: 1, the dynamic-mig mode works fine. Is this behavior expected? Does hami support a pod requesting two MIG instances on the same GPU card?
**What do you think about this question?**:
Is this a bug or not ?
**Environment**: 1 NVIDIA A100-PCIE-40GB
- HAMi version:2.9.0
- Kubernetes version:1.29.6
- Others:

saiyam1814:
**What type of PR is this?**

/kind bug

A Dynamic MIG request for two instances on a node with one physical GPU was rejected before NVIDIA's MIG placement logic ran. The scheduler compared the logical request count with the physical GPU count and returned `NodeInsufficientDevice`.

This change lets Dynamic MIG requests reach the existing profile and placement checks. Non-MIG requests keep the physical-device-count check.

The tests cover two legal placements on one GPU, insufficient MIG placement capacity, and the non-MIG regression path.

Fixes #2403

**Testing**:

```text
go test ./pkg/scheduler/... -short -race -count=1
make verify
```

### RTX PRO 6000 before/after proof

Validated on August 19, 2026:

- 8 x NVIDIA RTX PRO 6000 Blackwell Server Edition
- driver 610.43.02
- Kubernetes v1.35.6
- one target physical GPU: `GPU-30512c46-708b-f374-5698-ee24be6cd626`

The same Pod requested:

```yaml
nvidia.com/gpu: 2
nvidia.com/gpumem: 8000
```

**Before the fix**, using commit `634bf2b32e68`, the Pod remained unbound and the scheduler emitted:

```text
0/1 nodes are available: 1 NodeInsufficientDevice
```

No MIG instance was created.

**With this PR**, using image commit `a1879dd`, the same Pod became Ready and received two distinct `1g.24gb` instances on the same parent GPU:

```text
device 0: placement start=9 size=3, MIG-bee9dda1-d43e-5a20-8f78-49fde08dd64f
device 1: placement start=6 size=3, MIG-6c1bc072-8b8f-50c9-8f10-90c689a01339
```

`vectorAdd` passed independently with `CUDA_VISIBLE_DEVICES=0` and `CUDA_VISIBLE_DEVICES=1` inside that container.

A second Pod requesting two devices also became Ready and occupied starts 3 and 0, filling all four legal `1g.24gb` placements. A fifth request stayed Pending with `CardInsufficientMemory`. Deleting the first two-device Pod reclaimed both of its MIG UUIDs while the second Pod retained both UUIDs and passed CUDA again on both devices.

Final marker:

```text
ALL_MULTI_MIG_HARDWARE_PROOFS_PASSED
```

The node was restored after the test: all eight GPUs are back in non-MIG mode, `nvidia.com/gpu` allocatable is 8, the NVIDIA device plugin is Ready, four Kubernetes model servers are Ready, and the accepted Docker model server is running on GPU 2. Five GPU workloads are active on five unique physical GPUs.

```release-note
Fixed Dynamic MIG scheduling when one Pod requests multiple MIG instances from a single physical GPU.
```

**AI assistance disclosure**:

AI assistance was used while exploring the code path and drafting tests. I reviewed the implementation and ran all listed unit, repository, CI, and real-hardware checks.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved scheduling for Multi-Instance GPU (MIG) requests, allowing multiple placements on a single physical GPU when capacity permits.
  * Preserved clear failure reporting when available MIG capacity is insufficient.
  * Maintained physical GPU count validation for standard, non-MIG GPU requests.


pkg/scheduler/mig_allocation_test.go
```diff
@@ -108,3 +112,109 @@ func TestGetNodesUsageRestoresUnconsumedMigReservations(t *testing.T) {
...
+	if !fit {
+		t.Fatalf("fitInDevices() = false, reason = %q; NVIDIA MIG Fit() accepted the request", reason)
+	}
+	got := allocated[nvidia.NvidiaGPUDevice]
+	if len(got) != 1 || len(got[0]) != 2 || got[0][0].UUID != "GPU-a" || got[0][1].UUID != "GPU-a" {
+		t.Fatalf("allocated devices = %+v, want two MIG placements on GPU-a", got)
+	}
+	firstPlacement, firstOK := got[0][0].CustomInfo[nvidia.MigPlacementCustomInfo].(device.MigPlacement)
+	secondPlacement, secondOK := got[0][1].CustomInfo[nvidia.MigPlacementCustomInfo].(device.MigPlacement)
+	if !firstOK || !secondOK || firstPlacement != (device.MigPlacement{Start: 6, Size: 1}) || secondPlacement != (device.MigPlacement{Start: 5, Size: 1}) {
```
mesutoezdil: asserting the exact starts 6 and 5 pins the internal placement order of Fit. a harmless ordering change breaks this test with no real regression. assert two distinct placements on GPU-a instead.

saiyam1814: Fixed in `d71ab71`. The test now verifies that both placement values are present and distinct without asserting their order or exact start values. I also added direct table coverage for each `isMIGRequest` branch.

pkg/scheduler/score.go
```diff
@@ -50,6 +51,23 @@ func getNodeResources(list NodeUsage, t string) []*device.DeviceUsage {
 	return l
 }
 
+func isMIGRequest(request device.ContainerDeviceRequest, devices []*device.DeviceUsage, pod *corev1.Pod) bool {
+	if request.Type != nvidia.NvidiaGPUDevice {
+		return false
+	}
+	if pod != nil {
+		if mode, ok := pod.GetAnnotations()[nvidia.AllocateMode]; ok && !strings.Contains(mode, nvidia.MigMode) {
+			return false
```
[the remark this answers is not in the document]
saiyam1814: I am not changing this key. `nvidia.com/vgpu-mode` is the existing public workload API and predates this PR; `nvidia.AllocateMode` is also used by the NVIDIA backend's normal mode filter. It is documented in the Dynamic MIG implementation and migration guide, used by the shipped example, and referenced throughout the website. Renaming it here would break existing workload manifests and turn this focused scheduler bug fix into an incompatible API migration. The new code intentionally reads the established constant rather than introducing another annotation key.

saiyam1814:
Review update in `d71ab71`:

- placement assertions are now order-independent;
- `isMIGRequest` has direct coverage for non-NVIDIA requests, explicit HAMi-core mode, MIG with and without a Pod, and non-MIG device state;
- the established `nvidia.com/vgpu-mode` API remains unchanged for compatibility.

Focused tests and the full scheduler race suite pass, along with `git diff --check` and `make verify`. The earlier RTX before/after hardware proof remains valid because this follow-up changes tests only.

archlitchi (approved):
/lgtm

PR #2733, opened 2026-08-19
fix(nodelock): serialize expired-lock recovery

### LINKED ISSUE #2681: fix(nodelock): serialize expired-lock recovery to prevent concurrent lock acquisition
### What happened

`LockNode()` can recover an expired or dangling node lock by first calling `ReleaseNodeLock()` and then `SetNodeLock()`.

Although these operations individually use the per-node mutex, the entire recovery sequence is not protected as one atomic operation. This creates a race when multiple schedulers attempt to recover the same node lock concurrently.

### Race scenario

Consider a node whose existing lock has expired:

Scheduler A                    Scheduler B
-----------                    -----------

LockNode(node)
  ↓
sees expired lock
                               LockNode(node)
                                 ↓
                               sees expired lock

ReleaseNodeLock()
  ↓
old lock removed

SetNodeLock(A)
  ↓
A acquires lock
                               ReleaseNodeLock()
                                 ↓
                               A's newly acquired lock removed

                               SetNodeLock(B)
                                 ↓
                               B acquires lock

A → LockNode() returns success
B → LockNode() returns success

In this situation, both callers can believe they successfully acquired the node lock, while the lock established by the first caller has already been replaced by the second caller.

### Why this matters

The node lock is intended to serialize scheduler operations on a node. If a caller loses its lock immediately after `LockNode()` reports success, it can continue scheduling/allocation work while another scheduler is operating on the same node.

This can undermine the synchronization guarantees provided by the node-lock mechanism.

### Current behavior

The expired/dangling recovery path effectively performs:

    ReleaseNodeLock(...)
    SetNodeLock(...)

as two separately synchronized operations.

The per-node mutex protects each operation individually, but does not protect the complete check → release → reacquire transaction.

### Expected behavior

Expired or dangling lock recovery should be serialized per node so that:

- only one scheduler can recover a particular node lock at a time;
- a scheduler cannot release a lock that another scheduler has just acquired;
- `LockNode()` only returns success when the caller has a valid lock;
- concurrent callers either observe lock contention or safely recover the lock without invalidating another caller's ownership.

### Suggested regression test

Add a deterministic concurrency test that:

1. Creates a node with an expired lock.
2. Starts two concurrent `LockNode()` calls for different owners.
3. Ensures both callers initially observe the expired lock.
4. Exercises the recovery path concurrently.
5. Verifies that the two callers cannot both successfully complete recovery while replacing each other's locks.
6. Verifies that exactly one caller owns the final node lock.

A similar test should cover recovery of a dangling lock where the previous owner pod no longer exists.

### Impact

This is a concurrency correctness issue in the node-lock recovery path. Under concurrent scheduling activity during expired/dangling lock recovery, the current implementation can potentially allow multiple callers to proceed as though they own the same node lock.

The fix should preserve the existing lock behavior while making the expired/dangling recovery operation atomic with respect to other lock attempts on the same node.

AyushSrivastava1818:
## What type of PR is this?

/kind bug

`LockNode()` can recover expired or dangling node locks by releasing the existing lock and acquiring a new one. Previously, the state check, stale-lock release, and new lock acquisition were not serialized as one atomic operation.

Although `SetNodeLock()` and `ReleaseNodeLock()` individually used the per-node mutex, multiple concurrent `LockNode()` callers could observe the same expired or dangling lock before either recovery completed.

This created a race where one scheduler could acquire a recovered lock and have that lock subsequently removed by another concurrent recovery attempt. Both callers could return success even though only the final lock remained valid.

This PR serializes the complete expired/dangling lock recovery transaction per node.

### Changes

- Protect the complete `LockNode()` state inspection and recovery flow with the existing per-node mutex.
- Extract internal locked helpers for setting and releasing node locks to avoid recursive mutex acquisition.
- Re-check the latest node lock state after waiting for a competing recovery operation.
- Ensure only one concurrent caller can successfully recover an expired or dangling lock.
- Preserve existing same-owner reentrancy behavior.
- Preserve contention behavior for valid unexpired locks.
- Preserve concurrency between locks on different nodes.
- Reject nil pod arguments before any lock state is inspected or modified.
- Add deterministic synchronization to the recovery concurrency test so it verifies that a competing caller has actually reached the mutex boundary.

Fixes #2681

The fix intentionally uses the existing per-node `nodeLockManager` rather than introducing a new synchronization mechanism.

The mutex is held for the complete `LockNode()` operation on the affected node, including:

1. Reading the current node lock state.
2. Checking expiration or dangling ownership.
3. Releasing a stale lock when necessary.
4. Acquiring the replacement lock.
5. Returning the final result.

The internal `setNodeLockLocked` and `releaseNodeLockLocked` helpers assume the caller already holds the per-node mutex. The existing public `SetNodeLock()` and `ReleaseNodeLock()` APIs continue to acquire the mutex themselves.

This prevents recursive locking while preserving the existing API behavior.

Concurrent callers targeting different nodes continue to use independent mutexes and are not unnecessarily serialized.

Nil pod validation is performed at the beginning of `LockNode()` and `SetNodeLock()`, before any node state or lock annotation can be modified.

### Regression coverage

The tests cover:

- Concurrent expired-lock recovery with exactly one successful owner.
- Concurrent dangling-lock recovery with exactly one successful owner.
- Preservation of an existing valid unexpired lock.
- Deterministic serialization of the release → acquire recovery sequence.
- Verification that a competing caller has reached the per-node mutex before the recovery sequence is released.
- Retryable contention behavior in `SetNodeLock()`.
- Same-pod reentrancy behavior.
- Concurrent locking of different nodes.
- Nil pod passed to `LockNode()` with an expired lock.
- Nil pod passed to `SetNodeLock()`.

## Validation

The following checks were completed:

- `go test ./pkg/util/nodelock/...` — passed
- `go test -v -run 'LockNode|NodeLock' -count=20` — passed, 20/20 iterations
- `go vet ./pkg/util/nodelock/...` — passed
- `golangci-lint run ./pkg/util/nodelock/...` — passed
- `gofmt` — passed
- `git diff --check` — passed

The race detector was attempted but is blocked by the local Windows CGO compiler environment (`cc1.exe: 64-bit mode not compiled in`). The normal concurrency test suite and 20-run stress test passed successfully.

No.

This is a concurrency correctness fix in the internal node-lock recovery path. Nil pod arguments are now rejected safely before any lock mutation instead of potentially reaching a recovery path that could dereference the nil pod.

## AI Assistance Disclosure

AI assistance from Claude and Antigravity was used during codebase investigation, root-cause analysis, implementation assistance, regression-test development, concurrency/race analysis, review-comment analysis, and validation of this change.

The resulting implementation and tests were reviewed against the existing HAMi codebase and node-lock semantics, including same-owner reentrancy, lock contention, retry behavior, nil-pod handling, and per-node concurrency. The final changes were validated by the contributor using the project's unit tests, concurrency stress tests, vet, lint, formatting, and diff-check workflows.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved node-lock handling to prevent recursive locking and race conditions during recovery.
  * Correctly recovers expired or dangling locks while preserving valid, unexpired locks.
  * Rejects nil pod requests immediately without changing existing lock state.
  * Ensures concurrent lock attempts are serialized consistently and safely.

* **Tests**
  * Added coverage for concurrent recovery, deterministic lock serialization, valid-lock preservation, and invalid pod requests.
  * Added verification that recovery behavior remains stable when multiple lock attempts occur simultaneously.

### DIFF
--- a/pkg/util/nodelock/nodelock.go
+++ b/pkg/util/nodelock/nodelock.go
@@ -132,6 +132,10 @@ func SetNodeLock(nodeName string, lockname string, pods *corev1.Pod) error {
 	nodeLock.Lock()
 	defer nodeLock.Unlock()
 
+	return setNodeLockLocked(nodeName, lockname, pods)
+}
+
+func setNodeLockLocked(nodeName string, lockname string, pods *corev1.Pod) error {
 	ctx := context.Background()
 	node, err := client.GetClient().CoreV1().Nodes().Get(ctx, nodeName, metav1.GetOptions{})
 	if err != nil {
@@ -181,6 +185,10 @@ func ReleaseNodeLock(nodeName string, lockname string, pod *corev1.Pod, skipNode
 	nodeLock.Lock()
 	defer nodeLock.Unlock()
 
+	return releaseNodeLockLocked(nodeName, lockname, pod, skipNodeLockOwnerCheck)
+}
+
+func releaseNodeLockLocked(nodeName string, lockname string, pod *corev1.Pod, skipNodeLockOwnerCheck bool) error {
 	ctx := context.Background()
 	node, err := client.GetClient().CoreV1().Nodes().Get(ctx, nodeName, metav1.GetOptions{})
 	if err != nil {
@@ -239,13 +247,18 @@ func ReleaseNodeLock(nodeName string, lockname string, pod *corev1.Pod, skipNode
 }
 
 func LockNode(nodeName string, lockname string, pods *corev1.Pod) error {
+	// Acquire per-node lock instead of global lock
+	nodeLock := nodeLocks.getLock(nodeName)
+	nodeLock.Lock()
+	defer nodeLock.Unlock()
+
 	ctx := context.Background()
 	node, err := client.GetClient().CoreV1().Nodes().Get(ctx, nodeName, metav1.GetOptions{})
 	if err != nil {
 		return err
 	}
 	if _, ok := node.Annotations[NodeLockKey]; !ok {
-		return SetNodeLock(nodeName, lockname, pods)
+		return setNodeLockLocked(nodeName, lockname, pods)
 	}
 	lockTime, ns, previousPodName, err := ParseNodeLock(node.Annotations[NodeLockKey])
 	if err != nil {
@@ -256,7 +269,7 @@ func LockNode(nodeName string, lockname string, pods *corev1.Pod) error {
 	if time.Since(lockTime) > NodeLockTimeout {
 		klog.InfoS("Node lock expired", "node", nodeName, "lockTime", lockTime, "timeout", NodeLockTimeout)
 		skipOwnerCheck = true
-	} else if ns == pods.Namespace && previousPodName == pods.Name {
+	} else if pods != nil && ns == pods.Namespace && previousPodName == pods.Name {
 		// The lock is already held by this exact pod. lockAllDevices calls
 		// LockNode once per device vendor a pod requests resources from, so
 		// a pod requesting resources from two or more vendors (e.g. both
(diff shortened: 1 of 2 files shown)


pkg/util/nodelock/nodelock.go
```diff
@@ -181,6 +191,10 @@ func ReleaseNodeLock(nodeName string, lockname string, pod *corev1.Pod, skipNode
 	nodeLock.Lock()
 	defer nodeLock.Unlock()
 
+	return releaseNodeLockLocked(nodeName, lockname, pod, skipNodeLockOwnerCheck)
+}
+
+func releaseNodeLockLocked(nodeName string, lockname string, pod *corev1.Pod, skipNodeLockOwnerCheck bool) error {
```
mesutoezdil, line 197: SetNodeLock and LockNode reject a nil pod, ReleaseNodeLock does not and the owner check derefs pod.Name. intentional?

AyushSrivastava1818: yes the asymmetry is intentional because ReleaseNodeLock() historically handles its own nil-pod semantics differently,while LockNode() and SetNodeLock() must reject nil before any lock state mutation/recovery

archlitchi (approved):
/lgtm

PR #2743, opened 2026-08-20
fix(scheduler): respect HAMI_NODELOCK_EXPIRE env when --node-lock-timeout flag is absent

### LINKED ISSUE #2692: scheduler.nodeLockExpire / HAMI_NODELOCK_EXPIRE has no effect: overwritten by --node-lock-timeout flag default

Siraryansingh:
**What type of PR is this?**

/kind bug

The Helm chart has `scheduler.nodeLockExpire` which sets `HAMI_NODELOCK_EXPIRE` env var on the scheduler pod. The nodelock package reads this in `init()` and sets `NodeLockTimeout` correctly. But then in `start()`, we do:

```go
```

`config.NodeLockTimeout` comes from the `--node-lock-timeout` cobra flag which defaults to 5m. Since cobra always fills in the default even when the flag isn't passed, this line overwrites whatever the env var set. So `scheduler.nodeLockExpire` in the Helm chart literally does nothing.

I used `cmd.Flags().Changed("node-lock-timeout")` to check if the flag was actually passed by the user. If it was, use the flag value. If not, keep whatever `init()` set from the env var.

After this fix the precedence is:
1. `--node-lock-timeout` flag (if explicitly passed)
2. `HAMI_NODELOCK_EXPIRE` env var
3. default 5m

Fixes #2692

**AI Disclosure**: I used Gemini to help explore the codebase and understand the cobra flag flow, but I identified the root cause myself by tracing through `nodelock.go` init -> `main.go` start. The fix and testing were done with AI assistance.

**How was this tested?**

- `go build ./cmd/scheduler/...` passes
- `go test ./pkg/util/nodelock/... -run TestSetupNodeLockTimeout` passes
- all existing nodelock tests still pass

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved scheduler handling of the node-lock timeout setting.
  - Explicit command-line values now take precedence, while environment-provided and default settings are preserved when no flag is specified.
  - Added logging to indicate which timeout source was selected.

- **Tests**
  - Added coverage for default, environment-derived, and explicitly configured timeout behavior.

### DIFF
--- a/cmd/scheduler/main.go
+++ b/cmd/scheduler/main.go
@@ -54,7 +54,7 @@ var (
 		Short: "kubernetes vgpu scheduler",
 		RunE: func(cmd *cobra.Command, args []string) error {
 			flag.PrintPFlags(cmd.Flags())
-			return start()
+			return start(cmd)
 		},
 	}
 )
@@ -109,10 +109,17 @@ func injectProfilingRoute(router *httprouter.Router) {
 	})
 }
 
-func start() error {
-	// Initialize node lock timeout from config
-	nodelock.NodeLockTimeout = config.NodeLockTimeout
-	klog.InfoS("Set node lock timeout", "timeout", nodelock.NodeLockTimeout)
+func resolveNodeLockTimeout(flagChanged bool) {
+	if flagChanged {
+		nodelock.NodeLockTimeout = config.NodeLockTimeout
+		klog.InfoS("Set node lock timeout from flag", "timeout", nodelock.NodeLockTimeout)
+	} else {
+		klog.InfoS("Using node lock timeout from environment or default", "timeout", nodelock.NodeLockTimeout)
+	}
+}
+
+func start(cmd *cobra.Command) error {
+	resolveNodeLockTimeout(cmd.Flags().Changed("node-lock-timeout"))
 	client.InitGlobalClient(
 		client.WithBurst(config.Burst),
 		client.WithQPS(config.QPS),
--- a/cmd/scheduler/main_test.go
+++ b/cmd/scheduler/main_test.go
@@ -1,0 +1,72 @@
+/*
+Copyright 2024 The HAMi Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+	"testing"
+	"time"
+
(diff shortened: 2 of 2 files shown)


Siraryansingh:
hey, just wanted to give some context on how i found this.

i was going through the open issues and #2692 caught my eye. i traced the flow from the helm chart (`values.yaml` -> deployment template -> env var) into the go code and noticed the problem pretty quickly - `nodelock.init()` reads the env var fine, but then `start()` just stomps on it with the cobra flag default.

the fix is small - just pass the `*cobra.Command` into `start()` and check `cmd.Flags().Changed()` before overwriting. this is the standard cobra pattern for "flag vs env var" precedence.

i saw that #2721 and #2722 tried to fix the same issue. i came across it independently while looking through open bugs - happy to answer any questions about the approach or the code.

also adding an AI disclosure since the guidelines require it: i used gemini to help navigate the codebase and understand the flag registration flow, but the root cause analysis and fix were done by me with AI assistance. i can explain any part of the change if needed.

cmd/scheduler/main.go
```diff
@@ -109,10 +109,17 @@ func injectProfilingRoute(router *httprouter.Router) {
...
-func start() error {
-	// Initialize node lock timeout from config
-	nodelock.NodeLockTimeout = config.NodeLockTimeout
-	klog.InfoS("Set node lock timeout", "timeout", nodelock.NodeLockTimeout)
+func start(cmd *cobra.Command) error {
+	// Only override the node lock timeout when the --node-lock-timeout flag was
+	// explicitly passed on the command line. When the flag is absent, the value
+	// parsed from the HAMI_NODELOCK_EXPIRE environment variable (set by the
+	// Helm chart's scheduler.nodeLockExpire) in nodelock.init() is preserved.
+	if cmd.Flags().Changed("node-lock-timeout") {
```
mesutoezdil: no test for the fix, the listed nodelock test does not cover this branch. extract it into a helper taking a bool and add a regression test for flag set, env only, neither.

mesutoezdil (changes requested):

Siraryansingh:
@mesutoezdil thanks for the review.

I extracted the resolution logic into `resolveNodeLockTimeout(flagChanged bool)` in `cmd/scheduler/main.go` and added a regression test in `cmd/scheduler/main_test.go` covering all three cases:
1. `neither flag nor env set` - preserves default (`5m`)
2. `env set, flag not set` - preserves env setting (`10m`)
3. `flag explicitly set` - overrides with flag value (`2m`)

All tests in `cmd/scheduler` pass cleanly!

mesutoezdil:

PR #2751, opened 2026-08-20
fix(nvidia): collapse duplicate MIG candidates in topology scheduling

### LINKED ISSUE #2750: MIG devices are allocated repeatedly on a single card in topology scheduling mode
**What happened**:

When a pod requests multiple GPUs with `hami.io/gpu-scheduler-policy: "topology"` on a node with MIG-enabled cards, all requested devices are allocated from a **single physical card** while other free cards sit idle. Scheduling latency also grows exponentially with the requested device count.

Root cause is in `NvidiaGPUDevices.Fit()` (`pkg/device/nvidia/device.go`). The candidate-collection loop re-examines a MIG card once per free MIG placement:

```go
if dev.Mode == "mig" {
    i++
}
```

In topology mode nothing short-circuits the loop, so every pass appends another candidate entry. An 8-card node with 7 free slices each yields **56 candidate entries describing only 8 physical cards** — seven identical structs per card.

That inflated pool causes two distinct problems:

1. **Wrong placement.** `generateCombinations` walks the candidate slice in order, so the first combination it emits is N copies of the same card. `computeBestCombination` replaces the best only on `totalScore > bestScore` (strictly greater), so on a node where `hami.io/node-nvidia-score` was never published — every score `0` — that first same-card combination wins the tie. A card also has no pair score against itself (`calculate_score.go` skips `indexI == indexJ`), so a same-card set can never lose on score either.
2. **Combinatorial blow-up.** `generateCombinations` enumerates C(56,5) = 3,819,816 combinations instead of C(8,5) = 56, per node, on the scheduling hot path.

The requested devices should be spread across distinct physical cards when the node has enough of them, and scheduling should complete in a reasonable time.

Schedule a pod requesting multiple GPUs onto a node with MIG-mode NVIDIA cards, using the topology policy:

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: mig-topology-repro
  annotations:
    hami.io/gpu-scheduler-policy: "topology"
spec:
  containers:
  - name: cuda
    resources:
      limits:
        nvidia.com/gpu: 5
        nvidia.com/gpumem: 10240
```

The resulting `hami.io/vgpu-devices-allocated` annotation contains five entries that all share one physical GPU UUID.

The same behaviour reproduces deterministically in a unit test against `Fit()`, with 8 MIG cards of 7 free single-slice placements each, the topology policy annotation, and no published pair scores:

| request | wall time | distinct cards selected |
|---|---|---|
| 1 GPU | 785 µs | 1 |
| 2 GPUs | 1.15 ms | **1** (expected 2) |
| 3 GPUs | 14.3 ms | **1** (expected 3) |
| 4 GPUs | 237.7 ms | **1** (expected 4) |
| 5 GPUs | **11.19 s** | **1** (expected 5) |
| 4 GPUs, non-topology | 53.8 µs | 4 |

- Non-MIG nodes are unaffected: without MIG there is no `i++`, so no duplicate candidates. This is why the existing topology tests — all of which use non-MIG cards — never caught it. The MIG path and the topology path are each covered; their interaction was not.
- The latency figure is **per node**. `Fit()` runs on every candidate node, so a 20-node cluster spends roughly 20x that on a single pod.
- The wrong-placement half only manifests when pair scores are absent or tied, which is the normal state for a PCIe-only machine where nothing publishes `hami.io/node-nvidia-score`. The latency half manifests regardless.
- I found this by code reading rather than from a production incident, so I have no `nvidia-smi -a` output or scheduler logs from a live cluster to attach. The unit-test reproduction above exercises `Fit()` directly against current master.

**Environment**:
- HAMi version: master @ `98c520e`
- nvidia driver or other AI device driver version: N/A — reproduced against `Fit()` in a unit test, not on live hardware
- Others: requires MIG-mode NVIDIA cards (A100/H100 class) and the per-pod `hami.io/gpu-scheduler-policy: topology` annotation

Eshiv-Pandey:
**What type of PR is this?**
/kind bug

`NvidiaGPUDevices.Fit()` re-examines a MIG card once per free MIG placement (`if dev.Mode == "mig" { i++ }`). In topology mode nothing short-circuits the collection loop, so each pass appends another candidate entry. An 8-card node with 7 free slices each produces 56 candidate entries describing only 8 physical cards — seven identical structs per card.

That inflated pool causes two defects:

1. **All devices land on one card.** `generateCombinations` walks the slice in order, so the first combination emitted is N copies of the same card. `computeBestCombination` replaces the best only on `totalScore > bestScore`, so when no pair scores are published (every score `0`) that first same-card combination wins the tie. A card has no pair score against itself, so a same-card set can never lose on score either. A 5-GPU request on an 8-card node is satisfied by 5 slices of one card, leaving 7 cards idle.

2. **Combinatorial blow-up.** C(56,5) = 3,819,816 combinations are enumerate instead of C(8,5) = 56, per node, on the scheduling hot path.

Measured on 8 MIG cards x 7 free slices with the topology policy and no published pair scores: 2 GPUs -> 1 card in 1.15ms, 3 -> 1 card in 14.3ms, 4 -> 1 card in 237.7ms, 5 -> 1 card in **11.19s**. The same 4-GPU request without the topology policy takes 53.8us and correctly picks 4 cards.

This PR collapses the candidate pool by physical card UUID before scoring, but only when the node has at least as many distinct cards as the pod requested. The duplicate entries are identical structs and pair scores are per physical
card, so they carry no information the scorer can use - dropping them cannot change which set of cards is best. The concrete MIG placement is still chosen later by `selectMigCandidate()` in `AddResourceUsage()`, so nothing is lost.

The `len(distinct) >= originReq` guard preserves packing multiple MIG instances onto one card when the node has no alternative - behaviour PR #2724 shows with hardware evidence is wanted.

After the fix all requests select the expected number of distinct cards, and the 5-GPU case drops from 11.19s to under 1ms.

Non-MIG nodes are unaffected: without MIG there are no duplicates, so the helper returns the slice unchanged and every path is byte-identical. The `== originReq exact-match branch is deliberately untouched - that pool is the only option, so
there is nothing to choose between.

Fixes #2750

- Existing topology tests are all non-MIG, so this interaction was untested; six tests added.
- Verified both spread tests fail without the fix: the 2-GPU case picks 1 card (0.03s) and the 5-GPU case picks 1 card (~12s); both pass in 0.00s with it.
- Known residual, out of scope: a request exceeding the node's card count keeps the full pool. `pkg/scheduler/score.go` rejects the node before `Fit()` runs, so it is unreachable unless #2724 changes that gate.
- Unrelated latent bug, untouched: `computeWorstSingleCard` dereferences `deviceScoreMap[dev1.UUID].Scores` with no nil check. Happy to file separately.
- **AI assistance disclosure**: I used Claude Code for code analysis, the implementation, and the tests; all changes reviewed and verified by me.

```release-note
Fixed a bug where MIG devices were repeatedly allocated from a single physical GPU under the `topology` scheduler policy, leaving other free cards idle and causing exponential scheduling latency as the requested device count grew.
```

### DIFF
--- a/pkg/device/nvidia/device_test.go
+++ b/pkg/device/nvidia/device_test.go
@@ -2830,3 +2830,202 @@ func TestNodeDeleted_ReuseAfterDeletion(t *testing.T) {
 	assert.Equal(t, healthy, true, "re-created node should be healthy")
 	assert.Equal(t, needUpdate, true, "re-created node must trigger an update (stale bookkeeping was cleared)")
 }
+
+// migTopologyDevice builds the NVIDIA backend used by the MIG topology tests.
+func migTopologyDevice() *NvidiaGPUDevices {
+	return InitNvidiaDevice(NvidiaConfig{
+		ResourceCountName:            "nvidia.com/gpu",
+		ResourceMemoryName:           "nvidia.com/gpumem",
+		ResourceCoreName:             "nvidia.com/gpucores",
+		ResourceMemoryPercentageName: "nvidia.com/gpumem-percentage",
+	})
+}
+
+// migTopologyDevices builds MIG cards with seven single-slice placements each,
+// so the Fit loop contributes seven interchangeable candidates per card.
+func migTopologyDevices(ids ...string) []*device.DeviceUsage {
+	placements := make([]device.MigPlacement, 0, 7)
+	for start := range uint32(7) {
+		placements = append(placements, device.MigPlacement{Start: start, Size: 1})
+	}
+	devices := make([]*device.DeviceUsage, 0, len(ids))
+	for idx, id := range ids {
+		devices = append(devices, &device.DeviceUsage{
+			ID: id, Index: uint(idx), Used: 0, Count: 7,
+			Totalmem: 81920, Totalcore: 100, Type: NvidiaGPUDevice, Health: true,
+			Mode: MigMode,
+			MigProfiles: []device.MigProfile{
+				{Name: "1g.10gb", MemoryMB: 10240, Core: 14, SliceCount: 1, Placements: placements},
+			},
+		})
+	}
+	return devices
+}
+
+// migTopologyNodeInfo registers cards with no pair score, as on a node where
+// hami.io/node-nvidia-score was never published.
+func migTopologyNodeInfo(ids ...string) *device.NodeInfo {
+	infos := make([]device.DeviceInfo, 0, len(ids))
+	for _, id := range ids {
+		infos = append(infos, device.DeviceInfo{ID: id})
+	}
+	return &device.NodeInfo{Devices: map[string][]device.DeviceInfo{NvidiaGPUDevice: infos}}
+}
+
+func migTopologyPod() *corev1.Pod {
+	return &corev1.Pod{
+		ObjectMeta: metav1.ObjectMeta{
+			Annotations: map[string]string{util.GPUSchedulerPolicyAnnotationKey: util.GPUSchedulerPolicyTopology.String()},
+		},
+	}
+}
+
+func distinctUUIDs(devices device.ContainerDevices) []string {
+	seen := map[string]struct{}{}
+	out := make([]string, 0, len(devices))
+	for _, dev := range devices {
+		if _, ok := seen[dev.UUID]; ok {
+			continue
+		}
+		seen[dev.UUID] = struct{}{}
+		out = append(out, dev.UUID)
+	}
+	return out
+}
+
+// Without pair scores every combination ties at zero, so the first one generated
+// wins; that used to be several slots of one card with the others left idle.
+func TestFit_TopologyMigSpreadsAcrossCardsWithoutScores(t *testing.T) {
+	nv := migTopologyDevice()
+	devices := migTopologyDevices("dev-0", "dev-1", "dev-2")
+	nodeInfo := migTopologyNodeInfo("dev-0", "dev-1", "dev-2")
+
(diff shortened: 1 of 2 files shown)


pkg/device/nvidia/device_test.go
```diff
@@ -2830,3 +2830,202 @@ func TestNodeDeleted_ReuseAfterDeletion(t *testing.T) {
 	assert.Equal(t, healthy, true, "re-created node should be healthy")
 	assert.Equal(t, needUpdate, true, "re-created node must trigger an update (stale bookkeeping was cleared)")
 }
+
+// migTopologyDevice builds the NVIDIA backend used by the MIG topology tests.
+func migTopologyDevice() *NvidiaGPUDevices {
+	return InitNvidiaDevice(NvidiaConfig{
```
mesutoezdil, line 2836: ran both spread tests, this 2 gpu case already passes there, only the 5 gpu one pins the fix. worth noting or dropping.

Eshiv-Pandey: Thanks for running those but i am unable to replicate this on my end idky. 
fail test
image
pass test
image
On my side the 2-GPU case are failing without the fix

mesutoezdil: you are right, retested on current master and both tests fail there now. my earlier run was against an older base. withdrawn.

archlitchi (approved):
/lgtm

PR #2754, opened 2026-08-20
Fixed Scheduler returning false

### LINKED ISSUE #2749: scheduler returns false on the bases of only one init-container status
**What happened**:
According to the workflow the HAMi scheduler needs to know when the pod has finished its setup, and that setups are performed by the init containers.

The issue is that HAMi only looks at the init container status it received, instead of confirming that it received a record for every init container the pod declares.

The scheduler should only treat init containers as complete, when all init containers in have a successful terminal status.

For an example, if 2 init containers exist but only 1 status entry is present, the function should return false instead of true.

- create a pod having 2 init containers
- only one of them reports a success status
- call the `AllInitContainersSucceeded()`

```bash
=== RUN   TestAllInitContainersSucceeded_MissingStatus_Reproducer
    util_test.go:948: spec init containers: 2
    util_test.go:949: reported init container statuses: 1
    util_test.go:950: reported successful container: "init-one"
    util_test.go:951: missing init-container status: "init-two"
    util_test.go:952: AllInitContainersSucceeded: true
    util_test.go:955: BUG REPRODUCED: expected false because init-two has no status, got true
--- FAIL: TestAllInitContainersSucceeded_MissingStatus_Reproducer (0.00s)
FAIL
FAIL    github.com/Project-HAMi/HAMi/pkg/util   1.282s
FAIL
```

**Environment**:
- HAMi version: b28d7d205d5447f4aaac1fc6f80e44ff62c28d89
- Kubernetes version: 
- Accelerator vendor and model:
- HAMi device backend and requested resource names (for example, `nvidia.com/gpu`):
- Accelerator driver and runtime version:
- Container runtime and version:
- Workload image and tag (redact private image names and state when redacted):
- Operating system and kernel version (for example, the output of `uname -a`):
- Others:

Kevindua26:
**What type of PR is this?**
/kind bug

Fixes the false result which schedular is generating on completion of only one init-container.

Fixes #2749 

image

Yes... for next `release-note`: fixed scheduler device, so init container resources are released only after statuses are reported for all init containers.

> This PR was authored with assistance from Codex for help me implementing the core fix logic for init-container validation in `util.go`, and `util_test.go` to cover missing init-container statuses.
> I have fully reviewed and tested all logics and AI-generated code. The implementation was validated to ensure the scheduler correctly evaluates all init containers before returning a result. All tests pass locally and attached with the PR.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation of init-container completion status.
  * Ensured only declared init containers are evaluated.
  * Pods with missing, incomplete, or unsuccessful init-container status now correctly report as unsuccessful.
  * Successfully completed init containers continue to be recognized correctly.

* **Tests**
  * Added coverage for missing statuses and successful init-container completion.

### DIFF
--- a/pkg/util/util.go
+++ b/pkg/util/util.go
@@ -160,7 +160,14 @@ func PatchNodeAnnotations(node *corev1.Node, annotations map[string]string) erro
 	return err
 }
 func AllInitContainersSucceeded(pod *corev1.Pod) bool {
-	if len(pod.Status.InitContainerStatuses) == 0 {
+	if pod == nil {
+		return false
+	}
+	initContainers := len(pod.Spec.InitContainers)
+	if initContainers == 0 {
+		return false
+	}
+	if len(pod.Status.InitContainerStatuses) != initContainers {
 		return false
 	}
 	for _, s := range pod.Status.InitContainerStatuses {
--- a/pkg/util/util.go
+++ b/pkg/util/util.go
@@ -170,6 +170,22 @@ func AllInitContainersSucceeded(pod *corev1.Pod) bool {
 	if len(pod.Status.InitContainerStatuses) != initContainers {
 		return false
 	}
+	remaining := make(map[string]struct{}, initContainers)
+	for _, container := range pod.Spec.InitContainers {
+		remaining[container.Name] = struct{}{}
+	}
+	for _, s := range pod.Status.InitContainerStatuses {
+		if _, ok := remaining[s.Name]; !ok {
+			return false
+		}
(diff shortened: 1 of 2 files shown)


mesutoezdil:

Kevindua26:
Hey @mesutoezdil, can you please take a look at the PR again, earlier i forgot to commits with signoffs... I force pushed the signoffs.

Kevindua26:
Hey @mesutoezdil, I've added the required disclosures to the PR description as per contributing guidelines, and force pushed signoffs for all three commits. Please review and reopen.

PR #2755, opened 2026-08-20
fix(scheduler): include init container requests in device locking

### LINKED ISSUE #2745: fix(scheduler): acquire device node locks for init-container-only GPU requests
## Problem

HAMi's resource request calculation explicitly includes GPU requests from both init containers and regular containers. `Resourcereqs()` processes `pod.Spec.InitContainers` first and includes their device requests in the scheduling allocation. This means a pod that requests a HAMi GPU exclusively from an init container is still treated as a GPU workload by the scheduler.

However, the node-locking path does not use the same container set.

For example, the NVIDIA implementation currently determines whether a node lock is required by iterating only over:

```go
for _, val := range p.Spec.Containers {
    if dev.GenerateResourceRequests(&val).Nums > 0 {
        found = true
        break
    }
}
```

`ReleaseNodeLock()` uses the same logic. The Hygon/DCU implementation follows the same pattern.

As a result, a pod with a GPU request only in an init container can be scheduled and have its GPU allocation calculated, while `LockNode()` concludes that the pod does not require a node lock and returns without acquiring one.

The relevant paths are therefore inconsistent:

```text
Pod with GPU request in initContainer
        |
        v
Resourcereqs()
        |
        +--> init containers are processed
        |
        v
GPU allocation is calculated
        |
        v
Bind / device locking
        |
        v
LockNode()
        |
        +--> only Spec.Containers checked
        |
        v
No node lock acquired
```

This is particularly problematic because the node lock exists to serialize allocation-related operations on the same node. If multiple scheduling/bind operations involving init-container-only GPU requests reach the same node concurrently, those operations can bypass the locking protection that is applied to equivalent requests made by regular containers.

## Expected behavior

The decision to acquire and release a device node lock should consider the same effective set of GPU requests used by scheduling, including init containers.

A pod requesting a GPU exclusively from an init container should acquire the corresponding device node lock during the allocation/bind path and release it normally afterward.

The behavior should be consistent across supported device backends.

## Suggested fix

Avoid independently checking only `pod.Spec.Containers` inside each device implementation.

Instead, use a shared helper or the existing resource-request calculation to determine whether the pod has a request for the current device type across both init and regular containers.

For example, the locking decision could effectively follow:

```text
initContainers + containers
        |
        v
GenerateResourceRequests()
        |
        v
Does this device have a request?
        |
     yes/no
        |
        v
LockNode / ReleaseNodeLock
```

This would also prevent different device backends from implementing different interpretations of whether an init-container-only GPU workload requires locking.

## Regression tests

Add tests covering at least:

1. A pod with a GPU request only in a regular container acquires the node lock.
2. A pod with a GPU request only in an init container acquires the node lock.
3. A pod with GPU requests in both init and regular containers acquires the node lock exactly once.
4. A pod with no GPU request in either init or regular containers does not acquire the node lock.
5. `ReleaseNodeLock()` follows the same behavior as `LockNode()`.
6. The behavior is covered for the affected device backends rather than only NVIDIA.

## Why this matters

HAMi already treats init-container GPU requests as part of the scheduler's resource accounting, so ignoring them at the node-lock boundary creates two different definitions of what constitutes a GPU allocation.

This can allow an init-container-only GPU workload to bypass the synchronization mechanism used to protect concurrent device allocation on a node.

The issue is therefore not about adding a new scheduling feature; it is about making the existing init-container resource accounting and node-locking behavior consistent.

AyushSrivastava1818:
## What type of PR is this?

/kind bug

HAMi's resource request calculation accounts for device requests from both init containers and regular containers. However, the device node-locking implementations were checking only `pod.Spec.Containers`.

As a result, a pod with a HAMi GPU/device request exclusively in an init container could be scheduled and have its resources accounted for without acquiring the corresponding device node lock.

This creates an inconsistency between resource accounting and node locking and can allow concurrent scheduling operations to bypass the synchronization intended for device allocation.

## Changes

- Added `PodRequiresDevice()` in `pkg/device/devices.go` to determine whether a pod requires a particular device.
- The helper checks both `InitContainers` and regular `Containers`.
- Updated the affected device backends to use the shared helper for `LockNode()` and `ReleaseNodeLock()`.
- Added regression coverage for init-container-only and combined init/regular container requests.
- Preserved existing backend-specific locking behavior.

## Affected device backends

- NVIDIA
- Hygon
- Cambricon
- Iluvatar
- Biren
- MetaX
- Kunlun
- AMD
- Ascend
- VASTAI

Fixes #2745

## Regression coverage

The tests cover:

- Regular-container-only device requests.
- Init-container-only device requests.
- Device requests in both init and regular containers.
- Pods without relevant device requests.
- `LockNode()` behavior.
- `ReleaseNodeLock()` behavior.
- Nil pod/device inputs for the shared helper.
- Multiple affected device backends.

The init-container-only regression cases fail against the previous `Spec.Containers`-only implementation and pass with this fix.

## Validation

- `go test ./pkg/device/...` — 12/13 packages passed.
- NVIDIA tests are blocked by the local NVML/CGO environment and are unrelated to this change.
- `go vet ./pkg/device/{amd,ascend,biren,cambricon,hygon,iluvatar,kunlun,metax,vastai}/...` — passed.
- `git diff --check` — passed.
- `gofmt` — verified.

No.

This is a scheduler/device-locking correctness fix. It ensures that init-container device requests receive the same node-lock treatment as equivalent regular-container requests.

## Special notes for reviewers

The fix intentionally centralizes device-request detection instead of maintaining separate `InitContainers`/`Containers` loops across each device backend.

`PodRequiresDevice()` uses the same `GenerateResourceRequests()` semantics as the existing resource-request calculation and checks init containers before regular containers.

Cambricon's existing unconditional `ReleaseNodeLock()` behavior is preserved because its locking mechanism differs from the pod-keyed node-lock implementation used by the other affected backends.

No changes were made to the node-lock recovery implementation from PR #2733.

## AI Assistance Disclosure

AI assistance from Claude and Antigravity was used during codebase investigation, root-cause analysis, implementation assistance, regression-test development, code review, and validation of this change.

The final implementation was reviewed against the existing HAMi codebase and device-locking behavior. The contributor independently verified the affected backends, confirmed that the shared helper matches the existing resource-request semantics, reviewed the resulting diff, and ran the relevant tests, vet, formatting, and diff checks.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved device requirement detection for pods using init containers, regular containers, or both.
  * Node locking and unlocking now consistently occur only when a pod requests the relevant device.

* **Tests**
  * Expanded coverage for device requests across container types, including pods without device requirements and mixed requests.

### DIFF
--- a/pkg/device/amd/device.go
+++ b/pkg/device/amd/device.go
@@ -131,28 +131,14 @@ func (dev *AMDDevices) PatchAnnotations(pod *corev1.Pod, annoinput *map[string]s
 }
 
 func (dev *AMDDevices) LockNode(n *corev1.Node, p *corev1.Pod) error {
-	found := false
-	for _, val := range p.Spec.Containers {
-		if (dev.GenerateResourceRequests(&val).Nums) > 0 {
-			found = true
-			break
-		}
-	}
-	if !found {
+	if !device.PodRequiresDevice(dev, p) {
 		return nil
 	}
 	return nodelock.LockNode(n.Name, NodeLockAMD, p)
 }
 
 func (dev *AMDDevices) ReleaseNodeLock(n *corev1.Node, p *corev1.Pod) error {
-	found := false
-	for _, val := range p.Spec.Containers {
-		if (dev.GenerateResourceRequests(&val).Nums) > 0 {
-			found = true
-			break
-		}
-	}
-	if !found {
+	if !device.PodRequiresDevice(dev, p) {
 		return nil
 	}
 	return nodelock.ReleaseNodeLock(n.Name, NodeLockAMD, p, false)
--- a/pkg/device/ascend/device.go
+++ b/pkg/device/ascend/device.go
@@ -256,14 +256,7 @@ func (dev *Devices) PatchAnnotations(pod *corev1.Pod, annoInput *map[string]stri
 }
 
 func (dev *Devices) LockNode(n *corev1.Node, p *corev1.Pod) error {
-	found := false
-	for _, val := range p.Spec.Containers {
-		if (dev.GenerateResourceRequests(&val).Nums) > 0 {
-			found = true
-			break
-		}
-	}
-	if !found {
+	if !device.PodRequiresDevice(dev, p) {
 		return nil
 	}
 
@@ -271,14 +264,7 @@ func (dev *Devices) LockNode(n *corev1.Node, p *corev1.Pod) error {
 }
 
 func (dev *Devices) ReleaseNodeLock(n *corev1.Node, p *corev1.Pod) error {
-	found := false
-	for _, val := range p.Spec.Containers {
-		if (dev.GenerateResourceRequests(&val).Nums) > 0 {
-			found = true
-			break
-		}
-	}
-	if !found {
+	if !device.PodRequiresDevice(dev, p) {
 		return nil
 	}
 
--- a/pkg/device/biren/device.go
+++ b/pkg/device/biren/device.go
@@ -96,28 +96,14 @@ func (dev *BirenDevices) MutateAdmission(ctr *corev1.Container, p *corev1.Pod) (
 }
 
 func (dev *BirenDevices) LockNode(n *corev1.Node, p *corev1.Pod) error {
-	found := false
-	for _, val := range p.Spec.Containers {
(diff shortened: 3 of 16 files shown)


archlitchi (approved):
/lgtm

PR #2765, opened 2026-08-21
fix(device): reject malformed and negative device allocation annotations

### LINKED ISSUE #2756: fix(device): reject malformed and negative device allocation annotations
## Problem

HAMi's device allocation annotations are decoded into `ContainerDevice` values and subsequently used by scheduler and device allocation/accounting paths.

The current `DecodeContainerDevices()` implementation does not fully validate the encoded device entries before accepting them.

In particular, two cases can result in invalid allocation state being accepted:

### 1. Malformed device segments can be silently skipped

`DecodeContainerDevices()` splits the annotation into device entries and only processes entries containing a comma.

A malformed non-empty segment that does not contain a comma can therefore be ignored instead of causing the complete annotation to be rejected.

For example, an annotation containing a valid device followed by a malformed entry can effectively be treated as a partially valid allocation:

```text
GPU-123,NVIDIA,4096,50:CORRUPTED
```

Instead of reporting that the allocation annotation is malformed, the valid portion can be decoded while the malformed segment is discarded.

For allocation metadata, silently accepting a partial decode can leave different components with different interpretations of the pod's assigned devices.

### 2. Negative memory/core quantities are not rejected

The decoder parses the memory and core fields using integer parsing, but successful parsing alone does not guarantee that the resulting resource quantities are valid.

Values such as:

```text
GPU-123,NVIDIA,-4096,50:
```

or:

```text
GPU-123,NVIDIA,4096,-50:
```

are syntactically valid integers and can therefore be represented as a `ContainerDevice` with negative resource usage.

These decoded values are later consumed by resource-accounting paths.

For example, scheduler usage reconciliation adds decoded device usage to the tracked device state:

```go
d.Device.Usedmem += udevice.Usedmem
d.Device.Usedcores += udevice.Usedcores
```

Allowing negative decoded quantities into this path can reduce the recorded memory/core usage rather than increase it.

## Why this matters

The device allocation annotation is part of the state exchanged between HAMi's scheduling and device-allocation components.

Invalid allocation metadata should therefore fail closed.

Currently, malformed or invalid values can potentially result in:

- partially decoded device allocations;
- scheduler/device-plugin disagreement about assigned devices;
- incorrect GPU memory accounting;
- incorrect GPU core accounting;
- invalid allocation state being treated as legitimate;
- confusing downstream allocation failures;
- scheduler capacity calculations based on corrupted allocation metadata.

The negative-value case is particularly problematic because invalid resource quantities can affect arithmetic used to reconstruct device usage.

This issue does not require assuming that malformed annotations are normally generated by HAMi. Even if they result only from stale, corrupted, manually modified, or otherwise unexpected pod state, the decoder should not convert invalid allocation metadata into apparently valid resource accounting.

## Expected behavior

`DecodeContainerDevices()` should reject malformed non-empty device entries instead of silently ignoring them.

Decoded resource quantities should also be validated before constructing a `ContainerDevice`.

At minimum:

- malformed non-empty device segments should return an error;
- required fields should be present;
- device UUID/ID should not be empty;
- device type should not be empty;
- memory usage should not be negative;
- core usage should not be negative;
- invalid or overflowing integer values should return an error;
- a mixture of valid and malformed device entries should fail the complete decode rather than returning a partial allocation.

Valid existing allocation annotations should continue to decode exactly as before.

## Example

Valid:

```text
GPU-123,NVIDIA,4096,50:
```

Expected:

```text
decode succeeds
```

Negative memory:

```text
GPU-123,NVIDIA,-4096,50:
```

Expected:

```text
decode returns an error
```

Negative cores:

```text
GPU-123,NVIDIA,4096,-50:
```

Expected:

```text
decode returns an error
```

Missing fields:

```text
GPU-123,NVIDIA,4096:
```

Expected:

```text
decode returns an error
```

Mixed valid and malformed entries:

```text
GPU-123,NVIDIA,4096,50:CORRUPTED
```

Expected:

```text
decode returns an error
```

rather than returning only the first valid allocation.

## Suggested fix

Strengthen validation in `DecodeContainerDevices()` while preserving the existing annotation format.

The decoder should validate every non-empty device entry before appending it to the returned `ContainerDevices`.

A malformed entry should cause the complete decode operation to fail.

Resource quantities should also be checked after parsing:

```text
Usedmem >= 0
Usedcores >= 0
```

The fix should remain localized to annotation decoding and its callers unless additional changes are required to propagate existing decode errors correctly.

It should not redesign the device annotation format or change valid allocation semantics.

## Regression tests

Add focused tests covering at least:

1. A valid single-device allocation decodes successfully.
2. Multiple valid device allocations decode successfully.
3. Existing valid annotations continue to preserve all fields.
4. A device entry with missing fields returns an error.
5. A malformed non-empty segment returns an error.
6. Negative memory returns an error.
7. Negative cores returns an error.
8. Invalid integer values return an error.
9. Integer overflow returns an error.
10. A mixed valid + malformed annotation fails completely instead of returning a partial allocation.
11. Empty separators/trailing delimiters continue to follow the existing valid annotation format where applicable.

Where practical, add a regression test demonstrating that invalid negative allocation values cannot reach resource accounting and reduce recorded `Usedmem` or `Usedcores`.

## Scope

This issue is specifically about validating HAMi's existing encoded device-allocation state.

The intended fix should:

- preserve the current annotation format;
- preserve valid device allocations;
- avoid unrelated scheduler/device refactoring;
- reject invalid allocation state before it can be treated as legitimate device usage.

## Impact

The current decoder is permissive in a path whose output is used as allocation state.

Failing closed on malformed entries and impossible negative resource quantities prevents corrupted annotation data from becoming partial or incorrect device accounting and makes scheduler/device state handling more deterministic.

AyushSrivastava1818:
# What type of PR is this?

/kind bug

`DecodeContainerDevices()` currently accepts malformed device allocation annotation data in several cases.

Non-empty malformed segments without the expected structure can be silently ignored, allowing a valid portion of an annotation to be decoded while invalid data is discarded. The decoder also accepts empty device UUID/type fields and negative memory/core values.

Negative resource values are particularly problematic because decoded `Usedmem` and `Usedcores` values are subsequently used by device resource-accounting paths. A negative value can therefore incorrectly reduce the tracked resource usage.

This PR makes device allocation annotation decoding fail closed by validating every device entry before it is accepted.

### Changes

- Reject non-empty malformed annotation segments instead of silently skipping them.
- Validate that required device UUID and device type fields are present.
- Reject negative GPU memory values.
- Reject negative GPU core values.
- Continue rejecting invalid/non-numeric resource values and integer overflow through `strconv.ParseInt`.
- Prevent partial device allocations from being returned when any annotation entry is invalid.
- Scope `ContainerDevice` construction to each decoded entry.
- Preserve existing handling of empty annotations and trailing/empty delimiters.
- Preserve the existing annotation format and valid allocation behavior.

Fixes #2756

## Regression coverage

The tests cover:

- Valid single-device annotations.
- Valid multiple-device annotations.
- Zero memory/core values.
- Existing optional fields.
- Empty annotations and trailing delimiters.
- Malformed segments without commas.
- Missing required fields.
- Empty device UUID.
- Empty device type.
- Negative memory.
- Negative cores.
- Non-numeric memory/core values.
- Integer overflow.
- Mixed valid and malformed annotations.
- Mixed valid and negative-resource annotations.
- Pod-level malformed allocation annotations.
- Protection against negative decoded values reaching resource accounting.
- Encode/decode fuzz coverage.

## Validation

The following checks were completed:

- `gofmt -w pkg/device/devices.go pkg/device/devices_test.go` — passed
- `git diff --check` — passed
- Targeted `go test` for decoder, pod decoder, resource-accounting, and fuzz tests — passed
- `CGO_ENABLED=0 go test ./pkg/device/...` — passed across the standalone device packages
- `go vet ./pkg/device` — passed
- `golangci-lint run ./pkg/device` — passed with no lint issues in the modified files

The NVIDIA package requires NVML/CGO headers that are unavailable in the local Windows environment. This is an environment limitation and is unrelated to the changes in this PR. The affected non-NVML device packages and root `pkg/device` package were validated successfully.

No.

This change only makes invalid device allocation annotations fail validation instead of allowing malformed or negative resource values to be interpreted as valid allocation state.

Valid existing annotations and their existing format remain unchanged.

## AI Assistance Disclosure

AI assistance from Claude and Antigravity was used during codebase investigation, root-cause analysis, implementation assistance, test development, review of the decoder and its callers, validation planning, and analysis of potential edge cases.

The final implementation was manually reviewed against the existing HAMi annotation format and resource-accounting behavior. The changes were validated by the contributor using targeted and package-level tests, `go vet`, `golangci-lint`, `gofmt`, and `git diff --check`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation when reading container device configurations.
  * Malformed entries, missing identifiers or types, negative resources, and invalid values are now rejected instead of silently ignored.
  * Prevented invalid multi-device configurations from being partially accepted or incorrectly affecting resource accounting.
  * Improved handling of device allocations for terminating pods and init-container resource adjustments.

* **Tests**
  * Added comprehensive coverage for valid, malformed, incomplete, negative, non-numeric, and overflowing device resource values.

### DIFF
--- a/pkg/device/devices.go
+++ b/pkg/device/devices.go
@@ -499,28 +499,45 @@ func DecodeContainerDevices(str string) (ContainerDevices, error) {
 	}
 	cd := strings.Split(str, OneContainerMultiDeviceSplitSymbol)
 	contdev := ContainerDevices{}
-	tmpdev := ContainerDevice{}
 	klog.V(5).Infof("Start to decode container device %s", str)
 	for _, val := range cd {
-		if strings.Contains(val, ",") {
-			tmpstr := strings.Split(val, ",")
-			if len(tmpstr) < 4 {
-				return nil, fmt.Errorf("pod annotation format error, missing fields, do not use nodeName in task spec")
-			}
-			tmpdev.UUID = tmpstr[0]
-			tmpdev.Type = tmpstr[1]
-			devmem, err := strconv.ParseInt(tmpstr[2], 10, 32)
-			if err != nil {
-				return nil, fmt.Errorf("invalid memory field: %w", err)
-			}
-			tmpdev.Usedmem = int32(devmem)
-			devcores, err := strconv.ParseInt(tmpstr[3], 10, 32)
-			if err != nil {
-				return nil, fmt.Errorf("invalid core field: %w", err)
-			}
-			tmpdev.Usedcores = int32(devcores)
-			contdev = append(contdev, tmpdev)
+		if val == "" {
+			continue
 		}
+		if !strings.Contains(val, ",") {
+			return nil, fmt.Errorf("malformed container device annotation segment: %q", val)
+		}
+		tmpstr := strings.Split(val, ",")
+		if len(tmpstr) < 4 {
+			return nil, fmt.Errorf("pod annotation format error, missing fields, do not use nodeName in task spec")
+		}
+		if tmpstr[0] == "" {
+			return nil, fmt.Errorf("pod annotation format error, missing device UUID")
+		}
+		if tmpstr[1] == "" {
+			return nil, fmt.Errorf("pod annotation format error, missing device type")
+		}
+		devmem, err := strconv.ParseInt(tmpstr[2], 10, 32)
+		if err != nil {
+			return nil, fmt.Errorf("invalid memory field: %w", err)
+		}
+		if devmem < 0 {
+			return nil, fmt.Errorf("memory field must not be negative: %d", devmem)
+		}
+		devcores, err := strconv.ParseInt(tmpstr[3], 10, 32)
+		if err != nil {
+			return nil, fmt.Errorf("invalid core field: %w", err)
+		}
+		if devcores < 0 {
+			return nil, fmt.Errorf("core field must not be negative: %d", devcores)
+		}
+		tmpdev := ContainerDevice{
+			UUID:      tmpstr[0],
+			Type:      tmpstr[1],
+			Usedmem:   int32(devmem),
+			Usedcores: int32(devcores),
+		}
+		contdev = append(contdev, tmpdev)
 	}
 	klog.V(5).Infof("Finished decoding container devices. Total devices: %d", len(contdev))
 	return contdev, nil
--- a/pkg/device/devices_test.go
+++ b/pkg/device/devices_test.go
@@ -46,6 +46,195 @@ func TestEmptyContainerDevicesCoding(t *testing.T) {
 	assert.DeepEqual(t, cd1, cd2)
 }
 
+func TestDecodeContainerDevices(t *testing.T) {
+	tests := []struct {
+		name        string
+		input       string
+		want        ContainerDevices
+		expectError bool
+	}{
+		{
+			name:        "empty string returns empty ContainerDevices",
+			input:       "",
+			want:        ContainerDevices{},
+			expectError: false,
+		},
+		{
+			name:        "trailing delimiters only returns empty ContainerDevices",
+			input:       ":::",
+			want:        ContainerDevices{},
+			expectError: false,
+		},
+		{
+			name:  "valid single device",
+			input: "GPU-8dcd427f-483b-b48f-d7e5-75fb19a52b76,NVIDIA,500,3:",
+			want: ContainerDevices{
+				{
+					UUID:      "GPU-8dcd427f-483b-b48f-d7e5-75fb19a52b76",
+					Type:      "NVIDIA",
+					Usedmem:   500,
+					Usedcores: 3,
+				},
+			},
+			expectError: false,
+		},
+		{
+			name:  "valid single device with zero resources",
+			input: "GPU-zero,NVIDIA,0,0:",
+			want: ContainerDevices{
+				{
+					UUID:      "GPU-zero",
+					Type:      "NVIDIA",
+					Usedmem:   0,
+					Usedcores: 0,
+				},
+			},
+			expectError: false,
+		},
+		{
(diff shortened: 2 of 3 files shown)


pkg/device/devices.go
```diff
@@ -499,28 +499,45 @@ func DecodeContainerDevices(str string) (ContainerDevices, error) {
...
+			continue
+		}
+		if !strings.Contains(val, ",") {
+			return nil, fmt.Errorf("malformed container device annotation segment: %q", val)
+		}
+		tmpstr := strings.Split(val, ",")
+		if len(tmpstr) < 4 {
+			return nil, fmt.Errorf("pod annotation format error, missing fields, do not use nodeName in task spec")
+		}
+		if tmpstr[0] == "" {
```
mesutoezdil, line 514: your own fixture fixes show this rejects entries older code produced. a running pod annotated by an older hami must still decode after upgrade, else its usage goes untracked and the card oversubscribes. did you audit the old release encoders?

AyushSrivastava1818: yesss thanks for pointing that out...I am gonna check older HAMi encoder/annotation format to confirm whether old running pods could have an empty device type before changing anything

AyushSrivastava1818: @mesutoezdil I checked the older HAMi versions too...THe empty Type cases were only in tests...real HAMi allocations have always inlcuded the device type....therefore I would keep the strict validation...wont weaken the fix for backward compatibility

archlitchi (approved):
/lgtm

PR #2768, opened 2026-08-21
ci: allow manual dispatch of the Trivy image scan

mesutoezdil:
Adds workflow_dispatch to the nightly Trivy scan so alert-clearing fixes can be verified right after merge instead of waiting for the next cron run.

## Summary by CodeRabbit

* **New Features**
  * Added the ability to manually trigger image security scans in addition to scheduled scans.


Shouren (approved):
/lgtm

PR #2769, opened 2026-08-21
fix(scheduler): keep the configured policy for unrecognized policy an…

### LINKED ISSUE #2767: scheduler: unrecognized gpu/node-scheduler-policy annotation is silently accepted and the configured defaultSchedulerPolicy is discarded
**What happened**:

`hami.io/gpu-scheduler-policy` and `hami.io/node-scheduler-policy` are user-set Pod annotations, documented as a way to "overlay scheduler config" (`docs/develop/scheduler-policy.md:10`). The value is never validated, and an unrecognized value is neither rejected nor resolved back to the configured policy — it is accepted, and the operator's configured default is discarded on the way.

Two things happen in sequence.

First, the annotation is an override rather than a fallback. Because the key is *present*, `ok` is true and the configured value is replaced, whatever the value is:

```go
// pkg/util/util.go:277-285
func GetGPUSchedulerPolicyByPod(defaultPolicy string, task *corev1.Pod) string {
	userGPUPolicy := defaultPolicy
	if task != nil && task.Annotations != nil {
		if value, ok := task.Annotations[GPUSchedulerPolicyAnnotationKey]; ok {
			userGPUPolicy = value
		}
	}
	return userGPUPolicy
}
```

`resolveNodeSchedulerPolicy` (`pkg/scheduler/score.go:139-146`) has the same shape for the node annotation.

Second, the comparator that receives the value applies a hardcoded fallback that is **not** the configured default. `DeviceUsageList.Less` (`pkg/scheduler/policy/gpu_policy.go:77`) tests for `binpack` at `:89` and for `mutex` at `:92`; anything else reaches the `// default policy is spread` branch at `:122`. That branch never consults `device.GPUSchedulerPolicy`.

```mermaid
flowchart LR
    A["values.yaml<br/>gpuSchedulerPolicy: binpack"] -->|"used only when the<br/>annotation is absent"| B
    C["Pod annotation<br/>'binpakc'"] -->|"key present -> overrides"| B
    B["GetGPUSchedulerPolicyByPod<br/>util.go:277"] --> D["Policy = 'binpakc'<br/>configured value discarded"]
    D --> E["gpu_policy.go:122<br/>hardcoded 'spread' branch"]
    E --> F["spread placement<br/>no error, no event, no log"]
```

The net effect is that on a cluster configured for binpack, a Pod with a misspelled annotation is scheduled with spread placement, and nothing anywhere reports a problem — no admission error, no Kubernetes event, no log line. The only observable signal is the resulting placement.

An unrecognized scheduler-policy value should either be rejected, or fall back to the policy the operator configured via `scheduler.defaultSchedulerPolicy` — not to a hardcoded value that ignores that configuration.

Found by code inspection on `master` at `ee92fc9` and reproduced with unit tests against the real scheduling path (`buildNodeUsage` -> `fitInDevices` -> `sort.Sort` -> the NVIDIA backend's `Fit()`). I have not reproduced this on a live cluster. Per CONTRIBUTING gate 2, changes scoped to the scheduler extender may be validated with unit tests rather than GPU hardware, and this is within that scope.

With the cluster default set to `binpack`, one node, and three cards that differ only in current load:

```
cluster default = "binpack"
  annotation "binpack" -> placed on gpu-busy
  annotation "spread"  -> placed on gpu-idle
  no annotation        -> placed on gpu-busy    (configured default honoured)
  annotation "binpakc" -> placed on gpu-idle    (unrecognized value; configured default ignored)
```

And confirming that the configured value is discarded rather than used as a fallback:

```
configured cluster default = "binpack", pod annotation = "binpakc", resolved = "binpakc"
```

On a cluster, the equivalent would be an installation with `scheduler.defaultSchedulerPolicy.gpuSchedulerPolicy: binpack`, a node holding both a loaded and an idle card, and:

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: policy-typo
  annotations:
    hami.io/gpu-scheduler-policy: "binpakc"   # note the misspelling
spec:
  containers:
  - name: cuda
    resources:
      limits:
        nvidia.com/gpu: 1
        nvidia.com/gpumem: 1024
```

Expected under `binpack`: the loaded card. Per the unit tests: the idle card.

I am happy to attach the test files, or to open them as part of a PR.

Scope, to be clear about what this is not: there is no crash, no overcommit, and no quota bypass. The Pod runs. This is placement correctness and the absence of any diagnostic, nothing more.

Known limitation of this report. The hardcoded fallbacks happen to coincide with the shipped defaults — `gpuSchedulerPolicy: spread` and `nodeSchedulerPolicy: binpack` (`charts/hami/values.yaml:77-78`, `cmd/scheduler/main.go:70-71`). On a default installation the behaviour is therefore indistinguishable from correct. It only becomes observable for an operator who changed the Helm value, which is also the operator most likely to care about the result.

`NodeScoreList.Less` (`pkg/scheduler/policy/node_policy.go:48`) resolves unrecognized values the same way, falling through to `// default policy is Binpack` at `:52`, so a fix will need to cover both paths.

Two possible directions, whichever fits the project's intent:

1. Reject at admission when the annotation is not a recognized policy, matching `validateMemoryPercentage` in `pkg/device/nvidia/device.go`, which already returns an error for an out-of-range `gpumem-percentage`.
2. Fall back to the configured `scheduler.defaultSchedulerPolicy` and emit a warning event. This preserves current scheduling behaviour for anyone relying on it.

I would lean towards 2, since it makes the annotation behave the way the documentation describes without changing admission behaviour, but I am happy to implement whichever the maintainers prefer.

Related precedent for validating user-controlled scheduling input rather than letting it fail silently: #1997 (gpumem-percentage range), #2388 (hygon, metax-sgpu bounds), #2601 (ascend bounds).

_This report was prepared with AI assistance (per CONTRIBUTING.md); all claims, line references and test results were reviewed and verified by me._

**Environment**:
- HAMi version: `master` @ ee92fc9
- nvidia driver or other AI device driver version: n/a - not exercised; this is in the scheduler extender
- Docker version from `docker version`: n/a
- Docker command, image and tag used: n/a
- Kernel version from `uname -a`: n/a
- Others: found by code inspection and reproduced with unit tests (Go 1.26.6, linux/amd64, CGO enabled). No cluster or GPU was involved.

IIITManjeet:
**What type of PR is this?**

/kind bug

`hami.io/gpu-scheduler-policy` and `hami.io/node-scheduler-policy` are user-set Pod annotations, documented as a way to overlay the scheduler configuration. Both were read as unconditional overrides: `GetGPUSchedulerPolicyByPod` (`pkg/util/util.go`) and `resolveNodeSchedulerPolicy` (`pkg/scheduler/score.go`) return the annotation value whenever the key is present, valid or not.

An unrecognized value therefore replaced the configured policy and reached the sort comparators, which apply their own defaults rather than `scheduler.defaultSchedulerPolicy` — `gpu_policy.go` falls through to spread, `node_policy.go` to binpack. On a cluster configured for binpack, a Pod with a misspelled `gpu-scheduler-policy` was placed with spread, and nothing was logged at any point.

This PR adds `IsValidGPUSchedulerPolicy` and `IsValidNodeSchedulerPolicy`, and applies the annotation only when it names a known policy. An unrecognized value is ignored with a warning, so the configured policy stands.

### Changes

* Add `IsValidGPUSchedulerPolicy`, which reads the value as a comma-separated ordered list the same way `PolicyContains` and `gpuSortKeyChain` do, and accepts `binpack`, `spread`, `numa`, `mutex` and `topology-aware`.
* Add `IsValidNodeSchedulerPolicy`, accepting `binpack` and `spread`. `NodeScoreList.Less` has no chain form, so only a single name is meaningful there.
* Gate both annotation lookups on the corresponding validator, logging a warning and keeping the configured policy when the value is not recognized.
* Add unit coverage for the validators, for the resolution behaviour, and for end-to-end placement through `fitInDevices` and the NVIDIA backend's `Fit()`.

Fixes #2767

The change is deliberately scoped to the two Pod annotations.

The `--gpu-scheduler-policy` and `--node-scheduler-policy` flags are not validated. They feed in as the fallback value and never pass through the validator, so operator-level typos behave exactly as they do today.

The comparators themselves are untouched. They only receive the resolved policy string and have no access to `device.GPUSchedulerPolicy` or `config.NodeSchedulerPolicy`, so resolving the value before it reaches them is the smaller change; making the comparators aware of the configured default would mean threading it through the sort path.

Over-rejection was the main regression risk, since `mutex` and `topology-aware` are filters consumed via `PolicyContains` in each backend's `Fit()` rather than sort keys, and chains such as `binpack,numa` are also valid. All of those forms are covered by tests. Every policy value that appears in the repository's docs, examples, charts and tests is `binpack`, `spread` or `mutex`, and all are accepted.

Two known limits, neither introduced here. A value with surrounding whitespace such as `" binpack"` passes validation, because entries are trimmed to match `PolicyContains`, but `DeviceUsageList.Less` compares the whole string exactly and still falls through — pre-existing behaviour that this PR does not change; not trimming would have rejected `binpack, numa`, which works today. And the warning is a log line rather than a Pod event, because `pkg/util` has no event recorder in scope; a misconfigured pending Pod will log it once per scheduling attempt. Happy to change either if preferred.

Yes. A Pod whose `hami.io/gpu-scheduler-policy` or `hami.io/node-scheduler-policy` annotation is not a recognized policy is now scheduled with the configured `scheduler.defaultSchedulerPolicy` instead of the comparator's own default, and the ignored value is logged. Pods with a valid annotation, and Pods with no annotation, are unaffected.

## Validation

Run on Linux x86_64 (WSL2, Ubuntu 22.04), Go 1.26.6, CGO enabled:

* `go test ./pkg/util/... ./pkg/scheduler/... ./pkg/device/... -count=1` — all packages pass
* `go test -race ./pkg/util/... ./pkg/scheduler/... -count=1` — all packages pass
* `go vet ./pkg/util/... ./pkg/scheduler/...` — clean
* `gofmt -l` on the changed files — clean
* `golangci-lint run --new-from-rev=upstream/master ./pkg/util/... ./pkg/scheduler/...` — 0 issues
* License-header and import-alias checks from `make verify` — pass

No GPU hardware was used. The change is scoped to the scheduler extender.

## AI Assistance Disclosure
This PR was written with AI assistance (Claude Code), covering the codebase
investigation, implementation, and tests. I reviewed
the change against the existing code and ran the validation listed above.

## Summary by CodeRabbit

* **Bug Fixes**
  * Invalid or unrecognized scheduler policy annotations now fall back to the configured default policy.
  * GPU scheduling overrides are applied only when valid policy values are provided.
  * Surrounding whitespace is handled correctly for valid node scheduling policies.
  * Warnings are issued when invalid policy annotations are detected.

* **Tests**
  * Added coverage for valid, invalid, missing, empty, whitespace-padded, and case-sensitive policy values.
  * Verified fallback behavior across supported bin-packing and spreading placement scenarios.


pkg/scheduler/policy_fallback_placement_test.go
```diff
@@ -0,0 +1,184 @@
+/*
+Copyright 2024 The HAMi Authors.
```
chetanmehra07: please make it 2026. as you are making this file in 2026 please keep it 2026.

pkg/scheduler/policy_fallback_placement_test.go
```diff
@@ -0,0 +1,184 @@
...
+	// With no annotation the configured cluster default (binpack) is honoured.
+	assert.Equal(t, noAnnotation, binpack,
+		"absent annotation should use the configured cluster default")
+
+	// An unrecognized value is ignored, so placement matches the configured
+	// default rather than the comparator's hardcoded spread fallback.
+	assert.Equal(t, unrecognized, noAnnotation,
+		"unrecognized policy should place identically to no annotation at all")
+	assert.Assert(t, unrecognized != spread,
+		"unrecognized policy must not silently produce spread placement")
```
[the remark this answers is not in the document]
IIITManjeet: have now added the both defaults (binpack and spread) are now exercised in all three places

pkg/util/util.go
```diff
@@ -274,11 +274,52 @@ func RemoveNodeAnnotation(node *corev1.Node, annotationKeys ...string) error {
...
+		}
+	}
+	return true
+}
+
+// IsValidNodeSchedulerPolicy reports whether policy names a node scheduling
+// policy HAMi recognizes. NodeScoreList.Less has no chain form, so only a
+// single name is meaningful here.
+func IsValidNodeSchedulerPolicy(policy string) bool {
+	switch SchedulerPolicyName(policy) {
```
mesutoezdil: the gpu validator trims each segment, this one does not trim at all, so a trailing space is valid on one annotation and rejected on the other. trim here too.

IIITManjeet: Sure @mesutoezdil will add here to. And will push again with the changes

IIITManjeet:
Hello @mesutoezdil sir have pushed the changes can you please review it once.
Thanks

pkg/util/util.go
```diff
@@ -274,11 +274,54 @@ func RemoveNodeAnnotation(node *corev1.Node, annotationKeys ...string) error {
...
+// with a warning, so defaultPolicy stands.
 func GetGPUSchedulerPolicyByPod(defaultPolicy string, task *corev1.Pod) string {
 	userGPUPolicy := defaultPolicy
 	if task != nil && task.Annotations != nil {
 		if value, ok := task.Annotations[GPUSchedulerPolicyAnnotationKey]; ok {
-			userGPUPolicy = value
+			if IsValidGPUSchedulerPolicy(value) {
+				userGPUPolicy = value
+			} else {
+				klog.Warningf("ignoring unrecognized %s=%q on pod %s/%s, using configured policy %q",
```
Shouren, line 322: @IIITManjeet I am considering rejecting the invalid policy instead of falling back. How do you think?

mesutoezdil: i would keep the fallback. the webhook can be disabled or bypassed, so the scheduler needs this safety net anyway. rejection can come as a follow up in the webhook, #2813 already proposed that part.

IIITManjeet: @Shouren @mesutoezdil sir what should be my next step for this PR can you please help here. Thanks

mesutoezdil: no change needed from your side. the fallback approach stands, we will handle admission time rejection as a separate follow up. waiting for @Shouren to confirm, then this can get lgtm.

IIITManjeet: Sure Thanks a lot

Shouren:
@IIITManjeet Please rename the `policy_fallback_placement_test.go` to `scheduler_policy_test.go`

IIITManjeet:
> @IIITManjeet Please rename the `policy_fallback_placement_test.go` to `scheduler_policy_test.go`

Hii @Shouren have renamed and pushed the changes. 
Best
Manjeet

Shouren (approved):
/lgtm

PR #2775, opened 2026-08-22
fix(device-plugin): remove panic(0) traps from getAPIDevices and dead util code

devGPP23:
**Problem**

While going through the device plugin code, I noticed that `getAPIDevices()` in `register.go` calls panic(0) whenever any NVML query fails,for example if NVML `init` fails, or a GPU drops off the PCIe bus.

Since this code runs inside a Kubernetes `DaemonSet`, I saw that one bad GPU query would crash the entire device plugin pod on that node. The pod then gets stuck in `CrashLoopBackOff`, and all the healthy GPUs on that node also become unusable until someone intervenes manually. I checked the official NVIDIA/k8s-device-plugin repo and noticed they handle these exact failures by returning normal Go errors instead of panicking.

I also noticed 4 more panic(0) traps in `GetIndexAndTypeFromUUID()` in `util.go`, but when I searched the whole repo, I found this function is never called anywhere ,it is dead code.

**Solution**

I did the following to fix this:

I changed getAPIDevices() to return an (*[]*device.DeviceInfo, error) instead of panicking. All 5 panic(0) calls are now replaced with proper wrapped errors using fmt.Errorf().
I updated RegisterInAnnotation() to catch this new error, log it with klog.ErrorS, and return it. The caller (WatchAndRegister) already retries every 5 seconds on error, so now the plugin safely retries instead of crashing.
I deleted the dead function GetIndexAndTypeFromUUID() from util.go, which removes the remaining 4 panic traps.
I updated the unit tests: instead of testing panics with recover(), they now assert on returned errors. I also added new tests covering every error path and a healthy-device success path.
After my change, I verified there are 0 instances of panic(0) left in the whole nvinternal package.

**New tests added:**

TestGetAPIDevicesErrorsOnNVMLInitFailure
TestRegisterInAnnotationPropagatesNVMLInitError
TestGetAPIDevicesErrorsOnPerDeviceNVMLFailures (handle / index / memory / name failures)
TestGetAPIDevicesRegistersHealthyDevice

**AI DISCLOSURE**

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved NVIDIA GPU discovery to report initialization, device lookup, memory, and naming errors instead of crashing.
  * Device registration now propagates discovery failures and records warning events when discovery or node updates fail.
  * Healthy devices continue to register successfully when other devices encounter errors.

* **Reliability**
  * Improved error handling and diagnostics during NVIDIA device plugin startup and registration.

### DIFF
--- a/pkg/device-plugin/nvidiadevice/nvinternal/plugin/register.go
+++ b/pkg/device-plugin/nvidiadevice/nvinternal/plugin/register.go
@@ -92,12 +92,12 @@ var nvmlInit = nvml.Init
 // calculateGPUScore is overridable in tests to simulate topology-score calculation failures.
 var calculateGPUScore = nvidia.CalculateGPUScore
 
-func (plugin *NvidiaDevicePlugin) getAPIDevices() *[]*device.DeviceInfo {
+func (plugin *NvidiaDevicePlugin) getAPIDevices() (*[]*device.DeviceInfo, error) {
 	devs := plugin.Devices()
 	klog.V(5).InfoS("getAPIDevices", "devices", devs)
 	if nvret := nvmlInit(); nvret != nvml.SUCCESS {
 		klog.Errorln("nvml Init err: ", nvret)
-		panic(0)
+		return nil, fmt.Errorf("nvml init failed: %v", nvret)
 	}
 	// Shutdown is deferred only after Init succeeds, since calling it after a failed Init crashes the process.
 	defer nvml.Shutdown()
@@ -112,13 +112,13 @@ func (plugin *NvidiaDevicePlugin) getAPIDevices() *[]*device.DeviceInfo {
 	for UUID := range devs {
 		ndev, ret := nvml.DeviceGetHandleByUUID(UUID)
 		if ret != nvml.SUCCESS {
-			klog.Errorln("nvml new device by index error uuid=", UUID, "err=", ret)
-			panic(0)
+			klog.Errorf("skipping device uuid=%s: nvml DeviceGetHandleByUUID failed: %v", UUID, ret)
+			continue
 		}
 		idx, ret := ndev.GetIndex()
 		if ret != nvml.SUCCESS {
-			klog.Errorln("nvml get index error ret=", ret)
-			panic(0)
+			klog.Errorf("skipping device uuid=%s: nvml GetIndex failed: %v", UUID, ret)
+			continue
 		}
 		memoryTotal := 0
 		memory, ret := ndev.GetMemoryInfo()
@@ -139,13 +139,13 @@ func (plugin *NvidiaDevicePlugin) getAPIDevices() *[]*device.DeviceInfo {
 				continue
 			}
 		default:
-			klog.Error("nvml get memory error ret=", ret)
-			panic(0)
+			klog.Errorf("skipping device uuid=%s: nvml GetMemoryInfo failed: %v", UUID, ret)
+			continue
 		}
 		Model, ret := ndev.GetName()
 		if ret != nvml.SUCCESS {
-			klog.Error("nvml get name error ret=", ret)
-			panic(0)
+			klog.Errorf("skipping device uuid=%s: nvml GetName failed: %v", UUID, ret)
+			continue
 		}
 
 		registeredmem := int32(memoryTotal / 1024 / 1024)
@@ -206,7 +206,7 @@ func (plugin *NvidiaDevicePlugin) getAPIDevices() *[]*device.DeviceInfo {
 		res = append(res, info)
 		klog.V(3).Infof("Registered device id=%v, memory=%vMB, type=%v, numa=%v, health=%v", idx, registeredmem, Model, numa, health)
 	}
-	return &res
+	return &res, nil
 }
 
 func (plugin *NvidiaDevicePlugin) discoverMigProfiles(dev nvml.Device, model string) []device.MigProfile {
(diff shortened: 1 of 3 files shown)


pkg/device-plugin/nvidiadevice/nvinternal/plugin/register.go
```diff
@@ -256,7 +256,11 @@ func (plugin *NvidiaDevicePlugin) discoverMigProfiles(dev nvml.Device, model str
 // RegisterInAnnotation scans devices and patches node annotations.
 // Returns (changed, error) where changed indicates whether the annotation was actually updated.
 func (plugin *NvidiaDevicePlugin) RegisterInAnnotation() (bool, error) {
-	devices := plugin.getAPIDevices()
+	devices, err := plugin.getAPIDevices()
+	if err != nil {
```
mesutoezdil, line 263: crashloop was ugly but visible. a forever retry is silent, the node just stops updating its annotation. emit a node event or metric when registration keeps failing?

devGPP23: Thanks for the feedback @mesutoezdil ! 

I've updated the PR to emit a deduplicated Warning event (RegistrationFailed) on the Node via util.EmitNodeWarningEvent whenever device discovery or annotation patching fails (with a 10-minute dedup window to avoid event spamming).

mesutoezdil:
pls resolve conflicts

pkg/device-plugin/nvidiadevice/nvinternal/plugin/register.go
```diff
@@ -256,7 +256,16 @@ func (plugin *NvidiaDevicePlugin) discoverMigProfiles(dev nvml.Device, model str
...
+		klog.ErrorS(err, "failed to get API devices")
+		if node, nodeErr := util.GetNode(util.NodeName); nodeErr == nil {
```
Shouren: @devGPP23 `RegisterInAnnotation` retries every 5 seconds if error is not nil, so there will be a lots of requests to get node object in API Server. I think we should take this into consideration.

pkg/device-plugin/nvidiadevice/nvinternal/plugin/register.go
```diff
@@ -110,12 +110,12 @@ func (plugin *NvidiaDevicePlugin) getAPIDevices() *[]*device.DeviceInfo {
...
 		if ret != nvml.SUCCESS {
 			klog.Errorln("nvml new device by index error uuid=", UUID, "err=", ret)
-			panic(0)
+			return nil, fmt.Errorf("failed to get device handle for uuid %s: %v", UUID, ret)
 		}
 		idx, ret := ndev.GetIndex()
 		if ret != nvml.SUCCESS {
 			klog.Errorln("nvml get index error ret=", ret)
-			panic(0)
+			return nil, fmt.Errorf("failed to get device index: %v", ret)
```
Shouren: @devGPP23 for this and the following errors, I think it is better to add `UUID` in error message.

pkg/device-plugin/nvidiadevice/nvinternal/plugin/register.go
devGPP23: Hi  @Shouren, I've updated the PR based on your feedback i.e.
Fixed the API server load issue. Now, if getting the devices fails, it just returns the error right away instead of calling util.GetNode() and spamming the API server every 5 seconds. (The event is only emitted if the actual node patch fails).
Please take a look whenever you are free

pkg/device-plugin/nvidiadevice/nvinternal/plugin/register.go
devGPP23: Hello @Shouren, I've Added the UUID to all the error messages so we know exactly which device fails
Please take a look whenever you are free

devGPP23:
Hi @mesutoezdil , I've updated the PR based on your feedback i.e.
I have fixed all the merge conflicts with the latest master,
Please take a look whenever you are free

Shouren:
@mesutoezdil @archlitchi How do you think about only calling `klog.ErrorS` when `plugin.getAPIDevices()` returns error?

mesutoezdil:
log alone is silent, nobody watches plugin pod logs and the annotation just goes stale. the event is visible on the node and deduplicated, i would keep it.

pkg/device-plugin/nvidiadevice/nvinternal/plugin/register.go
```diff
@@ -113,12 +113,12 @@ func (plugin *NvidiaDevicePlugin) getAPIDevices() *[]*device.DeviceInfo {
 		ndev, ret := nvml.DeviceGetHandleByUUID(UUID)
 		if ret != nvml.SUCCESS {
 			klog.Errorln("nvml new device by index error uuid=", UUID, "err=", ret)
-			panic(0)
+			return nil, fmt.Errorf("failed to get device handle for uuid=%s: %v", UUID, ret)
```
mesutoezdil: this still aborts the whole scan on one bad device, so the healthy GPUs on that node stay unregistered. the pr says that is the problem being fixed. should the loop skip the device instead?

devGPP23: Hello @mesutoezdil,
I've updated the loop so that per-device failures (GetHandleByUUID, GetIndex, GetMemoryInfo, GetName) now log the error and continue to skip just that device. Healthy GPUs on the same node will still get registered normally.
nvml.Init failure still returns an error since no device can be queried without a working NVML session.

 Please take another look whenever u have time

archlitchi (approved):
/lgtm

PR #2580, opened 2026-08-11
fix(vGPUmonitor): add missing node label to host GPU metrics

SK8-infi:
**What type of PR is this?**
/kind bug

**What this PR does / why we need it**: In cmd/vGPUmonitor/metrics.go, the metric descriptors for host-level GPU memory and utilization (hostGPUdesc and hostGPUUtilizationdesc) do not include the node label ([]string{"device_index", "device_uuid", "device_type"}).

Without node label, Prometheus metrics scraped across nodes cannot be grouped/filtered by node in Grafana. And PromQL joins between scheduler and monitor host metrics fail due to mismatched label keys.

I guess host GPU metrics in vGPUmonitor should include the node label to align with the scheduler metrics.

**Which issue(s) this PR fixes**:
Fixes #2578 

**Special notes for your reviewer**: Nope

image

disclosure: Antigravity was used to audit.

**Does this PR introduce a user-facing change?**:
```release-note
fix(vGPUmonitor): add missing node label to host GPU metrics (hami_host_gpu_memory_used_bytes, hami_host_gpu_utilization_ratio)
```

## Summary by CodeRabbit

* **New Features**
  * Host GPU memory and utilization metrics now include the node name for improved identification.
  * Legacy GPU metrics now include the corresponding node identifier.
  * GPU utilization reporting now uses utilization-rate data for more accurate metrics.

* **Bug Fixes**
  * GPU metric collection now reports an error when the node name is unavailable, preventing incomplete metric data.

PR #2776, opened 2026-08-22
Add Missing Node Label to Memory Controller Metric in vGPUmonitor (Fixes #2771)

### LINKED ISSUE #2771: bug(vGPUmonitor): hami_host_gpu_memory_controller_utilization_ratio  missing  node  label
### What happened?

The `hami_host_gpu_memory_controller_utilization_ratio` metric introduced in #2580 is missing the `node` label in both its descriptor and its `sendMetric` call.

PR #2580 correctly added the `node` label to `hostGPUdesc` and `hostGPUUtilizationdesc`, but the new `hostGPUMemoryUtilizationdesc` descriptor that was introduced in the same PR was not updated to include it.

All `hami_host_gpu_*` metrics should consistently include the `node` label so that per-node aggregation works in PromQL and Grafana dashboards.

### How to reproduce

1. Deploy HAMi on a multi-node cluster with the latest `master` build
2. Scrape `/metrics` from the vGPU monitor
3. Observe that `hami_host_gpu_memory_controller_utilization_ratio` does not have a `node` label, while `hami_host_gpu_memory_used_bytes` and `hami_host_gpu_utilization_ratio` do

### Root cause

In `cmd/vGPUmonitor/metrics.go`:

**Descriptor (line ~70):**
```go
hostGPUMemoryUtilizationdesc = prometheus.NewDesc(
    "hami_host_gpu_memory_controller_utilization_ratio",
    "GPU memory controller utilization ratio (0-100)",
    // ^^^ missing "node" as the first label
)

if err := sendMetric(ch, hostGPUMemoryUtilizationdesc, prometheus.GaugeValue,
    float64(utilRates.Memory),
    // ^^^ missing nodeName as the first label value
); err != nil {

### PR DESCRIPTION
**Pull request by Gurkaran18**
### Description

This PR addresses issue #2771 by fixing a label inconsistency in the `vGPUmonitor` Prometheus collector.

The `hami_host_gpu_memory_controller_utilization_ratio` metric, which was recently introduced, was missing the `node` label in both its descriptor and its emission call. This omission breaks per-node aggregation in PromQL (e.g., `sum by (node)`) and dashboard filtering, as all other `hami_host_gpu_*` metrics require the `node` label.

This fix:
1. Adds the `"node"` label to the `hostGPUMemoryUtilizationdesc` descriptor.
2. Passes the `nodeName` variable as the first label value in the `sendMetric` call.

### Testing

- Verified `go vet` and `go test ./cmd/vGPUmonitor/...` pass cleanly.
- Expanded the existing `TestHostGPUMetricsDescriptorsIncludeNodeLabel` unit test to explicitly assert that `hostGPUMemoryUtilizationdesc` contains the `"node"` label, ensuring this regression doesn't happen again.
- 
### Release Note

```release-note
`hami_host_gpu_memory_controller_utilization_ratio` now includes the `node` label, matching all other `hami_host_gpu_*` metrics. This changes the series identity for anyone already scraping this metric — recording rules, alerts, or dashboards that reference it without the `node` label will need to be updated.

> This PR was authored with the assistance of Antigravity AI. I have fully reviewed the implementation to ensure the metric labels are emitted correctly according to the established codebase patterns and have validated that it conforms to all HAMi contribution guidelines.


## Summary by CodeRabbit

## Summary by CodeRabbit

* **Bug Fixes**
  * Host GPU memory-controller utilization metrics now include the node label, improving identification and filtering across multiple nodes.
  * Node information is now correctly included when collecting these metrics, providing more accurate monitoring data.
* **Tests**
  * Expanded host GPU metric validation to confirm that the node label is present on applicable memory-utilization metrics and matches labels exactly.

### CODE REVIEW

File: cmd/vGPUmonitor/metrics.go [hunk 1]
```diff
@@ -366,17 +366,18 @@ func (cc ClusterManagerCollector) collectGPUUtilizationMetrics(ch chan<- prometh
...
-		nodeName, fmt.Sprint(index), uuid, deviceName,
-	)
-
 	if err := sendMetric(ch, hostGPUMemoryUtilizationdesc, prometheus.GaugeValue,
 		float64(utilRates.Memory),
-		fmt.Sprint(index), uuid, deviceName,
+		nodeName, fmt.Sprint(index), uuid, deviceName,
 	); err != nil {
-		return fmt.Errorf("nvml send memory controller utilization: %w", err)
+		klog.Errorf("Failed to send host memory utilization metric: %v", err)
```
mesutoezdil: keep this pr label only, two lines. the return to log change and the legacy reorder belong to #2716, which already rewrites this function and will conflict with you.

mesutoezdil (dismissed):

cmd/vGPUmonitor/metrics.go
Gurkaran18: I have  pushed an update to  isolate this PR to only the two necessary lines , So it won't conflict with #2716.

FouoF:
This is a small but user face change, you should update the doc on the https://github.com/Project-HAMi/website but the website v2.10.0 has not released so i'm afraid you need to wait some time.

Gurkaran18:
@FouoF Since Project-HAMi/website is still tracking the older release, I'll hold off and open a docs PR there once v2.10.0 ships so it lands with the right version. I'll keep a note on my end so it doesn't get lost.

cmd/vGPUmonitor/metrics_test.go
```diff
@@ -80,6 +80,12 @@ func TestHostGPUMetricsDescriptorsIncludeNodeLabel(t *testing.T) {
 		t.Errorf("hostGPUUtilizationdesc does not contain 'node' label: %s", hostGPUUtilString)
 	}
 
+	hostGPUMemoryUtilString := hostGPUMemoryUtilizationdesc.String()
+	if !strings.Contains(hostGPUMemoryUtilString, `"node"`) && !strings.Contains(hostGPUMemoryUtilString, `node`) {
```
mesutoezdil: any string with "node" also contains node, so the first check never decides anything. matching the label list would be tighter than a substring.

mesutoezdil:
adding a label changes the series identity for anyone already scraping this metric. could you add a /kind and a release-note block?

cmd/vGPUmonitor/metrics_test.go
Gurkaran18: Added descHasLabel helper that parses the descriptor's actual label list instead of doing a substring check

Gurkaran18:
@mesutoezdil rebased onto master to resolve the conflict, added the /kind bug + release-note, and addressed the two CodeRabbit findings (docstrings, slices.Contains)

mesutoezdil:
dco pls

Gurkaran18:
@mesutoezdil DCO is fixed, all 5 commits now carry Signed-off-by trailers.

mesutoezdil:
lint is red from an unrelated master conflict (duplicate effectivePodDeviceUsage), already fixed upstream, rebasing onto master should clear it

cmd/vGPUmonitor/metrics.go
```diff
@@ -69,7 +69,7 @@ var (
 	hostGPUMemoryUtilizationdesc = prometheus.NewDesc(
```
mesutoezdil, line 69: what happens if label values are out of order?

cmd/vGPUmonitor/metrics.go
```diff
@@ -406,7 +408,7 @@ func (cc ClusterManagerCollector) collectGPUUtilizationMetrics(ch chan<- prometh
 
 	if err := sendMetric(ch, hostGPUMemoryUtilizationdesc, prometheus.GaugeValue,
 		float64(utilRates.Memory),
-		fmt.Sprint(index), identity.uuid, identity.deviceName,
+		identity.nodeName, fmt.Sprint(index), identity.uuid, identity.deviceName,
 	); err != nil {
 		return fmt.Errorf("nvml send memory controller utilization: %w", err)
```
mesutoezdil, line 413: did you check for existing duplicate prs first?

cmd/vGPUmonitor/metrics.go
Gurkaran18: It wouldn't throw an error, the values just get matched by position, so the metric would end up with the wrong labels. I kept the same order the other host metrics use (node, device_index, device_uuid, device_type). I also updated the test to check each label value by name, so a swapped order fails now.

cmd/vGPUmonitor/metrics.go
Gurkaran18: Yes, I checked. #2580 added the node label to the other two host metrics but missed this one, and I couldn't find any other PR fixing it.

archlitchi (approved):
/lgtm

PR #2780, opened 2026-08-22
[kind/bug] Restore model-aware scheduling for Cambricon MLU devices

### LINKED ISSUE #2774: Restore model-aware scheduling for Cambricon MLU devices

I would like HAMi to support model-based scheduling for Cambricon MLU devices.

HAMi defines these pod annotations:

- `cambricon.com/use-mlutype`
- `cambricon.com/nouse-mlutype`

However, the Cambricon scheduler does not currently use them.

The `checkType` function receives the pod annotations and device information, but it does not check the requested or excluded MLU model. It accepts every request whose device type is `MLU`.

There is another limitation: `GetNodeDevices` registers every Cambricon device with the generic type `MLU`. Therefore, the scheduler cannot distinguish between models such as MLU370 and MLU590.

### Current behavior

For example, a user may add:

```yaml
metadata:
  annotations:
    cambricon.com/nouse-mlutype: "MLU370"
```

The user expects HAMi to avoid MLU370 devices. However, all Cambricon devices are treated as the same generic `MLU` type, so the annotation does not affect scheduling.

### Expected behavior

HAMi should know the real model of each Cambricon device.

It should then apply the model-selection annotations as follows:

- `cambricon.com/use-mlutype` allows only the listed MLU models.
- `cambricon.com/nouse-mlutype` excludes the listed MLU models.
- When neither annotation is present, existing scheduling behavior remains unchanged.
- Empty annotation values should not block scheduling.
- Matching behavior should follow the existing `device.CheckType` helper used by other device backends.

### Behavior diagram

~~~mermaid
flowchart LR
    subgraph Current
        A[Pod requests an MLU model] --> B[HAMi reads the request]
        B --> C[Every device has type MLU]
        C --> D[Model annotation is ignored]
        D --> E[Any MLU can be selected]
    end

    subgraph Expected
        F[Cambricon device reports its real model] --> G[HAMi stores the model]
        H[Pod contains use-mlutype or nouse-mlutype] --> I[HAMi checks the model]
        G --> I
        I --> J{Model allowed?}
        J -->|Yes| K[Device can be selected]
        J -->|No| L[Device is excluded]
    end
~~~

## Why is this needed?

Clusters may contain different Cambricon models, such as MLU370 and MLU590.

Some workloads may support only a specific model. Other workloads may need to avoid a model because of compatibility, performance, or availability requirements.

At present, users can set the model-selection annotations, but the scheduler cannot enforce them. This may allow a workload to be scheduled onto an unwanted or incompatible MLU model.

Supporting these annotations would make Cambricon scheduling consistent with model-selection support in other HAMi device backends.

## Proposed solution

The exact source of the MLU model should be confirmed with maintainers before implementation.

A possible implementation is:

1. Obtain the real model name for every Cambricon device from a reliable device-plugin, node-annotation, or node-resource source.
2. Store that model in the device's `DeviceInfo.Type`.
3. Update the Cambricon `checkType` function to call:

   ```go
   device.CheckType(annos, d.Type, MLUInUse, MLUNoUse)
   ```

4. Keep the generic backend/vendor identifier as `MLU` where it is needed for internal device routing.
5. Add unit tests for allowed, excluded, empty, comma-separated, and missing annotations.
6. Confirm that pods without these annotations keep the current scheduling behavior.

Adding only the `device.CheckType` call is not enough because every Cambricon device is currently registered with the same generic `MLU` type.

Relevant code:

- `pkg/device/cambricon/device.go`
- `MLUInUse`
- `MLUNoUse`
- `CambriconDevices.GetNodeDevices`
- `CambriconDevices.checkType`
- `CambriconDevices.Fit`
- `pkg/device/cambricon/device_test.go`
- `pkg/device/devices.go`
- `device.CheckType`

### Historical context

Older HAMi code contained Cambricon model-filtering logic. The current implementation still contains the annotation constants, but the model check is no longer connected to the scheduling path.

Therefore, this may be a regression or unfinished restoration of an older capability rather than a completely new feature.

### Related roadmap work

Roadmap issue #923 includes:

> Support multiple cambricon types (370,590,etc..)

Related issue:

- https://github.com/Project-HAMi/HAMi/issues/923

This proposal is a focused description of the model-discovery and annotation-filtering behavior. Please confirm whether maintainers prefer tracking it in #923 or in a separate issue.

## Environment

- HAMi source revision reviewed: `9c93c8442259496499d079c12f7664c1d955cf72`
- Component: HAMi scheduler and Cambricon device backend
- Cambricon model: Not tested on hardware
- Cambricon driver version: Not tested
- Kubernetes version: Not tested
- Operating system: Not tested
- Validation status: Source-code and history analysis only

## AI assistance disclosure

I used OpenAI Codex to help analyze the source code and organize this description.

asadjan4611:
/kind bug

This PR fixes Cambricon MLU model-aware scheduling.

HAMi defines the following pod annotations:

- `cambricon.com/use-mlutype`
- `cambricon.com/nouse-mlutype`

Previously, the Cambricon scheduler did not use these annotations. Every Cambricon device was recorded with the generic type `MLU`. Therefore, HAMi could not distinguish models such as `MLU370-X4` and `MLU590`.

This could allow a pod to be scheduled on an MLU model that it did not request or explicitly excluded.

This PR:

- Reads the MLU model from the node's `Model` label.
- Stores the real model in the HAMi device information when the label contains `MLU`.
- Falls back to the generic `MLU` type when the label is missing, empty, or does not contain `MLU`.
- Uses the existing `device.CheckType` helper to check the model annotations.
- Allows devices that match `use-mlutype`.
- Rejects devices that match `nouse-mlutype`.
- Preserves existing scheduling behavior when no model constraint is provided.
- Rejects constrained requests when the device model cannot be verified.

### Behavior before this fix

```text
Node label: Model=MLU370-X4
              |
              v
HAMi records Type="MLU"
              |
              v
Model annotations are ignored
              |
              v
Any healthy MLU may be selected
```

### Behavior after this fix

```text
Node label: Model=MLU370-X4
              |
              v
HAMi records Type="MLU370-X4"
              |
              v
Check the pod's model annotations
          /               \
         /                 \
Model is allowed       Model is excluded
       |                       |
       v                       v
Schedule device           Reject device
```

### Behavior when the model label is missing or invalid

```text
Model label contains "MLU" -> Use the real model
Model label is missing     -> Fall back to "MLU"
Model label lacks "MLU"    -> Fall back to "MLU"

No model constraint        -> Preserve existing scheduling behavior
Model constraint set       -> Reject the device because its model cannot be verified
```

This change only affects Cambricon model filtering in the scheduler. It does not change device resource counts, memory accounting, core accounting, allocation annotations, or other accelerator backends.

Fixes #2774

Cambricon device-plugin `v2.0.20` introduced automatic MLU node labels. It publishes the bare `Model` label when `--node-label` is enabled:

- [Cambricon device-plugin v2.0.20 documentation](https://github.com/Cambricon/cambricon-k8s-device-plugin/blob/v2.0.20/device-plugin/README.md#mlu-device-label-management)
- [Cambricon device-plugin v2.0.20 label definition](https://github.com/Cambricon/cambricon-k8s-device-plugin/blob/v2.0.20/device-plugin/pkg/nodeLabel/nodelabel.go#L26-L31)

The implementation reuses the existing `device.CheckType` helper used by other device backends.

Validation completed:

- `go test -count=1 ./pkg/device/cambricon` — PASS
- `go test -race -count=1 ./pkg/device/cambricon` — PASS
- `make test` — PASS
- `make verify` — PASS
- `make build` — PASS
- `make tidy` — PASS with no module-file changes

Physical Cambricon MLU hardware was not tested because this change is limited to scheduler logic and is covered by unit and mock tests.

AI assistance disclosure:

I used OpenAI Codex to help analyze the code, implement and review the fix and tests, darfting this pull request description. I reviewed the changes and take responsibility for the contribution. 

Does this PR introduce a user-facing change?**:

```release-note
Cambricon scheduling now honors the `cambricon.com/use-mlutype` and `cambricon.com/nouse-mlutype` annotations. Pods using these annotations may remain pending on nodes whose `Model` label is missing or does not contain `MLU`. Cambricon device-plugin v2.0.20 and later can publish this label when `--node-label` is enabled.
```

## Summary by CodeRabbit

* **New Features**
  * Cambricon devices now identify their specific model from node labels when available.
  * Device selection supports model-specific allow and deny rules, including case-insensitive matching and multiple model values.

* **Bug Fixes**
  * Improved fallback behavior when model information is missing, empty, or unsupported.
  * Corrected device-fit decisions for constrained workloads using model-specific requirements.

### DIFF
--- a/pkg/device/cambricon/device.go
+++ b/pkg/device/cambricon/device.go
@@ -42,11 +42,14 @@ import (
 const (
 	CambriconMLUDevice     = "MLU"
 	CambriconMLUCommonWord = "MLU"
-	MluMemSplitLimit       = "CAMBRICON_SPLIT_MEMS"
-	MluMemSplitIndex       = "CAMBRICON_SPLIT_VISIBLE_DEVICES"
-	MluMemSplitEnable      = "CAMBRICON_SPLIT_ENABLE"
-	MLUInUse               = "cambricon.com/use-mlutype"
-	MLUNoUse               = "cambricon.com/nouse-mlutype"
+	// MLUModelLabel is published by cambricon-k8s-device-plugin when node label
+	// reporting is enabled.
+	MLUModelLabel     = "Model"
+	MluMemSplitLimit  = "CAMBRICON_SPLIT_MEMS"
+	MluMemSplitIndex  = "CAMBRICON_SPLIT_VISIBLE_DEVICES"
+	MluMemSplitEnable = "CAMBRICON_SPLIT_ENABLE"
+	MLUInUse          = "cambricon.com/use-mlutype"
+	MLUNoUse          = "cambricon.com/nouse-mlutype"
 	// MLUUseUUID annotation specifies a comma-separated list of MLU UUIDs to use.
 	MLUUseUUID = "cambricon.com/use-gpuuuid"
 	// MLUNoUseUUID annotation specifies a comma-separated list of MLU UUIDs to exclude.
@@ -214,6 +217,10 @@ func (dev *CambriconDevices) GetNodeDevices(n corev1.Node) ([]*device.DeviceInfo
 		return []*device.DeviceInfo{}, fmt.Errorf("device not found %s", MLUResourceCores)
 	}
 	memoryTotal, _ := n.Status.Capacity.Name(corev1.ResourceName(MLUResourceMemory), resource.DecimalSI).AsInt64()
+	mluType := strings.TrimSpace(n.Labels[MLUModelLabel])
+	if mluType == "" {
+		mluType = CambriconMLUDevice
+	}
 	for int64(i)*100 < cards {
 		nodedevices = append(nodedevices, &device.DeviceInfo{
 			Index:        uint(i),
@@ -221,7 +228,7 @@ func (dev *CambriconDevices) GetNodeDevices(n corev1.Node) ([]*device.DeviceInfo
 			Count:        100,
 			Devmem:       int32(memoryTotal * MemoryFactor * 100 / cards),
 			Devcore:      100,
-			Type:         CambriconMLUDevice,
+			Type:         mluType,
 			Numa:         0,
 			Health:       true,
 			DeviceVendor: CambriconMLUCommonWord,
@@ -242,7 +249,12 @@ func (dev *CambriconDevices) MutateAdmission(ctr *corev1.Container, p *corev1.Po
 
 func (dev *CambriconDevices) checkType(annos map[string]string, d device.DeviceUsage, n device.ContainerDeviceRequest) (bool, bool, bool) {
 	if strings.Compare(n.Type, CambriconMLUDevice) == 0 {
-		return true, true, false
+		model := strings.TrimSpace(d.Type)
+		constrained := strings.TrimSpace(annos[MLUInUse]) != "" || strings.TrimSpace(annos[MLUNoUse]) != ""
+		if constrained && (model == "" || strings.EqualFold(model, CambriconMLUDevice)) {
+			return true, false, false
+		}
+		return true, device.CheckType(annos, d.Type, MLUInUse, MLUNoUse), false
 	}
 	return false, false, false
 }
--- a/pkg/device/cambricon/device.go
+++ b/pkg/device/cambricon/device.go
@@ -42,8 +42,9 @@ import (
 const (
 	CambriconMLUDevice     = "MLU"
 	CambriconMLUCommonWord = "MLU"
-	// MLUModelLabel is published by cambricon-k8s-device-plugin when node label
-	// reporting is enabled.
+	// MLUModelLabel is published by cambricon-k8s-device-plugin v2.0.20 and later
+	// when --node-label is enabled. See:
+	// https://github.com/Cambricon/cambricon-k8s-device-plugin/blob/v2.0.20/device-plugin/README.md#mlu-device-label-management
 	MLUModelLabel     = "Model"
 	MluMemSplitLimit  = "CAMBRICON_SPLIT_MEMS"
 	MluMemSplitIndex  = "CAMBRICON_SPLIT_VISIBLE_DEVICES"
@@ -218,7 +219,7 @@ func (dev *CambriconDevices) GetNodeDevices(n corev1.Node) ([]*device.DeviceInfo
 	}
 	memoryTotal, _ := n.Status.Capacity.Name(corev1.ResourceName(MLUResourceMemory), resource.DecimalSI).AsInt64()
 	mluType := strings.TrimSpace(n.Labels[MLUModelLabel])
-	if mluType == "" {
+	if !strings.Contains(strings.ToUpper(mluType), CambriconMLUDevice) {
 		mluType = CambriconMLUDevice
 	}
 	for int64(i)*100 < cards {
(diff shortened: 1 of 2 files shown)


pkg/device/cambricon/device.go
```diff
@@ -42,11 +42,14 @@ import (
...
 	CambriconMLUDevice     = "MLU"
 	CambriconMLUCommonWord = "MLU"
-	MluMemSplitLimit       = "CAMBRICON_SPLIT_MEMS"
-	MluMemSplitIndex       = "CAMBRICON_SPLIT_VISIBLE_DEVICES"
-	MluMemSplitEnable      = "CAMBRICON_SPLIT_ENABLE"
-	MLUInUse               = "cambricon.com/use-mlutype"
-	MLUNoUse               = "cambricon.com/nouse-mlutype"
+	// MLUModelLabel is published by cambricon-k8s-device-plugin when node label
+	// reporting is enabled.
+	MLUModelLabel     = "Model"
```
mesutoezdil, line 48: which plugin version publishes a bare Model node label? link the doc, the whole feature hangs on that key.

pkg/device/cambricon/device.go
```diff
@@ -214,14 +217,18 @@ func (dev *CambriconDevices) GetNodeDevices(n corev1.Node) ([]*device.DeviceInfo
 		return []*device.DeviceInfo{}, fmt.Errorf("device not found %s", MLUResourceCores)
 	}
 	memoryTotal, _ := n.Status.Capacity.Name(corev1.ResourceName(MLUResourceMemory), resource.DecimalSI).AsInt64()
+	mluType := strings.TrimSpace(n.Labels[MLUModelLabel])
```
mesutoezdil, line 221: if a model label ever lacks the substring mlu, getNodeResources stops matching and the card vanishes from scheduling. fall back to MLU unless the label contains it.

pkg/device/cambricon/device.go
```diff
@@ -214,14 +217,18 @@ func (dev *CambriconDevices) GetNodeDevices(n corev1.Node) ([]*device.DeviceInfo
...
+	if mluType == "" {
```
mesutoezdil: user facing: pods with use-mlutype on unlabeled nodes go pending after this. add a release note.

mesutoezdil (dismissed):

pkg/device/cambricon/device.go
asadjan4611: I resolved that Cambricon device-plugin v2.0.20 introduced node-label support and publishes the bare model label when `--node-label` is enabled. I updated the code comment with a link to the versioned official documentation.

link of the offical documentation is :
https://github.com/Cambricon/cambricon-k8s-device-plugin/blob/v2.0.20/device-plugin/README.md#mlu-device-label-management

pkg/device/cambricon/device.go
asadjan4611: i  have fixed it . HAMi now uses the model label only when it contains MLU, using a case insensitive check. Otherwise, it falls back to the generic MLU type, so the device remains available to the scheduler. I also added a regression test using the label 370-X4.

pkg/device/cambricon/device.go
asadjan4611: updated the release  note

asadjan4611:
@mesutoezdil  can you please review it now i fixed all the issue's.

FouoF (approved):

archlitchi (approved):
/lgtm

PR #2789, opened 2026-08-23
fix(nvidia): reject fractional gpu memory percentages

### LINKED ISSUE #2788: fix(nvidia): reject fractional GPU memory percentage requests
## Describe the bug

HAMi's NVIDIA admission validation can silently round fractional
`nvidia.com/gpumem-percentage` values instead of rejecting them.

The admission path currently obtains the percentage through
`resource.Quantity.Value()`, which converts the quantity to an integer value.

As a result, fractional requests such as:

nvidia.com/gpumem-percentage: "99.1"

can be interpreted as `100` instead of being rejected as an invalid
percentage request.

Similarly, quantities such as:

nvidia.com/gpumem-percentage: "50m"

can be converted to an integer value instead of being rejected as a
non-integer percentage.

## Why this matters

`nvidia.com/gpumem-percentage` represents a percentage and should be an
exact integer value between 0 and 100.

Silently rounding the user's requested value changes the actual scheduling
request.

For example:

Requested:
99.1%

Effective:
100%

This means HAMi can accept a resource request different from what the user
declared, making invalid configurations difficult to detect and debug.

## Expected behavior

GPU memory percentage requests should satisfy:

0 <= gpumem-percentage <= 100

and must be represented as an exact integer.

Valid values such as:

- 0
- 50
- 100

must continue to work unchanged.

Fractional values such as:

- 50m
- 50.5
- 99.1

should be rejected instead of being rounded.

## Suggested fix

Update the NVIDIA GPU memory percentage validation path to use exact
integer conversion through `resource.Quantity.AsInt64()`.

The validation should:

1. Retrieve the configured `gpumem-percentage` quantity.
2. Require `AsInt64()` to succeed.
3. Reject the request if the conversion is not exact.
4. Apply the existing 0-100 range validation.
5. Preserve existing behavior for valid integer percentages.

The validation should be applied at the admission boundary so invalid
fractional requests are rejected before they reach scheduling.

The existing request-generation behavior should also remain consistent with
this validation.

## Regression tests

Add tests covering:

- `0` -> accepted
- `50` -> accepted
- `100` -> accepted
- `50m` -> rejected
- `50.5` -> rejected
- `99.1` -> rejected
- `101` -> rejected
- negative values -> rejected

A regression test should specifically verify that:

requested = 99.1

does not become:

effective = 100

## Compatibility

This change only affects invalid fractional percentage requests.

Existing integer values from 0 through 100 retain their current behavior.

## Impact

This is a resource-validation correctness issue.

HAMi should not silently transform a fractional GPU memory percentage into a
different integer allocation. Rejecting invalid values at admission makes
the resource contract deterministic and gives users a clear configuration
error.

AyushSrivastava1818:
<html>
<body>
<html><head></head><body><h2>What type of PR is this?</h2><p>/kind bug</p><h2>What this PR does / why we need it:</h2><p>Fixes #2788.</p><p>Rejects fractional and non-integer values for <code inline="">nvidia.com/gpumem-percentage</code>, ensuring that only integer percentages in the valid <code inline="">[0, 100]</code> range are accepted.</p><h3>Changes</h3><ul><li><p><strong><code inline="">MutateAdmission</code></strong>: Replaces <code inline="">resourceValue</code>-based handling with <code inline="">resource.Quantity</code> and <code inline="">AsInt64()</code> validation, rejecting fractional/non-integer values such as <code inline="">50m</code>, <code inline="">50.5</code>, and <code inline="">99.1</code>, as well as negative and out-of-range values.</p></li><li><p><strong><code inline="">GenerateResourceRequests</code></strong>: Uses <code inline="">AsInt64()</code> for exact integer validation. Invalid or unparseable memory percentage requests return an empty <code inline="">device.ContainerDeviceRequest{}</code> rather than encoding invalid values through artificial sentinels.</p></li><li><p><strong><code inline="">Fit()</code></strong>: Validates <code inline="">MemPercentagereq</code> as a request-level invariant before entering the per-device loop. Invalid values cause <code inline="">Fit()</code> to fail immediately and increment <code inline="">CardInsufficientMemory</code> exactly once per request. The existing internal <code inline="">101</code> sentinel continues to represent an unset percentage.</p></li></ul><h3>Expected behavior</h3>
Requested value | Result
-- | --
0 | Accepted
50 | Accepted
100 | Accepted
50m | Rejected
50.5 | Rejected
99.1 | Rejected
101 | Rejected as an external request
150 | Rejected
-1 | Rejected

<p>In particular, a request for <code inline="">99.1%</code> must never silently become a <code inline="">100%</code> request.</p><h2>Regression Tests</h2><p>Coverage includes:</p><ul><li><p>Valid boundary values (<code inline="">0</code>, <code inline="">50</code>, <code inline="">100</code>)</p></li><li><p>Fractional values (<code inline="">50m</code>, <code inline="">50.5</code>, <code inline="">99.1</code>)</p></li><li><p>Negative values (<code inline="">-1</code>)</p></li><li><p>Out-of-range values (<code inline="">101</code>, <code inline="">150</code>)</p></li><li><p>Excessively large values</p></li><li><p>Validation through both container <code inline="">Limits</code> and <code inline="">Requests</code></p></li><li><p>GPU memory percentage validation during <code inline="">GenerateResourceRequests</code></p></li><li><p>Invalid requests returning an empty <code inline="">device.ContainerDeviceRequest{}</code></p></li><li><p>Exclusive-core defaulting with fractional memory percentage values</p></li><li><p>Defensive <code inline="">MemPercentagereq</code> validation in <code inline="">Fit()</code></p></li><li><p>Multiple-device <code inline="">Fit()</code> regression coverage verifying <code inline="">CardInsufficientMemory</code> is incremented only once per invalid request</p></li></ul><h2>Validation</h2><ul><li><p><code inline="">gofmt</code> — passed</p></li><li><p><code inline="">git diff --check</code> — passed</p></li><li><p><code inline="">go test ./pkg/device/nvidia</code> — passed</p></li><li><p><code inline="">go test ./pkg/scheduler/...</code> — passed</p></li></ul><h2>User-facing impact</h2><p>Valid integer percentages from <code inline="">0</code> through <code inline="">100</code> continue to behave as before.</p><p>Invalid fractional, negative, or out-of-range values are now rejected rather than being silently rounded or propagated through the scheduling path.</p><h2>Which issue(s) this PR fixes:</h2><p>Fixes #2788</p></body></html>
</body>
</html>


pkg/device/nvidia/device.go
```diff
@@ -555,18 +565,16 @@ func (dev *NvidiaGPUDevices) GenerateResourceRequests(ctr *corev1.Container) dev
...
-					if mempnums > 0 {
-						mempnum = int32(mempnums)
-					} else {
-						// 0 would inject CUDA_DEVICE_MEMORY_LIMIT=0m, which hami-core reads as "no limit", so keep the "unset" sentinel and let the default below apply, like nvidia.com/gpumem: 0.
-						klog.ErrorS(nil, "memory percentage request is not positive, ignoring it", "container", ctr.Name, "requested", mempnums)
-						mempnum = 101
-					}
+				if !ok || mempnums < 0 || mempnums > 100 {
+					klog.ErrorS(nil, "memory percentage request out of range", "container", ctr.Name, "requested", mem.String())
+					return device.ContainerDeviceRequest{}
```
maishivamhoo123: here returning device.ContainerDeviceRequest{} doesn't reject the pod, it just looks like the container never asked for a GPU. so Resourcereqs will drop it, Filter passes every node, and the pod gets bound with no HAMi allocation (possibly even on a non-GPU node, since nvidia.com/* are ignoredByScheduler: true). Then it fails at the device plugin with an error that has nothing to do with the percentage.
and Also, fractional and negative values can't actually reach the scheduler, the API server already rejects those for nvidia.com/* resources. So this only changes what happens for out-of-range values when the webhook is bypassed.

maishivamhoo123 (changes requested):

pkg/device/nvidia/device.go
```diff
@@ -449,14 +452,21 @@ func (dev *NvidiaGPUDevices) defaultExclusiveCoreIfNeeded(ctr *corev1.Container)
...
-		return qty.Value(), true
+		return qty, true
 	}
 	if qty, ok := ctr.Resources.Requests[name]; ok {
+		return qty, true
+	}
+	return resource.Quantity{}, false
+}
+
+func resourceValue(ctr *corev1.Container, name corev1.ResourceName) (int64, bool) {
```
mesutoezdil: resourceValue is not called from any production code path now, both call sites moved to resourceQuantity. only its own test calls it. can this be removed, or is it kept for something else?

mesutoezdil:
pls resolve conflicts

pkg/device/nvidia/device.go
AyushSrivastava1818: pushed the change have a look!

pkg/device/nvidia/device.go
AyushSrivastava1818: yess looked at that later pushed the change removed the unused helper and updated the test that now covers resourceQuantity

AyushSrivastava1818:
@mesutoezdil resolved the conflicts take a look whenever you are free

AyushSrivastava1818:
@mesutoezdil resolved the conflicts have a look...there's security analysis failing which am not able to get the info about...if theres anything I have to work upon please tell

pkg/device/nvidia/device.go
```diff
@@ -566,18 +569,28 @@ func (dev *NvidiaGPUDevices) GenerateResourceRequests(ctr *corev1.Container) dev
...
-					if mempnums > 100 {
-						klog.ErrorS(nil, "memory percentage request out of range, clamping to 100", "container", ctr.Name, "requested", mempnums)
-						mempnums = 100
-					}
-					if mempnums > 0 {
-						mempnum = int32(mempnums)
+				if !ok {
+					// Fractional or invalid quantity: pass an out-of-range percentage so Fit rejects the request rather than silently ignoring the GPU.
+					klog.ErrorS(nil, "fractional or invalid memory percentage request", "container", ctr.Name, "requested", mem.String())
+					mempnum = math.MaxInt32
```
mesutoezdil: the error is routed through the value itself here: 101 means unset, 102 invalid, MaxInt32 fractional, -1 negative, all in the same int32 that also carries real percentages. the plain memory path returns an empty request instead. would that work here too, or is there something that rules it out?

pkg/device/nvidia/device.go
```diff
@@ -787,6 +800,11 @@ func (nv *NvidiaGPUDevices) Fit(devices []*device.DeviceUsage, request device.Co
 			reason[common.CardInsufficientCore]++
 			continue
 		}
+		if (k.MemPercentagereq < 0 || k.MemPercentagereq > 100) && k.MemPercentagereq != 101 {
```
mesutoezdil, line 791: this reads like a request level invariant, but it sits inside the per device loop, so it runs once per card and counts CardInsufficientMemory each time. would checking it once before the loop, with its own reason, describe the failure better?

pkg/device/nvidia/device.go
```diff
@@ -566,18 +569,28 @@ func (dev *NvidiaGPUDevices) GenerateResourceRequests(ctr *corev1.Container) dev
...
-						mempnum = int32(mempnums)
+				if !ok {
+					// Fractional or invalid quantity: pass an out-of-range percentage so Fit rejects the request rather than silently ignoring the GPU.
+					klog.ErrorS(nil, "fractional or invalid memory percentage request", "container", ctr.Name, "requested", mem.String())
+					mempnum = math.MaxInt32
+				} else if mempnums < 0 || mempnums > 100 {
+					klog.ErrorS(nil, "memory percentage request out of range", "container", ctr.Name, "requested", mem.String())
+					if mempnums == 101 {
+						// 101 is the internal sentinel for "unset", so use an explicit out-of-range value to ensure rejection.
+						mempnum = 102
```
mesutoezdil: validateMemoryPercentage already rejects these at admission. could you walk me through the path that reaches Fit with a fractional value, and whether a test covers it?

AyushSrivastava1818: The admission path normally rejects fractional values,but Fit can still get a malformed ContainerDevuceRequest through internal calls...So I have kept the validation in Fit as a defensive invariant and added coverage for invalid MemPercentagereq values there as well

pkg/device/nvidia/device.go
AyushSrivastava1818: Yes I checked it too...now i have moved the invalid percentage validation outside the per-device loop so the request level error is evaluated only once instead of incrementing cardInsufficientMemory once per GPU..its better way I guess

pkg/device/nvidia/device.go
AyushSrivastava1818: I have changed the invalid-value handling to return an empty device containerdeviceRequest{} instead of encoding different failure cases...

AyushSrivastava1818:
@mesutoezdil pushed the changes have a look sir again please

AyushSrivastava1818:
@mesutoezdil @archlitchi have a look Sir if any changes needed I will do that

pkg/device/nvidia/device.go
```diff
@@ -787,6 +788,11 @@ func (nv *NvidiaGPUDevices) Fit(devices []*device.DeviceUsage, request device.Co
 			reason[common.CardInsufficientCore]++
 			continue
 		}
+		if (k.MemPercentagereq < 0 || k.MemPercentagereq > 100) && k.MemPercentagereq != 101 {
```
mesutoezdil: is this still inside the per-device loop instead of running once per request?

AyushSrivastava1818: yes,I have moved the MemPercentageReq validation outside the per-device loop.It is now evaluated once at the request level....

pkg/device/nvidia/device.go
Shouren: @AyushSrivastava1818 GenerateResourceRequests still returns an empty request for invalid
  percentages. That can make the scheduler interpret an invalid GPU request as “no HAMi resource requested,” rather than rejecting it. The valid-limit/invalid-request masking issue also remains in [resourceQuantity](https://github.com/Project-HAMi/HAMi/blob/7d4c0a1ccfca6707ee199e786cfc0690d47966fb/pkg/device/nvidia/device.go#L462-L470)

Shouren: @AyushSrivastava1818 It seems that the changes is missing.

archlitchi:
please resolve the comments regarding Shouren's review

PR #2793, opened 2026-08-24
test(e2e): add scheduler policy E2E suite

haitwang-cloud:
**What type of PR is this?**
/kind feature

  Adds a Ginkgo E2E suite asserting deterministic placement for HAMi's scheduler policies: node/GPU `binpack` & `spread`, comma-separated policy chains (incl. mutex), and per-Pod device scoring weight overrides. Run via `make e2e-policy-test`.

  The suite waits on the scheduler cache (via /metrics) so assertions never see stale utilization — pods are confirmed released from every replica after cleanup, metrics are scraped from all scheduler replicas (HA/leader-change safe), and allocation decoding tolerates the trailing-separator quirk of EncodePodSingleDevice.

Fixes #

  Developed with AI assistance (Claude Code), per CONTRIBUTING.md. Requires a GPU cluster with ≥2 idle hami-core GPUs; skips otherwise. Not wired into CI — on-demand target only.

NO

## Summary by CodeRabbit

* **New Features**
  * Added end-to-end coverage for scheduler policies, including GPU sharing and release, binpacking, spreading, combined policies, mutex policies, and per-Pod device scoring.
  * Added Kubernetes-based scenarios to validate policy behavior and resource allocation.

* **Tests**
  * Added a dedicated command and test suite for running policy end-to-end tests.
  * Added checks for Kubernetes connectivity, scheduler synchronization, GPU allocation, and eligible-node handling.

### DIFF
--- a/test/utils/node.go
+++ b/test/utils/node.go
@@ -19,9 +19,12 @@ package utils
 import (
 	"context"
 	"fmt"
+	"os"
+	"time"
 
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/util/retry"
 	"k8s.io/klog/v2"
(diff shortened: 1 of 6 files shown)


test/utils/node.go
```diff
@@ -67,28 +68,42 @@ func UpdateNode(clientSet *kubernetes.Clientset, node *v1.Node) (*v1.Node, error
...
 		return nil, err
 	}
-
-	if node.Labels == nil {
-		node.Labels = make(map[string]string)
-	}
-	node.Labels[labelKey] = labelValue
-
-	return UpdateNode(clientSet, node)
+	time.Sleep(time.Second * 30)
```
mesutoezdil, line 87: this sleep 30s runs for every caller of AddNodeLabel, not only the new policy suite. test/e2e/node and test/e2e/pod use this too and they run in ci. does this slow down those suites now? can the wait be scoped to only the new policy test instead?

FouoF:
@haitwang-cloud  Note that we only have one node and one GPU for e2e now, will node/GPU binpack/spread permanent fail in this e2e environment?

test/utils/node.go
haitwang-cloud: @mesutoezdil Good catch! The 30s sleep was redundant — all callers already have their own wait mechanisms . I've removed the sleep from the shared utility functions so it no longer impacts CI runtime for existing suites.

haitwang-cloud:
> @haitwang-cloud Note that we only have one node and one GPU for e2e now, will node/GPU binpack/spread permanent fail in this e2e environment?

@FouoF No, binpack/spread tests won't permanently fail. The requireNodes function (line 664) gracefully skips tests when topology requirements aren't met. On a single-node/single-GPU setup, tests requiring ≥2 nodes or ≥2 GPUs are skipped via `ginkgo.Skip`. Only tests designed for single-GPU (e.g., vGPU sharing, per-Pod weights) run. The `HAMI_E2E_REQUIRE_POLICY_TOPOLOGY=true `env var is only set in the dedicated policy CI job where the topology is available

archlitchi:
> > @haitwang-cloud Note that we only have one node and one GPU for e2e now, will node/GPU binpack/spread permanent fail in this e2e environment?
> 
> @FouoF No, binpack/spread tests won't permanently fail. The requireNodes function (line 664) gracefully skips tests when topology requirements aren't met. On a single-node/single-GPU setup, tests requiring ≥2 nodes or ≥2 GPUs are skipped via `ginkgo.Skip`. Only tests designed for single-GPU (e.g., vGPU sharing, per-Pod weights) run. The `HAMI_E2E_REQUIRE_POLICY_TOPOLOGY=true `env var is only set in the dedicated policy CI job where the topology is available

it stucks at the schedule policy e2e, looks like it doesn't skip this test case

haitwang-cloud:
@archlitchi I have fixed the failure E2E test, PTAL

archlitchi (approved):
/lgtm

PR #2796, opened 2026-08-24
fix(device): acquire node lock for initContainer device requests

### LINKED ISSUE #2795: fix(scheduler): acquire device node locks for initContainer-only requests
## Description

HAMi's device node locking logic does not account for device resources requested exclusively by `initContainers`.

The scheduler correctly recognizes and allocates device resources from `initContainers`, but the device backends' `LockNode()` and `ReleaseNodeLock()` implementations only inspect `pod.Spec.Containers`.

For example, NVIDIA's implementation checks:

```go
for _, val := range p.Spec.Containers {
    if dev.GenerateResourceRequests(&val).Nums > 0 {
        found = true
        break
    }
}

if !found {
    return nil
}

return nodelock.LockNode(n.Name, NodeLockNvidia, p)
```

Therefore, a Pod with:

```yaml
initContainers:
  - name: gpu-init
    resources:
      limits:
        nvidia.com/gpu: "1"

containers:
  - name: app
```

has a valid device allocation, but `LockNode()` sees no device request because it never examines `InitContainers` and returns without acquiring the node lock.

## Impact

This breaks the node-level synchronization expected during device allocation/binding.

Two Pods being bound concurrently to the same node can therefore take different locking paths:

- A Pod with a device request only in an `initContainer` bypasses `LockNode()` and acquires no node lock.
- Another Pod with a device request in a regular container acquires the node lock normally.
- The `initContainer`-only Pod can proceed through the binding/allocation path without waiting for the node lock.

This creates a race around node/device allocation state and **can potentially result in conflicting device assignments or inconsistent node allocation annotations** — not a guaranteed failure on every run, but a real race window that depends on scheduling timing.

The issue is particularly important because HAMi already treats `initContainers` as device-consuming workloads during resource calculation and allocation, so the locking path is inconsistent with the scheduling path.

## Affected Backends

The same `Spec.Containers`-only locking pattern exists in multiple device backends, including:

- NVIDIA
- Ascend
- AMD
- Hygon
- Cambricon
- Kunlun
- MetaX
- VastAI
- Biren
- Iluvatar

Backends whose `LockNode()` implementation is currently a no-op should be reviewed separately rather than changed unnecessarily.

## Expected Behavior

If a Pod requests device resources in either:

- `spec.initContainers`, or
- `spec.containers`

then the corresponding device backend should acquire the node lock during the binding/allocation phase.

`ReleaseNodeLock()` should use the same detection logic so that the lock is also correctly released.

## Proposed Fix

Update the device locking logic to consider both `InitContainers` and regular `Containers`.

For example:

```go
func podRequestsDevice(dev Devices, pod *corev1.Pod) bool {
    for i := range pod.Spec.InitContainers {
        if dev.GenerateResourceRequests(&pod.Spec.InitContainers[i]).Nums > 0 {
            return true
        }
    }

    for i := range pod.Spec.Containers {
        if dev.GenerateResourceRequests(&pod.Spec.Containers[i]).Nums > 0 {
            return true
        }
    }

    return false
}
```

Then use the same detection for both `LockNode()` and `ReleaseNodeLock()`.

The implementation should avoid changing backends where node locking is intentionally a no-op.

## Regression Tests

Add coverage for at least:

- Device request in regular container → node lock acquired.
- Device request only in `initContainer` → node lock acquired.
- No device request → node lock is not acquired.
- Device request in both `initContainer` and regular container → node lock acquired.
- `initContainer`-only request → `ReleaseNodeLock()` correctly releases the lock.

The NVIDIA test suite should contain the initial regression test, with equivalent coverage added to the other affected backends where appropriate.

## Why This Matters

HAMi already accounts for `initContainer` device usage during scheduling and allocation. The locking layer must maintain the same resource model; otherwise a workload can be correctly allocated while bypassing the synchronization mechanism protecting that allocation.

Fixing this keeps device detection consistent across admission, scheduling, allocation, binding, and node locking.

AyushSrivastava1818:
## What type of PR is this?

/kind bug

## What does this PR do / why do we need it?

HAMi's device resource accounting supports device requests in both `initContainers` and regular `containers`, but the node-locking logic in several device backends only inspected `pod.Spec.Containers`.

As a result, a pod requesting device resources exclusively through an `initContainer` could reach the scheduling/bind path without acquiring the corresponding node lock.

This could allow concurrent scheduling operations on the same node to proceed without the expected node-level mutual exclusion, potentially causing device allocation or node annotation races.

This PR updates the affected device backends so that node-lock detection consistently considers both `InitContainers` and regular `Containers`.

## Changes

- Update `LockNode()` to inspect `pod.Spec.InitContainers` in addition to `pod.Spec.Containers`.
- Update `ReleaseNodeLock()` to use the same detection logic.
- Apply the fix to the affected device backends:
  - NVIDIA
  - Ascend
  - AMD
  - Hygon
  - Cambricon
  - Kunlun vDevice
  - MetaX SDevice
  - VastAI
  - Biren
  - Iluvatar
- Keep intentionally no-op `LockNode()` implementations unchanged.
- Add regression coverage for initContainer-only device requests.

## Expected behavior

| Pod device request | Node lock |
|---|---|
| No device request | Not acquired |
| Regular container only | Acquired as before |
| InitContainer only | Acquired |
| InitContainer + regular container | Acquired once |

## Regression Tests

Added coverage for the affected backends covering:

- Device request in `initContainers` only.
- Device request in regular containers only.
- Device requests in both `initContainers` and regular containers.
- Pods with no device requests.
- Both `LockNode()` and `ReleaseNodeLock()` behavior.

## Validation

- `go test -count=1 ./pkg/device/...` — passed
- `golangci-lint` v2.12.2 — passed with 0 issues
- `gofmt -s -w` — passed
- `git diff --check` — passed
- No unrelated changes introduced.

## User-facing impact

Pods that request HAMi-managed device resources exclusively through `initContainers` will now correctly participate in the device backend's node-locking mechanism.

Existing behavior for regular-container device requests is preserved.

## AI Assistance Disclosure

AI assistance from Claude and Antigravity was used for codebase investigation, identifying and tracing the initContainer node-locking gap, implementation assistance, test development, and review of the affected logic.

The final implementation was manually reviewed, and the relevant unit tests, formatting checks, `git diff --check`, and repository-configured `golangci-lint` validation were performed to verify the changes.

## Related Issue

Fixes #2795 

## Summary by CodeRabbit

* **Bug Fixes**
  * Node locking and unlocking now correctly detect GPU and accelerator requests in both init containers and regular containers.
  * Prevented missed locks for workloads requesting devices exclusively during initialization.
  * Pods without accelerator requests remain unaffected.

* **Tests**
  * Expanded coverage across supported accelerator types for regular, init-only, mixed, and CPU-only workloads.
  * Added validation for matching lock release behavior.


pkg/device/cambricon/device.go
```diff
@@ -124,12 +124,20 @@ func (dev *CambriconDevices) setNodeLock(node *corev1.Node) error {
...
 		}
 	}
+	if !found {
+		for _, val := range p.Spec.Containers {
+			if (dev.GenerateResourceRequests(&val).Nums) > 0 {
+				found = true
+				break
+			}
+		}
+	}
```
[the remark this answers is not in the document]
AyushSrivastava1818: resolved

pkg/device/amd/device.go
```diff
@@ -132,12 +132,20 @@ func (dev *AMDDevices) PatchAnnotations(pod *corev1.Pod, annoinput *map[string]s
...
 	found := false
-	for _, val := range p.Spec.Containers {
+	for _, val := range p.Spec.InitContainers {
 		if (dev.GenerateResourceRequests(&val).Nums) > 0 {
 			found = true
 			break
 		}
 	}
+	if !found {
+		for _, val := range p.Spec.Containers {
```
maishivamhoo123, line 142: @AyushSrivastava1818  the same loop is copy pasted multiple times. Please move it into a single helper in pkg/device or a suitable file. which will check InitContainers then Containers and you can call that from each LockNode/ReleaseNodeLock.

maishivamhoo123 (changes requested):

pkg/device/amd/device.go
AyushSrivastava1818: Appreciate the review...will move it into a single helper soon

pkg/device/amd/device.go
```diff
@@ -132,12 +132,20 @@ func (dev *AMDDevices) PatchAnnotations(pod *corev1.Pod, annoinput *map[string]s
...
 func (dev *AMDDevices) LockNode(n *corev1.Node, p *corev1.Pod) error {
 	found := false
-	for _, val := range p.Spec.Containers {
+	for _, val := range p.Spec.InitContainers {
 		if (dev.GenerateResourceRequests(&val).Nums) > 0 {
 			found = true
 			break
 		}
 	}
+	if !found {
```
mesutoezdil, line 141: this block is copied in all 10 device files, same code twice per file. can this move to one shared helper function instead? less risk of copy paste bugs later.

mesutoezdil:
superseded by #2755, which moved the init container check into the shared PodRequiresDevice helper and removed the per backend loops this PR edits. thanks for the work here.

PR #2809, opened 2026-08-24
fix(awsneuron): cap whole-device core accounting to addressable cores

mesutoezdil:
/kind bug

A whole device request copied the raw per device core count into Coresreq and the core mask was built from the same value. On inf1 hardware with four cores per device this bypassed the two bit cap used by core accounting, so a fully owned device still reported free cores and a second pod could be packed onto it. Both paths now go through the existing coresPerDevice cap.


archlitchi (approved):
/lgtm

PR #2810, opened 2026-08-24
fix(enflame): keep DRS profile core costs within the device budget

mesutoezdil:
/kind bug

Profile core costs were rounded up, so on a six slice device every 1g slice cost 17 and six of them summed to 102. The sixth advertised slice was rejected with CardInsufficientCore while slice count and memory were still free. Rounding down keeps a fully sliced device within its 100 core budget.


archlitchi (approved):
/lgtm

PR #2811, opened 2026-08-24
fix(kunlun): make topology selection independent of device slice order

mesutoezdil:
/kind bug

graghSelect reads topology from slice positions, but the scheduler sorts devices by score before Fit runs. A used XPU that sorts to the front shifts both wings and the picker can return a set that crosses wings and is not a valid interconnect group. Sorting a copy by index restores the physical layout.


archlitchi (approved):
/lgtm

PR #2812, opened 2026-08-24
fix(scheduler): account terminating pods replayed by informer sync

mesutoezdil:
/kind bug

Terminating pods were short circuited with UpdatePod, which is a no op when the pod is not cached. After a scheduler restart the informer replays such pods as adds, so a running pod with a long grace period lost its device usage and its GPU looked free. The handlers now fall through to normal accounting when the pod is missing from the cache.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of terminating workloads during scheduler restarts and informer resyncs.
  * Preserved GPU device allocations and quota usage for replayed terminating pods, preventing resource overcommit.
  * Ensured previously uncached terminating pods are processed correctly during updates.
  * Improved reservation consistency during concurrent scheduling and NUMA resource recalculation.

* **Tests**
  * Added coverage for pod lifecycle recovery and resource accounting after scheduler restarts.

### DIFF
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -159,9 +159,15 @@ func (s *Scheduler) onAddPod(obj any) {
 		return
 	}
 	if util.IsPodTerminating(pod) {
-		klog.V(5).InfoS("Pod is terminating but holding locks, preserving cache", "pod", pod.Name)
-		s.podManager.UpdatePod(pod)
-		return
+		// A terminating pod still holds its devices. When it is already
+		// cached, refresh the object; when it is not (the informer's initial
+		// sync after a scheduler restart replays it as an add), fall through
+		// so its usage is accounted instead of silently dropped.
+		if _, cached := s.podManager.GetPod(pod); cached {
+			klog.V(5).InfoS("Pod is terminating but holding locks, preserving cache", "pod", pod.Name)
+			s.podManager.UpdatePod(pod)
+			return
+		}
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/scheduler/scheduler.go
```diff
@@ -159,9 +159,15 @@ func (s *Scheduler) onAddPod(obj any) {
...
 	}
 	if util.IsPodTerminating(pod) {
-		klog.V(5).InfoS("Pod is terminating but holding locks, preserving cache", "pod", pod.Name)
-		s.podManager.UpdatePod(pod)
-		return
+		// A terminating pod still holds its devices. When it is already
+		// cached, refresh the object; when it is not (the informer's initial
+		// sync after a scheduler restart replays it as an add), fall through
+		// so its usage is accounted instead of silently dropped.
+		if _, cached := s.podManager.GetPod(pod); cached {
```
Shouren, line 166: @mesutoezdil I think we can add a new method for checking the existence of Pod in `podManager` to avoid deepcopy in `GetPod`.

Shouren (commented):
@mesutoezdil Can we rename the `scheduler_terminating_replay_test.go` to somethinkg like `scheduler_pod_lifecycle_test.go` so it'll be easier to add similar test cases in the future?

mesutoezdil:
> @mesutoezdil Can we rename the `scheduler_terminating_replay_test.go` to somethinkg like `scheduler_pod_lifecycle_test.go` so it'll be easier to add similar test cases in the future?

sure, renamed

Shouren (approved):
/lgtm

archlitchi (approved):
/lgtm

PR #2729, opened 2026-08-19
feat: add groundwork for NUMA-aware device refit

saiyam1814:
**What type of PR is this?**

/kind feature

**What this PR does / why we need it**:

Groundwork for the NUMA refit from #2080 (Phase 2 of the NUMA work; Phase 1 was #2065). When kubelet's Topology Manager restricts an allocation to a GPU the scheduler did not pick, HAMi's runtime and accounting can disagree. The fix is a scheduler-side refit (#2731, stacked on this PR); this PR lands the pieces it needs, with no behavior change:

- New pod annotation `hami.io/numa-alignment` with a `best-effort` mode. `strict` is not accepted yet — it arrives in #2731 together with the enforcement, so the annotation never promises something it doesn't do. This is a new annotation on purpose: `nvidia.com/numa-bind` means GPU-to-GPU co-location and is unrelated.
- Request/response types for the refit API.
- A scheduler helper that runs the existing device fit against a restricted set of physical GPUs. Policy behavior (binpack, spread, mutex, policy chains, numa-bind, the pod's own policy annotation) is covered by unit tests.
- The device plugin logs the mismatch for opted-in pods instead of silently swallowing it. The response to kubelet is unchanged.

**Which issue(s) this PR fixes**:

Part of #2080. The refit itself is #2731.

**Special notes for your reviewer**:

- Scheduler parts are unit-tested. The device-plugin change is log-only and was also checked on real hardware (8x NVIDIA RTX PRO 6000 Blackwell Server Edition, driver 610.43.02, Kubernetes v1.35.6, Topology Manager `single-numa-node`): pods without the annotation behave exactly as before; opted-in pods get one extra log line on a mismatch.
- User docs for the annotation land with #2731, which enables the behavior.

AI assistance disclosure: written primarily with Claude Code, directed and reviewed by me. The design follows the #2080 discussion.

**Does this PR introduce a user-facing change?**:

```release-note
Added the hami.io/numa-alignment pod annotation (best-effort): HAMi logs when kubelet's Topology Manager restricts an allocation away from the scheduler-selected GPU. Detection only; reconciliation lands with the NUMA refit (#2080).
```

## Summary by CodeRabbit

- **New Features**
  - Added optional per-Pod NUMA alignment configuration using the `hami.io/numa-alignment` annotation.
  - Added validation that rejects invalid NUMA alignment values during admission.
  - Added NUMA-aware device refitting that honors kubelet-approved assignments while preserving scheduling policies and topology information.
[...]

PR #2813, opened 2026-08-25
fix(scheduler): validate scheduler-policy annotations

### LINKED ISSUE #2767: scheduler: unrecognized gpu/node-scheduler-policy annotation is silently accepted and the configured defaultSchedulerPolicy is discarded

saiyam1814:
**What type of PR is this?**

/kind bug

An unrecognized `hami.io/gpu-scheduler-policy` or `hami.io/node-scheduler-policy` value was silently accepted: the annotation overrode the configured policy whatever its value, and the comparator's hardcoded fallback then replaced the operator's configured default.

Two layers of fix, following the same pattern as the numa-alignment validation in #2729:

- The admission webhook denies invalid values, so a typo surfaces at `kubectl apply` with a clear message listing the accepted values.
- `GetGPUSchedulerPolicyByPod` and `resolveNodeSchedulerPolicy` keep the configured default with a warning when the value is unrecognized, covering pods created before upgrade or bypassing the webhook.

Accepted GPU policy values: `binpack`, `spread`, `numa`, `mutex`, `topology-aware`, or a comma-separated chain of them. Node policy: `binpack` or `spread`.

Fixes #2767

Scheduler-extender scoped; covered by unit tests (validators, resolver fallbacks, webhook admission table). The sort comparator is left untouched: with validation at the sources, unrecognized values no longer reach it.

AI assistance disclosure: written primarily by Claude Code, directed and reviewed by me.

```release-note
Pods with an unrecognized hami.io/gpu-scheduler-policy or hami.io/node-scheduler-policy annotation are now rejected at admission; previously the invalid value silently replaced the configured scheduler policy.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Added validation for GPU and node scheduling policies.
  * Invalid policy annotations are rejected with clear errors during admission.
  * Invalid annotations now fall back to configured default policies instead of being applied.
  * Valid policy annotations continue to be accepted, including multiple comma-separated GPU policies.
* **Tests**
  * Expanded coverage for valid, invalid, whitespace-trimmed, and unsupported policy values.

### DIFF
--- a/pkg/scheduler/score.go
+++ b/pkg/scheduler/score.go
@@ -139,7 +139,11 @@ func fitInDevices(node *NodeUsage, requests device.ContainerDeviceRequests, pod 
 func resolveNodeSchedulerPolicy(task *corev1.Pod) string {
 	if task.GetAnnotations() != nil {
 		if value, ok := task.GetAnnotations()[util.NodeSchedulerPolicyAnnotationKey]; ok {
-			return value
+			if util.ValidNodeSchedulerPolicy(value) {
+				return value
+			}
+			klog.Warningf("ignoring unrecognized %s annotation %q on pod %s/%s, keeping configured policy %q",
+				util.NodeSchedulerPolicyAnnotationKey, value, task.Namespace, task.Name, config.NodeSchedulerPolicy)
 		}
 	}
 	return config.NodeSchedulerPolicy
--- a/pkg/scheduler/webhook.go
+++ b/pkg/scheduler/webhook.go
@@ -74,6 +74,20 @@ func (h *webhook) Handle(_ context.Context, req admission.Request) admission.Res
 		klog.Warningf(template+" - Denying admission: %v", pod.Namespace, pod.Name, pod.UID, err)
 		return admission.Denied(err.Error())
 	}
+	// Reject unrecognized scheduler-policy values the same way, instead of
+	// silently discarding the operator's configured default (#2767).
+	if pod.Annotations != nil {
+		if value, ok := pod.Annotations[util.GPUSchedulerPolicyAnnotationKey]; ok && !util.ValidGPUSchedulerPolicy(value) {
+			reason := fmt.Sprintf("invalid %s annotation %q: expected binpack, spread, numa, mutex, topology-aware, or a comma-separated chain of them", util.GPUSchedulerPolicyAnnotationKey, value)
+			klog.Warningf(template+" - Denying admission: %s", pod.Namespace, pod.Name, pod.UID, reason)
+			return admission.Denied(reason)
+		}
+		if value, ok := pod.Annotations[util.NodeSchedulerPolicyAnnotationKey]; ok && !util.ValidNodeSchedulerPolicy(value) {
+			reason := fmt.Sprintf("invalid %s annotation %q: expected binpack or spread", util.NodeSchedulerPolicyAnnotationKey, value)
+			klog.Warningf(template+" - Denying admission: %s", pod.Namespace, pod.Name, pod.UID, reason)
+			return admission.Denied(reason)
+		}
+	}
 	klog.V(5).Infof(template, pod.Namespace, pod.Name, pod.UID)
 	privilegedName, hasPrivileged := privilegedContainerName(pod)
 	hasResource := false
--- a/pkg/scheduler/webhook_test.go
+++ b/pkg/scheduler/webhook_test.go
@@ -1276,3 +1276,102 @@ func TestHandleNumaAlignmentAnnotation(t *testing.T) {
 		})
 	}
 }
+
+func TestHandleSchedulerPolicyAnnotations(t *testing.T) {
+	tests := []struct {
+		name        string
+		annotations map[string]string
+		wantDenied  bool
+	}{
+		{
+			name:        "valid gpu policy chain is admitted",
+			annotations: map[string]string{util.GPUSchedulerPolicyAnnotationKey: "binpack,numa"},
+		},
+		{
+			name:        "valid node policy is admitted",
+			annotations: map[string]string{util.NodeSchedulerPolicyAnnotationKey: "spread"},
+		},
+		{
+			name:        "unrecognized gpu policy is denied",
+			annotations: map[string]string{util.GPUSchedulerPolicyAnnotationKey: "topology"},
+			wantDenied:  true,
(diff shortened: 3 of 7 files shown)


mesutoezdil:
closing as duplicate of #2769, which was opened earlier for the same issue #2767. that PR also guards the gpu policy read path in GetGPUSchedulerPolicyByPod, so the fallback works even when the webhook is disabled, and its placement test covers the behavior end to end. the admission time error message from this PR is a nice idea, feel free to propose it as a follow up once #2769 lands.

PR #2814, opened 2026-08-25
fix(plugin): pair per device limits by UUID when aligning kubelet IDs

mesutoezdil:
/kind bug

alignContainerDevicesWithAllocatedIDs replaced UUIDs positionally while keeping Usedmem and Usedcores in annotation order. Kubelet reports DevicesIds in its own order, so on nodes where per device memory differs the CUDA_DEVICE_MEMORY_LIMIT env of one GPU was applied to another. Entries are now paired by physical UUID, and the positional behavior is kept only for devices the annotation does not mention.


archlitchi (approved):
/lgtm

PR #2815, opened 2026-08-25
fix(amd): keep allocation annotations decodable for empty registered types

mesutoezdil:
/kind bug

The AMD backend writes the registered product type into the allocation annotation so consumers can identify the model. That type is read verbatim from the external node registration annotation and can be empty, and since the decoder rejects an empty type field the scheduler could no longer read the pod allocation and silently dropped its device usage. The backend now falls back to the logical AMD type when the registered type is empty.


archlitchi (approved):
/lgtm

PR #2827, opened 2026-08-25
webhook: validate device-scoring-weights at admission

### LINKED ISSUE #2826: webhook: validate device-scoring-weights at admission

Validate the `hami.io/device-scoring-weights` Pod annotation in the HAMi admission webhook.
If the value is invalid, the webhook should reject the Pod and return the existing parser error. Valid values and Pods without this annotation should continue to be accepted.

The existing scheduler validation should remain as a safety check.

### Current and expected behavior

```text
Current behavior

Create Pod
    |
    v
Admission webhook accepts the annotation
    |
    v
Scheduler checks the annotation
    |
    v
Invalid value found
    |
    v
Scheduling fails and the Pod remains Pending


Expected behavior

Create Pod
    |
    v
Admission webhook checks the annotation
    |
    +---- Valid or missing ----> Pod is accepted
    |
    +---- Invalid ------------> Pod is rejected immediately
                                 with a clear error message
```

**Why is this needed**:

Currently, the annotation is checked only during scheduling. For example:

`hami.io/device-scoring-weights: "slot=1,core=1"`

This value is invalid because the required `memory` weight is missing. The Pod is accepted when it is created, but scheduling later fails and the Pod remains Pending.

Validating the annotation during admission gives the user an immediate and clear error. It also matches the existing admission validation for `hami.io/numa-alignment`.

The existing `GetDeviceScoringWeightsByPod` function can be reused in the webhook. The schedule-time validation should remain as an additional safety check.

asadjan4611:
/kind feature

This PR validates the `hami.io/device-scoring-weights` annotation in the HAMi admission webhook.

Previously, an invalid value was accepted when the Pod was created. The scheduler found the error later, causing the Pod to remain Pending.

After this change, the webhook rejects an invalid value immediately and returns the existing parser error. Valid values and Pods without the annotation continue to be accepted.

The existing schedule-time validation remains as a safety check when admission validation is unavailable.

```text
Before

Create Pod
    |
    v
Webhook accepts invalid value
    |
    v
Scheduler finds the error
    |
    v
Pod remains Pending


After

Create Pod
    |
    v
Webhook checks the value
    |
    +---- Valid or missing ----> Pod is accepted
    |
    +---- Invalid ------------> Pod is rejected immediately
                                 with a clear error
```

Fixes #2826

The implementation reuses `GetDeviceScoringWeightsByPod`; no new parser or validation rules were added.

The change does not affect device allocation, scoring calculations, resource accounting, or other device backends.

The scheduler-policy documentation was updated to explain admission validation and the retained schedule-time safety check.

AI assistance disclosure:

I used AI Assistance to review and refine the webhook change, update tests and documentation and run validation. I reviewed the final implementation and understand the changes.

Yes. When the admission webhook is enabled, Pods with an invalid `hami.io/device-scoring-weights` annotation are rejected during creation instead of failing later during scheduling. Valid and missing annotations keep their existing behavior.

## Summary by CodeRabbit

- **Bug Fixes**
  - Invalid device-scoring weight annotations are now rejected during admission only for Pods requesting HAMi-managed resources.
  - Pods without HAMi-managed resources are no longer rejected because of these annotations.
  - Scheduler validation continues when admission validation is unavailable, helping prevent invalid configurations from remaining unschedulable.
  - Valid and missing annotations continue to be accepted.

### DIFF
--- a/pkg/scheduler/webhook.go
+++ b/pkg/scheduler/webhook.go
@@ -74,6 +74,13 @@ func (h *webhook) Handle(_ context.Context, req admission.Request) admission.Res
 		klog.Warningf(template+" - Denying admission: %v", pod.Namespace, pod.Name, pod.UID, err)
 		return admission.Denied(err.Error())
 	}
+	// Validate device-scoring weights at admission so malformed workload input
+	// is reported before scheduling. The scheduler retains the same validation
+	// as a defense-in-depth check.
+	if _, err := util.GetDeviceScoringWeightsByPod(pod); err != nil {
+		klog.Warningf(template+" - Denying admission: %v", pod.Namespace, pod.Name, pod.UID, err)
+		return admission.Denied(err.Error())
+	}
 	klog.V(5).Infof(template, pod.Namespace, pod.Name, pod.UID)
 	privilegedName, hasPrivileged := privilegedContainerName(pod)
 	hasResource := false
--- a/pkg/scheduler/webhook.go
+++ b/pkg/scheduler/webhook.go
@@ -74,13 +74,6 @@ func (h *webhook) Handle(_ context.Context, req admission.Request) admission.Res
 		klog.Warningf(template+" - Denying admission: %v", pod.Namespace, pod.Name, pod.UID, err)
 		return admission.Denied(err.Error())
 	}
-	// Validate device-scoring weights at admission so malformed workload input
-	// is reported before scheduling. The scheduler retains the same validation
-	// as a defense-in-depth check.
-	if _, err := util.GetDeviceScoringWeightsByPod(pod); err != nil {
-		klog.Warningf(template+" - Denying admission: %v", pod.Namespace, pod.Name, pod.UID, err)
-		return admission.Denied(err.Error())
-	}
 	klog.V(5).Infof(template, pod.Namespace, pod.Name, pod.UID)
 	privilegedName, hasPrivileged := privilegedContainerName(pod)
 	hasResource := false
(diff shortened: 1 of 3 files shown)


pkg/scheduler/webhook.go
```diff
@@ -74,6 +74,13 @@ func (h *webhook) Handle(_ context.Context, req admission.Request) admission.Res
 		klog.Warningf(template+" - Denying admission: %v", pod.Namespace, pod.Name, pod.UID, err)
 		return admission.Denied(err.Error())
 	}
+	// Validate device-scoring weights at admission so malformed workload input
+	// is reported before scheduling. The scheduler retains the same validation
+	// as a defense-in-depth check.
+	if _, err := util.GetDeviceScoringWeightsByPod(pod); err != nil {
+		klog.Warningf(template+" - Denying admission: %v", pod.Namespace, pod.Name, pod.UID, err)
```
mesutoezdil: this runs before hasResource is known, so a pod with no GPU request is now denied for a typo in a hami annotation. the numa check above does the same, so is that the intended precedent?

archlitchi: good point, i think it should run after hasResource is known

asadjan4611: @archlitchi @mesutoezdil thanks for valuabe feedback on it.i have updated it. Validation now runs after hasResource is known and only applies to Pods requesting HAMi resources. I also added a test confirming that non-HAMi Pods remain accepted.

Shouren (approved):
/lgtm

FouoF (approved):

PR #2836, opened 2026-08-26
fix(scheduler): validate NUMA refit quota with container positions

### LINKED ISSUE #2828: NUMA refit can under-count ResourceQuota when container positions are compacted
**What happened**:

During a NUMA refit, HAMi removes the target container and empty entries from the temporary device-allocation list.

This changes the container positions. Later, `CollapseInitContainerUsage` uses those positions to decide whether an allocation belongs to an init container or an app container. An app container can therefore be treated as an init container, causing the quota check to count less memory than the final allocation uses.

NUMA refit should preserve the original container positions during quota calculation.

If the final allocation would exceed the namespace `ResourceQuota`, the refit should fail without changing the pod annotations or scheduler accounting.

Use a namespace GPU-memory quota of `12000`.

Create a pod with these allocations:

| Index | Container | Initial GPU | GPU memory |
|---|---|---|---:|
| 0 | Init container | GPU-A | 5000 |
| 1 | App container | GPU-A | 5000 |
| 2 | App container | GPU-A | 5000 |

The current usage is:

```text
max(init 5000, apps 5000 + 5000) = 10000
```

The pod is within the `12000` quota.

Next, request a NUMA refit that moves the init container from GPU-A to GPU-B.

The correct usage after the refit is:

```text
GPU-A app usage  = 10000
GPU-B init usage = 5000
Total             = 15000
```

The refit should fail because `15000` exceeds the quota.

However, the seed loop removes container 0 and compacts the list:

```text
Original:

[0: init] [1: app] [2: app]
     |
     | remove refit target
     v
Temporary list:

[0: app] [1: app] [2: refitted init]
     |
     v
Treated as an init container
```

The temporary quota calculation becomes `10000`, so the refit succeeds. HAMi then rebuilds the final allocation and records `15000` as used against the `12000` quota.

I confirmed this behavior with a focused scheduler unit test.

Relevant code:

- `pkg/scheduler/numa_refit_handler.go`
- `pkg/device/initContainer.go`
- `pkg/device/nvidia/device.go`
- `pkg/scheduler/numa_refit_handler_test.go`

A possible fix is to preserve the original container positions, including empty entries, and evaluate the refitted allocation at its original index.

**Environment**:

- HAMi version: `master` at `4bfa94371b3dd3adc2cad8265289a123e99d2c4f`
- Kubernetes version: Not applicable to the unit test
- Accelerator vendor and model: Mock NVIDIA GPU-A and GPU-B
- HAMi backend and resources: NVIDIA, `nvidia.com/gpu`, `nvidia.com/gpumem`
- NVIDIA driver and runtime version: Not applicable
- Workload image: Not applicable
- Operating system and kernel: Not applicable
- Real GPU validation: Not yet performed

asadjan4611:
**What type of PR is this?**

/kind bug

This PR fixes ResourceQuota accounting during a NUMA device refit.

The restricted fit can temporarily remove empty and target container entries. This changes container positions and may cause an application container to be counted as an init container. As a result, the quota check can report less GPU usage than the final allocation uses.

This PR preserves the original container positions when validating the new allocation. If the correct usage exceeds the namespace quota, the refit is rejected before pod annotations or scheduler accounting are changed.

### Behavior

```text
Before

Original allocation:
[init: GPU-A] [app: GPU-A] [app: GPU-A]
                         |
                         v
Compacted quota input:
[app: GPU-A] [new init: GPU-B]
                         |
                         v
Incorrect quota usage: 10000


After

Original positions preserved:
[init: GPU-B] [app: GPU-A] [app: GPU-A]
                         |
                         v
Correct collapsed usage:
GPU-A applications: 10000
GPU-B init:          5000
                         |
                         v
Correct quota usage: 15000
```

A refit within the quota continues normally. A rejected refit leaves the existing annotations, device reservation, and quota usage unchanged.

Fixes #2828

Regression tests cover:

- Rejecting a refit when the correctly calculated usage exceeds the quota.
- Keeping annotations, cached allocation, and quota usage unchanged after rejection.
- Allowing a refit when the resulting usage fits within the quota.
- Recording the correct collapsed usage after a successful refit.

AI assistance disclosure:

I used AI Assistance  to help review the final code  and refine the PR description. 

```release-note
Fixed NUMA refit quota validation so init and application container usage is counted using the correct container positions.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved NUMA allocation refits for workloads with init containers.
  * Preserved per-container device allocation annotations during refits.
  * Enforced device-specific namespace memory quotas before applying allocation changes.
  * Prevented over-quota refits from modifying annotations or recorded usage.
  * Correctly collapsed init-container usage into total device accounting after release.
  * Serialized allocation updates to prevent inconsistent resource accounting.
  * Added a timeout to allocation annotation updates to prevent stalled operations.

### DIFF
--- a/pkg/scheduler/numa_refit_handler.go
+++ b/pkg/scheduler/numa_refit_handler.go
@@ -241,7 +241,7 @@ func (s *Scheduler) RefitNumaAllocation(req device.NumaRefitRequest) device.Numa
 	maps.Copy(hypothetical, allocated)
 	hypothetical[req.DeviceType] = refitted
 	var quotaMem, quotaCores int64
-	for _, ctrDevs := range device.CollapseInitContainerUsage(pod, hypothetical)[req.DeviceType] {
+	for _, ctrDevs := range effectivePodDeviceUsage(pod, hypothetical, pi.InitContainerResourceReleased)[req.DeviceType] {
 		for _, d := range ctrDevs {
 			quotaMem += int64(d.Usedmem)
 			quotaCores += int64(d.Usedcores)
@@ -283,10 +283,7 @@ func (s *Scheduler) RefitNumaAllocation(req device.NumaRefitRequest) device.Numa
 		// init-container usage has been released, collapsing again would
 		// re-inflate the reservation back to the init peak, the same hazard
 		// PodManager.AddPod guards against on a re-add.
-		effective := device.CollapseInitContainerUsage(pod, rawDevices)
-		if pi.InitContainerResourceReleased {
-			effective = device.SteadyStateDeviceUsage(pod, rawDevices)
-		}
+		effective := effectivePodDeviceUsage(pod, rawDevices, pi.InitContainerResourceReleased)
 		if _, ok := s.podManager.ReplacePodDevices(key, effective); ok {
 			s.quotaManager.AddUsage(pod, effective)
 		} else {
@@ -303,6 +300,16 @@ func (s *Scheduler) RefitNumaAllocation(req device.NumaRefitRequest) device.Numa
 	klog.InfoS(message, "pod", klog.KObj(pod), "node", req.NodeName)
 	s.recordNumaRefitResultEvent(pod, message, nil)
 	return device.NumaRefitResponse{Succeeded: true, ContainerDevices: device.EncodeContainerDevices(newDevices)}
+}
+
+// effectivePodDeviceUsage mirrors the accounting shape stored by PodManager.
+// Non-sidecar init-container usage is part of the phase peak until the
+// informer records its release, and excluded from steady-state usage after it.
+func effectivePodDeviceUsage(pod *corev1.Pod, raw device.PodDevices, initReleased bool) device.PodDevices {
+	if initReleased {
+		return device.SteadyStateDeviceUsage(pod, raw)
+	}
+	return device.CollapseInitContainerUsage(pod, raw)
 }
 
 // replaceContainerDeviceEntry swaps one container's entry inside an encoded
--- a/pkg/scheduler/numa_refit_handler.go
+++ b/pkg/scheduler/numa_refit_handler.go
@@ -89,7 +89,8 @@ func (s *Scheduler) RefitNumaAllocation(req device.NumaRefitRequest) device.Numa
 	if len(req.AllowedDeviceUUIDs) > maxAllowedDeviceUUIDs {
 		return numaRefitFailure(nil, "refit request carries %d allowed devices, limit is %d", len(req.AllowedDeviceUUIDs), maxAllowedDeviceUUIDs)
 	}
-	if _, ok := device.GetDevices()[req.DeviceType]; !ok {
+	refitDevice, ok := device.GetDevices()[req.DeviceType]
+	if !ok {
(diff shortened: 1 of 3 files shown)


mesutoezdil:
resolve the conflicts pls

asadjan4611:
@mesutoezdil  i  have fixed  the conflicts can you plz review  it now

pkg/scheduler/numa_refit_handler.go
```diff
@@ -230,6 +231,27 @@ func (s *Scheduler) RefitNumaAllocation(req device.NumaRefitRequest) device.Numa
...
+	// quota check cannot preserve the container indexes used to distinguish init
+	// and app usage. Validate the selected devices in the original annotation
+	// layout before changing annotations or cached accounting.
+	refitted := append(device.PodSingleDevice{}, allocatedSingle...)
+	refitted[req.ContainerIndex] = newDevices
+	hypothetical := make(device.PodDevices, len(allocated))
+	maps.Copy(hypothetical, allocated)
+	hypothetical[req.DeviceType] = refitted
+	var quotaMem, quotaCores int64
+	for _, ctrDevs := range device.CollapseInitContainerUsage(pod, hypothetical)[req.DeviceType] {
```
mesutoezdil: line 287 picks SteadyStateDeviceUsage when pi.InitContainerResourceReleased. this always collapses. for a released pod the check uses the init peak and could reject a refit the commit path then accounts as steady state. should it branch too?

pkg/scheduler/numa_refit_handler.go
```diff
@@ -230,6 +231,27 @@ func (s *Scheduler) RefitNumaAllocation(req device.NumaRefitRequest) device.Numa
...
+		for _, d := range ctrDevs {
+			quotaMem += int64(d.Usedmem)
+			quotaCores += int64(d.Usedcores)
+		}
+	}
+	resourceNames := refitDevice.GetResourceNames()
+	if !s.quotaManager.FitQuota(pod.Namespace, quotaMem, resourceNames.MemoryFactor, quotaCores, req.DeviceType) {
```
mesutoezdil, line 251: #2868 rewrites this same block and adds effectivePodDeviceUsage for the branch above. worth coordinating on which lands first?

mesutoezdil:
/lgtm cancel

the quota check always collapses, but the commit path branches on InitContainerResourceReleased. holding until that is settled.

pkg/scheduler/numa_refit_handler.go
asadjan4611: fixed

pkg/scheduler/numa_refit_handler.go
asadjan4611: fixed

asadjan4611:
Yes, the two PRs overlap in this block. I kept #2836 focused on the quota bug. I suggest merging #2836 first, then rebasing #2868 and reusing the shared helper there. I can follow the opposite order if you prefer.

mesutoezdil:
/lgtm

both points are addressed

asadjan4611:
thanks @mesutoezdil

FouoF (approved):

PR #2837, opened 2026-08-26
fix(metax): let multi container pods spread over exclusive GPUs

mesutoezdil:
/kind bug

customFilterRule required every container of a pod to land on the card a previous container already took. Metax whole GPUs register with Count 1, so that card is always exhausted for the next container and every other card fails the filter, leaving pods with two or more GPU containers unschedulable on any node. The rule was copied from the shareable mthreads backend and cannot be satisfied on exclusive cards, so it is removed and each container takes its own card.


archlitchi (approved):
/lgtm

PR #2838, opened 2026-08-26
fix(scheduler): keep candidate nodes for non HAMi pods on the simulation path

mesutoezdil:
/kind bug

The no HAMi resource early return in Filter echoed only NodeNames. Simulation callers such as the cluster autoscaler send Nodes with NodeNames unset, so a pod without HAMi resources got an empty result and the extender vetoed scale up for every ordinary workload. The early return now echoes Nodes as well.


archlitchi (approved):
/lgtm

PR #2848, opened 2026-08-27
feat(device plugin): add auto detecting nvidiaDriverRoot

### LINKED ISSUE #867: 安装hami后，设置limit nvidia.com/gpu: `1`时，pod启动失败
1，先安装的gpu-operator，成功后安装的hami
```
helm install gpu-operator nvidia/gpu-operator \
  --namespace gpu-operator \
  --set devicePlugin.enabled=false \
  --set driver.enabled=true \
  --set toolkit.enabled=true \
  --set toolkit.env[0].name=CONTAINERD_CONFIG \
  --set toolkit.env[0].value=/var/lib/rancher/k3s/agent/etc/containerd/config.toml \
  --set toolkit.env[1].name=CONTAINERD_SOCKET \
  --set toolkit.env[1].value=/run/k3s/containerd/containerd.sock \
  --set toolkit.env[2].name=CONTAINERD_RUNTIME_CLASS \
  --set toolkit.env[2].value=nvidia

helm install hami hami-charts/hami --set devicePlugin.runtimeClassName=nvidia -n kube-system
```
Image
Image

2，创建pod的时候给环境变量NVIDIA_VISIBLE_DEVICES=all可以执行nvidia-smi，加上limit nvidia.com/gpu: `1`时就不行了，pod报错
Image

3，显卡驱动版本：
Image

4，版本
hami-2.5.0
gpu-operator-v24.9.2

FouoF:
**What type of PR is this?**

/kind feature

As I known nvidiaDriverRoot is frequently misconfigured by users. So config automation can significantly improve the user experience.

Fixes #

This detecting depend on gpu-operator validation ready. If it hasn't, the device plugin can not detecting what the driver exact path and fallback to host driver which may not correct. But when the gpu-operator ready it will recover.

This PR draft mostly by Codex, but has been tested in real gpu node.

Changed the default helm values: devicePlugin.nvidiaDriverRoot="auto" and passDeviceSpecsEnabled="true" which was changed to "false" to fix #867 which makes regression for #690. The root cause for it is incorrect devicePlugin.nvidiaDriverRoot.

## Summary by CodeRabbit

* **New Features**
  * Added optional host-PID broker support for NVIDIA workloads.
  * Added scheduler PodDisruptionBudget configuration.
  * Added scheduler leader-election and cache-sync metrics.
  * NVIDIA driver paths can be detected automatically from GPU Operator readiness information.
* **Bug Fixes**
  * Improved GPU allocation validation, topology handling, and multi-container placement.
  * Improved scheduler recovery and simulation behavior.
  * Unsupported GPU utilization metrics are now skipped gracefully.
  * NUMA refit operations avoid stale resource reservations.
* **Documentation**
  * Expanded configuration and host-PID broker guidance.


cmd/device-plugin/nvidia/main.go
```diff
@@ -423,6 +434,76 @@ func startPlugins(c *cli.Context, o *options) ([]plugin.Interface, bool, error)
...
+			"path", gpuOperatorDriverReadyFile,
+			"driverRoot", driverRoot,
+			"devRoot", devRoot)
+	}
+
+	// Assign distinct pointers because the NVIDIA configuration loader makes
+	// NvidiaDevRoot point to NvidiaDriverRoot when no dev root is specified.
+	config.Flags.NvidiaDriverRoot = &driverRoot
+	if config.Flags.NvidiaDevRoot != nil && *config.Flags.NvidiaDevRoot == autoNvidiaDriverRoot {
+		config.Flags.NvidiaDevRoot = &devRoot
```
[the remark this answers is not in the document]
FouoF: A node without gpu-operator will never create driver-ready.

charts/hami/values.yaml
```diff
@@ -404,10 +404,13 @@ devicePlugin:
   createRuntimeClass: false
   migStrategy: "none"
   disablecorelimit: "false"
-  passDeviceSpecsEnabled: false
+  passDeviceSpecsEnabled: true
```
mesutoezdil, line 416: this default flip hits every existing install on upgrade and is independent of the auto feature. was the #867 scenario retested on a real node with the new driver root, or should this move to its own pr?

charts/hami/values.yaml
```diff
@@ -404,10 +404,13 @@ devicePlugin:
...
   deviceListStrategy: "envvar"
   nvidiaHookPath: null
-  nvidiaDriverRoot: null
+  # Root of the NVIDIA driver on the host. Set to "auto" to read the driver
+  # and device roots from GPU Operator's driver-ready contract. If the
+  # contract is absent, both roots default to / for a host-installed driver.
+  nvidiaDriverRoot: "auto"
```
mesutoezdil, line 422: when hami starts before the operator, auto silently picks / and needs a manual daemonset restart. could the gpuOperatorToolkitReady init container also wait for driver-ready when auto is set, so the fallback never happens on operator nodes?

charts/hami/values.yaml
FouoF: I successfully reproduce #867 when testing and ensure correct driver root can solve it. Strictly this change should be moved to independent PR, but if you trace the #671 to #867 you will find they are related in changelog.

charts/hami/values.yaml
FouoF: Yes, gpuOperatorToolkitReady init container can ensure the gpu-operator is ready. But it will fall into infinity loop in no gpuOperator node. So the user must judge is he using gpu-operator. On the other word, it's still not automation enough. Actually we finally return to the same problem. By the way, auto can satisfy mixed nodes with gpu-operator and no gpu-operator.

archlitchi (approved):
/lgtm

Issue #2829, opened 2026-08-26
Close the ResourceQuota TOCTOU window in Fit() for the 11 backends that still lack it


im-Toqeer-506:
Problem Statement: pkg/scheduler/webhook.go enforces ResourceQuota centrally at admission time via device.GetLocalCache().FitQuota() for all backends generically. But namespace usage isn't recorded until the pod actually schedules, so between admission and Filter there's a race: concurrent pod creation can burst past a namespace's quota. I found that nvidia and cambricon each carry a package-local fitQuota() re-check inside their Fit() path specifically to close this window — but the other 11 backends (amd, ascend, awsneuron, biren, enflame, hygon, iluvatar, kunlun, metax, mthreads, vastai) don't. This is a known, previously-attempted gap: PR #2397 ("enforce ResourceQuota in Fit() for all backends", 12 backends, size/L, dco-signoff: no) and PR #2377 ("re-check ResourceQuota in Fit() for non-NVIDIA backends", 5 backends) were both closed without merging. Cambricon's own re-check landed successfully via a much smaller, single-backend PR (#2536, merged). The pattern is clear: the gap is real, the fix is understood, but it needs to ship the way cambricon's did — small and single-backend at a time — not as one large sweep.

Proposed Solution: Don't repeat #2397/#2377's mistake. File one tracking issue describing the race and the fix shape (matching cambricon's merged approach: reuse the fitQuota-style re-check, ideally promoted to a shared helper in pkg/device as #2377 attempted with FitQuotaWithPodDevices), then land it as 11 separate small PRs, one backend per PR, each with its own tests — exactly the cambricon precedent.

Expected Benefits: Removes a genuine multi-tenancy correctness gap for 11 of 13 vendors; directly addresses something two prior contributors already tried and failed to land, so fixing the scoping problem itself has clear value to maintainers who had to close both those PRs.

Affected Components: pkg/device/{amd,ascend,awsneuron,biren,enflame,hygon,iluvatar,kunlun,metax,mthreads,vastai}/device.go; potentially a new shared function in pkg/device/quota.go alongside the existing QuotaManager.

Implementation Complexity: High overall (11 backends), but each individual PR is Low-Medium — that's the point of scoping it this way.

Breaking Changes: None for well-behaved usage; pods that were relying on the race to exceed quota will now be correctly rejected under contention — same category of change as the merged cambricon fix, so not a new precedent.

Acceptance Criteria:

[...]

PR #2849, opened 2026-08-27
Feat/quotacheck ci gate

### LINKED ISSUE #2841: add AST-based check that all device backends re-check ResourceQuota in Fit()

A Go AST-based static verification tool — `hack/tools/quotacheck/`, built
the same way as the existing `hack/tools/rbaccheck` (added in #2567) — that
walks each `pkg/device/<vendor>/device.go`'s `Fit()` implementation and
fails CI if it doesn't call the shared `ResourceQuota` re-check helper.
Add a matching `hack/verify-quota.sh` and wire it into `hack/verify-all.sh`
so `make verify` runs it on every PR, the same way `hack/verify-rbac.sh`
already does for RBAC.

**Why is this needed**:

`device.Devices.Fit()` (`pkg/device/devices.go:47`) is implemented
separately by every vendor backend. The ResourceQuota admission-vs-schedule
race (TOCTOU) fix landed for nvidia and cambricon (precedent PR #2536), but
as of `master` today, none of the other 10 backends (amd, ascend,
awsneuron, biren, enflame, hygon, iluvatar, kunlun, metax, mthreads,
vastai) call the re-check — verified directly by grepping
`pkg/device/*/device.go` on master. This gap has already been tracked
twice — #2829 (per-backend fix) and #2379 (a proposed runtime conformance
suite) — and both were closed as completed without the corresponding code
landing on master. Nothing currently catches a backend silently missing
this check, so reviewers have to manually diff every `Fit()` against
nvidia/cambricon to notice, and the gap can silently reappear even after
it's fixed.

- Scope: this issue adds the CI gate only; it does not itself fix the 9
  backends currently missing the check — each backend still gets its own
  PR, per #2829's original scoping.
- Related: #2829, #2379, #2536 (precedent fix), #2567 (precedent tool —
  Go AST-based RBAC check, direct template for this one).
- Acceptance criteria: tool flags all backends currently missing the
  check as its own test fixture; passes with nvidia/cambricon as
  baseline; wired into `make verify`; unit tests for the AST-walking
  logic itself (mirroring `hack/tools/rbaccheck`'s tests).

im-Toqeer-506:
**What type of PR is this?**
/kind feature

Adds `hack/tools/quotacheck/`, an AST-based static check (built the same way
as `hack/tools/rbaccheck`, #2567) that verifies every `pkg/device/<vendor>/device.go`'s
`Fit()` reaches the shared `device.QuotaManager.FitQuota` re-check — closing
the gap where a backend can silently miss the admission-vs-schedule
ResourceQuota (TOCTOU) fix that landed for nvidia and cambricon in #2536.

This PR was written primarily by Claude Code, working from the issue text and
`rbaccheck` as a template. I reviewed and understand the AST-walking, the
call-graph traversal, and the allowlist/stale-entry design, and can answer
questions about any part of it. It's a static-analysis tool over source files
only — no device allocation, device plugin, or in-container isolation code is
touched or executed, so hardware validation doesn't apply here; correctness is
covered by the fixture and integration tests below.

Fixes #2841

**What's included**:
- `hack/tools/quotacheck/main.go` — parses each vendor's `device.go`, finds
  `Fit()`, and walks the package's call graph (following local wrappers like
  the existing `fitQuota()` helper) to confirm it reaches `FitQuota()`.
- `hack/tools/quotacheck/main_test.go` — fixture cases (direct call, same-file
  wrapper, cross-file wrapper, missing check, missing `Fit()`), plus an
  integration test against the real backends confirming nvidia/cambricon/amd
  pass and the rest currently fail.
- `hack/verify-quota.sh` — CI entry point, mirrors `hack/verify-rbac.sh`.
- `hack/verify-all.sh` — wires `verify-quota.sh` into `make verify`.

**Scope note — allowlist for backends not yet fixed**:
10 of 13 backends (ascend, awsneuron, biren, enflame, hygon, iluvatar, kunlun,
metax, mthreads, vastai) don't call the re-check yet; those fixes are separate
PRs per #2829's scoping. `verify-quota.sh` passes `quotacheck` an explicit
`-allow` list of those ten so `make verify` doesn't break on master before
they land. An unlisted backend that fails still fails the build immediately,
and a listed backend that starts passing without being removed from the
allowlist also fails the build — each backend's fix PR just deletes its own
name from `ALLOWED_VENDORS`, and the gate tightens on its own.

**Test plan**:
- `go build ./...`
- `go test ./hack/tools/quotacheck/... -race -count=1 -v`
- `golangci-lint run ./hack/tools/quotacheck/...`
- `gofmt -l` / import-aliases clean
- `bash hack/verify-quota.sh` exits 0 on current master (10 allowed failures reported)
- Manually confirmed regression detection: an unlisted failing backend fails
  the run, and a stale allowlist entry (a listed backend now passing) also
  fails the run

```release-note
NONE
```

## Summary by CodeRabbit

* **New Features**
  * Added automated verification that device fitting logic performs required quota checks.
  * Added support for allowlisting known exceptions and detecting outdated allowlist entries.
  * Integrated quota verification into the standard project verification workflow.
* **Tests**
  * Added comprehensive coverage for compliant, missing, asynchronous, and allowlisted quota-check scenarios.
* **Chores**
  * Added repository tooling and configuration for running quota verification.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,7 +10,7 @@ libvgpu.so
 vendor
 license
 vgpuvalidator
-quotacheck
+/quotacheck
 _output/
 coverage.out
 .DS_Store
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@ libvgpu.so
 vendor
 license
 vgpuvalidator
+quotacheck
 _output/
 coverage.out
 .DS_Store
--- a/hack/tools/quotacheck/main.go
+++ b/hack/tools/quotacheck/main.go
@@ -241,6 +241,11 @@ func parsePackage(fset *token.FileSet, dir string) ([]*ast.File, error) {
 // the FuncDecls of the local functions/methods it calls directly. Only
 // same-package callees can be resolved this way, which is sufficient for
 // following a local wrapper like fitQuota() back to FitQuota().
+//
+// Calls inside a function literal are only followed when that literal is
+// statically invoked (an immediately-invoked func expression); a closure
+// that is merely declared and never called cannot influence Fit()'s
+// behaviour and so must not count towards reachability.
 func buildCallGraph(files []*ast.File) map[string][]*ast.FuncDecl {
 	byName := make(map[string]*ast.FuncDecl)
 	for _, f := range files {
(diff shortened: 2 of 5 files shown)


mesutoezdil:
closing this: the allowlist exempts 11 of 13 vendors, so the gate only covers the two backends that already call FitQuota.

PR #2852, opened 2026-08-27
fix(dashboard): use native node labels for host panels

### LINKED ISSUE #2843: [Bug][Dashboard] Host GPU panels require redundant per-device scheduler metric matches
**What happened**:

The “Host GPU memory used” and “Host GPU utilization” dashboard panels filter by node through a scheduler metric join:

```promql
hami_host_gpu_memory_used_bytes
and on (device_uuid)
hami_gpu_memory_limit_bytes{node=~"$node"}
```

This join was introduced in #2301 because the host metrics did not have a `node` label at that time. However, #2580 subsequently added a native `node` label to both host metrics.

The join now creates an unnecessary dependency on scheduler metrics. Valid vGPUmonitor data may be hidden when the corresponding scheduler series is unavailable or stale.

The host panels should filter directly using the native `node` label:

```promql
hami_host_gpu_memory_used_bytes{node=~"$node"}
```

```promql
hami_host_gpu_utilization_ratio{node=~"$node"}
```

This would avoid requiring a matching scheduler series for every host GPU series while preserving node filtering through the native label.

1. Open `dashboards/hami-vgpu-dashboard.json`.
2. Locate the “Host GPU memory used” and “Host GPU utilization” panels.
3. Observe that both queries join their vGPUmonitor metric with `hami_gpu_memory_limit_bytes`.
4. Compare the host metric series with the scheduler metric series and consider a `device_uuid` present only in the host metric, such as during a missing or stale scheduler series.
5. Observe that the `and on (device_uuid)` expression removes that host series even though it already contains the selected `node` label.

This is a follow-up to #2580 and is related to the observability work in #2126.

If this direction is acceptable, I would like to submit a focused PR updating both queries and adding a dashboard regression test.

The requested workload manifests, accelerator diagnostics, device-plugin logs, scheduler logs, kubelet logs and kernel output are not applicable because this report concerns the PromQL expressions shipped in the static Grafana dashboard.

**Environment**:
- HAMi version: Current `master` after #2580
- Kubernetes version: Not applicable
- Accelerator vendor and model: Not vendor-specific
- HAMi device backend and requested resource names: Not applicable
- Accelerator driver and runtime version: Not applicable
- Workload image and tag: Not applicable
- Operating system and kernel version: Not applicable
- Others: `dashboards/hami-vgpu-dashboard.json`

blackdragoon26:
**What type of PR is this?**

/kind bug

After #2580, the host GPU memory and utilization metrics include a native `node` label. The dashboard still filters these metrics through a `device_uuid` join with scheduler metrics.

This PR updates both host GPU panels to filter directly using their native `node` label. This prevents valid vGPUmonitor series from being removed when a matching scheduler series is unavailable.

A regression test verifies the query contract for both panels.

Fixes #2843

Verified with:

- `make verify`
- `go test ./dashboards`
- `go test ./cmd/vGPUmonitor`

Yes. The host GPU memory and utilization panels now use the metrics’ native node labels without requiring matching per-device scheduler series.

AI assistance disclosure: OpenAI Codex assisted with repository inspection, implementation, test preparation, and verification. I reviewed the changes and take responsibility for the contribution.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved host GPU memory and utilization dashboards to display native host metrics more reliably.
  * Removed an unnecessary filtering condition that could exclude valid host GPU data.
* **Tests**
  * Added coverage to verify host GPU panels use the expected native node metrics.
  * Streamlined dashboard fixture loading and panel validation in tests.

### DIFF
--- a/dashboards/hami-vgpu-dashboard.json
+++ b/dashboards/hami-vgpu-dashboard.json
@@ -273,7 +273,7 @@
         {
           "datasource": { "type": "prometheus", "uid": "${datasource}" },
           "editorMode": "code",
-          "expr": "hami_host_gpu_memory_used_bytes and on (device_uuid) hami_gpu_memory_limit_bytes{node=~\"$node\"}",
+          "expr": "hami_host_gpu_memory_used_bytes{node=~\"$node\"}",
           "legendFormat": "{{device_type}} idx{{device_index}} ({{device_uuid}})",
           "refId": "A"
         }
@@ -319,7 +319,7 @@
         {
           "datasource": { "type": "prometheus", "uid": "${datasource}" },
           "editorMode": "code",
-          "expr": "hami_host_gpu_utilization_ratio and on (device_uuid) hami_gpu_memory_limit_bytes{node=~\"$node\"}",
+          "expr": "hami_host_gpu_utilization_ratio{node=~\"$node\"}",
           "legendFormat": "{{device_type}} idx{{device_index}} ({{device_uuid}})",
           "refId": "A"
         }
--- a/dashboards/dashboard_test.go
+++ b/dashboards/dashboard_test.go
@@ -45,7 +45,10 @@ type panel struct {
 	} `json:"targets"`
 }
 
-func TestNodeGPUMemoryAllocatedRatioPanelUsesFractionScale(t *testing.T) {
+// loadDashboard reads and parses the importable HAMi dashboard fixture.
+func loadDashboard(t *testing.T) dashboard {
+	t.Helper()
+
 	data, err := os.ReadFile("hami-vgpu-dashboard.json")
 	if err != nil {
 		t.Fatalf("read dashboard: %v", err)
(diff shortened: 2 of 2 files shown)


dashboards/hami-vgpu-dashboard.json
```diff
@@ -273,7 +273,7 @@
         {
           "datasource": { "type": "prometheus", "uid": "${datasource}" },
           "editorMode": "code",
-          "expr": "hami_host_gpu_memory_used_bytes and on (device_uuid) hami_gpu_memory_limit_bytes{node=~\"$node\"}",
+          "expr": "hami_host_gpu_memory_used_bytes{node=~\"$node\"}",
```
mesutoezdil, line 276: the join was not just a node filter. it also limited the panel to GPUs the scheduler knows. now every GPU on the node shows, HAMi managed or not. intended?

dashboards/hami-vgpu-dashboard.json
```diff
@@ -319,7 +319,7 @@
         {
           "datasource": { "type": "prometheus", "uid": "${datasource}" },
           "editorMode": "code",
-          "expr": "hami_host_gpu_utilization_ratio and on (device_uuid) hami_gpu_memory_limit_bytes{node=~\"$node\"}",
+          "expr": "hami_host_gpu_utilization_ratio{node=~\"$node\"}",
```
mesutoezdil, line 322: $node comes from label_values(hami_gpu_memory_limit_bytes, node), a scheduler metric. did you confirm vGPUmonitor writes the same node value?

dashboards/hami-vgpu-dashboard.json
blackdragoon26: Yes, this is intended. As these are host GPU panels, so, i believe they should show all physical gpu's
 on the selected hami node without depending on scheduler metrics

dashboards/hami-vgpu-dashboard.json
blackdragoon26: Yes, NODE_NAME comes from [spec.nodeName](https://github.com/Project-HAMi/HAMi/blob/master/charts/hami/templates/device-plugin/daemonsetnvidia.yaml#L92-L96), and vGPUmonitor uses the same value for the [node metric label](https://github.com/Project-HAMi/HAMi/blob/master/cmd/vGPUmonitor/metrics.go#L320-L325). so, it matches the k8s node name used by scheduler metrics.

mesutoezdil:
/lgtm

verified both: NODE_NAME really is spec.nodeName end to end, and scheduling the panel by physical presence rather than scheduler-known GPUs is a reasonable call for a host panel.

FouoF (approved):

PR #2853, opened 2026-08-27
fix(ascend): gate 910C module pair allocation on SuperPod

mesutoezdil:
/kind bug

Module pair allocation was added for SuperPod environments in #1610, and #2005 introduced the `superPod` config flag but gated only `MutateAdmission` on it, leaving `Fit` forcing pairing for every 910C request larger than one. In split mode an odd request therefore passed admission and was then rejected on every node, because the pair combination selects whole cards and returns more devices than requested, so such pods stayed Pending forever.

release-note:
Fixed Ascend 910C scheduling in non-SuperPod deployments, where an odd device request passed admission and then failed to schedule on every node.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Ascend 910C multi-device scheduling to correctly account for effective device requests.
  * Ensured multi-device requests use topology-aware pairing only when SuperPod mode is enabled.
  * Improved scheduling of split-mode requests, including odd device counts.

* **Tests**
  * Added regression coverage for successful three-device allocation across healthy Ascend 910C devices.

### DIFF
--- a/pkg/device/ascend/device.go
+++ b/pkg/device/ascend/device.go
@@ -174,6 +174,9 @@ func (dev *Devices) MutateAdmission(ctr *corev1.Container, p *corev1.Pod) (bool,
 			}
 		}
 	}
+	// count, not reqNum: the 910C SuperPod rewrite to 2 is HAMi's module
+	// packaging rule, not a multi device request, and #2005 added 910C vNPU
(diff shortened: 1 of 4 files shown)


archlitchi (approved):
/lgtm

PR #2855, opened 2026-08-27
feat(nvidia): score MIG placements and add backtracking fallback.

### LINKED ISSUE #2831: Enhance Dynamic MIG: Optimize the MIG template allocation algorithm
Improve MIG profile placement and template selection so that each allocation preserves the largest possible contiguous free region and reduces fragmentation.

For common seven-slice GPUs, such as an A100 configured as `7 × 1g`, a greedy strategy can be used:

1. Process profiles that require more slices first.
2. Evaluate the remaining contiguous space after each candidate placement.
3. Prefer placements that do not split a large free region into smaller fragments.
4. Use a stable and deterministic placement order when multiple candidates have the same score.
5. Fall back to backtracking or exhaustive search when the greedy strategy cannot find a valid solution.

For example, a `1g` instance should preferably be allocated from the edge of a free region so that the remaining contiguous slices can still accommodate `2g`, `3g`, or larger profiles.

The algorithm should cover:

- Multiple allocations of the same profile.
- Mixed small and large profiles.
- GPUs with existing active MIG instances.
- Cases where greedy selection fails but backtracking can still find a valid placement.
- Different GPU models, slice counts, and placement constraints reported by NVML.

maishivamhoo123:
**What type of PR is this?**
/kind feature

Before this PR MIG  takes the first free slot based on the profile name example :- 1g and 3g scan from the top, the rest from the bottom, and it never look back what is left.
This PR changes check Whenever a container asks for several slices on one card, the profile with the fewest free positions is placed first. Each and every candidate position is scored by the largest free run it leaves, after then by how many legal placements survive.
Fixes #2831 
AI Disclosure :- This pr is used AI for understanding the code  , optimizing the code quality and writing comment (for increasing the readblity )and writing the test.

## Summary by CodeRabbit

- **New Features**
  - Improved NVIDIA MIG allocation by planning multiple slices on the same GPU together.
  - Selects suitable profiles and conflict-free placements to use GPU capacity more efficiently.
  - Makes placement decisions that better preserve capacity in fragmented layouts.
  - Retains valid allocation plans through resource updates.

- **Bug Fixes**
  - Added fallback handling when joint MIG planning is unavailable.
  - Prevents stale or conflicting allocation plans from being committed.
  - Improves reliability when requested MIG slices cannot be placed together.

### DIFF
--- a/pkg/device/nvidia/device.go
+++ b/pkg/device/nvidia/device.go
@@ -21,6 +21,7 @@ import (
 	"flag"
 	"fmt"
 	"math"
+	"slices"
 	"sort"
 	"strconv"
 	"strings"
@@ -586,25 +587,44 @@ func (dev *NvidiaGPUDevices) GenerateResourceRequests(ctr *corev1.Container) dev
 }
 
 func (dev *NvidiaGPUDevices) CustomFilterRule(allocated *device.PodDevices, request device.ContainerDeviceRequest, toAllocate device.ContainerDevices, devusage *device.DeviceUsage) bool {
-	if devusage.Mode == MigMode {
-		occupied := occupiedMigPlacements(devusage.MigAllocationsInUse)
-		for _, existing := range toAllocate {
-			if existing.UUID != devusage.ID {
-				continue
-			}
-			_, placement, ok := selectMigCandidate(devusage.MigProfiles, occupied, existing.Usedmem)
-			if !ok {
-				return false
-			}
-			occupied = append(occupied, placement)
-		}
-		_, _, ok := selectMigCandidate(devusage.MigProfiles, occupied, request.Memreq)
-		return ok
+	if devusage.Mode != MigMode {
+		return true
 	}
-	return true
+	occupied := occupiedMigPlacements(devusage.MigAllocationsInUse)
+	// Plan every slice of this container on this card together; reject only when no layout exists.
+	memories := queuedMigMemories(toAllocate, devusage.ID)
+	memories = append(memories, request.Memreq)
+	if _, _, ok := planMigAllocations(devusage.MigProfiles, occupied, memories); ok {
+		return true
+	}
+	// Fall back to the sequential path, which may upgrade a request to a larger profile.
+	for _, existing := range toAllocate {
+		if existing.UUID != devusage.ID {
+			continue
+		}
+		_, placement, ok := selectMigCandidate(devusage.MigProfiles, occupied, existing.Usedmem)
+		if !ok {
+			return false
+		}
+		occupied = append(occupied, placement)
+	}
+	_, _, ok := selectMigCandidate(devusage.MigProfiles, occupied, request.Memreq)
+	return ok
 }
 
-func selectMigCandidate(profiles []device.MigProfile, occupied []device.MigPlacement, memory int32) (device.MigProfile, device.MigPlacement, bool) {
+// queuedMigMemories returns the memory requests already queued for this container on the given card.
+func queuedMigMemories(toAllocate device.ContainerDevices, uuid string) []int32 {
+	memories := make([]int32, 0, len(toAllocate)+1)
+	for _, existing := range toAllocate {
+		if existing.UUID == uuid {
+			memories = append(memories, existing.Usedmem)
+		}
+	}
+	return memories
+}
+
+// migProfilesByMemory returns the profiles ordered smallest first.
+func migProfilesByMemory(profiles []device.MigProfile) []device.MigProfile {
 	candidates := append([]device.MigProfile(nil), profiles...)
 	sort.SliceStable(candidates, func(i, j int) bool {
 		if candidates[i].MemoryMB != candidates[j].MemoryMB {
@@ -612,7 +632,22 @@ func selectMigCandidate(profiles []device.MigProfile, occupied []device.MigPlace
 		}
 		return candidates[i].SliceCount < candidates[j].SliceCount
 	})
-	for _, profile := range candidates {
+	return candidates
+}
+
+// migProfileForMemory returns the smallest profile whose memory covers the request.
+func migProfileForMemory(profiles []device.MigProfile, memory int32) (device.MigProfile, bool) {
+	for _, profile := range migProfilesByMemory(profiles) {
+		if profile.MemoryMB >= memory {
+			return profile, true
+		}
+	}
+	return device.MigProfile{}, false
+}
+
+// selectMigCandidate picks the smallest covering profile that still has a free placement, and its best placement.
+func selectMigCandidate(profiles []device.MigProfile, occupied []device.MigPlacement, memory int32) (device.MigProfile, device.MigPlacement, bool) {
+	for _, profile := range migProfilesByMemory(profiles) {
 		if profile.MemoryMB < memory {
 			continue
 		}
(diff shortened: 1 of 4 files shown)


pkg/device/nvidia/device.go
```diff
@@ -590,33 +587,67 @@ func (dev *NvidiaGPUDevices) GenerateResourceRequests(ctr *corev1.Container) dev
...
-		for _, existing := range toAllocate {
-			if existing.UUID != devusage.ID {
-				continue
-			}
-			_, placement, ok := selectMigCandidate(devusage.MigProfiles, occupied, existing.Usedmem)
-			if !ok {
-				return false
-			}
-			occupied = append(occupied, placement)
+	if devusage.Mode != MigMode {
```
FouoF: Only recommendation: This change enhance the readability but better not include in this PR to keep it smallest and focus.

pkg/device/nvidia/mig_topology.go
```diff
@@ -55,78 +71,230 @@ func occupiedMigPlacements(allocations []device.MigAllocation) []device.MigPlace
...
+	}
+	return result, true
+}
 
-	used := append([]device.MigPlacement(nil), occupied...)
-	var place func(int) bool
-	place = func(idx int) bool {
-		if idx == len(ordered) {
+// backtrackMigProfiles searches every assignment, trying candidates best score first.
+func backtrackMigProfiles(profiles []device.MigProfile, occupied []device.MigPlacement, requested []string) ([]device.MigPlacement, bool) {
```
FouoF: The backtrace seems useless according to current supported MIG profile. I think we can safely remove it.

FouoF:
image
Also I think this allocation method can be further optimized. Observing this supported MIG profiles for NVIDIA A, H, and B series GPU, we can notice that the config 5, 6, 7 is better than 2, 3, 4, so we should find a way to give preference for them. Also we can notice that if config 2, 3, 4 is excluded the slots can be divided to 2 individual parts. We can treat them different to achieve the preference. This divide can simplify the following score too.

maishivamhoo123:
> image Also I think this allocation method can be further optimized. Observing this supported MIG profiles for NVIDIA A, H, and B series GPU, we can notice that the config 5, 6, 7 is better than 2, 3, 4, so we should find a way to give preference for them. Also we can notice that if config 2, 3, 4 is excluded the slots can be divided to 2 individual parts. We can treat them different to achieve the preference. This divide can simplify the following score too.

Thanks for the review So first, config 2, 3, 4 all need a 4g, and 4g is not in our allowlist, so I simply ignore them. Once they are out of the picture, every 1g, 2g and 3g placement stays inside one half of the card, either slot 0-3 or slot 4-7. Only the 7g covers both halves. Then I noticed the two halves are not the same. The lower half can fit 4 small instances, but the upper half only 3, because slot 7 can only be used by a 3g. So I reserve the upper half for the 3g, and the smaller profiles go to the lower half first. Inside each half I fill from the outer edge towards the middle, so the 2g span next to the middle stays free as long as possible. Because of this split, the score became very simple. It is now just 4 comparisons: how many halves are still empty, is the profile in its own half, distance from the middle, and finally lowest start. The largest-free-run, the survivor counting and the backtracking are all removed.

pkg/device/nvidia/mig_topology.go
```diff
@@ -18,26 +18,248 @@ package nvidia
...
+	if sliceCount < 2 {
+		return whole
+	}
+	mid := uint32(sliceCount / 2)
+	for _, profile := range profiles {
+		for _, p := range profile.Placements {
+			if p.Size == 0 || p.Size >= uint32(sliceCount) {
+				continue
+			}
+			if p.Start < mid && p.Start+p.Size > mid {
```
mesutoezdil, line 64: on a 7 slice A100 mid is 3, and 4g at start 0 size 4 straddles it. so this returns the whole card and score() skips every zone branch. which hardware engages the two region path?

pkg/device/nvidia/mig_topology.go
```diff
@@ -18,26 +18,248 @@ package nvidia
...
+		if remaining == 0 {
+			return true
+		}
+		idx := nextMigRequest(used, requested, placed)
+		if idx < 0 {
+			return false
+		}
+		placed[idx] = true
+		for _, candidate := range l.rank(used, requested[idx]) {
+			if budget == 0 {
```
mesutoezdil, line 267: the doc says a valid layout is never rejected. budget is shared across the recursion, so hitting 0 returns false. could you reconcile?

pkg/device/nvidia/device.go
```diff
@@ -588,6 +589,13 @@ func (dev *NvidiaGPUDevices) GenerateResourceRequests(ctr *corev1.Container) dev
 func (dev *NvidiaGPUDevices) CustomFilterRule(allocated *device.PodDevices, request device.ContainerDeviceRequest, toAllocate device.ContainerDevices, devusage *device.DeviceUsage) bool {
 	if devusage.Mode == MigMode {
 		occupied := occupiedMigPlacements(devusage.MigAllocationsInUse)
+		// Plan every slice of this container on this card together; reject only when no layout exists.
+		memories := queuedMigMemories(toAllocate, devusage.ID)
+		memories = append(memories, request.Memreq)
+		if _, _, ok := planMigAllocations(devusage.MigProfiles, occupied, memories); ok {
```
mesutoezdil, line 595: this backtracks per candidate device per node in Filter, and recordMigPlans repeats it per card. any timing on a busy node?

pkg/device/nvidia/device.go
```diff
@@ -624,13 +659,103 @@ func selectMigCandidate(profiles []device.MigProfile, occupied []device.MigPlace
...
+
 func (dev *NvidiaGPUDevices) ScoreNode(node *corev1.Node, podDevices device.PodSingleDevice, previous []*device.DeviceUsage, policy string) float32 {
 	return 0
 }
 
 func (dev *NvidiaGPUDevices) AddResourceUsage(pod *corev1.Pod, n *device.DeviceUsage, ctr *device.ContainerDevice) error {
 	if n.Mode == MigMode {
-		profile, placement, ok := selectMigCandidate(n.MigProfiles, occupiedMigPlacements(n.MigAllocationsInUse), ctr.Usedmem)
+		occupied := occupiedMigPlacements(n.MigAllocationsInUse)
+		profile, placement, ok := plannedMigAllocation(n.MigProfiles, occupied, ctr)
```
mesutoezdil, line 755: gate 2 asks for hardware validation here, with device type and driver version. could you add it? also, what was preferHighMigPlacement for?

pkg/device/nvidia/mig_topology.go
maishivamhoo123: Thank you @mesutoezdil you are Right Added the case to TestMigRegionsSplitTheCardInHalves.

pkg/device/nvidia/device.go
maishivamhoo123: Measured: worst case ~35 µs per plan on my laptop (BenchmarkPlaceMigProfilesWorstFit / WorstRejection, Ryzen AI 7), and the exhaustive test caps any plan at 15 candidates on A100/A30/H100. Fit does under 25 plans for 8 cards × 7 slices, so under 1 ms per node even in the worst case. recordMigPlans has to replan because in topology mode the final combination is a subset of what Filter saw.

pkg/device/nvidia/device.go
maishivamhoo123: I don't have MIG hardware and CI runs on a Tesla P4, which can't do MIG. If someone with an A100/H100 can run three pods with gpumem 20000/10000/5000 in mig mode, the expected layout is 3g@4, 2g@0, 1g@2, and I'll add the output to the description. preferHighMigPlacement came from #2378 it scanned 1g and 3g placements from the top and everything else from the bottom, so the low spans stayed free for 2g/4g/7g. The zone split does the same thing explicitly now (3g pinned to the upper half), except 1g packs from slot 0, since filling from the top kills the 3g span with the first 1g and 4g isn't in our allowlist anyway.

FouoF:
@maishivamhoo123 Excellent work！But still a important thing should correct:

> 4g is not in our allowlist

should be "4g is not in **default** allowlist", we can't just ignore it. Actually after looking into the supported MIG profiles, I think we should change the default allowlist and add 4g to it.

maishivamhoo123:
> @maishivamhoo123 Excellent work！But still a important thing should correct:
> 
> > 4g is not in our allowlist
> 
> should be "4g is not in **default** allowlist", we can't just ignore it. Actually after looking into the supported MIG profiles, I think we should change the default allowlist and add 4g to it.

Thanks @FouoF, you're right, default allowlist is the correct wording, my mistake. The algorithm doesn't ignore 4g though. There's no profile-name logic left in mig_topology.go every profile goes through the same placement-table path, and the request with the fewest free placements is placed first:
```
free := len(freeMigPlacements(profile, occupied))

case best == -1,
	free < bestFree,
```
A 4g has exactly one placement (slot 0), so it always goes first when requested with other slices. It's in the H100 table, the slot-7-less A100 allowlist and now the A100 fixture too , TestPlaceMigProfilesHandles4gWhenRequestedTogether checks 1g + 4g and 4g + 3g end to end, and the exhaustive test covers every request that includes it. With pods arriving one at a time, small profiles consume the 4g span before the upper 3g span, which is the 5/6/7 over 2/3/4 preference you described. one caveat on the default allowlist a 20 GB request maps to 3g.20gb before 4g.20gb, so adding 4g won't change placements while 3g is allowed. I can add it here or in a follow-up, your call?.

FouoF:
> one caveat on the default allowlist a 20 GB request maps to 3g.20gb before 4g.20gb, so adding 4g won't change placements while 3g is allowed.

@maishivamhoo123 Yes, this is the limitation of our memory-only match strategy. Assuming the user don't want to waste the core, I have 3 propose:
1. The first 3g in one GPU will auto upgrade to 4g, simple but not stable.
2. Allowing annotate a Pod to prefer 4g instance, controllable but complex.
3. Both of above two and can be configed.

What do you think about it?

maishivamhoo123:
> What do you think about it?

@FouoF , i think option 2 is good one because if we use the existing cpucores request as the signal instead as of a new anotation match on memory and cores, so gpumem: 20000 alone still resolves to 3g and adding gpucores above the 3g share resolves to 4g, Option 1 worries me beyond instability , Usedcores changes silently on the upgrade, and the 4g lands in the lower half the split keeps for small profiles. Option 3 can wrap both in config later if needed.  

I think we should keep this out of the this PR since it's profile selection, not placement.
is it sound good?

FouoF:
@maishivamhoo123 You are right. We should discuss it in another issue or PR. Drop a quick reply here: MIG naturally couples cores and memory together that why we only evaluate memory, better not introduce core request in MIG.

FouoF (approved):

maishivamhoo123:
> @maishivamhoo123 You are right. We should discuss it in another issue or PR. Drop a quick reply here: MIG naturally couples cores and memory together that why we only evaluate memory, better not introduce core request in MIG.

You are right ,  because MIG profiles couple cores and memory, the signal should be the annotation as you proposed. Opened #2950  .

PR #2866, opened 2026-08-27
feat(scheduler): bound Filter's per-node goroutine fan-out

### LINKED ISSUE #2858: Bound goroutine fan-out in scheduler Filter

Limit the number of goroutines created by `calcScoreWithOptions` in `pkg/scheduler/score.go`.

Currently, it creates one goroutine per candidate node. I propose using a bounded worker pool, with the default worker count set to `runtime.GOMAXPROCS(0)` and exposed as a scheduler flag alongside `--kube-qps` and `--kube-burst`.

The scoring logic and result aggregation should remain unchanged.

**Why is this needed**:

For a large number of candidate nodes, the current implementation can create a large number of goroutines at once. Each goroutine also performs `NodeUsage.DeepCopy`, which can result in high memory usage and allocation pressure during a Filter call.

For example, a local benchmark with 1000 nodes and 3 init containers showed:

```text
41.3 ms/op
43.7 MB/op
487,416 allocs/op
1000 peak goroutines

### PR DESCRIPTION
**Pull request by im-Toqeer-506**
**What type of PR is this?**
/kind feature

`calcScoreWithOptions` spawned one goroutine per candidate node, and each call to `scoreNode` deep-copies the node's `NodeUsage` once per app/init container — so goroutine count and simultaneous copies both scaled with cluster size inside a synchronous extender call kube-scheduler blocks on.

This routes scoring through a worker pool bounded by a new `--filter-parallelism` flag, defaulting to `runtime.GOMAXPROCS` and capped at the node count (so small clusters keep existing parallelism, empty clusters spawn nothing). Scoring, aggregation, and the Filter response are unchanged.

**AI disclosure:**
I used AI assistance to verify the implementation and draft the PR description. I reviewed the implementation and final changes myself.

The changes are limited to the scheduler extender. No device allocation, device plugin, or in-container isolation code was modified. Therefore, the changes are covered by unit and mock testing in accordance with Contribution Gate 2.

Fixes #2858

```release-note
Add `--filter-parallelism` to the scheduler, bounding how many candidate nodes are scored concurrently during Filter. Defaults to `runtime.GOMAXPROCS`.
```

## New Features
- Added a command-line option to control scheduler filtering parallelism.
- Scheduler filtering now supports bounded concurrent node scoring, with automatic CPU-based defaults and safeguards for small workloads.

## Bug Fixes
- Improved consistency and reliability of parallel node scoring across different parallelism settings, including empty workloads.

### DIFF
--- a/cmd/scheduler/main.go
+++ b/cmd/scheduler/main.go
@@ -69,7 +69,7 @@ func init() {
 	rootCmd.Flags().StringVar(&config.SchedulerName, "scheduler-name", "", "the name to be added to pod.spec.schedulerName if not empty")
 	rootCmd.Flags().StringVar(&config.NodeSchedulerPolicy, "node-scheduler-policy", util.NodeSchedulerPolicyBinpack.String(), "node scheduler policy")
 	rootCmd.Flags().StringVar(&device.GPUSchedulerPolicy, "gpu-scheduler-policy", util.GPUSchedulerPolicySpread.String(), "GPU scheduler policy")
-	rootCmd.Flags().IntVar(&config.FilterParallelism, "filter-parallelism", 0, "maximum number of candidate nodes scored concurrently during Filter (0 means runtime.GOMAXPROCS)")
+	rootCmd.Flags().IntVar(&config.FilterParallelism, "filter-parallelism", 0, "maximum number of candidate nodes scored concurrently during Filter (zero or negative means runtime.GOMAXPROCS)")
 	rootCmd.Flags().StringVar(&config.MetricsBindAddress, "metrics-bind-address", ":9395", "The TCP address that the scheduler should bind to for serving prometheus metrics(e.g. 127.0.0.1:9395, :9395)")
 	rootCmd.Flags().StringToStringVar(&config.NodeLabelSelector, "node-label-selector", nil, "key=value pairs separated by commas")
 
--- a/cmd/scheduler/main.go
+++ b/cmd/scheduler/main.go
@@ -69,6 +69,7 @@ func init() {
 	rootCmd.Flags().StringVar(&config.SchedulerName, "scheduler-name", "", "the name to be added to pod.spec.schedulerName if not empty")
 	rootCmd.Flags().StringVar(&config.NodeSchedulerPolicy, "node-scheduler-policy", util.NodeSchedulerPolicyBinpack.String(), "node scheduler policy")
 	rootCmd.Flags().StringVar(&device.GPUSchedulerPolicy, "gpu-scheduler-policy", util.GPUSchedulerPolicySpread.String(), "GPU scheduler policy")
+	rootCmd.Flags().IntVar(&config.FilterParallelism, "filter-parallelism", 0, "maximum number of candidate nodes scored concurrently during Filter (0 means runtime.GOMAXPROCS)")
 	rootCmd.Flags().StringVar(&config.MetricsBindAddress, "metrics-bind-address", ":9395", "The TCP address that the scheduler should bind to for serving prometheus metrics(e.g. 127.0.0.1:9395, :9395)")
 	rootCmd.Flags().StringToStringVar(&config.NodeLabelSelector, "node-label-selector", nil, "key=value pairs separated by commas")
 
--- a/pkg/scheduler/score.go
+++ b/pkg/scheduler/score.go
@@ -417,7 +417,21 @@ func filterWorkerCount(nodeCount int) int {
 
 // runNodeWorkers calls fn once for every node, running at most workers of them
 // at a time. Callers must make fn safe for concurrent use.
+//
+// The worker count is normalised here rather than only in the caller: with no
+// worker no one ever receives from jobs, so a non-positive value handed in by
+// any future caller would block on the first send instead of failing. Such a
+// value falls back to runtime.GOMAXPROCS, the same meaning
+// config.FilterParallelism gives it, and an empty node set returns before any
+// goroutine is started.
 func runNodeWorkers(nodes map[string]*NodeUsage, workers int, fn func(nodeID string, node *NodeUsage)) {
+	if len(nodes) == 0 {
+		return
+	}
+	if workers <= 0 {
+		workers = runtime.GOMAXPROCS(0)
+	}
+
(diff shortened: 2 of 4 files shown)


cmd/scheduler/main.go
```diff
@@ -69,6 +69,7 @@ func init() {
 	rootCmd.Flags().StringVar(&config.SchedulerName, "scheduler-name", "", "the name to be added to pod.spec.schedulerName if not empty")
 	rootCmd.Flags().StringVar(&config.NodeSchedulerPolicy, "node-scheduler-policy", util.NodeSchedulerPolicyBinpack.String(), "node scheduler policy")
 	rootCmd.Flags().StringVar(&device.GPUSchedulerPolicy, "gpu-scheduler-policy", util.GPUSchedulerPolicySpread.String(), "GPU scheduler policy")
+	rootCmd.Flags().IntVar(&config.FilterParallelism, "filter-parallelism", 0, "maximum number of candidate nodes scored concurrently during Filter (0 means runtime.GOMAXPROCS)")
```
Eshiv-Pandey: filterWorkerCount treats any value <= 0 (including negatives) as GOMAXPROCS, and the config.go doc-comment already says "zero or neg." This usage string says only "0 means"  worth aligning so --help matches the actual contract.

pkg/scheduler/score.go
```diff
@@ -399,6 +400,48 @@ func (s *Scheduler) recordFilteringFailures(task *corev1.Pod, failureReason map[
...
+	workers := config.FilterParallelism
+	if workers <= 0 {
+		workers = runtime.GOMAXPROCS(0)
+	}
+	return min(workers, nodeCount)
+}
+
+// runNodeWorkers calls fn once for every node, running at most workers of them
+// at a time. Callers must make fn safe for concurrent use.
+func runNodeWorkers(nodes map[string]*NodeUsage, workers int, fn func(nodeID string, node *NodeUsage)) {
```
Eshiv-Pandey, line 427: errgroup (x/sync, already in go.mod) replaces this: g.SetLimit(workers) + g.Go(...)  fewer lines, buffered, and no workers<=0 deadlock. if keeping the manual pool, guard workers<=0 + test it.

pkg/scheduler/score.go
```diff
@@ -399,6 +400,48 @@ func (s *Scheduler) recordFilteringFailures(task *corev1.Pod, failureReason map[
...
+// candidate node makes both the goroutine count and the concurrent copy
+// footprint grow with cluster size. config.FilterParallelism caps that;
+// zero or negative selects runtime.GOMAXPROCS. The result never exceeds
+// nodeCount, so clusters smaller than the cap keep the parallelism they had.
+func filterWorkerCount(nodeCount int) int {
+	workers := config.FilterParallelism
+	if workers <= 0 {
+		workers = runtime.GOMAXPROCS(0)
+	}
+	return min(workers, nodeCount)
```
mesutoezdil, line 415: this serializes scoring into GOMAXPROCS batches. kube-scheduler blocks on Filter with a timeout. what is the wall time on 1000 nodes?

pkg/scheduler/score.go
```diff
@@ -399,6 +400,48 @@ func (s *Scheduler) recordFilteringFailures(task *corev1.Pod, failureReason map[
...
+		go func() {
+			defer wg.Done()
+			for job := range jobs {
+				fn(job.nodeID, job.node)
+			}
+		}()
+	}
+
+	for nodeID, node := range nodes {
+		jobs <- nodeJob{nodeID: nodeID, node: node}
```
mesutoezdil, line 453: workers 0 with a non empty map blocks here forever. the guard is only in the caller. why not guard in the helper?

cmd/scheduler/main.go
im-Toqeer-506: Right, fixed. Usage string now reads "zero or negative means
runtime.GOMAXPROCS", matching filterWorkerCount and the config.go comment.

pkg/scheduler/score.go
im-Toqeer-506: Two reasons for keeping the manual pool. golang.org/x/sync is "// indirect" in go.mod today, so errgroup would promote it to a direct dependency. And SetLimit(0) has the same block — g.Go sends on a zero-capacity sem channel with nothing draining it — so the guard is needed either way; errgroup would only save the WaitGroup boilerplate, not the deadlock. The pool also starts exactly `workers` goroutines rather than one per node, which is the thing the PR is trying to bound.

pkg/scheduler/score.go
im-Toqeer-506: Agreed, moved it. runNodeWorkers now returns early on an empty map and falls back to GOMAXPROCS for a non-positive count, so it can't block regardless of caller. 

Added TestRunNodeWorkersNonPositiveWorkers. The failure mode is a hang, not a wrong answer, so the test runs the fan-out in a goroutine and fails on a timeout instead. I checked it actually catches the bug: it times out against the pre-guard helper and passes against this one.

mesutoezdil:

pkg/scheduler/score.go
im-Toqeer-506: Measured it rather than guessing. 12 cores, calcScoreWithOptions over the test fixture (8 GPUs/node, one whole-card request per container), best of 3:

nodes | init | p=1      | default (GOMAXPROCS=12) | unbounded (pre-PR)
------|------|----------|-------------------------|-------------------
1000  | 0    | 23.1 ms  | 8.7 ms                  | 10.4 ms
1000  | 3    | 73.9 ms  | 29.4 ms                 | 47.9 ms
5000  | 3    | 499.1 ms | 132.1 ms                | 196.9 ms

So 1000 nodes is ~9 ms with no init containers, ~29 ms with three. Filter's timeout is in seconds, so there's a lot of headroom.

Worth flagging that the bounded version is faster than the unbounded one it replaces, not slower. Each in-flight goroutine holds its own NodeUsage deep copy, so past a point the extra goroutines cost more in GC and scheduler pressure than the added concurrency buys. --filter-parallelism is there if an operator on much wider hardware wants to raise it.

Synthetic fixture and my laptop, so treat the absolute numbers as indicative  the relative ordering held across every run.

PR #2868, opened 2026-08-27
fix(scheduler): use phase-aware accounting during NUMA refit

### LINKED ISSUE #2856: NUMA refit can reject a valid placement by treating init and app memory as concurrent
**What happened**:

NUMA refit can reject a GPU that has enough memory because it treats an init container and an app container as if they run at the same time.

For example:

- GPU-a and GPU-b each have 40 GB of memory.
- A non-sidecar init container reserves 30 GB on GPU-a.
- The app container reserves 20 GB on GPU-b.
- NUMA refit tries to move the init container to GPU-b.

HAMi currently sees 20 GB already reserved on GPU-b and adds the init container's 30 GB request:

```text
Before refit:

GPU-a                         GPU-b
init: 30 GB                   app: 20 GB
    \                            |
     \------ move init ---------->

Current check: 20 + 30 = 50 GB -> rejected
Correct peak: max(30, 20) = 30 GB -> fits
```

The refit fails with `CardInsufficientMemory`, even though init and app containers run in different phases and their actual peak usage is only 30 GB.

The problem happens because the scheduler cache stores collapsed init/app usage, but `RefitNumaAllocation` removes only the raw allocation being moved. The NVIDIA capacity check then treats the new request as fully additional usage.

NUMA refit should use the same phase-aware memory accounting as normal scheduling.

For a non-sidecar init container and an app container, memory usage should be:

```text
max(init-container peak, app-container usage)
```

In this example, the refit should succeed because `max(30 GB, 20 GB)` is 30 GB, which fits on the 40 GB GPU.

This was confirmed with a focused scheduler unit test on current `master`.

1. Create two modeled NVIDIA GPUs with 40,000 MB each.
2. Give the non-sidecar init container a 30,000 MB allocation on GPU-a.
3. Give the app container a 20,000 MB allocation on GPU-b.
4. Store the pod's collapsed usage in `PodManager`.
5. Request a NUMA refit of the init container with only GPU-b allowed.
6. Call `RefitNumaAllocation`.

Current result:

```text
no allowed device fits: 1/1 CardInsufficientMemory
```

Expected result:

The refit succeeds and moves the init-container allocation to GPU-b.

Accelerator requests used by the test:

- Init container: `hami.io/gpu: 1`, `hami.io/gpumem: 30000`
- App container: `hami.io/gpu: 1`, `hami.io/gpumem: 20000`
- NUMA mode: `hami.io/numa-alignment: strict`

PR #2817 identified the same raw-versus-collapsed accounting limitation but left it for follow-up work.

**Environment**:

- HAMi version: `master` at `e40ede755a4a8f12b482bc956a5e0c8bc4528826`
- Kubernetes version: Not tested; scheduler unit reproduction only
- Accelerator vendor and model: NVIDIA; two modeled 40 GB GPUs
- HAMi device backend and requested resource names: NVIDIA, `hami.io/gpu`, `hami.io/gpumem`
- Accelerator driver and runtime version: Not applicable to the unit reproduction
- Container runtime and version: Not applicable to the unit reproduction
- Workload image and tag: Not applicable to the unit reproduction
- Operating system and kernel version: Not applicable to the unit reproduction
- Others: Live GPU validation is still pending

asadjan4611:
**What type of PR is this?**

/kind bug

NUMA refit previously treated a pod's regular init-container memory and application-container memory as concurrent usage. This could reject a valid placement with `CardInsufficientMemory`.

This PR uses HAMi's existing init-container accounting rules while checking allowed NUMA devices. Regular init and application usage now use their phase peak, while sidecars and application containers remain concurrent.

~~~text
GPU-b capacity: 40 GB

Regular init: 30 GB ─┐
                     ├── max(30, 20) = 30 GB ──> placement fits
Application:  20 GB ─┘

Sidecar:      30 GB ─┐
                     ├── 30 + 20 = 50 GB ──> placement rejected
Application:  20 GB ─┘
~~~

The candidate calculation reuses `CollapseInitContainerUsage` and `SteadyStateDeviceUsage`. It preserves the accounting of other pods and updates the cached reservation after a successful refit.

Fixes #2856

Regression tests cover:

- A 30-GB regular init allocation and 20-GB application allocation fitting on a 40-GB GPU.
- Usage belonging to another pod remaining accounted.
- A sidecar starting after a larger regular init container.
- Sidecar and application memory remaining concurrent.
- Successful annotation and cached-accounting updates.

Only the NUMA-refit implementation and its focused scheduler tests are changed.

AI assistance disclosure:

I used OpenAI Codex to review the implementation and for this PR description.

~~~release-note
NUMA refit now uses phase-aware init and application container accounting, preventing valid placements from being rejected by double-counted GPU memory.
~~~

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved NUMA-aware resource recalculation as workloads transition from initialization to steady state.
  - Refined device capacity calculations for init containers, sidecars, reserved devices, and concurrent resource usage.
  - Corrected resource release tracking to restore the appropriate allocated capacity.
  - Device type matching is now more precise, and unsupported registered device types are rejected clearly.

- **Tests**
  - Added coverage for phase-aware NUMA allocation, concurrent sidecars, multiple application containers, and unsupported device types.

### DIFF
--- a/pkg/scheduler/numa_refit_handler.go
+++ b/pkg/scheduler/numa_refit_handler.go
@@ -182,9 +182,11 @@ func (s *Scheduler) RefitNumaAllocation(req device.NumaRefitRequest) device.Numa
 			return s.numaRefitFailureEvent(pod, "allowed device %s is in MIG mode; the NUMA refit does not support MIG", deviceList.Device.ID)
 		}
 	}
-	// The snapshot includes this container's own reservation; release it so
-	// capacity checks do not double count the pod against itself.
-	releaseContainerUsage(nodeUsage, current)
+	// The snapshot contains this pod's phase-collapsed reservation. Prepare
+	// each allowed candidate so adding the moved container produces the same
+	// effective usage as normal scheduling (for example max(init, app), not
+	// init+app, for a non-sidecar init container).
+	preparePhaseAwareRefitUsage(nodeUsage, pod, pi.Devices, allocated, req.DeviceType, req.ContainerIndex, current[0], allowed, pi.InitContainerResourceReleased)
 
 	weights, err := util.GetDeviceScoringWeightsByPod(pod)
 	if err != nil {
@@ -261,10 +263,7 @@ func (s *Scheduler) RefitNumaAllocation(req device.NumaRefitRequest) device.Numa
 		// init-container usage has been released, collapsing again would
 		// re-inflate the reservation back to the init peak, the same hazard
 		// PodManager.AddPod guards against on a re-add.
-		effective := device.CollapseInitContainerUsage(pod, rawDevices)
-		if pi.InitContainerResourceReleased {
-			effective = device.SteadyStateDeviceUsage(pod, rawDevices)
-		}
+		effective := effectivePodDeviceUsage(pod, rawDevices, pi.InitContainerResourceReleased)
 		if _, ok := s.podManager.ReplacePodDevices(key, effective); ok {
 			s.quotaManager.AddUsage(pod, effective)
 		} else {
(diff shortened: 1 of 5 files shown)


pkg/scheduler/numa_refit_handler.go
```diff
@@ -311,17 +310,98 @@ func containerNameAt(pod *corev1.Pod, index int) (string, bool) {
...
+		_, candidateAllowed := allowed[candidate.ID]
+		candidateTypeMatches := strings.Contains(strings.ToLower(candidate.Type), strings.ToLower(deviceType))
+		if candidateAllowed && candidateTypeMatches {
+			withUsage := effectiveUsageOnCandidate(pod, withoutMoved[deviceType], deviceType, containerIndex, moved, candidate.ID, initReleased)
+			preload = device.ContainerDevice{
+				UUID:      candidate.ID,
+				Type:      deviceType,
+				Usedmem:   max(withUsage.Usedmem-moved.Usedmem, 0),
+				Usedcores: max(withUsage.Usedcores-moved.Usedcores, 0),
+				Slots:     max(withUsage.Slots-1, 0),
```
mesutoezdil, line 392: mem and cores subtract the moved row's value, slots subtracts a hardcoded 1. if moved.Slots is 3, Fit adds the extra 2 again. any test with Slots > 1?

pkg/scheduler/numa_refit_handler.go
```diff
@@ -311,17 +310,98 @@ func containerNameAt(pod *corev1.Pod, index int) (string, bool) {
...
+		}
+	}
+	moved.UUID = candidateID
+	candidateRows[containerIndex] = append(candidateRows[containerIndex], moved)
+	effective := effectivePodDeviceUsage(pod, device.PodDevices{deviceType: candidateRows}, initReleased)
+	return aggregateDeviceUsage(effective[deviceType])[candidateID]
+}
+
+// preparePhaseAwareRefitUsage replaces this pod's cached contribution with a
+// candidate-specific preload. Nvidia.Fit then adds the moved container's raw
```
mesutoezdil: the doc says Nvidia.Fit, but refit takes any deviceType and enflame charges drsSlice slices. does the slots math hold there?

pkg/scheduler/numa_refit_handler.go
```diff
@@ -311,17 +310,98 @@ func containerNameAt(pod *corev1.Pod, index int) (string, bool) {
...
+
+	withoutMoved := raw.DeepCopy()
+	withoutMoved[deviceType][containerIndex] = nil
+	withoutUsage := aggregateDeviceUsage(effectivePodDeviceUsage(pod, withoutMoved, initReleased)[deviceType])
+
+	for _, deviceList := range node.Devices.DeviceLists {
+		candidate := deviceList.Device
+		preload, exists := withoutUsage[candidate.ID]
+		_, candidateAllowed := allowed[candidate.ID]
+		candidateTypeMatches := strings.Contains(strings.ToLower(candidate.Type), strings.ToLower(deviceType))
```
mesutoezdil: this is a third copy of the same match, alongside score.go:47 and numa_refit.go:61. worth pulling into one helper?

pkg/scheduler/numa_refit_handler.go
asadjan4611: great concern from your side  @mesutoezdil. Clarified that annotations contain one raw slot, prevented cached Slots from entering this calculation, and added a three-concurrent-slot test

pkg/scheduler/numa_refit_handler.go
asadjan4611: Not applicable currently: NUMA refit only supports NVIDIA. I did not introduce unsafe Enflame behavior; instead, I documented and tested that boundary......

pkg/scheduler/numa_refit_handler.go
asadjan4611: fixed now

archlitchi (approved):
/lgtm

PR #2869, opened 2026-08-27
test(scheduler): verify NUMA refit PATCH timeout releases allocLock

### LINKED ISSUE #2857: Add regression test for NUMA-refit PATCH timeout and allocLock release

Add a deterministic scheduler test for the NUMA-refit annotation PATCH timeout.

The test should block the PATCH before it is applied, start a concurrent `Filter` request, and verify this sequence:

~~~text
NUMA refit: lock -> PATCH waits -> timeout -> restore accounting -> unlock
Filter:                 waits -------------------------------> continues
~~~

The test should use a short configurable timeout instead of waiting for the production two-second timeout.

The test should verify that:

- the blocked PATCH stops at the test timeout;
- the failed refit releases `allocLock`;
- the concurrent `Filter` request continues;
- Pod annotations, device reservations, and quota usage remain unchanged when the PATCH is cancelled before being applied; and
- existing scheduler tests continue to pass.

**Why is this needed**:

`RefitNumaAllocation` holds the scheduler-wide `allocLock` while updating Pod annotations. PR #2817 added a timeout to prevent this operation from blocking scheduling indefinitely.

Existing tests cover an immediate PATCH failure, but they do not verify timeout handling, lock release, or progress of a concurrent `Filter` request. Without this regression test, a future change could allow an API server failure to block scheduler requests.

Likely affected files:

- `pkg/scheduler/numa_refit_handler.go`
- `pkg/scheduler/numa_refit_handler_test.go`

This issue covers a PATCH that is cancelled before the API server applies it.

This change introduces no breaking behavior.

asadjan4611:
**What type of PR is this?**

/kind feature

This PR adds concurrency coverage for the NUMA-refit annotation PATCH timeout.

The scheduler holds `allocLock` while performing a NUMA refit. If the Kubernetes API PATCH blocks, concurrent `Filter` requests must not remain blocked indefinitely.

The test uses a short injectable timeout and verifies that:

- The blocked PATCH returns `context deadline exceeded`.
- The refit releases `allocLock`.
- A waiting `Filter` request continues.
- Pod annotations remain unchanged.
- The original device reservation remains unchanged.
- ResourceQuota usage is restored.

The production timeout remains two seconds.

~~~text
NUMA refit acquires allocLock
             |
             v
      Annotation PATCH blocks
             |
             v
       PATCH timeout expires
             |
             v
Refit restores quota and releases allocLock
             |
             v
      Waiting Filter continues
~~~

Fixes #2857

The change moves creation of the existing timeout context to `RefitNumaAllocation` so the PATCH test seam receives the same context used in production. No scheduling or timeout behavior is changed.

AI assistance disclosure:

I used an  AI Assistance to review the implementation and for the PR description.

~~~release-note
NONE
~~~

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved scheduler behavior when an annotation update times out.
  * Ensured filtering can continue after a timed-out NUMA allocation refit.
  * Preserved device assignment and quota usage when the update fails.

### DIFF
--- a/pkg/scheduler/numa_refit_handler.go
+++ b/pkg/scheduler/numa_refit_handler.go
@@ -40,7 +40,7 @@ import (
 // newer update to the same annotations (for example Allocate consuming a
 // to-allocate entry). Conflicts fail the refit; there is no retry with
 // cached values. Also a test seam.
-var patchPodAnnotations = func(ctx context.Context, pod *corev1.Pod, annotations map[string]string) error {
+var patchPodAnnotations = func(pod *corev1.Pod, annotations map[string]string) error {
 	metadata := map[string]any{"annotations": annotations}
 	if pod.ResourceVersion != "" {
 		metadata["resourceVersion"] = pod.ResourceVersion
@@ -49,6 +49,11 @@ var patchPodAnnotations = func(ctx context.Context, pod *corev1.Pod, annotations
 	if err != nil {
 		return err
 	}
+	// The refit holds the allocation lock that Filter also takes, so this
+	// call must not block scheduling on an unreachable API server. The
+	// device plugin gives up after numaRefitTimeout anyway.
+	ctx, cancel := context.WithTimeout(context.Background(), refitPatchTimeout)
+	defer cancel()
 	_, err = client.GetClient().CoreV1().Pods(pod.Namespace).
 		Patch(ctx, pod.Name, k8stypes.MergePatchType, payload, metav1.PatchOptions{})
 	return err
@@ -60,9 +65,8 @@ const maxAllowedDeviceUUIDs = 512
 // refitPatchTimeout bounds the annotation patch. It is deliberately no longer
 // than the device plugin's own refit budget: a patch that has not landed by
 // then cannot be used, and waiting longer would hold the allocation lock
-// against every concurrent Filter call. It is a variable so the timeout path
-// can be tested without making unit tests wait for the production duration.
-var refitPatchTimeout = 2 * time.Second
+// against every concurrent Filter call.
+const refitPatchTimeout = 2 * time.Second
 
 // RefitNumaAllocation moves one container's device reservation onto a device
 // from the caller-supplied allowed set, re-running the pod's normal
@@ -242,12 +246,7 @@ func (s *Scheduler) RefitNumaAllocation(req device.NumaRefitRequest) device.Numa
 		device.InRequestDevices[req.DeviceType]: pendingValue,
 		device.SupportDevices[req.DeviceType]:   allocatedValue,
 	}
-	// The refit holds the allocation lock that Filter also takes, so this
-	// call must not block scheduling on an unreachable API server. The
-	// device plugin gives up after numaRefitTimeout anyway.
-	ctx, cancel := context.WithTimeout(context.Background(), refitPatchTimeout)
-	defer cancel()
-	if err := patchPodAnnotations(ctx, pod, annotations); err != nil {
+	if err := patchPodAnnotations(pod, annotations); err != nil {
 		return failWithQuotaRestore("cannot patch pod annotations: %v", err)
 	}
 
--- a/pkg/scheduler/numa_refit_handler.go
+++ b/pkg/scheduler/numa_refit_handler.go
@@ -40,7 +40,7 @@ import (
 // newer update to the same annotations (for example Allocate consuming a
 // to-allocate entry). Conflicts fail the refit; there is no retry with
 // cached values. Also a test seam.
-var patchPodAnnotations = func(pod *corev1.Pod, annotations map[string]string) error {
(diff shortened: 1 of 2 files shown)


pkg/scheduler/numa_refit_handler_test.go
```diff
@@ -88,6 +92,85 @@ func stubRefitPatch(t *testing.T, fail error) (map[string]string, *int) {
...
+	filterStarted := make(chan struct{})
+	filterReturned := make(chan struct{})
+	filterDone := make(chan error, 1)
+	filterBlocked := make(chan bool, 1)
+	previousPatch := patchPodAnnotations
+	patchPodAnnotations = func(ctx context.Context, _ *corev1.Pod, _ map[string]string) error {
+		close(patchStarted)
+		<-filterStarted
+		select {
+		case <-filterReturned:
```
mesutoezdil: line 146 closes filterStarted before it calls s.Filter, so line 114 unblocks before Filter runs and this select always takes default. that makes filterBlocked always true, and line 159 would pass even if allocLock were never held. could you double check what it is asserting?

pkg/scheduler/numa_refit_handler_test.go
```diff
@@ -88,6 +92,85 @@ func stubRefitPatch(t *testing.T, fail error) (map[string]string, *int) {
...
+	}
+	assert.Equal(t, <-filterBlocked, true, "Filter completed while the refit held allocLock")
+	select {
+	case err := <-filterDone:
+		assert.NilError(t, err)
+	case <-time.After(time.Second):
+		t.Fatal("Filter did not continue after the refit released allocLock")
+	}
+
+	assert.DeepEqual(t, pod.Annotations, originalAnnotations)
```
mesutoezdil: the stub never writes annotations, so this comparison holds regardless. is there a path here that could change them?

pkg/scheduler/numa_refit_handler.go
```diff
@@ -65,8 +60,9 @@ const maxAllowedDeviceUUIDs = 512
 // refitPatchTimeout bounds the annotation patch. It is deliberately no longer
 // than the device plugin's own refit budget: a patch that has not landed by
 // then cannot be used, and waiting longer would hold the allocation lock
-// against every concurrent Filter call.
-const refitPatchTimeout = 2 * time.Second
+// against every concurrent Filter call. It is a variable so the timeout path
+// can be tested without making unit tests wait for the production duration.
+var refitPatchTimeout = 2 * time.Second
```
mesutoezdil: this turns a production constant into a mutable package var so one test can run faster. would passing the timeout in, or having the stub return DeadlineExceeded itself, keep the seam out of production code?

pkg/scheduler/numa_refit_handler_test.go
asadjan4611: Fixed in d1f860c. The test now directly checks allocLock with TryLock while the PATCH stub is blocked, removes the ineffective filterBlocked assertion, starts Filter before releasing the PATCH, and verifies that Filter completes after the refit returns.

pkg/scheduler/numa_refit_handler_test.go
asadjan4611: Fixed in d1f860c. I removed the annotation equality assertion because the stub cannot mutate API-server state. The test now keeps only meaningful rollback checks for the tracked reservation and quota usage.

pkg/scheduler/numa_refit_handler.go
asadjan4611: Fixed in d1f860c. refitPatchTimeout is a production constant again, and the production PATCH path and signature are unchanged. The test stub now returns context.DeadlineExceeded after a controlled release, as suggested.

pkg/scheduler/numa_refit_handler_test.go
asadjan4611: I removed the ineffective filterBlocked check and now use TryLock to directly verify that allocLock is held while the PATCH is blocked. The test also confirms that Filter completes after the refit releases the lock. Fixed in d1f860c.

pkg/scheduler/numa_refit_handler_test.go
asadjan4611: you are right and i have fixed in https://github.com/Project-HAMi/HAMi/commit/d1f860cad4fe7ace610eecbf794529dc83481325.

pkg/scheduler/numa_refit_handler.go
asadjan4611: i restored refitPatchTimeout as a constant and kept the production PATCH implementation unchanged. The test stub now returns context.DeadlineExceeded directly, so the test remains fast without modifying production configuration.

archlitchi (approved):
/lgtm

PR #2870, opened 2026-08-28
fix(scheduler): score devices against the request on model-typed backends

adity1raut:
**What type of PR is this?**

/kind bug

- ComputeScore filters requests with `container.Type != ds.Device.Type`, but the
device type on the node isn't the vendor name. NVIDIA stores the card model
(`NVIDIA A100-SXM4-40GB`, from nvml `GetName()` at `register.go:188`) and
Cambricon copies the node's `Model` label (`MLU370-X8`), while the request type is
`NVIDIA` or `MLU`. Those never compare equal, so every request is skipped and the
score collapses to how full the card already is, instead of how full it will be
once the request lands. That's the number binpack and spread both rank on.

Same idle 40GB card, same request (1 GPU, 20480MB, 30 cores):

```
Type "NVIDIA A100-SXM4-40GB"  ->  score 0
Type "NVIDIA"                 ->  score 9
```

`getNodeResources` (`score.go:47`) already picks candidates with a case-insensitive
substring match, so selection and scoring were answering the same question two
different ways. I pulled that match into `device.MatchesRequestType` and used it in
`ComputeScore`, `getNodeResources` and `restrictNodeUsage`. It's a superset of `==`,
so nothing that counted before stops counting.

Fixes #

AI disclosure, per CONTRIBUTING.md: I used Claude Code here. It helped me spot the
mismatch and draft the change and tests.

Nodes with identical cards don't move, since the missing term was the same for all
of them. Mixed-capacity nodes do: for a 20GB request against an idle 40GB card and
an 80GB card with 1 slot / 5 cores / 4096MB used, binpack goes from the 80GB card
(0.20 before) to the 40GB card (0.90 after), which leaves the big card free.

Four new tests, all verified to fail with the `!=` put back.

```release-note
Fixed GPU scoring dropping the pending request on backends that register a model-specific device type (NVIDIA, Cambricon), which left binpack and spread ranking cards by current utilisation instead of utilisation after placement. Nodes with identical cards are unaffected.
```

## Summary by CodeRabbit

- **New Features**
  - GPU requests can now match devices by vendor, including devices identified by specific model names.
  - Device usage records now retain vendor information for more accurate scheduling decisions.

- **Bug Fixes**
  - Improved GPU bin-packing on nodes with cards of different capacities, helping place workloads on the most appropriate available card.
  - Prevented requests for different vendors from being treated as interchangeable.

### DIFF
--- a/pkg/device/devices.go
+++ b/pkg/device/devices.go
@@ -754,6 +754,27 @@ func CheckType(annos map[string]string, cardType, useKey, noUseKey string) bool 
 	return true
 }
 
+// MatchesRequestType reports whether a registered device type satisfies a
+// container request type. A backend may register a model-specific type
+// ("NVIDIA A100-SXM4-40GB", "MLU370-X8") while the ContainerDeviceRequest it
+// generates carries the vendor's common word ("NVIDIA", "MLU"), so the two are
+// compared as a case-insensitive substring rather than for equality.
+//
+// An empty request type matches nothing. strings.Contains would match every
+// device instead, which is the one input where this is not a superset of the
+// equality test it replaces. Resourcereqs only records a request once
+// GenerateResourceRequests reports Nums > 0, and every backend sets a non-empty
+// Type in that case, so an empty type does not reach these call sites; the
+// guard is here so scoring a request against every device on the node cannot
+// become reachable by accident.
+func MatchesRequestType(deviceType, requestType string) bool {
+	requestType = strings.TrimSpace(requestType)
+	if requestType == "" {
+		return false
+	}
+	return strings.Contains(strings.ToLower(deviceType), strings.ToLower(requestType))
+}
+
 // PodRequiresDevice returns true if any container (init container or regular container)
 // in the pod requests resources from the specified device generator.
 func PodRequiresDevice(dev Devices, p *corev1.Pod) bool {
--- a/pkg/scheduler/policy/gpu_policy.go
+++ b/pkg/scheduler/policy/gpu_policy.go
@@ -191,16 +191,57 @@ func (ds *DeviceListsScore) DeepCopy() *DeviceListsScore {
 	}
 }
 
+// mostSpecificMatchingType returns the request type in requests that applies to
+// a device of deviceType, and whether any does.
+//
+// A device is served by exactly one backend, so at most one request *type* may
+// be scored against it. Comparing types for equality gave that for free. A
+// substring match does not, because registered types nest: the chart ships both
+// "Ascend910B4" and "Ascend910B4-1" as Ascend common words, and the longer type
+// contains the shorter one, so a pod requesting both would otherwise have such a
+// device scored twice. The longest matching type is the most specific one; ties
+// break on the string itself so the choice is deterministic.
+//
+// This selects the type only. ComputeScore still sums every request carrying it,
+// which is the existing behaviour for a pod whose containers request the same
+// device type more than once.
+func mostSpecificMatchingType(deviceType string, requests device.ContainerDeviceRequests) (string, bool) {
+	best := ""
+	found := false
+	for _, container := range requests {
+		if !device.MatchesRequestType(deviceType, container.Type) {
+			continue
+		}
+		if !found || len(container.Type) > len(best) ||
+			(len(container.Type) == len(best) && container.Type < best) {
+			best = container.Type
+			found = true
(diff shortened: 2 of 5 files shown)


pkg/scheduler/policy/gpu_policy.go
```diff
@@ -197,10 +197,15 @@ func (ds *DeviceListsScore) ComputeScore(requests device.ContainerDeviceRequests
...
+	// A node registers model-specific device types ("NVIDIA A100-SXM4-40GB",
+	// "MLU370-X8") while a request carries the vendor common word ("NVIDIA",
+	// "MLU"), so match the way getNodeResources selects candidate devices.
+	// Comparing for equality dropped the request from the score entirely on
+	// those backends, leaving devices ranked by current usage instead of usage
+	// after placement.
 	for _, container := range requests {
-
-		if container.Type != ds.Device.Type {
+		if !device.MatchesRequestType(ds.Device.Type, container.Type) {
```
mesutoezdil: checking the scope here: if this is right, binpack and spread were never scoring NVIDIA, which is the main backend. the existing ComputeScore tests all use type1/type1 or NVIDIA/NVIDIA, so none of them covers the model typed case. is that the intended reading?

pkg/device/devices.go
```diff
@@ -754,6 +754,20 @@ func CheckType(annos map[string]string, cardType, useKey, noUseKey string) bool
...
 
+// MatchesRequestType reports whether a registered device type satisfies a
+// container request type. A backend may register a model-specific type
+// ("NVIDIA A100-SXM4-40GB", "MLU370-X8") while the ContainerDeviceRequest it
+// generates carries the vendor's common word ("NVIDIA", "MLU"), so the two are
+// compared as a case-insensitive substring rather than for equality. An empty
+// request type matches nothing.
+func MatchesRequestType(deviceType, requestType string) bool {
+	requestType = strings.TrimSpace(requestType)
+	if requestType == "" {
```
mesutoezdil, line 772: the two Contains call sites had no empty check, and Contains(x, "") is true, so an empty type used to select every device and now selects none. the pr says nothing that counted before stops counting. could you reconcile those two?

pkg/device/devices.go
```diff
@@ -754,6 +754,20 @@ func CheckType(annos map[string]string, cardType, useKey, noUseKey string) bool
...
+		return false
+	}
+	return strings.Contains(strings.ToLower(deviceType), strings.ToLower(requestType))
```
mesutoezdil, line 775: equality guaranteed at most one request matched a device, and substring does not. ComputeScore sums over every match, so two request keys hitting one device would both add. is something upstream preventing that?

Shouren:
@adity1raut There is a `DeviceVendor` inside `DeviceInfo` struct as the following code shows:
https://github.com/Project-HAMi/HAMi/blob/d96e7e8ccaf85b15cb47709b319509b0a9a53eda/pkg/device/devices.go#L127-L144
So I think we can add it to DeviceUsage when calling [buildNodeUsage](https://github.com/Project-HAMi/HAMi/blob/d96e7e8ccaf85b15cb47709b319509b0a9a53eda/pkg/scheduler/scheduler.go#L681). Later we can use it for comparing with `container.Type` in `ComputeScore`.
How do you think?

mesutoezdil:
closing per the issue/pr lifecycle policy: no author response in two weeks to the outstanding review comments (last review 2026-08-30).

adity1raut:
> closing per the issue/pr lifecycle policy: no author response in two weeks to the outstanding review comments (last review 2026-08-30).

Sorry, but I can get the answer to each comment. Could you please reopen this PR?

Shouren:
> > closing per the issue/pr lifecycle policy: no author response in two weeks to the outstanding review comments (last review 2026-08-30).
> 
> Sorry, but I can get the answer to each comment. Could you please reopen this PR?

@adity1raut Let me reopen it

mesutoezdil:
remove the claude code from the history pls

adity1raut:
> remove the claude code from the history pls

Yeah!

Shouren (commented):
@adity1raut Please move the test cases in `pkg/scheduler/policy/gpu_policy_model_type_test.go` and `pkg/scheduler/score_model_type_test.go` into the existing test files

Shouren (approved):
/lgtm

PR #2877, opened 2026-08-28
fix(scheduler): keep multi-slot device usage across annotation round trip

### LINKED ISSUE #2876: Enflame DRS slice usage is undercounted after a scheduler restart
**What happened**:

`ContainerDevice.Slots` records how many concurrent tasks a collapsed allocation occupies on one device, but `EncodeContainerDevices` never wrote the field into the pod annotation. `DecodeContainerDevices` therefore always returns `Slots == 0`, and `getNodesUsage` clamps that to 1 when it rebuilds the node cache from pod annotations.

Enflame DRS is the concrete case. `Fit()` selects a profile of `profile.Size` slices and `AddResourceUsage` charges that many slices live, so a 6 slice card filled by two 3 slice profiles reports `Used == 6` while the scheduler is running. After a restart the same card rebuilds as `Used == 2`. The slice guard in `Fit()` is `dev.Count <= dev.Used`, so the scheduler then admits DRS instances for which the card has no slices left.

Per card slice usage rebuilt from pod annotations should equal the live usage, and a card whose slices are fully consumed should be rejected regardless of whether the scheduler has restarted.

1. Register an Enflame node with a DRS capacity of 6 slices per card.
2. Schedule two pods that each land on a `3g.20gb` profile of the same card. The live `DeviceUsage.Used` is 6 and further DRS pods are correctly rejected.
3. Restart the scheduler so the node cache is rebuilt from the pod annotations. `Used` comes back as 2 and a third 3 slice pod is admitted onto the same card.

The round trip alone is enough to show the loss: encode a `ContainerDevice` with `Slots: 3`, decode it, and the decoded entry has `Slots == 0`.

The same loss affects any future backend that relies on `Slots`, not only Enflame DRS, because the field simply never reaches the annotation.

**Environment**:

- HAMi version: master
- Affected component: scheduler extender, device allocation annotation encoding
- Others: no hardware needed to observe the encode/decode loss

Eshiv-Pandey:
**What type of PR is this?**

/kind bug

`ContainerDevice.Slots` is the number of concurrent tasks a collapsed entry occupies on a device, but `EncodeContainerDevices` never wrote it. The pod annotation lost the value, `DecodeContainerDevices` always returned `Slots == 0`, and `getNodesUsage` clamps that to 1 when it rebuilds the node cache. Any allocation worth more than one slot was undercounted after a scheduler restart.

Enflame DRS is the concrete case. `Fit()` picks a profile of `profile.Size` slices and `AddResourceUsage` charges that many live, so a card with 6 slices filled by two 3 slice profiles reported `Used 6` while running and `Used 2` after the cache rebuild. The slice guard `dev.Count <= dev.Used` then admitted DRS instances the card had no slices left for.

Changes:

- Encode `Slots` as an optional 5th field, written only when it exceeds 1, and decode it back when present with validation for negative and non numeric values.
- Set `Slots` from `profile.Size` in the Enflame DRS `Fit()` path, since `CustomInfo["drsSlice"]` does not survive the annotation.
- Sum the recorded slot count in `CollapseInitContainerUsage` and `SteadyStateDeviceUsage` instead of counting one per container occurrence.

Fixes #2876

The annotation change is compatible in both directions. Single slot allocations keep the historical 4 field encoding byte for byte, and the decoder's pre existing guard is `len(fields) < 4`, so an older parser reads a 5 field entry without error and a new parser reads old annotations unchanged.

New unit tests: `pkg/device/enflame/drs_reconstruction_test.go` asserts that the slice count survives the round trip and that a rebuilt card still rejects slice oversubscription; `pkg/device/devices_test.go` covers the optional field, invalid and negative values, and extends the encode/decode fuzz target with a slots argument; `pkg/device/initContainer_test.go` covers collapsing two multi slot app containers.

`go vet` and `go test -short -count=1` pass for `pkg/device/...` except `pkg/device/nvidia`, and `pkg/scheduler/...` could not be built on my machine because go-nvml requires cgo and my local toolchain cannot build 64 bit cgo objects. Those packages are untouched by this PR, so I am relying on CI for `make verify` and the full `make test`.

This PR was prepared with AI assistance.

The device allocation annotation gains an optional 5th field carrying the slot count for multi slot allocations. It is only written when the count exceeds 1, so existing annotations are unchanged and both old and new parsers read either form.

## Summary by CodeRabbit

- **New Features**
  - Multi-slot device allocations are preserved when assignments are encoded and restored.
  - Device usage accounting reflects each allocation’s declared slot count.
  - Enflame DRS slice usage and allocation state are accurately reconstructed after restarts.

- **Bug Fixes**
  - Prevented oversubscription during multi-slot allocations.
  - Maintained compatibility with existing annotations without slot information.
  - Invalid or negative slot values are safely ignored during restoration.

### DIFF
--- a/pkg/device/devices.go
+++ b/pkg/device/devices.go
@@ -543,17 +543,14 @@ func DecodeContainerDevices(str string) (ContainerDevices, error) {
 			Usedmem:   int32(devmem),
 			Usedcores: int32(devcores),
 		}
-		// The 5th field is optional and only written for multi-slot entries,
-		// so annotations produced by older versions stay readable.
+		// The 5th field is optional. An unusable value is ignored rather than
+		// rejected: dropping the whole annotation undercounts far more than one slot.
 		if len(tmpstr) >= 5 {
-			devslots, err := strconv.ParseInt(tmpstr[4], 10, 32)
-			if err != nil {
-				return nil, fmt.Errorf("invalid slots field: %w", err)
+			if devslots, err := strconv.ParseInt(tmpstr[4], 10, 32); err == nil && devslots >= 0 {
+				tmpdev.Slots = int32(devslots)
+			} else {
+				klog.V(5).Infof("ignoring unusable slots field %q in segment %q", tmpstr[4], val)
 			}
-			if devslots < 0 {
-				return nil, fmt.Errorf("slots field must not be negative: %d", devslots)
-			}
-			tmpdev.Slots = int32(devslots)
 		}
 		contdev = append(contdev, tmpdev)
 	}
--- a/pkg/device/devices.go
+++ b/pkg/device/devices.go
@@ -451,6 +451,12 @@ func UnMarshalNodeDevices(str string) ([]*DeviceInfo, error) {
 func EncodeContainerDevices(cd ContainerDevices) string {
 	var builder strings.Builder
 	for _, val := range cd {
+		// Slots is appended as an optional 5th field only when it exceeds 1, so
+		// single-slot allocations keep the historical 4-field encoding.
+		if val.Slots > 1 {
+			fmt.Fprintf(&builder, "%s,%s,%d,%d,%d%s", val.UUID, val.Type, val.Usedmem, val.Usedcores, val.Slots, OneContainerMultiDeviceSplitSymbol)
+			continue
+		}
 		fmt.Fprintf(&builder, "%s,%s,%d,%d%s", val.UUID, val.Type, val.Usedmem, val.Usedcores, OneContainerMultiDeviceSplitSymbol)
 	}
 	tmp := builder.String()
@@ -536,6 +542,18 @@ func DecodeContainerDevices(str string) (ContainerDevices, error) {
 			Type:      tmpstr[1],
 			Usedmem:   int32(devmem),
 			Usedcores: int32(devcores),
+		}
+		// The 5th field is optional and only written for multi-slot entries,
+		// so annotations produced by older versions stay readable.
+		if len(tmpstr) >= 5 {
+			devslots, err := strconv.ParseInt(tmpstr[4], 10, 32)
+			if err != nil {
+				return nil, fmt.Errorf("invalid slots field: %w", err)
+			}
+			if devslots < 0 {
+				return nil, fmt.Errorf("slots field must not be negative: %d", devslots)
+			}
+			tmpdev.Slots = int32(devslots)
 		}
 		contdev = append(contdev, tmpdev)
 	}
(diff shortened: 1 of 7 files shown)


pkg/device/devices.go
```diff
@@ -537,6 +543,18 @@ func DecodeContainerDevices(str string) (ContainerDevices, error) {
 			Usedmem:   int32(devmem),
 			Usedcores: int32(devcores),
 		}
+		// The 5th field is optional and only written for multi-slot entries,
+		// so annotations produced by older versions stay readable.
+		if len(tmpstr) >= 5 {
```
mesutoezdil, line 548: the "extra optional fields" test case is gone, and a non numeric 5th field that used to be ignored now errors. the pr says the change is compatible both ways. could you reconcile those?

pkg/device/enflame/device.go
```diff
@@ -454,6 +460,9 @@ func (enf *EnflameDevices) Fit(devices []*device.DeviceUsage, request device.Con
 				Type:      k.Type,
 				Usedmem:   profileMemoryMiB,
 				Usedcores: profileCorePercent,
+				// A DRS profile consumes profile.Size slices; recording it here keeps
+				// the count across the annotation round trip that drops CustomInfo.
+				Slots: profileSlices,
```
mesutoezdil, line 474: AddResourceUsage at line 372 still reads CustomInfo drsSlice. the same number now lives in two fields. why not read Slots there?

pkg/device/enflame/device.go
```diff
@@ -425,16 +429,18 @@ func (enf *EnflameDevices) Fit(devices []*device.DeviceUsage, request device.Con
...
-			continue
-		}
 		if isMutex && dev.Used > 0 {
 			reason[common.ExclusiveDeviceAllocateConflict]++
 			klog.V(5).InfoS(common.ExclusiveDeviceAllocateConflict, "pod", klog.KObj(pod), "device", dev.ID, "device index", i, "used", dev.Used)
 			continue
 		}
+		// The whole profile has to fit: a 3 slice profile needs 3 free slices,
+		// not just one.
+		if dev.Count-dev.Used < profileSlices {
```
mesutoezdil, line 443: this also moved below the mutex check, so on a full card the reported reason changes. was that part of the fix?

Eshiv-Pandey: Unintentional, reverted

pkg/device/enflame/device.go
Eshiv-Pandey: it now reads Slots first ..CustomInfo only as fallback.

pkg/device/devices.go
Eshiv-Pandey: my bad. reverted to ignoring it, case restored and decoder lenient again

archlitchi (approved):
/lgtm

PR #2881, opened 2026-08-29
fix(scheduler): keep bare reasons in FilteringFailed events

### LINKED ISSUE #2880: scheduler: ParseReason drops NodeInsufficientDevice from FilteringFailed events
**What happened**:

`common.GenReason` writes failure reasons as `"<count>/<cards> <Reason>"`, and `common.ParseReason` reads them back with `fmt.Sscanf(reason, "%d/%d %s", ...)`, skipping any entry that fails to parse:

https://github.com/Project-HAMi/HAMi/blob/ebcd8ae000d0ded373cad0ebfabb8289f2c5810a/pkg/device/common/common.go#L55-L70

Four of the five failure returns in `fitInDevices` do not use that format. They return a bare string — [score.go:95](https://github.com/Project-HAMi/HAMi/blob/ebcd8ae000d0ded373cad0ebfabb8289f2c5810a/pkg/scheduler/score.go#L95), [:102](https://github.com/Project-HAMi/HAMi/blob/ebcd8ae000d0ded373cad0ebfabb8289f2c5810a/pkg/scheduler/score.go#L102), [:122](https://github.com/Project-HAMi/HAMi/blob/ebcd8ae000d0ded373cad0ebfabb8289f2c5810a/pkg/scheduler/score.go#L122), [:128](https://github.com/Project-HAMi/HAMi/blob/ebcd8ae000d0ded373cad0ebfabb8289f2c5810a/pkg/scheduler/score.go#L128). Only line 107, which forwards a backend's `Fit()` reason, is in the parseable form.

Both kinds of reason enter the same pipeline in `calcScoreWithOptions`, but only one survives it:

```mermaid
flowchart TD
    R["fitInDevices returns a failure reason"]
    R --> Q{"Matches 'N/M Reason' ?"}

    Q -->|"YES — built by GenReason<br/>'2/4 CardInsufficientMemory'"| A1["ParseReason returns<br/>map[CardInsufficientMemory:2]"]
    Q -->|"NO — bare string at score.go 95, 102, 122, 128<br/>'NodeInsufficientDevice'"| B1["Sscanf fails, loop hits 'continue'<br/>ParseReason returns map[]"]

    A1 --> A2["failureReason[reason] records the node"]
    B1 --> B2["range over empty map<br/>nothing is recorded"]

    A2 --> A3["recordFilteringFailures emits<br/>FilteringFailed naming the reason ✅"]
    B2 --> B3["recordFilteringFailures has nothing<br/>to emit — reason is lost ❌"]
```

The practical effect is on `NodeInsufficientDevice` (score.go:102), which fires whenever a pod requests more devices of a type than a node has — one of the most common reasons a GPU pod fails to schedule, and it takes the right-hand path.

When every candidate node is rejected for too few devices, the pod should carry a `FilteringFailed` event naming `NodeInsufficientDevice` and the affected nodes, as `CardInsufficientMemory` does today.

No cluster needed. In `pkg/device/common/`:

```go
for _, s := range []string{
	NodeInsufficientDevice,
	"Device type not found",
	GenReason(map[string]int{CardInsufficientMemory: 2}, 4),
} {
	t.Logf("%-30q -> %v", s, ParseReason(s))
}
```

Output on `ebcd8ae`:

```
"NodeInsufficientDevice"       -> map[]
"Device type not found"        -> map[]
"2/4 CardInsufficientMemory"   -> map[CardInsufficientMemory:2]
```

On a cluster: request `nvidia.com/gpu: 8` where every node has 4, and compare `kubectl describe pod` against the same pod failing for insufficient device memory. Only the memory case carries a reason.

Scope — the pod is not left without any event. `Filter` still records the generic `no available node, %d nodes do not meet`, and the raw string still reaches kube-scheduler via `ExtenderFilterResult.FailedNodes`. What is lost is the per-reason breakdown `recordFilteringFailures` exists to produce. Backend `Fit()` reasons are unaffected.

Smallest fix: make `ParseReason` total — on `Sscanf` failure, record the whole string as one reason with count 1 instead of `continue`. Four lines, covers all four sites and any added later, cannot change well-formed input. Optionally also return `GenReason(map[string]int{NodeInsufficientDevice: len(typeDevices)}, int(k.Nums))` at score.go:102 so the event carries real counts.

Caveat: that promotes the raw errors at score.go:95/122/128 to reason keys, and they embed a device UUID and error text — they should be truncated to a bounded prefix to avoid unbounded event cardinality.

Happy to send a PR, with a round-trip test over the `common` reason vocabulary.

On the diagnostic information the template asks for: this is a defect in the extender's reason-aggregation code path, not a device or runtime failure, so pod manifests, node annotations, `nvidia-smi` output, runtime configuration, device-plugin logs, kubelet logs and `dmesg` are not applicable. The reproduction above is self-contained. The relevant scheduler behaviour is the absence of a per-reason `FilteringFailed` event, shown by `kubectl describe pod` in the cluster reproduction.

**Environment**:
- HAMi version: master @ `ebcd8ae000d0ded373cad0ebfabb8289f2c5810a`
- Kubernetes version: n/a — reproduced from source and a unit test
- Accelerator vendor and model: vendor-independent; `NodeInsufficientDevice` is raised in `fitInDevices` before any backend runs
- HAMi device backend and requested resource names (for example, `nvidia.com/gpu`): any backend; most visible with `nvidia.com/gpu`
- Accelerator driver and runtime version: n/a
- Container runtime and version: n/a
- Workload image and tag (redact private image names and state when redacted): n/a
- Operating system and kernel version (for example, the output of `uname -a`): n/a; reproduction run with go1.26.6 linux/amd64
- Others: scoped to the scheduler extender, so it is unit-testable without GPU hardware

im-Toqeer-506:
**What type of PR is this?**

/kind bug

`common.GenReason` writes failure reasons as `"<count>/<cards> <Reason>"`, and `common.ParseReason` reads them back with `fmt.Sscanf`, skipping any entry that fails to parse. One reason never goes through `GenReason`: `fitInDevices` returns a bare `common.NodeInsufficientDevice` at `score.go:102`, raised before any device backend runs. `Sscanf` fails on it, the loop hits `continue`, and `ParseReason` returns an empty map.

Its only caller ranges over that map's keys:

```go
for reasonType := range common.ParseReason(result.reason) {
    failureReason[reasonType] = append(failureReason[reasonType], nodeID)
}
```

An empty map means zero iterations, so `recordFilteringFailures` has nothing to emit and the pod carries no `FilteringFailed` event naming why nodes were rejected. This path fires whenever a pod requests more devices of a type than a node has, which is one of the more common reasons a GPU pod fails to schedule.

This PR makes `ParseReason` recognise the bare form, using the invariant that separates the two: every reason constant is a single word, and `GenReason` always writes a space. A reason with no space is a bare constant; everything else falls through to the existing loop unchanged, so well-formed input is unaffected.

Before / after:

```
"NodeInsufficientDevice"      map[]                         ->  map[NodeInsufficientDevice:1]
"Device type not found"       map[]                         ->  map[]
"2/4 CardInsufficientMemory"  map[CardInsufficientMemory:2] ->  map[CardInsufficientMemory:2]
```

Fixes #2880

```release-note
Fixed a bug where a pod rejected because no node had enough devices carried no FilteringFailed event. The event now names NodeInsufficientDevice and the affected nodes.
```

---

**AI assistance disclosure**

I am using an AI assistant to verify the implementation and review the test cases to ensure the changes are properly tested and meet the required criteria.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected handling of device insufficiency failures when reported as a standalone reason.
  * Ensured these failures are counted and surfaced in scheduler filtering results and failure events.
  * Preserved support for detailed reason lists and improved handling of empty or free-form error messages.

* **Tests**
  * Added coverage for standalone reason tokens, generated reason lists, error messages, and scheduler failure reporting.

### DIFF
--- a/pkg/device/common/common.go
+++ b/pkg/device/common/common.go
@@ -52,6 +52,11 @@ func GenReason(reasons map[string]int, cards int) string {
 	return strings.Join(reason, ", ")
 }
 
+// ParseReason reads a failure reason back as a count per reason type. It reads
+// only the "<count>/<cards> <Reason>" list GenReason writes, so callers that
+// report a rejection must go through GenReason to have it counted. Free-form
+// internal error text reaches the same field and stays unparsed, keeping the
+// device UUIDs and node names it embeds out of event reason keys.
 func ParseReason(reason string) map[string]int {
 	reasons := strings.Split(reason, ", ")
 
--- a/pkg/device/common/common_test.go
+++ b/pkg/device/common/common_test.go
@@ -38,6 +38,28 @@ func TestParseReason(t *testing.T) {
 				"CardNotHealth":          3,
 			},
 		},
+		{
+			name:   "node-level rejection reported before any device backend runs",
+			reason: GenReason(map[string]int{NodeInsufficientDevice: 1}, 4),
+
+			expectedReasonMap: map[string]int{
+				NodeInsufficientDevice: 1,
+			},
+		},
+		{
+			// fitInDevices reports internal failures as plain error text on the same
+			// field. Parsing it would turn the UUID it carries into an event reason.
+			name:   "free-form internal error is not a reason type",
+			reason: "AddResourceUsage failed for device GPU-0fc3eda5: device is full",
+
+			expectedReasonMap: map[string]int{},
+		},
+		{
+			name:   "empty reason",
+			reason: "",
+
+			expectedReasonMap: map[string]int{},
+		},
 	} {
 		t.Run(ts.name, func(t *testing.T) {
 			result := ParseReason(ts.reason)
--- a/pkg/scheduler/mig_allocation_test.go
+++ b/pkg/scheduler/mig_allocation_test.go
@@ -259,7 +259,7 @@ func TestFitInDevicesRetainsPhysicalCountCheckForNonMigGPU(t *testing.T) {
 	allocated := make(device.PodDevices)
 
 	fit, reason := fitInDevices(&node, requests, pod, node.NodeInfo, &allocated, util.DefaultDeviceScoringWeights())
-	if fit || reason != common.NodeInsufficientDevice {
+	if fit || common.ParseReason(reason)[common.NodeInsufficientDevice] != 1 {
 		t.Fatalf("fitInDevices() = %v, reason = %q, devices = %+v; want physical-device-count rejection", fit, reason, allocated)
 	}
 }
--- a/pkg/scheduler/score.go
+++ b/pkg/scheduler/score.go
@@ -99,7 +99,10 @@ func fitInDevices(node *NodeUsage, requests device.ContainerDeviceRequests, pod 
 		if int(k.Nums) > len(typeDevices) && !isMIGRequest(k, typeDevices, pod) {
 			klog.V(5).InfoS(common.NodeInsufficientDevice, "pod", klog.KObj(pod),
 				"request devices nums", k.Nums, "node device nums (type)", len(typeDevices), "type", k.Type)
-			return false, common.NodeInsufficientDevice
+			// Report it in the same form every device backend uses. common.ParseReason
+			// reads only "<count>/<total> <Reason>", so the bare constant this used to
+			// return was dropped and the pod got no event naming why the node was rejected.
+			return false, common.GenReason(map[string]int{common.NodeInsufficientDevice: len(typeDevices)}, int(k.Nums))
 		}
 
 		fit, tmpDevs, reason := devPlugin.Fit(typeDevices, k, pod, nodeInfo, devinput)
--- a/pkg/scheduler/score_test.go
+++ b/pkg/scheduler/score_test.go
@@ -25,6 +25,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/tools/record"
 	"k8s.io/klog/v2"
 
 	"github.com/Project-HAMi/HAMi/pkg/device"
@@ -3986,7 +3987,8 @@ func Test_fitInDevices(t *testing.T) {
 				devinput: &device.PodDevices{},
 			},
 			want1: false,
-			want2: "NodeInsufficientDevice",
+			// One of the two requested GPUs is present on the node.
+			want2: common.GenReason(map[string]int{common.NodeInsufficientDevice: 1}, 2),
 		},
 		{
 			name: "device type the different from request type",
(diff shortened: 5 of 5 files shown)


pkg/device/common/common.go
```diff
@@ -52,7 +52,19 @@ func GenReason(reasons map[string]int, cards int) string {
...
+// ParseReason reads a failure reason back as a count per reason type. It accepts
+// both forms a caller can produce: the "<count>/<cards> <Reason>" list GenReason
+// writes, and a bare reason constant, which is how fitInDevices reports
+// NodeInsufficientDevice before any device backend runs. Free-form internal error
+// text is not a reason type and stays unparsed, so the device UUIDs and node names
+// it embeds never become event reason keys.
 func ParseReason(reason string) map[string]int {
+	// Reason constants are single words and GenReason always writes a space, so a
+	// reason with no space is a bare constant rather than a generated list.
+	if reason != "" && !strings.Contains(reason, " ") {
```
mesutoezdil: score.go:102 looks like the only bare producer in the tree. would fixing that one line be simpler than teaching the parser a second format?

pkg/device/common/common.go
```diff
@@ -52,7 +52,19 @@ func GenReason(reasons map[string]int, cards int) string {
 	return strings.Join(reason, ", ")
 }
 
+// ParseReason reads a failure reason back as a count per reason type. It accepts
+// both forms a caller can produce: the "<count>/<cards> <Reason>" list GenReason
+// writes, and a bare reason constant, which is how fitInDevices reports
+// NodeInsufficientDevice before any device backend runs. Free-form internal error
+// text is not a reason type and stays unparsed, so the device UUIDs and node names
+// it embeds never become event reason keys.
```
mesutoezdil: only multi word text is filtered out. a one word error string would become a reason key. how is that prevented?

pkg/device/common/common_test.go
```diff
@@ -48,3 +75,48 @@ func TestParseReason(t *testing.T) {
 		})
 	}
 }
+
+// TestParseReasonRoundTripsVocabulary checks every reason this package defines
+// against both forms a caller can produce: GenReason output, and the bare token
+// fitInDevices returns when it rejects a node before a device backend runs.
+func TestParseReasonRoundTripsVocabulary(t *testing.T) {
+	for _, reason := range []string{
```
mesutoezdil: the whole vocabulary is listed right here. why does ParseReason use a space check instead of this set?

im-Toqeer-506: Fixed the producer in `score.go` to use `GenReason` matching all device backends.
Removed the parser change entirely, so the space check and vocabulary issue no longer apply.

archlitchi (approved):
/lgtm

Issue #2080, opened 2026-07-16
NUMA Phase 2: reconcile kubelet-selected GPU with HAMi accounting


saiyam1814:
## TL;DR for maintainers

Phase 1 (#2065) works: on real hardware, kubelet used the new topology information and selected a vGPU replica on the same NUMA node as the pod's CPUs. The remaining problem is that the container can use kubelet's GPU B while HAMi's annotation and scheduler accounting still record the scheduler's original GPU A.

**Proposed Phase 2:** behind an opt-in flag and outside MIG mode, treat kubelet's selected physical GPU as final, verify that GPU has enough HAMi memory/core capacity, preserve the requested limits, and update both HAMi allocation annotations together before returning success.

### Maintainers only need to review these four decisions

1. **Final authority:** When NUMA reconciliation is enabled, should kubelet's selected physical GPU UUID override the scheduler's first GPU choice?
2. **Accounting update:** May the device plugin atomically rewrite `hami.io/vgpu-devices-allocated` and consume `hami.io/vgpu-devices-to-allocate` so runtime and scheduler accounting agree?
3. **Capacity and locking:** Is the existing node allocation lock sufficient to protect the target GPU memory/core capacity check and annotation update from concurrent allocations?
4. **No-capacity behavior:** If kubelet's NUMA-local GPU lacks HAMi memory/core capacity, should allocation fail clearly, or should HAMi implement a retry/fallback mechanism?

If these four decisions are agreed, implementation can start using the existing `alignContainerDevicesWithAllocatedIDs` path. The remaining sections contain the evidence, detailed workflow, and test plan.

## What should be added?

Add Phase 2 of NUMA-aware vGPU allocation after #2065.

Phase 1 lets kubelet see the NUMA node of each HAMi vGPU replica. Phase 2 should make HAMi's runtime allocation and scheduler accounting follow the replica that kubelet actually selects.

/kind feature
/kind design

## Phase 1 in simple English

A large GPU server can have several NUMA nodes. CPUs communicate faster with GPUs on the same NUMA node.

Before #2065, HAMi told kubelet about its vGPU replicas, but did not tell kubelet where their physical GPUs were connected. Kubelet could therefore give a pod CPUs from one NUMA node and a GPU from another.

Phase 1 added the opt-in `enableNumaTopology` setting. When enabled:

1. HAMi reads each physical GPU's NUMA node.
2. Each vGPU replica reports the same NUMA node as its physical GPU.
[...]

PR #2882, opened 2026-08-29
fix(scheduler): require caller authentication on /refit 

### LINKED ISSUE #2878: /refit scheduler endpoint has no caller authentication — any network client can reassign another pod's GPU device
**What happened**:

The scheduler's HTTP server (cmd/scheduler/main.go) registers /filter,
/bind, /refit, and /webhook on one http.Server. When TLS is on (the
chart's own default: --http_bind=0.0.0.0:443 with cert_file/key_file),
the tls.Config only sets GetCertificate — ClientAuth is never set, so it
stays tls.NoClientCert. TLS authenticates the server to the caller, never
the other way around. There's no token or header check anywhere in
pkg/scheduler or in NumaRefitRequest (pkg/device/numa_refit.go) either.

RefitNumaAllocation (pkg/scheduler/numa_refit_handler.go) validates a lot
about the request — pod exists, container index in range, node matches,
allocation still pending, quota rechecked — but never checks that the
caller is actually the device plugin running on req.NodeName. The
comment above it assumes "the device plugin calls it," but nothing
enforces that.

This is reachable in practice, not just on paper: the chart defaults the
scheduler Service to type: NodePort (schedulerPort 31998) with no
NetworkPolicy shipped, so the port is open to any pod in the cluster by
default and possibly to off-cluster hosts on bare-metal setups. Everything
a /refit request needs — podUID, podNamespace, podName, nodeName — is
ordinary info any namespace-scoped user or pod's default service account
can already read from the API, not something unique to the real device
plugin.

**Impact**:

Anyone who can reach the port can POST /refit naming a pod they don't
own and, while that pod's allocation is still pending, redirect its GPU
device reservation to a device of their choosing — no RBAC on that pod
required.

**How to reproduce it**:

Read-through, not a live exploit against a running cluster:
1. Deploy the chart with its own defaults (NodePort service, TLS on via
   --cert_file/--key_file, no client auth, no NetworkPolicy).
2. From any pod in the cluster (or, on bare-metal, any host that can
   reach a node IP on 31998), list a target pod's podUID/podNamespace/
   podName/nodeName via normal pod-list RBAC.
3. POST that data plus a chosen allowedDeviceUUIDs set to /refit before
   the target pod's device plugin Allocate() call consumes its pending
   allocation.

**Anything else we need to know**:

This is new code — numa_refit_handler.go / numa_refit.go are 2026-dated,
added in PR #2731 for issue #2080 — and I couldn't find any existing
issue/PR discussing the missing auth. (/filter and /bind sharing the same
unauthenticated server is the standard kube-scheduler extender pattern
and predates this, so I'm not flagging that part as new — just that
/refit adds a mutating, single-pod-targeted RPC on top of it.)

**Suggested fix**:

- Require mutual TLS (ClientAuth: tls.RequireAndVerifyClientCert) and
  check the client cert's CN/SAN matches req.NodeName, or issue the
  device plugin a per-node token the scheduler validates against
  req.NodeName.
- Default the chart's scheduler Service to ClusterIP instead of NodePort,
  and ship a NetworkPolicy restricting the port to the device-plugin
  DaemonSet and kube-scheduler.

**Environment**:
- HAMi version: main branch, numa_refit_handler.go/numa_refit.go (2026),
  PR #2731
- Kubernetes version: n/a — applies to any version, not a K8s bug
- Component: hami-scheduler

AyushSrivastava1818:
Fixes #2878.

## Background

#2878 reported that the scheduler's `/refit` endpoint (added in #2731 for
#2080) accepts requests with no caller authentication: TLS is
server-only (`ClientAuth` is never set, so it stays `tls.NoClientCert`),
and `NumaRefitRequest` has no token/signature field. Combined with the
chart's default `NodePort` Service and no bundled `NetworkPolicy`, any
network caller that can read a target pod's `podUID`/`podNamespace`/
`podName`/`nodeName` (ordinary info visible via normal pod-list RBAC)
could POST `/refit` and redirect that pod's pending GPU device
reservation to a device of its choosing, without holding any RBAC on the
pod itself.

## Fix

**Caller authentication (`numa_refit_auth.go`, `numa_refit_handler.go`,
`routes/route.go`)**

Chose a `TokenReview`-based check over mutual TLS: every pod, including
the device plugin's, already gets a bound ServiceAccount token by
default, so this needed no new PKI or cert-manager work.

- Device plugin (`numa_refit_client.go`) reads its own bound SA token
  from `/var/run/secrets/kubernetes.io/serviceaccount/token` and sends it
  as `Authorization: Bearer <token>` on `/refit` calls. If the token
  can't be read, the request is still sent without the header (fails
  closed on the server side, not silently dropped on the client side).
- `routes.NumaRefit` extracts the bearer token via a new `bearerToken()`
  helper.
- `authenticateRefitCaller` (new, `numa_refit_auth.go`) rejects an empty
  token the same way as a failed `TokenReview` — one code path. On
  success it checks the token's username matches
  `system:serviceaccount:<device-plugin-namespace>:<device-plugin-sa>`,
  then pulls the bound pod's identity out of
  `status.User.Extra["authentication.kubernetes.io/pod-name"/"pod-uid"]`
  and confirms that pod's `spec.nodeName` matches the request's
  `NodeName`. Any failure returns through the existing
  `numaRefitFailure` helper, so rejections are logged with a specific
  reason the same way other refit failures already are.
- New scheduler flags (`--device-plugin-namespace`,
  `--device-plugin-service-account`) carry the expected identity,
  populated by the chart from the same helpers `daemonsetnvidia.yaml`
  already uses for its own `serviceAccountName`, so they can't drift.
- Scheduler `ClusterRole` gets a new `tokenreviews.authentication.k8s.io`
  `create` rule.

Scope: only `/refit`. `/filter`/`/bind` are the standard kube-scheduler
extender protocol (kube-scheduler is the caller by convention) and are
unrelated to this issue.

**Defense in depth (chart)**

- Scheduler Service default flipped `NodePort` → `ClusterIP`
  (`values.yaml`, `service.yaml`); still overridable.
- Added `charts/hami/templates/scheduler/networkpolicy.yaml`, gated by
  `scheduler.networkPolicy.enabled`. **Defaults to `false`** — see below.

## Why the NetworkPolicy is opt-in, not opt-in-by-default

Originally added it default-on, then caught a real problem before
pushing: `/webhook` shares the same port as `/refit`/`/filter`/`/bind`
(NetworkPolicy can't split by path), and `/webhook` is called by
kube-apiserver, not kube-scheduler or the device plugin. kube-apiserver
is commonly a `hostNetwork: true` static pod, and whether a given CNI
applies NetworkPolicy ingress matching to hostNetwork traffic is
implementation-specific — the chart doesn't commit to a particular CNI,
so this is genuinely undefined here, not something `helm template` can
prove either way.

Traced the failure mode through `webhook.go` rather than assuming it's
safe: with `admissionWebhook.failurePolicy: Ignore` (the chart's
default), a blocked webhook doesn't error or stick pods pending — it
silently skips HAMi's mutation, so a pod requesting `nvidia.com/gpu`
resources can get admitted with **zero GPU enforcement applied and no
error surfaced**. That's a worse silent-failure mode than the NodePort
exposure this PR is fixing, so the policy ships opt-in with that context
noted in `values.yaml`, rather than carved down to a webhook-only
exception that isn't achievable given the shared port.

## Tests

- `numa_refit_handler_test.go`: table test over the four auth outcomes
  (valid token + matching node, wrong SA, wrong node, missing token)
  using a fake clientset `TokenReview` reactor and pod lister.
- `bearerToken()` unit test.
- `go build ./...`, `go vet` clean on touched packages; full
  `pkg/scheduler/...` and device-plugin suites pass.
- `golangci-lint` clean on both commits.
- `helm lint` / `helm template` confirm the chart renders correctly and
  the NetworkPolicy's rendered `podSelector` matches the DaemonSet's and
  Deployment's actual rendered pod labels (checked directly, not just by
  source inspection).

Built/tested via WSL — this environment can't build the cgo-dependent
device-plugin packages natively.

## AI disclosure

This PR, including the code changes, tests, and this description, was
produced with AI assistance (Claude) under my direction and review. I
reviewed the diff, reasoning, and test coverage before committing either
commit, and take responsibility for the change as submitted.

## Summary by CodeRabbit

* **New Features**
  * Added optional network policies to restrict scheduler access.
  * Secured NUMA refit requests with authenticated device-plugin ServiceAccount tokens and caller validation.
  * Added configurable scheduler service ports and Kubernetes client limits.

* **Bug Fixes**
  * Prevented unauthorized or insecure refit requests from modifying allocations.

* **Changes**
  * Scheduler services now default to `ClusterIP`.
  * Improved GPU driver and device specification defaults.
  * Added warnings for network policies that may affect admission webhooks.


pkg/scheduler/numa_refit_handler.go
```diff
@@ -79,7 +79,14 @@ const refitPatchTimeout = 2 * time.Second
 // and hami.io/vgpu-devices-allocated together in one merge patch, and only
 // then moves the in-memory reservation. Failures are reported in-band and
 // leave both annotations and accounting untouched.
-func (s *Scheduler) RefitNumaAllocation(req device.NumaRefitRequest) device.NumaRefitResponse {
+//
+// token is the bearer token the caller presented (from the request's
+// Authorization header, see routes.NumaRefit); it must belong to the
+// device-plugin pod running on req.NodeName, see issue #2878.
+func (s *Scheduler) RefitNumaAllocation(req device.NumaRefitRequest, token string) device.NumaRefitResponse {
+	if err := s.authenticateRefitCaller(context.Background(), token, req.NodeName); err != nil {
```
mesutoezdil: why context.Background() and not r.Context()? this TokenReview has no timeout.

pkg/scheduler/config/config.go
```diff
@@ -71,6 +71,14 @@ var (
 	LeaderElect                  bool
 	LeaderElectResourceName      string
 	LeaderElectResourceNamespace string
+
+	// DevicePluginNamespace and DevicePluginServiceAccount identify the
+	// device-plugin's bound ServiceAccount. The /refit endpoint authenticates
+	// callers by TokenReview and only accepts a token whose username is
+	// exactly system:serviceaccount:<DevicePluginNamespace>:<DevicePluginServiceAccount>.
+	// See issue #2878.
+	DevicePluginNamespace      string
```
mesutoezdil, line 80: both flags default to empty, and then expectedUsername is system:serviceaccount:: and every refit fails. is there a startup check?

pkg/device-plugin/nvidiadevice/nvinternal/plugin/numa_refit_client.go
```diff
@@ -80,23 +90,33 @@ func numaRefitTLSConfig() (*tls.Config, error) {
...
 	}
 	return &http.Client{
-		Timeout:   numaRefitTimeout,
-		Transport: &http.Transport{TLSClientConfig: tlsConfig},
+		Timeout: numaRefitTimeout,
+		Transport: &http.Transport{
+			TLSClientConfig: tlsConfig,
+		},
+		CheckRedirect: func(req *http.Request, via []*http.Request) error {
+			req.Header.Del("Authorization")
```
mesutoezdil: the header is deleted and then the next line returns an error, so the redirect is never followed. is the Del doing something i am missing?

pkg/device-plugin/nvidiadevice/nvinternal/plugin/numa_refit_client.go
```diff
@@ -146,6 +166,30 @@ func (plugin *NvidiaDevicePlugin) tryNumaRefit(ctx context.Context, pod *corev1.
...
+	parsedURL, err := url.Parse(rawURL)
+	if err != nil {
+		return nil, fmt.Errorf("invalid scheduler endpoint URL: %w", err)
+	}
+
+	var token string
+	if tokenBytes, tokenErr := os.ReadFile(serviceAccountTokenFile); tokenErr != nil {
+		// Older scheduler versions accept unauthenticated refits, so degrade
+		// to that behavior rather than failing the allocation outright.
+		klog.InfoS("cannot read service account token; sending refit without caller authentication", "err", tokenErr)
```
mesutoezdil: the comment says it degrades to unauthenticated, but the new scheduler rejects an empty token. what is the intended degrade path?

archlitchi:
please resolve this conflict

AyushSrivastava1818:
@archlitchi pushed the change have a look Sir!

archlitchi (dismissed):
/lgtm

archlitchi:
please resolve these conflicts plz

AyushSrivastava1818:
@archlitchi resolved the conflicts have a look Sir plsssss

archlitchi (approved):
/lgtm

PR #2887, opened 2026-08-29
pkg/device: use sentinel for decoded ContainerDevice.Idx

### LINKED ISSUE #2875: ContainerDevice.Idx does not survive the pod-annotation round trip (EncodeContainerDevices/DecodeContainerDevices drop it)
**What happened**:
ContainerDevice.Idx (pkg/device/devices.go) is set during scheduling, but EncodeContainerDevices (pkg/device/devices.go:451) never writes it into the pod-annotation wire format (UUID,Type,Usedmem,Usedcores), and DecodeContainerDevices (pkg/device/devices.go:496) never restores it. Any ContainerDevice obtained by decoding a pod annotation always has Idx == 0. This is already flagged by a TODO at pkg/device/devices.go:210 but was never turned into a tracked issue.

Either Idx survives the encode/decode round trip, or the field is clearly documented as scheduling-time-only so nothing downstream can rely on it after decode.

This is a code-inspection finding, not a reproduced runtime symptom — no pod manifest or logs apply. To see it directly: call EncodeContainerDevices on a slice with nonzero Idx, then DecodeContainerDevices the result — every decoded entry's Idx is 0.
I checked every current reader of .Idx in the tree (pkg/device/mthreads/device.go:163, cambricon/device.go:332, kunlun/device.go:99,175, ascend/device.go:666, awsneuron/device.go:218-226) — all six read it inside PatchAnnotations/ScoreNode, on the fresh pre-encode scheduling result, so no backend is silently misallocating today.

This is a landmine, not an active bug: anything added later that reads .Idx off a ContainerDevice obtained after decode (e.g. inside the device plugin's Allocate()) will silently get 0 for every device, with no error. Suggested fix directions: (a) include Idx in the wire format, or (b) keep it scheduling-only but document that boundary clearly. Happy to send a PR for whichever direction maintainers prefer.

AI assistance disclosure: I used Claude to scan the codebase and trace every .Idx call site; I've verified the file/line references and reasoning myself before filing.

Gurkaran18:
**What type of PR is this?**

/kind bug

`ContainerDevice.Idx` is not part of the pod-annotation wire format   (`EncodeContainerDevices `/`DecodeContainerDevices`),  so a value set before encoding does not survive a decode — the decoded value always comes back as the zero value, which is indistinguishable from a real device index 0. This PR documents that behavior on the field and introduces an explicit sentinel, `UnsetContainerDeviceIdx = -1`, which `DecodeContainerDevices` now sets instead of leaving `Idx` at its implicit zero value. This makes the gap visible at the type level instead of silently produced 0s, without changing the wire format itself.

Updated the existing test assertions in `pkg/device/devices_test.go` and `pkg/device/numa_refit_test.go` that build `ContainerDevices` via decode to expect the sentinel.

Fixes #2875

This is intentionally scoped as a doc/guard-only fix — no change to the annotation wire format or to how devices are actually allocated. Any code that reads `.Idx` off a *decoded* `ContainerDevice` and previously treated 0 as "valid" should now check for `UnsetContainerDeviceIdx` instead.

This PR was drafted with AI (Claude) assistance; I reviewed the diff and understand the change before opening this PR.

No. Internal field semantics only — no change to allocation behavior or the annotation wire format.

## Summary by CodeRabbit

* **Bug Fixes**
  * Clarified that decoded container device indexes are not preserved through serialization.
  * Decoded devices now use an explicit unset index value instead of defaulting to device 0.
  * Added a public sentinel constant for identifying unset indexes.

* **Tests**
  * Added coverage confirming index values are reset after encoding and decoding.

### DIFF
--- a/pkg/device/devices.go
+++ b/pkg/device/devices.go
@@ -207,7 +207,12 @@ type ResourceNames struct {
 }
 
 type ContainerDevice struct {
-	// TODO current Idx cannot use, because EncodeContainerDevices method not encode this filed.
+	// Idx is the device's index at scheduling time. It is NOT part of the pod
+	// annotation wire format written by EncodeContainerDevices, so it does not
+	// survive a decode: DecodeContainerDevices always sets it to
+	// UnsetContainerDeviceIdx. Only read Idx on a ContainerDevice you built or
+	// received directly from scheduling (e.g. in PatchAnnotations/ScoreNode);
+	// never on one obtained via DecodeContainerDevices/DecodePodDevices.
 	Idx       int
 	UUID      string
 	Type      string
@@ -241,6 +246,13 @@ const (
 
 	// OnePodMultiContainerSplitSymbol this is when one pod having multi container and more than one container use device, use ; symbol to join device info.
 	OnePodMultiContainerSplitSymbol = ";"
+
+	// UnsetContainerDeviceIdx is the Idx value DecodeContainerDevices assigns to
+	// every decoded ContainerDevice, since Idx is not part of the wire format.
+	// It is deliberately an invalid device index (real indices are >= 0) rather
+	// than 0, so code that mistakenly reads Idx off a decoded ContainerDevice
+	// fails obviously instead of silently behaving as if device 0 was meant.
+	UnsetContainerDeviceIdx = -1
 )
 
 var (
@@ -532,6 +544,10 @@ func DecodeContainerDevices(str string) (ContainerDevices, error) {
 			return nil, fmt.Errorf("core field must not be negative: %d", devcores)
 		}
 		tmpdev := ContainerDevice{
+			// Idx is not part of the wire format (see the ContainerDevice.Idx
+			// doc comment): use the sentinel rather than the zero value, which
+			// would be indistinguishable from a real device index 0.
+			Idx:       UnsetContainerDeviceIdx,
 			UUID:      tmpstr[0],
 			Type:      tmpstr[1],
 			Usedmem:   int32(devmem),
--- a/pkg/device/numa_refit_test.go
+++ b/pkg/device/numa_refit_test.go
@@ -74,7 +74,12 @@ func TestNumaRefitResponseCarriesEncodedContainerDevices(t *testing.T) {
 
 	roundTripped, err := DecodeContainerDevices(decoded.ContainerDevices)
 	assert.NilError(t, err)
-	assert.DeepEqual(t, roundTripped, devices)
+	// Idx is not part of the wire format, so it comes back as the sentinel
+	// UnsetContainerDeviceIdx rather than whatever devices set it to (here, 0).
+	want := ContainerDevices{
+		{Idx: UnsetContainerDeviceIdx, UUID: "GPU-aaaa", Type: "NVIDIA", Usedmem: 20000, Usedcores: 30},
+	}
+	assert.DeepEqual(t, roundTripped, want)
 }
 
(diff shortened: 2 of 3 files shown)


pkg/device/devices.go
```diff
@@ -532,6 +544,10 @@ func DecodeContainerDevices(str string) (ContainerDevices, error) {
 			return nil, fmt.Errorf("core field must not be negative: %d", devcores)
 		}
 		tmpdev := ContainerDevice{
+			// Idx is not part of the wire format (see the ContainerDevice.Idx
+			// doc comment): use the sentinel rather than the zero value, which
+			// would be indistinguishable from a real device index 0.
+			Idx:       UnsetContainerDeviceIdx,
```
mesutoezdil: scheduler.go calls CollapseInitContainerUsage right after decode. it rebuilds ContainerDevice without Idx, so it is 0 again. same in SteadyStateDeviceUsage. does the sentinel survive that?

pkg/device/devices.go
```diff
@@ -207,7 +207,12 @@ type ResourceNames struct {
 }
 
 type ContainerDevice struct {
-	// TODO current Idx cannot use, because EncodeContainerDevices method not encode this filed.
+	// Idx is the device's index at scheduling time. It is NOT part of the pod
+	// annotation wire format written by EncodeContainerDevices, so it does not
+	// survive a decode: DecodeContainerDevices always sets it to
+	// UnsetContainerDeviceIdx. Only read Idx on a ContainerDevice you built or
+	// received directly from scheduling (e.g. in PatchAnnotations/ScoreNode);
+	// never on one obtained via DecodeContainerDevices/DecodePodDevices.
```
mesutoezdil: which caller reads Idx off a decoded device today? i could not find one.

pkg/device/devices.go
```diff
@@ -241,6 +246,13 @@ const (
 
 	// OnePodMultiContainerSplitSymbol this is when one pod having multi container and more than one container use device, use ; symbol to join device info.
 	OnePodMultiContainerSplitSymbol = ";"
+
+	// UnsetContainerDeviceIdx is the Idx value DecodeContainerDevices assigns to
+	// every decoded ContainerDevice, since Idx is not part of the wire format.
+	// It is deliberately an invalid device index (real indices are >= 0) rather
+	// than 0, so code that mistakenly reads Idx off a decoded ContainerDevice
+	// fails obviously instead of silently behaving as if device 0 was meant.
+	UnsetContainerDeviceIdx = -1
```
mesutoezdil: this is a public const. it is only ever set by one decoder. why export it?

pkg/device/numa_refit_test.go
```diff
@@ -74,7 +74,12 @@ func TestNumaRefitResponseCarriesEncodedContainerDevices(t *testing.T) {
 
 	roundTripped, err := DecodeContainerDevices(decoded.ContainerDevices)
 	assert.NilError(t, err)
-	assert.DeepEqual(t, roundTripped, devices)
+	// Idx is not part of the wire format, so it comes back as the sentinel
+	// UnsetContainerDeviceIdx rather than whatever devices set it to (here, 0).
+	want := ContainerDevices{
```
mesutoezdil: this hardcodes the fixture again. why not copy devices and set Idx? the two can drift.

mesutoezdil:
closing this: CollapseInitContainerUsage rebuilds ContainerDevice without Idx right after the decode, so the sentinel does not survive the scheduler path it is meant to guard.

PR #2889, opened 2026-08-29
fix(ascend): guard against empty ResourceMemoryName injection in admission

### LINKED ISSUE #2888: fix(ascend): prevent mutating container with empty ResourceMemoryName key during admission
**What happened**:
In `pkg/device/ascend/device.go:187-188`, the `MutateAdmission` webhook handler unconditionally writes to container `Limits` and `Requests` using `corev1.ResourceName(dev.config.ResourceMemoryName)` without checking whether `ResourceMemoryName` is configured:

```go
// Current implementation in pkg/device/ascend/device.go (lines 184-188):
if ctr.Resources.Requests == nil {
    ctr.Resources.Requests = corev1.ResourceList{}
}
ctr.Resources.Limits[corev1.ResourceName(dev.config.ResourceMemoryName)] = resource.MustParse(fmt.Sprint(trimMem))
ctr.Resources.Requests[corev1.ResourceName(dev.config.ResourceMemoryName)] = resource.MustParse(fmt.Sprint(trimMem))
```

When an Ascend chip configuration does not use memory slicing (e.g. `ResourceMemoryName: ""` for full-chip or count-only NPU devices), `corev1.ResourceName(dev.config.ResourceMemoryName)` evaluates to an empty string `""`.

This mutates the container specification by injecting an empty resource key `""`:
- `ctr.Resources.Limits[""] = 0`
- `ctr.Resources.Requests[""] = 0`

When the mutated Pod is admitted and processed by Kubernetes, the API server rejects it with a field validation error:
```text
spec.containers[0].resources.limits[""]: Invalid value: "": name part must not be empty
```

`MutateAdmission` should only inject memory resource limits and requests when `dev.config.ResourceMemoryName` is non-empty (`dev.config.ResourceMemoryName != ""`). If memory slicing is unconfigured, it should leave `ctr.Resources.Limits` and `ctr.Resources.Requests` untouched.

1. Configure an Ascend device in HAMi with count-only resource (`ResourceName: "huawei.com/Ascend910A"`, `ResourceMemoryName: ""`).
2. Submit a Pod requesting `huawei.com/Ascend910A: 1`.
3. `MutateAdmission` mutates the container spec to include `limits[""] = 0` and `requests[""] = 0`.

- **Pod manifest snippet**:
  ```yaml
  apiVersion: v1
  kind: Pod
  metadata:
    name: ascend-test
  spec:
    containers:
    - name: test-ctr
      image: ubuntu:22.04
      resources:
        limits:
          huawei.com/Ascend910A: "1"
  ```
- **Proposed Code Fix in `pkg/device/ascend/device.go`**:
  ```go
  if dev.config.ResourceMemoryName != "" {
      if ctr.Resources.Requests == nil {
          ctr.Resources.Requests = corev1.ResourceList{}
      }
      ctr.Resources.Limits[corev1.ResourceName(dev.config.ResourceMemoryName)] = resource.MustParse(fmt.Sprint(trimMem))
      ctr.Resources.Requests[corev1.ResourceName(dev.config.ResourceMemoryName)] = resource.MustParse(fmt.Sprint(trimMem))
  }
  ```

**Environment**:
- HAMi version: `master` / `v2.10.x`
- Kubernetes version: `v1.28+`
- Accelerator vendor and model: Huawei Ascend (e.g. Ascend 910 series)
- HAMi device backend and requested resource names (for example, `nvidia.com/gpu`): `huawei.com/Ascend910A`
- Accelerator driver and runtime version: N/A (logic in webhook admission)
- Container runtime and version: containerd
- Workload image and tag (redact private image names and state when redacted): Any workload requesting Ascend count resources
- Operating system and kernel version (for example, the output of `uname -a`): Linux
- Others: Tested via unit test on CPU (`go test ./pkg/device/ascend/...`), no hardware GPU required.

KunwarSidhu47:
**What type of PR is this?**

/kind bug

In `pkg/device/ascend/device.go`, `MutateAdmission` unconditionally writes to container `Limits` and `Requests` with `corev1.ResourceName(dev.config.ResourceMemoryName)`:

```go
// Before:
if ctr.Resources.Requests == nil {
    ctr.Resources.Requests = corev1.ResourceList{}
}
ctr.Resources.Limits[corev1.ResourceName(dev.config.ResourceMemoryName)] = resource.MustParse(fmt.Sprint(trimMem))
ctr.Resources.Requests[corev1.ResourceName(dev.config.ResourceMemoryName)] = resource.MustParse(fmt.Sprint(trimMem))
```

When an Ascend configuration does not specify a `ResourceMemoryName` (e.g. count-only NPU devices where memory slicing is not configured), `corev1.ResourceName(dev.config.ResourceMemoryName)` evaluates to an empty string `""`. This causes the webhook to inject an invalid resource key (`limits[""] = 0` and `requests[""] = 0`), which fails Kubernetes API server validation.

This PR fixes it by wrapping the memory limit/request injection with a non-empty check:

```go
// After:
if dev.config.ResourceMemoryName != "" {
    if ctr.Resources.Requests == nil {
        ctr.Resources.Requests = corev1.ResourceList{}
    }
    ctr.Resources.Limits[corev1.ResourceName(dev.config.ResourceMemoryName)] = resource.MustParse(fmt.Sprint(trimMem))
    ctr.Resources.Requests[corev1.ResourceName(dev.config.ResourceMemoryName)] = resource.MustParse(fmt.Sprint(trimMem))
}
```

And adds unit test `Test_MutateAdmission_EmptyResourceMemoryName` in `pkg/device/ascend/device_test.go`.

Fixes #2888 

- Authored and verified manually with AI pair-programming assistance (Antigravity).
- Tests are pure Go unit tests and do not require physical GPU hardware.
- Verified locally via `go test -v ./pkg/device/ascend/...`.

NONE

## Summary by CodeRabbit

- **Tests**
  - Added coverage confirming admission processing succeeds when no memory resource name is configured.
  - Verified that empty-named memory limits and requests are not added to containers.
  - Improved regression protection for admission behavior under an empty memory resource configuration.

### DIFF
--- a/pkg/device/ascend/device.go
+++ b/pkg/device/ascend/device.go
@@ -179,9 +179,9 @@ func (dev *Devices) MutateAdmission(ctr *corev1.Container, p *corev1.Pod) (bool,
 			return true, errors.New("vNPU not supported for multiple devices")
 		}
 	}
-	// Requests may be nil when the pod declares only limits; writing to a
-	// nil map panics.
 	if dev.config.ResourceMemoryName != "" {
+		// Requests may be nil when the pod declares only limits; writing to a
+		// nil map panics.
 		if ctr.Resources.Requests == nil {
 			ctr.Resources.Requests = corev1.ResourceList{}
 		}
--- a/pkg/device/ascend/device.go
+++ b/pkg/device/ascend/device.go
@@ -181,11 +181,13 @@ func (dev *Devices) MutateAdmission(ctr *corev1.Container, p *corev1.Pod) (bool,
 	}
 	// Requests may be nil when the pod declares only limits; writing to a
 	// nil map panics.
-	if ctr.Resources.Requests == nil {
-		ctr.Resources.Requests = corev1.ResourceList{}
+	if dev.config.ResourceMemoryName != "" {
+		if ctr.Resources.Requests == nil {
+			ctr.Resources.Requests = corev1.ResourceList{}
+		}
+		ctr.Resources.Limits[corev1.ResourceName(dev.config.ResourceMemoryName)] = resource.MustParse(fmt.Sprint(trimMem))
+		ctr.Resources.Requests[corev1.ResourceName(dev.config.ResourceMemoryName)] = resource.MustParse(fmt.Sprint(trimMem))
 	}
-	ctr.Resources.Limits[corev1.ResourceName(dev.config.ResourceMemoryName)] = resource.MustParse(fmt.Sprint(trimMem))
-	ctr.Resources.Requests[corev1.ResourceName(dev.config.ResourceMemoryName)] = resource.MustParse(fmt.Sprint(trimMem))
 
 	// Set runtime class name if it is not set by user and the runtime class name is configured
 	if p.Spec.RuntimeClassName == nil && dev.config.RuntimeClassName != "" {
--- a/pkg/device/ascend/device_test.go
+++ b/pkg/device/ascend/device_test.go
@@ -794,8 +794,8 @@ func Test_MutateAdmission_EmptyResourceMemoryName(t *testing.T) {
 			CommonWord:         "Ascend910A",
 			ResourceName:       "huawei.com/Ascend910A",
 			ResourceMemoryName: "",
-			MemoryCapacity:     32768,
-			MemoryAllocatable:  32768,
+			MemoryCapacity:     0,
+			MemoryAllocatable:  0,
 		},
 	}
 	ctr := corev1.Container{
--- a/pkg/device/ascend/device_test.go
+++ b/pkg/device/ascend/device_test.go
@@ -787,6 +787,40 @@ func Test_MutateAdmission_NilRequests(t *testing.T) {
 	}
 }
 
+func Test_MutateAdmission_EmptyResourceMemoryName(t *testing.T) {
+	dev := Devices{
+		config: VNPUConfig{
+			ChipName:           "Ascend910A",
+			CommonWord:         "Ascend910A",
+			ResourceName:       "huawei.com/Ascend910A",
+			ResourceMemoryName: "",
+			MemoryCapacity:     32768,
+			MemoryAllocatable:  32768,
+		},
+	}
+	ctr := corev1.Container{
+		Resources: corev1.ResourceRequirements{
+			Limits: corev1.ResourceList{
+				"huawei.com/Ascend910A": resource.MustParse("1"),
+			},
+		},
+	}
+	pod := corev1.Pod{}
+	result, err := dev.MutateAdmission(&ctr, &pod)
+	if err != nil {
+		t.Fatalf("MutateAdmission returned unexpected error: %v", err)
+	}
+	if !result {
+		t.Fatalf("MutateAdmission expected true, got false")
+	}
+	if _, ok := ctr.Resources.Limits[corev1.ResourceName("")]; ok {
+		t.Fatalf("MutateAdmission should not inject empty resource name into Limits")
+	}
+	if _, ok := ctr.Resources.Requests[corev1.ResourceName("")]; ok {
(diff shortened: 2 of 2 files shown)


pkg/device/ascend/device.go
```diff
@@ -181,11 +181,13 @@ func (dev *Devices) MutateAdmission(ctr *corev1.Container, p *corev1.Pod) (bool,
 	}
 	// Requests may be nil when the pod declares only limits; writing to a
 	// nil map panics.
```
mesutoezdil: this comment is about nil requests. now it sits on the name check. move it inside?

pkg/device/ascend/device.go
```diff
@@ -181,11 +181,13 @@ func (dev *Devices) MutateAdmission(ctr *corev1.Container, p *corev1.Pod) (bool,
...
-	if ctr.Resources.Requests == nil {
-		ctr.Resources.Requests = corev1.ResourceList{}
+	if dev.config.ResourceMemoryName != "" {
```
mesutoezdil, line 182: which config leaves this empty? every ascend block in charts/hami/templates/scheduler/device-configmap.yaml sets resourceMemoryName.

pkg/device/ascend/device.go
```diff
@@ -181,11 +181,13 @@ func (dev *Devices) MutateAdmission(ctr *corev1.Container, p *corev1.Pod) (bool,
...
+		if ctr.Resources.Requests == nil {
+			ctr.Resources.Requests = corev1.ResourceList{}
+		}
+		ctr.Resources.Limits[corev1.ResourceName(dev.config.ResourceMemoryName)] = resource.MustParse(fmt.Sprint(trimMem))
```
mesutoezdil, line 188: line 166 reads limits with the same empty key, and only the write is guarded. is the read safe to leave as is?

pkg/device/ascend/device_test.go
```diff
@@ -787,6 +787,40 @@ func Test_MutateAdmission_NilRequests(t *testing.T) {
...
 
+func Test_MutateAdmission_EmptyResourceMemoryName(t *testing.T) {
+	dev := Devices{
+		config: VNPUConfig{
+			ChipName:           "Ascend910A",
+			CommonWord:         "Ascend910A",
+			ResourceName:       "huawei.com/Ascend910A",
+			ResourceMemoryName: "",
+			MemoryCapacity:     32768,
+			MemoryAllocatable:  32768,
```
mesutoezdil: this config claims 32768 allocatable memory but has no memory resource name. is that valid? should config load reject it instead?

pkg/device/ascend/device.go
KunwarSidhu47: Good point, moved the comment inside the block where it belongs.

pkg/device/ascend/device.go
KunwarSidhu47: The helm chart always sets it, but the field has no validation so a custom config that omits it would default to "" and hit this bug. The guard covers that case.

pkg/device/ascend/device.go
KunwarSidhu47: The read is safe. In Go, reading from a map with an empty key just returns the zero value and ok = false, it won't panic or persist anything. Only the write needed the guard.

pkg/device/ascend/device_test.go
KunwarSidhu47: You're right, having memory capacity set with no resource name is contradictory. Fixed the test to use MemoryCapacity: 0 and MemoryAllocatable: 0 to match a real count only device config.

KunwarSidhu47:
@mesutoezdil , I have made the requested changes. Kindly review it once again. Thanks

archlitchi (approved):
/lgtm

PR #2892, opened 2026-08-30
fix(device-plugin): propagate IMEX channels into the NVIDIA plugin

### LINKED ISSUE #2891: NVIDIA device plugin discards discovered IMEX channels before allocation
**What happened**:

On NVIDIA systems where IMEX channels are discovered (for example GB200 / NVL72 with an IMEX domain configured), the device plugin discovers the channels at startup but never applies them when allocating a device to a container. The container is created without the IMEX channel device nodes, without the IMEX channel mounts, and without the IMEX channel environment variable, so IMEX (internode memory exchange over NVLink) does not work inside HAMi-managed pods.

**Root cause**:

`cmd/device-plugin/nvidia/plugin-manager.go` discovers the channels and passes them into the plugin options:

```go
imexChannels, err := imex.GetChannels(config.Config, driverRoot.getDevRoot())
...
plugin.WithImexChannels(imexChannels),   // stores on options.imexChannels
```

`WithImexChannels` stores the value on the `options` struct. However, the constructor that builds the `NvidiaDevicePlugin` (`devicePluginForResource` in `pkg/device-plugin/nvidiadevice/nvinternal/plugin/server.go`) assembles the struct literal without copying `o.imexChannels` into the plugin's `imexChannels` field, so the field stays nil for the life of the plugin.

Four allocation-time methods then range over that nil field and silently produce nothing: `updateResponseForCDI`, `updateResponseForImexChannelsEnvVar`, `updateResponseForDeviceMounts`, and `apiDeviceSpecs`. The bug is easy to miss because the CDI handler receives its own separate copy of the channels (`cdi.WithImexChannels(imexChannels)` on the same call site), so only the plugin-side handling is affected.

**What I expected**:

The channels passed to `plugin.WithImexChannels` should reach `plugin.imexChannels` so the four methods emit the IMEX channel device nodes, mounts, CDI edits, and environment variable to the container.

**Environment**:

Affects any NVIDIA node where `imex.GetChannels` returns a non-empty set (IMEX-capable systems such as GB200 / NVL72). Independent of Kubernetes version.

Eshiv-Pandey:
**What type of PR is this?**

/kind bug

The NVIDIA device plugin discovers IMEX channels at startup and passes them into the plugin options via `plugin.WithImexChannels(...)`. `WithImexChannels` stores the channels on the `options` struct, but `devicePluginForResource` assembled the `NvidiaDevicePlugin` struct literal without copying `o.imexChannels` into the plugin's `imexChannels` field, so the field was always nil.

As a result the four allocation-time methods that range over `plugin.imexChannels` (`updateResponseForCDI`, `updateResponseForImexChannelsEnvVar`, `updateResponseForDeviceMounts`, and `apiDeviceSpecs`) silently produced nothing. On systems where IMEX channels are discovered (for example GB200 / NVL72), containers were created without the IMEX channel device nodes, mounts, and environment variable. The defect was easy to miss because the CDI handler receives its own separate copy of the channels on the same call site.

This PR:
- Adds the missing `imexChannels: o.imexChannels` field so discovered channels reach the four methods above.
- Extracts the struct assembly into a small `newNvidiaDevicePlugin` method so the wiring from options into the plugin can be unit tested without initializing devices, MIG, or NVML (the untestable constructor is why the omission went unnoticed).
- Adds `TestNewNvidiaDevicePluginPropagatesImexChannels`, which fails on the current code and passes with the fix.

There is no behavior change other than the IMEX field; the extracted method is a straight move of the existing struct assembly.

Fixes #2891

Verified locally on Linux (LF, matching CI): `gofmt -l` clean, `go vet`, `make lint`, `make verify`, and `make test` (full `./pkg/... ./cmd/...` with `-race -short`) all pass. As a sensitivity check, removing only the new `imexChannels: o.imexChannels` line makes the new test fail, confirming it guards the exact bug.

AI assistance: this change was prepared with the help of an AI coding assistant, and all code and tests were reviewed and verified by me.

```release-note
Fixed the NVIDIA device plugin dropping discovered IMEX channels, which left the IMEX device nodes, mounts, and environment variable out of allocated containers.
```

## Summary by CodeRabbit

* **New Features**
  * NVIDIA device allocations now expose configured IMEX channel identifiers through the allocation response environment.

* **Refactor**
  * Streamlined NVIDIA device plugin initialization while preserving existing behavior.
  * Improved propagation of configured IMEX channels to the initialized plugin.

* **Tests**
  * Added coverage verifying that multiple IMEX channel settings are retained and returned correctly.

### DIFF
--- a/pkg/device-plugin/nvidiadevice/nvinternal/plugin/server.go
+++ b/pkg/device-plugin/nvidiadevice/nvinternal/plugin/server.go
@@ -177,8 +177,6 @@ func getPluginSocketPath(resource spec.ResourceName) string {
 
 // NewNvidiaDevicePlugin returns an initialized NvidiaDevicePlugin
 func (o *options) devicePluginForResource(ctx context.Context, nvconfig *nvidia.DeviceConfig, resourceManager rm.ResourceManager, sConfig *config.Config, mode string) (Interface, error) {
-	_, name := resourceManager.Resource().Split()
-
 	deviceListStrategies, _ := spec.NewDeviceListStrategies(*nvconfig.Flags.Plugin.DeviceListStrategy)
 
 	klog.Infoln("reading config=", nvconfig, "resourceName", nvconfig.ResourceName, "configfile=", *ConfigFile, "sconfig=", sConfig)
@@ -197,6 +195,24 @@ func (o *options) devicePluginForResource(ctx context.Context, nvconfig *nvidia.
 			return nil, fmt.Errorf("init MIG instance manager: %w", err)
 		}
 	}
+	return o.newNvidiaDevicePlugin(ctx, nvconfig, resourceManager, deviceListStrategies, sConfig.NvidiaConfig, mode, migMgr), nil
+}
+
+// newNvidiaDevicePlugin assembles an NvidiaDevicePlugin from the options and the
+// values devicePluginForResource resolves at startup. Keeping the assembly in its
+// own method lets the wiring from options into the plugin (for example the IMEX
+// channels) be unit tested without initializing devices, MIG, or NVML.
+func (o *options) newNvidiaDevicePlugin(
+	ctx context.Context,
+	nvconfig *nvidia.DeviceConfig,
+	resourceManager rm.ResourceManager,
+	deviceListStrategies spec.DeviceListStrategies,
+	schedulerConfig nvidia.NvidiaConfig,
+	mode string,
+	migMgr *MigInstanceManager,
+) *NvidiaDevicePlugin {
+	_, name := resourceManager.Resource().Split()
+
 	return &NvidiaDevicePlugin{
 		ctx:                        ctx,
 		rm:                         resourceManager,
@@ -211,9 +227,10 @@ func (o *options) devicePluginForResource(ctx context.Context, nvconfig *nvidia.
 		socket:                     kubeletdevicepluginv1beta1.DevicePluginPath + "nvidia-" + name + ".sock",
 		cdiHandler:                 o.cdiHandler,
 		cdiAnnotationPrefix:        *o.config.Flags.Plugin.CDIAnnotationPrefix,
-		schedulerConfig:            sConfig.NvidiaConfig,
+		schedulerConfig:            schedulerConfig,
 		operatingMode:              mode,
 		migMgr:                     migMgr,
+		imexChannels:               o.imexChannels,
 		deviceCache:                "",
 
 		// These will be reinitialized every
@@ -221,7 +238,7 @@ func (o *options) devicePluginForResource(ctx context.Context, nvconfig *nvidia.
 		server: nil,
 		health: nil,
 		stop:   nil,
-	}, nil
+	}
 }
 
 func (plugin *NvidiaDevicePlugin) initialize() {
--- a/pkg/device-plugin/nvidiadevice/nvinternal/plugin/server.go
+++ b/pkg/device-plugin/nvidiadevice/nvinternal/plugin/server.go
@@ -195,7 +195,7 @@ func (o *options) devicePluginForResource(ctx context.Context, nvconfig *nvidia.
 			return nil, fmt.Errorf("init MIG instance manager: %w", err)
 		}
 	}
-	return o.newNvidiaDevicePlugin(ctx, nvconfig, resourceManager, deviceListStrategies, sConfig.NvidiaConfig, mode, migMgr), nil
+	return o.newNvidiaDevicePlugin(ctx, resourceManager, deviceListStrategies, sConfig.NvidiaConfig, mode, migMgr), nil
 }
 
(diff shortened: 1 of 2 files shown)


pkg/device-plugin/nvidiadevice/nvinternal/plugin/server.go
```diff
@@ -177,8 +177,6 @@ func getPluginSocketPath(resource spec.ResourceName) string {
 
 // NewNvidiaDevicePlugin returns an initialized NvidiaDevicePlugin
 func (o *options) devicePluginForResource(ctx context.Context, nvconfig *nvidia.DeviceConfig, resourceManager rm.ResourceManager, sConfig *config.Config, mode string) (Interface, error) {
-	_, name := resourceManager.Resource().Split()
-
 	deviceListStrategies, _ := spec.NewDeviceListStrategies(*nvconfig.Flags.Plugin.DeviceListStrategy)
```
mesutoezdil, line 180: o.deviceListStrategies is set in plugin-manager.go next to WithImexChannels and never read. same gap, one option away. left out on purpose?

pkg/device-plugin/nvidiadevice/nvinternal/plugin/server.go
```diff
@@ -197,6 +195,24 @@ func (o *options) devicePluginForResource(ctx context.Context, nvconfig *nvidia.
...
+	return o.newNvidiaDevicePlugin(ctx, nvconfig, resourceManager, deviceListStrategies, sConfig.NvidiaConfig, mode, migMgr), nil
+}
+
+// newNvidiaDevicePlugin assembles an NvidiaDevicePlugin from the options and the
+// values devicePluginForResource resolves at startup. Keeping the assembly in its
+// own method lets the wiring from options into the plugin (for example the IMEX
+// channels) be unit tested without initializing devices, MIG, or NVML.
+func (o *options) newNvidiaDevicePlugin(
+	ctx context.Context,
+	nvconfig *nvidia.DeviceConfig,
```
mesutoezdil: factory.go passes o.config as nvconfig, so nvconfig is always o.config here, and the body still reads o.config for cdiAnnotationPrefix. is there a case where the two differ?

pkg/device-plugin/nvidiadevice/nvinternal/plugin/server.go
```diff
@@ -211,17 +227,18 @@ func (o *options) devicePluginForResource(ctx context.Context, nvconfig *nvidia.
 		socket:                     kubeletdevicepluginv1beta1.DevicePluginPath + "nvidia-" + name + ".sock",
```
mesutoezdil: getPluginSocketPath builds this same path and currently has no callers. the new method carries Resource().Split() just to produce name. would calling that helper work here?

pkg/device-plugin/nvidiadevice/nvinternal/plugin/server_test.go
```diff
@@ -207,6 +207,48 @@ func TestCDIAllocateResponse(t *testing.T) {
...
+		context.Background(),
+		o.config,
+		resourceManager,
+		deviceListStrategies,
+		nvidia.NvidiaConfig{},
+		"hami-core",
+		nil,
+	)
+
+	require.Equal(t, channels, plugin.imexChannels,
```
mesutoezdil, line 247: this checks the field only. the reported bug was channels missing in the container. does any test cover updateResponseForImexChannelsEnvVar with channels set?

pkg/device-plugin/nvidiadevice/nvinternal/plugin/server.go
Eshiv-Pandey: Yes, on purpose unlike IMEX this field is filled from a param, so it's a dead option rather than a broken feature. tbh could not decide to keep it or remove it.

pkg/device-plugin/nvidiadevice/nvinternal/plugin/server.go
Eshiv-Pandey: no they can't differ actually factory.go is the only caller and always passes o.config. Fixed in 79a8e8b by dropping the redundant nvconfig param and reading o.config throughout.

pkg/device-plugin/nvidiadevice/nvinternal/plugin/server_test.go
Eshiv-Pandey: it does now.. 79a8e8b adds TestUpdateResponseForImexChannelsEnvVarExposesChannels covering wht you mentioned, thanks.

pkg/device-plugin/nvidiadevice/nvinternal/plugin/server.go
Eshiv-Pandey: yes, 79a8e8b calls getPluginSocketPath and drops the local Split().

archlitchi (approved):
/lgtm

PR #2894, opened 2026-08-30
fix(nvidia): keep device registration order stable across scans

### LINKED ISSUE #2893: nvidia device-plugin: sort devices in getAPIDevices so the node registration annotation is stable across scans
**Task description**:

On nodes with two or more NVIDIA GPUs, the device plugin rewrites the `hami.io/node-nvidia-register` annotation and logs `Updating node annotations with N device(s)` on every scan, even when nothing has changed. The `deviceCache` check in `RegisterInAnnotation` that is meant to skip the update almost never matches.

The reason: `getAPIDevices` builds the device list by ranging over a map and returns it unsorted (`register.go:112` ranges a `map[string]*Device`, returns at `register.go:209`). Go randomizes map order, so the marshaled string differs each scan and the cache comparison at `register.go:275` fails, forcing a node PATCH every loop. Single-GPU nodes are unaffected.

**Solution**:

1. Sort `res` by `Index` before `return &res` in `getAPIDevices` (`register.go:209`).
2. Add a test in `register_test.go` asserting the encoded output is identical across two calls.

Small change, NVIDIA device plugin only.

**Who can join or take the task**:

The good first issue is intended for `first-time contributors` to get started on his/her contributor journey.

After a contributor has successfully completed 1-2 good first issue's,
they should be ready to move on to `help wanted` items, saving the remaining `good first issue` for other new contributors.

**How to join or take the task**:

Just reply on the issue with the message `/assign` in a separate line.

Then, the issue will be assigned to you.

**How to ask for help**:

If you need help or have questions, please feel free to ask on this issue.
The issue author or other members of the community will guide you through the contribution process.

libaojiang:
**What type of PR is this?**

/kind bug

Sorts the NVIDIA devices returned by `getAPIDevices` by their stable NVML index before they are encoded for the node registration annotation. This prevents Go map iteration order from changing the annotation payload and triggering an unnecessary node patch on every scan.

Adds a mocked two-GPU regression test that calls `getAPIDevices` twice, verifies index ordering, and asserts that both encoded payloads are identical.

Fixes #2893

- `go fmt` and `git diff --check` pass for the touched files.
- Linux verification passed on Ubuntu 22.04: the focused regression test ran 20 times, and the complete `nvinternal/plugin` package test suite passed ([workflow run](https://github.com/libaojiang/HAMi/actions/runs/33316986244)).
- No GPU hardware was used. NVML is mocked by the regression test, and this change does not affect device allocation or in-container isolation behavior.
- AI assistance disclosure: I used OpenAI Codex to inspect the relevant code, prepare the small implementation and regression test, and run the available checks. I reviewed the resulting diff and understand the map-ordering bug, the index sort, and the test setup.

```release-note
NONE
```

### DIFF
--- a/pkg/device-plugin/nvidiadevice/nvinternal/plugin/register.go
+++ b/pkg/device-plugin/nvidiadevice/nvinternal/plugin/register.go
@@ -37,6 +37,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"sort"
 	"strconv"
 	"strings"
 	"time"
@@ -206,6 +207,9 @@ func (plugin *NvidiaDevicePlugin) getAPIDevices() *[]*device.DeviceInfo {
 		res = append(res, info)
 		klog.V(3).Infof("Registered device id=%v, memory=%vMB, type=%v, numa=%v, health=%v", idx, registeredmem, Model, numa, health)
 	}
+	sort.Slice(res, func(i, j int) bool {
+		return res[i].Index < res[j].Index
+	})
 	return &res
 }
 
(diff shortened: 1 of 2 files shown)


mesutoezdil:

PR #2895, opened 2026-08-30
fix(chart): target the scheduler metrics port by name

mesutoezdil:
The scheduler monitor service hardcoded targetPort 9395 while the container port is derived from `scheduler.metricsBindAddress`, so changing only that value left the service pointing at a closed port and the ServiceMonitor target down with no error. This targets the named `metrics` port instead, so the service follows the configured bind address.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the scheduler monitoring Service to target the named `metrics` port by default.
  * Monitoring traffic now follows the configured metrics endpoint rather than relying on a fixed port number.

* **Documentation**
  * Updated Helm chart values and monitoring guidance to reflect the new default monitoring target.
  * Clarified that the Service target follows the scheduler’s configured metrics bind address.

```release-note
The scheduler monitoring Service now targets the named metrics container port, so it follows scheduler.metricsBindAddress instead of a fixed 9395. Setting scheduler.service.monitorTargetPort still overrides it.
```

### DIFF
--- a/charts/hami/README.md
+++ b/charts/hami/README.md
@@ -139,7 +139,7 @@ This document provides detailed descriptions of all configurable values paramete
 | `scheduler.service.httpPort` | HTTP port | `443` |
 | `scheduler.service.schedulerPort` | Scheduler NodePort | `31998` |
 | `scheduler.service.monitorPort` | Monitor port | `31993` |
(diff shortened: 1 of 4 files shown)


archlitchi (approved):
/lgtm

PR #2897, opened 2026-08-30
fix(monitor): report metrics server listen failures

mesutoezdil:
A failed metrics bind was only logged, so `initMetrics` kept blocking on the context and the container stayed up with nothing to scrape and no probe to catch it. This binds before wiring up the collectors so the failure is caught at startup, and lets `start()` return the error so the process exits non-zero instead of Completed.

## Summary by CodeRabbit

* **Bug Fixes**
  * The monitoring process now exits with an error when startup or runtime failures occur.
  * Metrics service startup failures, including unavailable network ports, are now detected and reported instead of being silently logged.
  * Metrics shutdown is handled more gracefully when the application is canceled.

```release-note
vGPUmonitor now exits with an error when the metrics server cannot bind, instead of staying up with no metrics endpoint.
```

### DIFF
--- a/cmd/vGPUmonitor/main.go
+++ b/cmd/vGPUmonitor/main.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"net"
 	"net/http"
 	"os"
 	"os/signal"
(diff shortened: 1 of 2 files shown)


archlitchi (approved):
/lgtm

PR #2898, opened 2026-08-30
fix(metrics): drop the constant container index label

mesutoezdil:
`PodManager` only stores each pod's collapsed device usage, one entry per device type, so `container_index` and `containeridx` were always 0 and two containers sharing a GPU landed in a single series that looked per container. This drops the label from the four vGPU allocation metrics and corrects the skill doc that told operators to pick a container by it.


archlitchi (approved):
/lgtm

PR #2900, opened 2026-08-30
fix(scheduler): preserve allocation slots for model-specific types

### LINKED ISSUE #2899: v2.10.0 drops init-container allocation slots on model-specific NVIDIA nodes
### What happened

HAMi v2.10.0 drops the leading allocation slot for an init container that does
not request a device when the node registers a model-specific NVIDIA type.

With one ordinary init container and two app containers requesting NVIDIA
vGPUs, the scheduler sees the correct three request rows:

```text
[{}, {"NVIDIA": ...}, {"NVIDIA": ...}]
```

but writes only two rows to the Pod annotation:

```text
hami.io/vgpu-devices-allocated=GPU-A,NVIDIA,2500,25:;GPU-B,NVIDIA,2500,25:;
```

The expected positional encoding is:

```text
hami.io/vgpu-devices-allocated=;GPU-A,NVIDIA,2500,25:;GPU-B,NVIDIA,2500,25:;
```

### Why this matters

The annotation is positional. Dropping the empty init slot shifts every app
container:

- container/device identity reported by `vgpu-monitor` is wrong;
- the last app container is absent from container metrics;
- after init completion, app-only scheduler accounting retained only one of the
  two app allocations in this reproduction, under-reporting live usage.

This was reproduced with real T4 workloads, not a hand-written annotation.
Both app containers were Running and each received a different GPU, while
`hami_container_device_utilization_ratio` exposed only one of them under the
wrong device identity.

### Environment

- HAMi: v2.10.0 (`4707fb02c91c545bc7343ce26dba4c32919f9a3e`)
- Kubernetes: k3s v1.36.4+k3s1
- GPU: NVIDIA Tesla T4
- registered node device type: `NVIDIA-Tesla T4`

Minimal Pod shape:

```yaml
spec:
  initContainers:
    - name: prepare
      image: busybox:1.36.1
      command: [sh, -c, 'echo ready']
  containers:
    - name: worker-a
      image: nvcr.io/nvidia/k8s/cuda-sample:vectoradd-cuda12.5.0-ubuntu22.04
      resources:
        limits:
          nvidia.com/gpu: 1
          nvidia.com/gpumem: 2500
          nvidia.com/gpucores: 25
    - name: worker-b
      image: nvcr.io/nvidia/k8s/cuda-sample:vectoradd-cuda12.5.0-ubuntu22.04
      resources:
        limits:
          nvidia.com/gpu: 1
          nvidia.com/gpumem: 2500
          nvidia.com/gpucores: 25
```

### Root cause and current master

In v2.10.0, empty init rows are padded using `nodeDeviceBaseTypes`, which
contains the model-specific key (`NVIDIA-Tesla T4`). `fitInDevices` stores the
real allocations under the request key (`NVIDIA`). `PatchAnnotations` encodes
the latter, so its init placeholder is missing.

Current master already avoids the mismatch: #2723 / `d961856d` introduced
`allocationTypeKeys`, the union of registered model types and request types.
This issue is therefore about guarding the exact regression and making the fix
available to v2.10 users, not proposing a second implementation.

### Suggested follow-up

1. Add a regression test with a model-specific node type, one non-GPU init
   container, and two generic NVIDIA app requests; assert the leading empty
   annotation slot and post-init accounting.
2. Backport the small request-type/model-type key-union and padding portion to
   the v2.10 patch line; the unrelated native-sidecar feature work from #2723
   does not need to move with it.

Nimbus318:
**What type of PR is this?**

/kind bug

Backports the allocation-type key union from #2723 to `release-v2.10`.

On v2.10.0, nodes can register a model-specific type such as
`NVIDIA-Tesla T4`, while NVIDIA requests and fitted allocations use the
generic `NVIDIA` key. Empty init or app container slots were padded only under
the model key, so the encoded `NVIDIA` annotation dropped those positions and
shifted device ownership between containers.

This change pads both registered and requested device-type rows. It adds one
regression test covering empty init and app slots and the final annotation wire
format. The unrelated native-sidecar changes from #2723 are intentionally not
included.

Fixes #2899

Current `master` already contains the functional fix through #2723; this is the
minimal patch-release backport.

Verification:

- `go test ./pkg/scheduler -count=1`
- `go test -race ./pkg/scheduler -run '^TestScoreNodePreservesSlotsForModelSpecificDeviceType$' -count=1`

```release-note
Fix allocation annotations on model-specific NVIDIA nodes so init and app container slots remain aligned.
```

### DIFF
--- a/pkg/scheduler/score.go
+++ b/pkg/scheduler/score.go
@@ -223,7 +223,23 @@ func applyPeakUsage(node *NodeUsage, appNodeCopy *NodeUsage, peakUsage map[strin
 	}
 }
 
-func allocateInitContainers(node *NodeUsage, nodeID string, resourceReqs device.PodDeviceRequests, task *corev1.Pod, nodeInfo *device.NodeInfo, baseTypes map[string]struct{}, numInitContainers int, peakUsage map[string]peakUsageSnapshot, weights util.DeviceScoringWeights) (device.PodDevices, bool, string) {
+// fitInDevices stores allocations under the request type (for example,
+// "NVIDIA"), while nodes may register a model-specific device type. Include
+// both sets of keys so empty container slots are added to every allocation row.
+func allocationTypeKeys(resourceReqs device.PodDeviceRequests, baseTypes map[string]struct{}) map[string]struct{} {
+	allocTypes := make(map[string]struct{}, len(baseTypes))
+	for typ := range baseTypes {
+		allocTypes[typ] = struct{}{}
+	}
+	for _, ctrReqs := range resourceReqs {
+		for _, req := range ctrReqs {
+			allocTypes[req.Type] = struct{}{}
+		}
+	}
+	return allocTypes
+}
+
+func allocateInitContainers(node *NodeUsage, nodeID string, resourceReqs device.PodDeviceRequests, task *corev1.Pod, nodeInfo *device.NodeInfo, allocTypes map[string]struct{}, numInitContainers int, peakUsage map[string]peakUsageSnapshot, weights util.DeviceScoringWeights) (device.PodDevices, bool, string) {
 	initAllocs := make(device.PodDevices)
 
 	for i, req := range resourceReqs {
@@ -231,7 +247,7 @@ func allocateInitContainers(node *NodeUsage, nodeID string, resourceReqs device.
 			break
 		}
 		if len(req) == 0 {
-			for typ := range baseTypes {
+			for typ := range allocTypes {
 				initAllocs[typ] = append(initAllocs[typ], device.ContainerDevices{})
 			}
 			continue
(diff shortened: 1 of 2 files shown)


Nimbus318:
Real-GPU backport validation passed on k3s `v1.36.4+k3s1` with two T4 nodes (1 GPU + 2 GPUs), using the exact image artifact from CI run [33341810424](https://github.com/Project-HAMi/HAMi/actions/runs/33341810424). Scheduler, device plugin, and monitor all ran image ID `sha256:48cb3a81dfc97c1ba1c35052544ffe89154eb03349d2c1e0618d8aa08032ef15`.

The reproduction pod had one device-less init container followed by two NVIDIA app containers. Observed after the fix:

- the allocation annotation preserved the leading empty slot: `;GPU-cfa...,NVIDIA,1536,10:;GPU-daca...,NVIDIA,1536,10:;`
- `data-worker` and `trainer` were allocated to the expected distinct UUIDs
- after the init container completed, scheduler `oldUsage -> newUsage` retained both app allocations
- raw HAMi metrics and WebUI-derived core/memory metrics contained both containers with non-zero samples
- the WebUI API returned both rows with the correct UUID, 10 cores, and 1536 MiB

All UAT pods remained `Running` with zero restarts.

archlitchi (approved):
/lgtm

archlitchi (approved):
/lgtm

PR #2912, opened 2026-08-31
fix(scheduler): bind cache readiness to leadership terms

### LINKED ISSUE #2911: fix(scheduler): bind cache readiness to leadership terms after failover
/kind bug

## What happened

In an HA deployment with multiple `hami-scheduler` replicas, cache readiness is tracked by a process-wide boolean and is not bound to a specific leadership term.

During a leadership transition such as A -> B -> A, a `register()` operation started during the first A term may finish after the scheduler has reacquired leadership. Its final `IsLeader()` check returns true, so it publishes `synced=true` even though the refresh belongs to an older leadership term.

A similar issue can occur when the observed Lease expires and later becomes valid again with the same holder identity. The scheduler does not currently represent this as a new leadership term.

## Expected behavior

Cache readiness should belong to the current leadership term.

After losing leadership or after a Lease expiry, the scheduler should remain unready until a complete cache registration has successfully completed in the new leadership term.

An old `register()` operation must not be able to publish readiness for a newer term.

## Reproduction

1. Deploy multiple scheduler replicas with leader election enabled.
2. Start a cache registration and block it during device discovery or node usage retrieval.
3. Change the Lease holder from replica A to replica B.
4. Change the Lease holder back to replica A before the original registration finishes.
5. Release the blocked registration.
6. Observe that the old registration can publish `synced=true` for the new leadership period.

The same behavior should also be tested with an expired Lease followed by a real renewal using the same holder identity.

## Root cause

The scheduler currently uses:

- A boolean `synced` state without a leadership generation.
- `IsLeader()` checks that only report the current Lease state.
- A `register()` final check that cannot distinguish the current term from a previous term.
- Lease observation timestamps that must not be refreshed by informer resync or metadata-only updates.

## Proposed solution

Introduce a process-local, opaque leadership term or session:

- Advance the term whenever the scheduler enters a new observed leadership term.
- Clear cache readiness when a term starts or ends.
- Capture the term at the beginning of `register()`.
- Publish readiness only if the captured term is still the current term.
- Treat Lease expiry followed by a real renewal as a new term, even when the holder identity is unchanged.
- Do not refresh Lease validity for informer resync or metadata-only updates.
- Keep leadership callbacks independent from the long-running registration lock.
- Reject new filter, bind, and NUMA refit requests unless the current term has completed cache synchronization.

## Scope

This issue covers term-aware cache readiness and admission of new requests.

It does not attempt to provide distributed fencing for requests that were already admitted, nor transactional rollback of external side effects already performed by `register()`.

## Acceptance criteria

- A -> B -> A cannot allow an old `register()` operation to publish readiness.
- Lease expiry followed by a real same-holder renewal starts a new term.
- Informer resync does not revive an expired Lease.
- Leadership-loss callbacks do not wait for the full registration refresh.
- New `/filter`, `/bind`, and `/refit` requests are rejected until the current term is synchronized.
- Tests cover A -> B, A -> B -> A, same-holder expiry and renewal, and callback ordering.

## Related issues and PRs

- #1954: hami-scheduler multi-copy deployment problem
- #1957: fix_leaderNotify Block
- #1553: Implementing leader-follower high availability using leader-election
- #2758: Add Prometheus metrics for scheduler leader and cache sync state

This issue is related to the failover problem in #1954, but addresses a different remaining race: stale cache registration results being reused across leadership terms.

Lcos-000:
/kind bug

Fixes a leadership-term race where scheduler cache readiness can be published by a stale `register()` operation after a leadership transition.

In an HA deployment, a registration started during leadership term A may finish after an A -> B -> A transition. Since the scheduler only checks the current leader status, the old registration can incorrectly set `synced=true` for the new leadership term.

This PR binds cache readiness to an observed, process-local leadership term.

Fixes #2911

## Changes

- Add a term-aware leadership state to the leader election package.
- Advance the leadership term when leadership is acquired, the holder identity changes, or an expired Lease is renewed, including same-holder renewal.
- Prevent informer resync and metadata-only Lease updates from extending Lease validity.
- Ensure stop/start callbacks are paired and ordered across term transitions.
- Replace the scheduler-wide synced boolean with a term-bound synced state.
- Prevent an old `register()` operation from publishing readiness for a newer leadership term.
- Keep leadership callbacks independent from the long-running registration lock.
- Reject new `/filter`, `/bind`, and `/refit` requests until the current leadership term has completed cache synchronization.
- Preserve the existing `LeaderManager` interface and add `TermAwareLeaderManager` for term-aware consumers.
- Add regression tests for A -> B -> A, same-holder Lease expiry and renewal, informer resync, callback ordering, and hostname-prefix collisions.

## Scope

This PR guarantees that cache readiness and new-request admission belong to the current observed leadership term.

This PR does not provide distributed fencing for requests that were already admitted. It also does not transactionally roll back external side effects already performed by `register()`, such as Kubernetes label updates, device cleanup, or local cache mutations.

Those behaviors require a separate design involving request cancellation, ownership tokens, resource-version preconditions, or node-lock fencing.

## Testing

- `go test ./pkg/scheduler/... ./pkg/util/leaderelection ./cmd/scheduler -count=1`
- `go test -short -race ./pkg/scheduler/... ./pkg/util/leaderelection ./cmd/scheduler -count=1`
- `go vet ./pkg/scheduler/... ./pkg/util/leaderelection ./cmd/scheduler/...`
- Leadership term tests passed with `-race -count=50`.
- Scheduler leadership tests passed with `-race -count=30`.
- `gofmt` and `git diff --check` passed.

The full repository test run still has two unrelated host-environment failures in NVIDIA hostpid/MIG tests.

## Related issues and PRs

- #1954: hami-scheduler multi-copy deployment problem
- #1957: fix_leaderNotify Block
- #1553: Implementing leader-follower high availability using leader-election
- #2758: Add Prometheus metrics for scheduler leader and cache sync state

This PR is related to #1954 and extends the existing #1957 fix. It addresses a separate remaining race involving stale cache registration results being reused across leadership terms.

## AI assistance disclosure

AI assistance was used for repository research and regression test authoring. The implementation and final changes were reviewed and validated with unit tests, race detection, and static analysis.

## Summary by CodeRabbit

* **Bug Fixes**
  * Scheduling requests now fail fast with clear errors when the scheduler is not the active leader or its cache is not synchronized.
  * Improved readiness handling prevents stale synchronization state from being restored after leadership changes.
  * Leadership detection now avoids incorrectly matching similarly named instances.
* **Tests**
  * Added coverage for leadership transitions, cache synchronization, readiness, and unavailable-scheduler request handling.

### DIFF
--- a/pkg/scheduler/routes/numa_refit_route_test.go
+++ b/pkg/scheduler/routes/numa_refit_route_test.go
@@ -17,7 +17,6 @@ limitations under the License.
 package routes
 
 import (
-	"context"
 	"encoding/json"
 	"net/http/httptest"
 	"strings"
@@ -65,13 +64,10 @@ func TestNumaRefitRoute_CacheNotSynced(t *testing.T) {
 		t.Fatalf("failed to marshal request: %v", err)
 	}
 
-	ctx, cancel := context.WithCancel(context.Background())
-	cancel() // cancel immediately so WaitForCacheSync fails fast instead of polling forever
-
-	req := httptest.NewRequest("POST", "/refit", strings.NewReader(string(body))).WithContext(ctx)
+	req := httptest.NewRequest("POST", "/refit", strings.NewReader(string(body)))
 	w := httptest.NewRecorder()
 
-	handler := NumaRefit(&scheduler.Scheduler{}) // zero value: synced defaults to false
+	handler := NumaRefit(newTestScheduler(t, false))
 	handler(w, req, nil)
 
 	if w.Code != 200 {
@@ -81,7 +77,30 @@ func TestNumaRefitRoute_CacheNotSynced(t *testing.T) {
 	if err := json.Unmarshal(w.Body.Bytes(), &response); err != nil {
 		t.Fatalf("failed to unmarshal response: %v", err)
 	}
-	if response.Succeeded || !strings.Contains(response.FailureReason, "context cancelled") {
+	if response.Succeeded || !strings.Contains(response.FailureReason, "scheduler cache is not synced") {
 		t.Errorf("expected cache-not-synced failure, got %+v", response)
+	}
+}
+
+func TestNumaRefitRoute_NotLeaderFailsFast(t *testing.T) {
+	body, err := json.Marshal(device.NumaRefitRequest{PodUID: "uid"})
+	if err != nil {
+		t.Fatalf("failed to marshal request: %v", err)
+	}
+
+	req := httptest.NewRequest("POST", "/refit", strings.NewReader(string(body)))
+	w := httptest.NewRecorder()
+	handler := NumaRefit(newTestScheduler(t, true))
+	handler(w, req, nil)
+
+	if w.Code != 200 {
+		t.Fatalf("expected HTTP 200 with in-band error, got %d", w.Code)
+	}
+	var response device.NumaRefitResponse
+	if err := json.Unmarshal(w.Body.Bytes(), &response); err != nil {
+		t.Fatalf("failed to unmarshal response: %v", err)
+	}
+	if response.Succeeded || !strings.Contains(response.FailureReason, "scheduler is not leader") {
+		t.Errorf("expected not-leader failure, got %+v", response)
 	}
 }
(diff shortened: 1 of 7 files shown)


Lcos-000:
Superseded by #2913, which contains the same change on a signed commit so the DCO check can pass.

PR #2924, opened 2026-09-01
fix(kunlun): compare elements instead of indices in calcscore sort

Woolgathererer:

`calcscore` in `pkg/device/kunlun/topo.go` sorts its `prev`/`cur` inputs with
`sort.Slice(p, func(i, j int) bool { return i < j })`. The comparator compares
**slice indices** instead of elements: since `i < j` holds for every pair the
sort algorithm probes, both sorts report "already ordered" and are silent
no-ops — the slices come back unsorted.

Today's scores are unaffected because `countbubble` only counts elements
below/above 4 (order-independent). However, the sort is a documented contract:
`TestCalcscore` passes copies of its inputs with the comment *"calcscore sorts
its inputs in place; pass copies"*. That contract is silently broken, and any
future order-sensitive consumer of these slices would misbehave.

This PR fixes both comparators to `p[i] < p[j]` / `c[i] < c[j]` and adds a
regression test `TestCalcscoreSortsInputsInPlace` that feeds unsorted slices
and asserts they come back sorted. The test fails against the old comparator
(`calcscore left prev unsorted: [6 2 4 0]`) and passes with the fix.

## AI Assistance Disclosure

This PR was written with AI assistance (Claude) for bug identification; the
analysis was verified, and the fix rationale and test design were reviewed and
understood by me.

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected score calculation sorting so input data is ordered by value as expected.
  - Improved consistency and reliability when processing unsorted device topology data.
  - Ensured score calculations produce stable results regardless of the original ordering of topology inputs.

- **Tests**
  - Added coverage verifying that input data is sorted correctly during score calculation.
  - Added validation for processing unsorted inputs and confirming the resulting order.

### DIFF
--- a/pkg/device/kunlun/topo.go
+++ b/pkg/device/kunlun/topo.go
@@ -74,12 +74,8 @@ func countbubble(t []int) int {
 }
 
 func calcscore(p []int, c []int) float32 {
-	sort.Slice(p, func(i, j int) bool {
-		return i < j
-	})
-	sort.Slice(c, func(i, j int) bool {
-		return i < j
-	})
+	slices.Sort(p)
+	slices.Sort(c)
 	prev := countbubble(p)
 	cur := countbubble(c)
 	klog.V(5).Infoln("Score kunlun num prev=", prev, "cur=", cur)
--- a/pkg/device/kunlun/topo_test.go
+++ b/pkg/device/kunlun/topo_test.go
@@ -191,6 +191,24 @@ func TestCalcscore(t *testing.T) {
 	}
 }
 
+func TestCalcscoreSortsInputsInPlace(t *testing.T) {
+	// calcscore sorts its inputs in place (the contract TestCalcscore relies
+	// on when it passes copies). Feed unsorted slices and verify they come
+	// back sorted: a comparator returning "i < j" compares slice indices
+	// instead of elements and leaves the slices untouched.
+	prev := []int{6, 2, 4, 0}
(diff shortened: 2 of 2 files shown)


pkg/device/kunlun/topo.go
```diff
@@ -75,10 +75,10 @@ func countbubble(t []int) int {
 
 func calcscore(p []int, c []int) float32 {
 	sort.Slice(p, func(i, j int) bool {
```
mesutoezdil: lint flags this as slices.Sort(p) since the comparator is now just p[i] < p[j]. would that work here?

mesutoezdil:
DCO is failing too, this needs a signed-off commit (git commit --amend -s) before it can merge.

pkg/device/kunlun/topo.go
Woolgathererer: reasonable! TKS!!

Woolgathererer:
@mesutoezdil hi~ seems well performed~ can u merge it?

Woolgathererer:
@mesutoezdil thx~! but it's still waiting on ur review/approval~~ is that normal?

mesutoezdil:
> @mesutoezdil thx~! but it's still waiting on ur review/approval~~ is that normal?

Just waiting mate :)

archlitchi (approved):
/lgtm

PR #2926, opened 2026-09-01
 honor hami.io/device-cordon in all backends

### LINKED ISSUE #2925: Honor `hami.io/device-cordon` across all device backends
**What would you like to be added?**

Extend the existing `hami.io/device-cordon` support from NVIDIA to all device backends.

Currently, the cordon annotation is handled only by the NVIDIA backend. Other backends ignore the annotation, so an operator can cordon a device but workloads may still be scheduled on it.

**Why is this needed?**

`hami.io/device-cordon` is an operator-facing control, so it should behave consistently across supported device backends.

The shared `CardCordoned` reason already exists in `pkg/device/common`, and all backends already have their own `Fit()` checks for unavailable devices. This makes it possible to add the cordon check consistently without changing the existing behavior when the annotation is not set.

**Implementation notes:**

* Add shared handling for the `hami.io/device-cordon` annotation.
* Apply the cordon check in each device backend's `Fit()` path.
* Keep the check after the existing health check where applicable.
* Preserve device positions for topology-aware backends such as Kunlun.
* Reuse `common.CardCordoned` so cordoned devices are reported through the existing filtering reason mechanism.
* Add tests for the supported backends, following the existing NVIDIA `TestFit_DeviceCordon` coverage.

**Compatibility:**

This should not introduce a breaking change. If `hami.io/device-cordon` is not set, scheduling behavior remains unchanged.

Before implementing this, it would be useful to confirm whether the NVIDIA-only scope in #2298 was intentional or if extending it to all backends is the expected behavior.

I'd be happy to work on this once the scope is confirmed.

im-Toqeer-506:
**What type of PR is this?**

/kind feature

Adds `pkg/scheduler/score_bench_test.go` with benchmarks for the scheduler scoring path:

* `BenchmarkCalcScore` — benchmarks `calcScoreWithOptions` with 10, 100, and 1000 nodes.
* `BenchmarkScoreNode` — benchmarks `scoreNode` with 0, 1, and 4 non-sidecar init containers.
* Uses `b.ReportAllocs()` to track both execution time and allocations.

These benchmarks provide a baseline for the current scoring performance and will help evaluate future changes such as the bounded worker pool proposed in #2858.

This PR is test-only; no production code is changed.

Fixes #2925

* Benchmark fixtures are rebuilt for each iteration because scoring mutates `NodeUsage`.
* `klog` output is redirected to `io.Discard` so logging does not affect benchmark results.
* Real NVIDIA device scoring is used instead of `fitMockDevice` to benchmark the actual `Fit` path.
* A separate `newBenchmarkNodes` helper is used because the benchmarks require configurable node/GPU counts.

```release-note
NONE
```

**AI assistance disclosure:**

AI was used to assist with the implementation and review of the benchmark code and PR description. The final changes were reviewed and verified before submission.

## Summary by CodeRabbit

* **New Features**
  * Added node-level device cordoning to prevent new workloads from using selected devices while preserving running workloads.
  * Device scheduling now consistently skips cordoned devices across supported accelerator types and reports clear allocation reasons.
  * Improved device selection with UUID constraints and topology-aware handling.

* **Bug Fixes**
  * Added validation for invalid device counts, memory values, and core percentages.
  * Prevented unsupported or out-of-range requests from being silently adjusted.
  * Improved resource request handling and device allocation reliability.

### DIFF
--- a/pkg/device/amd/device.go
+++ b/pkg/device/amd/device.go
@@ -276,12 +276,18 @@ func (amddevice *AMDDevices) Fit(devices []*device.DeviceUsage, request device.C
 		return false, tmpDevs, "core limit out of range"
 	}
 	isMutex := util.PolicyContains(util.GetGPUSchedulerPolicyByPod(device.GPUSchedulerPolicy, pod), util.GPUSchedulerPolicyMutex)
+	cordoned := device.CordonedDevices(nodeinfo)
 	for i, v := range slices.Backward(devices) {
 		dev := v
 		klog.V(4).InfoS("scoring pod", "pod", klog.KObj(pod), "device", dev.ID, "Memreq", k.Memreq, "MemPercentagereq", k.MemPercentagereq, "Coresreq", k.Coresreq, "Nums", k.Nums, "device index", i)
 		if !dev.Health {
 			reason[common.CardNotHealth]++
 			klog.V(5).InfoS(common.CardNotHealth, "pod", klog.KObj(pod), "device", dev.ID, "health", dev.Health)
+			continue
+		}
+		if _, isCordoned := cordoned[dev.ID]; isCordoned {
+			reason[common.CardCordoned]++
+			klog.V(5).InfoS(common.CardCordoned, "pod", klog.KObj(pod), "device", dev.ID)
 			continue
 		}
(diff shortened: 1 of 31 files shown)


mesutoezdil:
For big changes, it's better to explain your idea in a meeting first

PR #2930, opened 2026-09-01
sync InitDefaultDevices with the chart device config

### LINKED ISSUE #2928: InitDefaultDevices has drifted from the chart's default device config

Sync `InitDefaultDevices()` in `pkg/scheduler/config/config.go` with the default device configuration used by the Helm chart.

The two configurations have drifted apart. The hardcoded test fixture is missing several vendors and device-specific fields, including Ascend `superPod` settings, NVIDIA MIG profiles, and vGPU resource fields.

Update `InitDefaultDevices()` to match the Helm chart fallback configuration and add a test/guard to prevent them from drifting again.

**Why is this needed**:

`InitDefaultDevices()` is only used by tests, so the current tests are running against a device configuration that does not match a default HAMi deployment.

Keeping the test configuration in sync ensures the tests cover the same device settings that users get from the Helm chart.

im-Toqeer-506:
**What type of PR is this?**

/kind feature

Sync `InitDefaultDevices()` in `pkg/scheduler/config/config.go` with the default device configuration in the Helm chart.

The two configurations had drifted apart. The test fixture was missing several vendor sections, Ascend chips, and device-specific fields such as `superPod`, MIG profiles, and resource names.

This meant the `config` and `nodes` tests were using a device configuration that did not match what a default Helm installation provides.

This PR:

* Updates `InitDefaultDevices()` to match the chart's default configuration.
* Adds the missing vendors, chips, and device fields.
* Adds a test guard to detect missing vendor sections if the chart and fixture drift again.

Fixes #2928

**AI assistance disclosure:**
This change was developed with AI assistance. I reviewed the changes, verified the diff, and ran the required checks locally.

**Hardware validation:**
Not applicable. `InitDefaultDevices()` is only used by unit tests and has no production caller. This PR does not change device allocation or isolation logic.

```release-note
NONE
```
**AI assistance disclosure:**

I used an AI assistant to verify the implementation and check that the test cases pass.

## Summary by CodeRabbit

* **New Features**
  * Expanded default accelerator support across Nvidia, Metax, Enflame, Hygon, AMD, Vastai, Biren, and Ascend hardware.
  * Added tuning options, resource mappings, memory settings, runtime configuration, and MIG profile support.
  * Added Ascend 910B4-1 and 910C configurations, including SuperPod support.
  * Improved Ascend 310P3 virtual device configuration.

* **Tests**
  * Added checks to keep default device settings aligned with chart configurations and supported hardware.

### DIFF
--- a/pkg/scheduler/config/device_config_drift_test.go
+++ b/pkg/scheduler/config/device_config_drift_test.go
@@ -17,6 +17,7 @@ limitations under the License.
 package config
 
 import (
+	"fmt"
 	"os"
 	"regexp"
 	"strings"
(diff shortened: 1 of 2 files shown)


pkg/scheduler/config/device_config_drift_test.go
```diff
@@ -0,0 +1,171 @@
...
+			return nil, false
+		}
+	}
+	return current, true
+}
+
+// renderChartFallback returns the device config the chart produces by default:
+// the `{{- else }}` branch with its .Values references substituted and its
+// ConfigMap indentation removed.
+func renderChartFallback(t *testing.T) string {
```
mesutoezdil, line 87: what happens when the chart adds an if?

pkg/scheduler/config/device_config_drift_test.go
```diff
@@ -0,0 +1,171 @@
...
+	// chartFallbackStart and chartFallbackEnd delimit the `{{- else }}` branch,
+	// the one used when a chart user supplies no device-config.content override.
+	chartFallbackStart = "{{- else }}"
+	chartFallbackEnd   = "{{ end }}"
+
+	// chartValueRef matches the only template construct the fallback branch
+	// uses: a .Values lookup with an optional `| default <literal>` fallback.
+	// The branch has no conditionals or loops, which is what makes rendering it
+	// here a substitution rather than a reimplementation of Helm.
+	chartValueRef = regexp.MustCompile(`\{\{\s*\.Values\.([A-Za-z0-9_.]+)\s*(?:\|\s*default\s+(\S+?)\s*)?\}\}`)
```
mesutoezdil, line 52: does this handle toYaml or piped functions?

pkg/scheduler/config/device_config_drift_test.go
im-Toqeer-506: The test fails instead of passing silently: only `{{ .Values.x }}` and `| default` are substituted. An `{{ if }}` remains in the output and breaks YAML parsing.
I confirmed this by wrapping a field in `{{- if … }}{{- end }}`. Both tests fail with `could not find expected ':'`.

pkg/scheduler/config/device_config_drift_test.go
im-Toqeer-506: Only `| default <literal>` is supported, which is all the fallback branch uses today. `| quote`, chained pipes, and `toYaml` are not rendered.
I tried each one, and the test fails every time with either a YAML parse error or a `DeepEqual` mismatch, so the fixture cannot drift silently.

mesutoezdil:

Point 4.

PR #2931, opened 2026-09-01
fix(monitor): count task priorities in a map, not a slice index

mesutoezdil:
`Observe` grew `utSwitchOn[uuid]` one entry at a time until it could be indexed by the container's task priority, and that priority is read from the shared memory region the device plugin mounts into the container read-write, so a single value sizes the slice to the value itself: one container at priority `1<<26` allocates 2800 MiB, and the loop runs per device per container every five seconds. Keying the counts by priority instead makes the allocation proportional to the number of distinct priorities rather than to the largest value, and keeps every priority class distinct so the contention checks compare exactly what they compared before.

### DIFF
--- a/cmd/vGPUmonitor/feedback.go
+++ b/cmd/vGPUmonitor/feedback.go
@@ -35,17 +35,20 @@ var errTemporaryClosed = errors.New("temporary closed")
 //	mtime      uint64
 //}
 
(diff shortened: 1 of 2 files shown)


archlitchi (approved):
/lgtm

PR #2932, opened 2026-09-01
fix(chart): render service labels at the right indent level

mesutoezdil:
Both Service templates emitted the optional label map with four literal spaces of template indentation followed by `indent 4`, so the first rendered key landed at column 8 and setting `scheduler.service.labels` or `devicePlugin.service.labels` made `helm template` and `helm install` fail with a YAML parse error. Switch both to `nindent 4`, matching the `annotations` line directly below them in the same files.

Does this PR introduce a user-facing change?

Setting scheduler.service.labels or devicePlugin.service.labels no longer breaks chart rendering with a YAML parse error.


archlitchi (approved):
/lgtm

PR #2939, opened 2026-09-01
fix(scheduler): prune printedLog when a node is deleted

### LINKED ISSUE #2937: cheduler: printedLog map is never pruned when nodes are deleted
## What happened

`RegisterFromNodeAnnotations` creates a `printedLog` map and passes it to every `register()` call:

```go
printedLog := map[string]bool{}
for {
    ...
    s.register(labelSelector, printedLog)
}
```

`register()` adds a node to this map the first time it logs the node:

```go
if printedLog[val.Name] {
    klog.V(5).InfoS("Node device updated", ...)
} else {
    klog.InfoS("Node device added", ...)
    printedLog[val.Name] = true
}
```

The problem is that entries are never removed.

`onDelNode` already cleans up other per-node state when a node is deleted, but it cannot clean `printedLog` because the map is local to `RegisterFromNodeAnnotations`.

### Why this is a problem

1. `printedLog` keeps growing for the lifetime of the scheduler. Each node name that is ever registered remains in the map.
2. If a node is deleted and later recreated with the same name, the scheduler treats it as already seen and logs `Node device updated` at `V(5)` instead of the expected info-level `Node device added`.

This can make it harder for operators to see that a replacement node has registered.

## Expected behavior

`printedLog` should be cleaned up when a node is deleted, just like the other per-node state.

## How to reproduce

1. Run `hami-scheduler` with the default log verbosity.
2. Register a GPU node and confirm that `Node device added` is logged.
3. Delete the node.
4. Recreate a node with the same name.
5. When it registers again, `Node device added` is not logged. Instead, it is logged as `Node device updated` at `V(5)`.

## Suggested fix

Move `printedLog` into the scheduler state and clean it up when the node is removed:

* Add `printedLog map[string]bool` to the scheduler and initialize it in `NewScheduler()`.
* Remove the `printedLog` parameter from `register()`.
* Delete the node from `s.printedLog` in `cleanupNodeUsage()`, which is already called by `onDelNode`.

This keeps the change limited to `pkg/scheduler/scheduler.go` and reuses the existing per-node cleanup path.

## Testing

Add a regression test covering:

`node added → node deleted → same node name added again`

The test should pass with `-race`.

## Scope

Only `pkg/scheduler/scheduler.go` and `pkg/scheduler/scheduler_test.go` are affected. No API, annotation, or configuration changes are needed.

I'm happy to take this if the approach looks reasonable.

im-Toqeer-506:
**What type of PR is this?**

/kind bug

`printedLog` was a loop-local map in `RegisterFromNodeAnnotations`, so `onDelNode` could not remove entries from it.

This caused two issues:

* The map kept growing for the lifetime of the scheduler, with one entry for every node name seen.
* If a node was deleted and recreated with the same name, it was logged as `Node device updated` at `V(5)` instead of `Node device added` at info level. This meant the replacement node was not visible in default-verbosity logs.

This PR moves `printedLog` into `Scheduler` state and protects it with the existing `s.lock`. Entries are removed in `cleanupNodeUsage`, which already handles per-node cleanup during `onDelNode`.

Fixes #2937

**Validation:**

* Added `Test_register_PrintedLogPrunedOnNodeDelete` covering add → delete → re-add and ensuring other nodes remain in the map.
* `Test_register_NodeCacheConcurrency` already tests concurrent `register()` and `onDelNode()` calls. It passes with `-race`.
* Full suite: `-short --race -count=1` — 38 packages pass.
* `golangci-lint` v2.13.1 — 0 issues.
* `gofmt`, `goimports`, and license checks pass.
* `TestPrepareHostPIDLockParent` still fails in my environment, but the same failure occurs on a clean `upstream/master`, so it is pre-existing and unrelated to this PR.

```release-note
NONE
```
**AI assistance disclosure:** AI was used to assist with the analysis, implementation, and tests. I reviewed the changes and verified the results.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved scheduler logging when nodes are removed and later recreated.
  * Ensured recreated nodes are recognized as new and their device information is logged again.
  * Preserved logging state for nodes that remain available.
* **Tests**
  * Added coverage for node deletion, recreation, and concurrent registration scenarios.

### DIFF
--- a/pkg/scheduler/register_race_test.go
+++ b/pkg/scheduler/register_race_test.go
@@ -104,11 +104,10 @@ func Test_register_NodeCacheConcurrency(t *testing.T) {
 	// indexer.Update never errors for the default store and require/FailNow must not
 	// run outside the test goroutine, so the error is ignored.
 	wg.Go(func() {
-		printed := map[string]bool{}
 		sel := labels.Everything()
 		for v := 1; v <= rounds; v++ {
 			_ = indexer.Update(mkNode(v%2 == 0))
-			s.register(sel, printed)
+			s.register(sel)
 		}
 	})
 
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -74,8 +74,13 @@ type Scheduler struct {
 	leaseLister coordinationv1.LeaseLister
 	//Node Overview
 	overviewstatus map[string]*NodeUsage
-	eventRecorder  record.EventRecorder
-	started        uint32 // 0 = false, 1 = true
+	// printedLog records the nodes whose devices have already been logged at
+	// info level, so a re-registration logs at V(5) instead. It is pruned when
+	// a node is deleted, both to keep it bounded under node churn and so a node
+	// that returns under the same name is logged as newly added again.
+	printedLog    map[string]bool
+	eventRecorder record.EventRecorder
+	started       uint32 // 0 = false, 1 = true
 
 	lock   sync.RWMutex
 	synced atomic.Bool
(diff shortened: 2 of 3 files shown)


Shouren (approved):
/lgtm

PR #2941, opened 2026-09-01
fixing the reject unknown fields in device-config.yml

### LINKED ISSUE #2940: fix(scheduler): reject unknown fields in device-config.yaml
## What happened?

`LoadConfig()` uses non-strict YAML decoding for `device-config.yaml`.

A misspelled field can be silently ignored, allowing the scheduler to start with an unintended default value. This is especially risky for custom device configuration supplied through Helm.

For example:

```yaml
nvidia:
  resourceCountNmae: nvidia.com/gpu
```

The misspelled `resourceCountNmae` field is currently ignored.

## What did you expect to happen?

The scheduler should fail at startup with a clear error when `device-config.yaml` contains:

- An unknown top-level field.
- An unknown nested device field.
- A duplicate YAML key.
- An invalid YAML value type.

Valid existing device configurations must continue to load unchanged.

## Proposed solution

Replace the non-strict YAML decode in `LoadConfig()` with `yaml.UnmarshalStrict`.

Add focused tests that load temporary configuration files through `LoadConfig()` and verify that valid configuration succeeds while unknown and duplicate fields return clear errors.

## Affected components

- `pkg/scheduler/config/config.go`
- `pkg/scheduler/config/config_test.go`

## Acceptance criteria

- [ ] Unknown top-level fields are rejected.
- [ ] Unknown nested backend fields are rejected.
- [ ] Duplicate YAML keys are rejected.
- [ ] Invalid YAML value types return an error.
- [ ] Error messages identify the invalid field or key.
- [ ] Existing valid configuration tests continue to pass.

## Compatibility

Valid configurations are unchanged.

Configurations containing ignored invalid fields will fail at startup instead of silently using default values.

## Additional information

- **Implementation complexity:** Low
- **Breaking change:** No for valid configurations
- **Documentation change:** No
- **Maintainer acceptance probability:** High

asadjan4611:
**What type of PR is this?**

/kind bug

This PR makes device configuration parsing strict.

Previously, an unknown or misspelled field in `device-config.yaml` could be ignored silently. The scheduler could then start with unintended default values.

This change uses strict YAML decoding so invalid configuration fails early with a clear error.

**What is included**:

- Reject unknown top-level configuration fields.
- Reject unknown nested device fields.
- Reject duplicate YAML fields.
- Reject invalid YAML value types.
- Add regression tests for valid and invalid device configurations.

Fixes #2940

Valid existing device configurations keep their current behavior.

Configurations with invalid or previously ignored fields will now fail during scheduler startup instead of silently using defaults.

Yes. Invalid `device-config.yaml` files now fail early with a clear validation error.

**AI assistance disclosure**:

I am using an AI assistance(Codex)  for drafting the PR description.

## Summary by CodeRabbit

* **Bug Fixes**
  * Configuration loading now reports errors for unknown, duplicate, or incorrectly typed YAML fields instead of silently accepting invalid settings.
  * Valid configurations continue to load successfully.
  * Configuration validation now provides more reliable feedback when YAML contains unsupported fields, repeated entries, or values with invalid types.

### DIFF
--- a/pkg/scheduler/config/config.go
+++ b/pkg/scheduler/config/config.go
@@ -489,7 +489,7 @@ func LoadConfig(path string) (*Config, error) {
 	}
 
 	var yamlData Config
-	if err := yaml.Unmarshal(data, &yamlData); err != nil {
+	if err := yaml.UnmarshalStrict(data, &yamlData); err != nil {
 		return nil, err
 	}
 	klog.Info("Successfully read and parsed config file")
--- a/pkg/scheduler/config/config_test.go
+++ b/pkg/scheduler/config/config_test.go
@@ -18,6 +18,8 @@ package config
 
 import (
 	"fmt"
+	"os"
+	"path/filepath"
 	"slices"
 	"testing"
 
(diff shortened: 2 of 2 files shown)


pkg/scheduler/config/config.go
```diff
@@ -489,7 +489,7 @@ func LoadConfig(path string) (*Config, error) {
 	}
 
 	var yamlData Config
-	if err := yaml.Unmarshal(data, &yamlData); err != nil {
+	if err := yaml.UnmarshalStrict(data, &yamlData); err != nil {
```
mesutoezdil, line 492: what breaks on upgrade with old config fields?

mesutoezdil, line 492: did you check this against every existing yaml key?

asadjan4611:
i checked the upgrade impact. Current  v2.10 configs pass strict loading while v2.9 `knownMigGeometries` requires migration to `migProfileAllowlist`. I also updated the multi-backend test to use the strict `LoadConfig` path.

archlitchi (approved):
/lgtm

PR #2902, opened 2026-08-31
feat(ascend): honor deviceCoreScaling in hami-core Fit budget

curry30000:
**What type of PR is this?**

/kind feature

**What this PR does / why we need it**:

Ascend hami-core Fit hardcodes `effectiveTotalCore = 100`. This wires the existing NVIDIA-style `deviceCoreScaling` knob for Ascend hami-core only:

- budget = `round(100 * deviceCoreScaling)`, default `1`
- `Coresreq=100` stays exclusive when the budget is oversold
- a card that already has one 100% occupant rejects later share requests

Helm: `devices.ascend.deviceCoreScaling` (default `1`) → `vnpus.deviceCoreScaling`. Memory is not scaled.

**Which issue(s) this PR fixes**:
Fixes #2901

**Special notes for your reviewer**:

AI assistance: I used Cursor to draft the patch and this description. I reviewed every changed line and can explain the Fit budget and exclusive checks. Commit messages were written by me.

Hardware validation (scheduler extender Fit path):
- Device: 8 × Ascend 910B3
- Driver / npu-smi: 25.5.0
- `deviceCoreScaling=1.5` (budget 150)
- 3 pods with `-core: "30"` plus 1 pod with `-core: "20"` on the same UUID
- Result: all four scheduled (110 ≤ 150). Default budget 100 rejects the fourth (`CardInsufficientCore`).

This is a new PR after #2883 was closed for contribution-process issues.

**Does this PR introduce a user-facing change?**:

Yes. Optional Helm value `devices.ascend.deviceCoreScaling` (default `1`). No behavior change unless set above 1.

## Summary by CodeRabbit

* **Bug Fixes**
  * Core-capacity checks now consistently use the capacity advertised by each device.
  * Improved handling of core oversubscription and exhaustion scenarios.
  * Exclusive device allocations continue to prevent conflicting shared allocations.

* **Tests**
  * Updated coverage for core-capacity limits, oversell behavior, and exclusive allocation conflicts.

PR #2947, opened 2026-09-02
feat(ascend): honor plugin-advertised hami-core percentage budget

### LINKED ISSUE #2946: [Ascend] honor plugin-advertised hami-core percentage budget

Honor a plugin-advertised hami-core percentage budget in Ascend Fit, without treating physical AICore as that budget.

Today hami-core Fit hardcodes `effectiveTotalCore = 100`. That is correct for `Coresreq` (0-100 percentage) when the plugin still registers hardware AICore (20/24/32). It also blocks compute oversell after the device plugin advertises `Devcore = 100 * deviceCoreScaling`.

Proposed behavior (default unchanged for existing deployments):

- hami-core Fit budget = `100` when advertised `Totalcore <= 100` (physical AICore or default percentage)
- hami-core Fit budget = advertised `Totalcore` when the plugin reports `Devcore > 100`
- Scaling itself lives in the device plugin (`vnpus.deviceCoreScaling`), not in HAMi Helm
- Coupled plugin PR: https://github.com/Project-HAMi/ascend-device-plugin/pull/132
- Memory is not scaled
- `Coresreq == 100` stays exclusive even when the budget is above 100
- A later share request must not land on a card that already has one 100% occupant

**Why is this needed**:

NVIDIA already oversells compute via `deviceCoreScaling` in the device plugin, then Fit uses the advertised capacity. Ascend hami-core uses the same 0-100 `-core` scale.

#2902 tried to drop the 100 override and use `dev.Totalcore` directly. That is wrong for Ascend: `Totalcore` is the physical AICore count on existing deployments, so removing the override would change `Coresreq` admission (for example `-core: 30` vs 20 cores). This issue keeps the 100-point budget and only honors an oversold percentage advertised by the coupled plugin.

Validated on 8 × Ascend 910B3, npu-smi 25.5.0, plugin `deviceCoreScaling=1.5` (Devcore 150):
three pods with `-core: "30"` plus one with `-core: "20"` on the same UUID Fit (110 ≤ 150).
Default `100` still rejects the fourth pod. Physical `Totalcore=20` still admits `-core: 30`.

Related: #2902 / #2901 were closed because Fit used physical `Totalcore` as the percentage budget.

curry30000:
**What type of PR is this?**

/kind feature

Follow-up to #2902, coupled with https://github.com/Project-HAMi/ascend-device-plugin/pull/132.

#2902 removed the hami-core `effectiveTotalCore = 100` override and used `dev.Totalcore` directly. That is incorrect on Ascend: `Totalcore` is the physical AICore count (8/20/24/30) on existing deployments, not a 0-100 percentage. Dropping the override would change `Coresreq` admission.

This PR keeps the 100-point percentage budget and only honors an oversold percentage advertised by the plugin:

- hami-core Fit budget = `100` when advertised `Totalcore <= 100` (physical AICore, or the default `Devcore=100`)
- hami-core Fit budget = advertised `Totalcore` when the plugin reports `Devcore > 100`
- scaling config is **not** added to HAMi Helm; it lives in the plugin as `hamiVnpuCore.deviceCoreScaling` (see the coupled PR)
- `Coresreq=100` stays exclusive for hami-core even when the budget is oversold
- a card already held by a single full-core occupant rejects every later request, whatever mode that request is in

Default deployments are unchanged: `-core: "30"` still fits against a 100-point budget even if `Totalcore` is 20.

Fixes #2946

AI assistance: I used Cursor to draft the patch and this description. I reviewed every changed line and can explain the Fit budget and exclusive checks. Commit messages were written by me.

This is a new PR after #2902 was closed. The close reasons are addressed as follows:

1. `dev.Totalcore` is physical AICore — Fit no longer uses it as the hami-core percentage when it is `<= 100`.
2. Removing the 100 override would change existing admission — the 100-point budget stays the default.
3. HAMi Helm `deviceCoreScaling` wiring is intentionally absent — the knob is in ascend-device-plugin#132.

The second commit fixes a bug I found while re-reviewing this against the plugin PR. The exclusivity guard was scoped to the requesting pod's mode (`isHAMiCore && dev.Used == 1 && ...`), but `Fit` only rejects hami-core pods on non-hami-core nodes, not the reverse. A legacy vNPU pod therefore skipped the guard and could land on a card already held exclusively by a 100%-core occupant: under oversell neither the capacity check (`150 - 100 = 50` still passes) nor `CardComputeUnitsExhausted` (`100 != 150`) catches that any more. Both did catch it before oversell existed, which is why it only surfaces with the feature on.

`dev.Used == 1` has to stay in the guard. With two or more tenants the used cores are a sum of sub-100 shares, and rejecting those would break exactly the oversell case this PR adds. There is a test for that direction too.

Hardware validation (scheduler extender Fit path):
- Device: 8 × Ascend 910B3
- Driver / npu-smi: 25.5.0
- plugin `deviceCoreScaling=1.5` (advertised Devcore 150)
- 3 pods with `-core: "30"` plus 1 pod with `-core: "20"` on the same UUID
- Result: all four scheduled (110 ≤ 150). Default budget 100 rejects the fourth (`CardInsufficientCore`).
- Physical `Totalcore=20` still admits `-core: "30"`.

Unit validation: `go test ./pkg/device/...` passes across all 15 device packages, with no reason-string changes for pre-existing cases. I checked the new regression test against the previous commit, where it fails with an empty reason string, i.e. the pod is admitted.

No HAMi Helm change. Behavior changes only when the coupled plugin advertises `Devcore > 100`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Ascend device core allocation in Hami-core mode.
  * Accurately honors advertised core capacities above the standard 100-core budget.
  * Prevents allocations that exceed the available device capacity.
  * Strengthened exclusive allocation handling when a device is fully occupied.
  * Allows valid shared allocations on devices with sufficient advertised capacity while rejecting conflicting requests.

### DIFF
--- a/pkg/device/ascend/device.go
+++ b/pkg/device/ascend/device.go
@@ -430,19 +430,25 @@ func (dev *Devices) GetResourceNames() device.ResourceNames {
 	}
 }
 
-// hamiCoreFitBudget is the 0-100 percentage capacity used by hami-core Fit.
-// Coresreq is a percentage, not physical AI cores. Existing plugins may still
-// register Devcore as the hardware AICore count (20/24/32). Keep today's 100-point
-// budget in that case so admission does not change.
+// hamiCorePercentBase is the fixed 0-100 percentage scale hami-core Coresreq
+// is expressed in.
+const hamiCorePercentBase = 100
+
+// hamiCoreFitBudget is the percentage capacity used by hami-core Fit.
 //
-// Coupled plugin change (https://github.com/Project-HAMi/ascend-device-plugin/pull/132)
-// advertises Devcore = round(100 * deviceCoreScaling) for hami-core, which is >= 100.
-// Honor that advertised percentage when it is larger than 100.
+// Coresreq is a percentage, not physical AI cores, so dev.Totalcore cannot be
+// used directly: a plugin that has not enabled hami-core registers Devcore as
+// the hardware AICore count (8/20/24/30). Keep the percentage base for any
+// advertised value at or below it so admission does not change.
+//
+// The coupled plugin change (Project-HAMi/ascend-device-plugin#132) advertises
+// Devcore = round(100 * deviceCoreScaling) for hami-core and never reports less
+// than the base, so any higher value is an intentional oversell budget.
 func hamiCoreFitBudget(advertisedTotalcore int32) int32 {
-	if advertisedTotalcore > 100 {
+	if advertisedTotalcore > hamiCorePercentBase {
 		return advertisedTotalcore
 	}
-	return 100
+	return hamiCorePercentBase
 }
 
 func (npu *Devices) Fit(devices []*device.DeviceUsage, request device.ContainerDeviceRequest, pod *corev1.Pod, nodeInfo *device.NodeInfo, allocated *device.PodDevices) (bool, map[string]device.ContainerDevices, string) {
@@ -545,12 +551,15 @@ func (npu *Devices) Fit(devices []*device.DeviceUsage, request device.ContainerD
 			klog.V(5).InfoS(common.CardInsufficientMemory, "pod", klog.KObj(pod), "device", dev.ID, "device index", i, "device total memory", dev.Totalmem, "device used memory", dev.Usedmem, "request memory", memreq)
 			continue
 		}
-		// hami-core Coresreq is a 0-100 percentage. Do not treat physical AICore
-		// (dev.Totalcore) as that budget. Default remains 100; honor plugin
-		// oversell only when advertised Devcore > 100.
+		// hami-core Coresreq is a percentage. Do not treat physical AICore
+		// (dev.Totalcore) as that budget; honor plugin oversell only when the
+		// advertised Devcore is above the percentage base.
 		effectiveTotalCore := dev.Totalcore
 		if isHAMiCore {
 			effectiveTotalCore = hamiCoreFitBudget(dev.Totalcore)
+			if effectiveTotalCore != dev.Totalcore {
+				klog.V(5).InfoS("hami-core budget clamped to the percentage base", "pod", klog.KObj(pod), "device", dev.ID, "advertised total core", dev.Totalcore, "budget", effectiveTotalCore)
+			}
 		}
 
 		if effectiveTotalCore-dev.Usedcores < k.Coresreq {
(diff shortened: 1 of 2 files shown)


curry30000:
Pushed a follow-up commit (`0ad7f97`). While re-reviewing this together with the coupled plugin PR I found the exclusivity guard in the first commit is not tight enough once oversell is actually on.

### Full-card exclusivity is lost when the budget is oversold

The guard was scoped to the requesting pod's mode:

```go
if isHAMiCore && dev.Used == 1 && dev.Usedcores >= 100 {
```

`Fit` only rejects hami-core pods on non-hami-core nodes; the reverse direction is not filtered. A legacy vNPU pod (no `hami.io/vnpu-mode` annotation) therefore skips this guard and can land on a card already held exclusively by a 100%-core occupant.

The two checks that used to catch that both stop working under oversell. With `deviceCoreScaling: 1.5` the budget is 150, so `effectiveTotalCore - Usedcores = 50` passes the capacity check, and `CardComputeUnitsExhausted` never fires because `Usedcores (100) != effectiveTotalCore (150)`. Without oversell `Totalcore` is 100 and both still reject, which is why this only surfaces with the feature enabled.

The fix drops the mode condition. `dev.Used == 1` has to stay: with `Used >= 2` the used cores are the sum of several sub-100 tenants, and rejecting those would break exactly the oversell case this PR adds.

I verified the new regression test actually catches it — reverting just the guard to the previous form makes `advertised 150 exclusive occupant rejects a legacy vnpu pod` fail with an empty reason string, i.e. the pod is admitted.

### Also in this commit

- `hamiCorePercentBase` replaces the repeated literal `100`.
- V(5) log when an advertised `Devcore` gets clamped back to the base. That means the plugin and the scheduler disagree about hami-core, and it used to be completely silent.
- An extra case covering two sub-100 tenants (60+50) still fitting a 150 budget, so the exclusivity fix cannot silently disable oversell.

### Validation

`go test ./pkg/device/...` is green across all 15 device packages, and no reason strings changed for pre-existing cases.

### Plugin side

Project-HAMi/ascend-device-plugin#132 also got a fix in the same round: `deviceCoreScaling` turned out to be unreachable from Helm values, and ratios below 1 were silently dropped. Details in that PR. This PR only consumes the advertised `Devcore`, so it is still safe to review independently, but the feature is only usable with both.

curry30000:
Pushed a third commit after re-auditing the branch end to end.

The exclusivity guard from the previous commit sat before the `CardComputeUnitsExhausted` check, which changed the rejection reason for a case that has nothing to do with oversell. On a full card with the default budget (`Totalcore` 100, `Usedcores` 100, one occupant) a `-core: 0` request used to be rejected as `CardComputeUnitsExhausted`; it started coming back as `ExclusiveDeviceAllocateConflict`. The admission decision was the same either way, but the reason shown to users changed, which contradicts the "default deployments are unchanged" claim in the description.

I measured this against master rather than reasoning about it: master returns `fit=false reason="1/1 CardComputeUnitsExhausted"`, and the previous commit returned `1/1 ExclusiveDeviceAllocateConflict`.

Moving the guard after that check restores master's reason and leaves oversell untouched, because with an advertised budget of 150 the occupant's 100 never equals `effectiveTotalCore`, so `CardComputeUnitsExhausted` cannot fire and the guard is still what rejects. There is now a test pinning the non-oversold reason so this cannot drift again.

`go test ./pkg/device/...` and `golangci-lint run ./pkg/device/ascend/...` are both clean.

mesutoezdil:
Point 4

curry30000:
> This is being closed because it does not comply with the contribution guidelines. Point 4

Sorry for the above AI‑sounding long comments . I will write my own replies for future reviews. Could you reopen this PR, or would you prefer me to create a new one instead?

PR #2952, opened 2026-09-02
feat(ascend): honor plugin-advertised hami-core percentage budget

### LINKED ISSUE #2951: [Ascend] honor plugin-advertised hami-core percentage budget
Ascend hami-core `Coresreq` is a 0-100 percentage. Existing plugin registrations advertise physical AICore as `Totalcore` (8/20/24/30), so Fit should not treat that number as the percentage budget.

Proposal:
- keep a 100-point hami-core Fit budget when advertised `Totalcore <= 100`
- honor advertised `Totalcore` only when the plugin reports `Devcore > 100` (compute oversell)

Plugin side that advertises the scaled `Devcore`: https://github.com/Project-HAMi/ascend-device-plugin/pull/132 (`hamiVnpuCore.deviceCoreScaling`).

curry30000:
**What type of PR is this?**

/kind feature

Ascend hami-core `Coresreq` is a 0-100 percentage. On current nodes the plugin advertises physical AICore as `Totalcore` (8/20/24/30), which is not that percentage.

This change keeps a 100-point Fit budget by default. It only uses the advertised `Totalcore` when the device plugin reports `Devcore > 100`.

It is meant to be used with https://github.com/Project-HAMi/ascend-device-plugin/pull/132. That plugin PR registers `Devcore = round(100 * deviceCoreScaling)`. This scheduler PR consumes that advertised value. The scaling knob stays on the plugin (`hamiVnpuCore.deviceCoreScaling`). HAMi Helm is not changed.

- advertised `Totalcore <= 100`: budget stays 100
- advertised `Totalcore > 100`: budget is the advertised value
- `Coresreq=100` stays exclusive when the budget is oversold
- a card already held by one full-core occupant rejects later requests, including legacy vNPU
- `-core: "30"` still fits when physical `Totalcore` is 20

Fixes #2951

AI assistance: Cursor helped draft the patch. I reviewed the Fit budget and exclusive checks. I wrote the commit messages.

Hardware (scheduler extender Fit):
- 8 x Ascend 910B3, npu-smi / driver 25.5.0
- plugin `deviceCoreScaling=1.5` (advertised Devcore 150)
- 3 pods with `-core: "30"` and 1 with `-core: "20"` on the same UUID: all scheduled (110 <= 150)
- default budget 100 rejects the fourth (`CardInsufficientCore`)
- physical `Totalcore=20` still admits `-core: "30"`

`go test ./pkg/device/...` and `golangci-lint run ./pkg/device/ascend/...` pass.

No. Behavior changes only when the coupled plugin advertises `Devcore > 100`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Ascend device allocation for hami-core workloads by correctly normalizing compute capacity.
  * Preserved advertised oversell capacity when allocating devices.
  * Corrected admission decisions for full-core requests and exclusive allocations.
  * Improved memory-capacity checks for workloads using oversold Ascend resources.
  * Prevented allocation conflicts when a card is already occupied by a full-core workload.

### DIFF
--- a/pkg/device/ascend/device.go
+++ b/pkg/device/ascend/device.go
@@ -550,7 +550,7 @@ func hamiCoreOthersOnDevice(dev *device.DeviceUsage, pod *corev1.Pod) int32 {
 		if podInfo == nil || podInfo.Pod == nil {
 			continue
 		}
-		if pod != nil && podInfo.Pod.UID != "" && pod.UID != "" && podInfo.Pod.UID == pod.UID {
+		if pod != nil && podInfo.UID != "" && pod.UID != "" && podInfo.UID == pod.UID {
 			continue
 		}
 		for _, podSingle := range podInfo.Devices {
--- a/pkg/device/ascend/device.go
+++ b/pkg/device/ascend/device.go
@@ -439,6 +439,21 @@ func (dev *Devices) GetResourceNames() device.ResourceNames {
 	}
 }
 
+// hamiCoreFitBudget is the 0-100 percentage capacity used by hami-core Fit.
+// Coresreq is a percentage, not physical AI cores. Existing plugins may still
+// register Devcore as the hardware AICore count (20/24/32). Keep today's 100-point
+// budget in that case so admission does not change.
+//
+// Coupled plugin change (https://github.com/Project-HAMi/ascend-device-plugin/pull/132)
+// advertises Devcore = round(100 * deviceCoreScaling) for hami-core, which is >= 100.
+// Honor that advertised percentage when it is larger than 100.
+func hamiCoreFitBudget(advertisedTotalcore int32) int32 {
+	if advertisedTotalcore > 100 {
+		return advertisedTotalcore
+	}
+	return 100
+}
+
 func (npu *Devices) Fit(devices []*device.DeviceUsage, request device.ContainerDeviceRequest, pod *corev1.Pod, nodeInfo *device.NodeInfo, allocated *device.PodDevices) (bool, map[string]device.ContainerDevices, string) {
 	k := request
 	originReq := k.Nums
@@ -545,10 +560,12 @@ func (npu *Devices) Fit(devices []*device.DeviceUsage, request device.ContainerD
 			klog.V(5).InfoS(common.CardInsufficientMemory, "pod", klog.KObj(pod), "device", dev.ID, "device index", i, "device total memory", dev.Totalmem, "device used memory", dev.Usedmem, "request memory", memreq)
 			continue
 		}
-		// Set dev.Totalcore to 100 if vnpuMode is hami-core
+		// hami-core Coresreq is a 0-100 percentage. Do not treat physical AICore
+		// (dev.Totalcore) as that budget. Default remains 100; honor plugin
+		// oversell only when advertised Devcore > 100.
 		effectiveTotalCore := dev.Totalcore
 		if isHAMiCore {
-			effectiveTotalCore = 100
+			effectiveTotalCore = hamiCoreFitBudget(dev.Totalcore)
 		}
 
 		if effectiveTotalCore-dev.Usedcores < k.Coresreq {
@@ -556,10 +573,16 @@ func (npu *Devices) Fit(devices []*device.DeviceUsage, request device.ContainerD
 			klog.V(5).InfoS(common.CardInsufficientCore, "pod", klog.KObj(pod), "device", dev.ID, "device index", i, "device total core", effectiveTotalCore, "device used core", dev.Usedcores, "request cores", k.Coresreq)
 			continue
 		}
-		// Coresreq=100 indicates it want this card exclusively
-		if effectiveTotalCore == 100 && k.Coresreq == 100 && dev.Used > 0 {
+		// Coresreq=100 stays exclusive for hami-core even when the plugin
+		// advertises a budget above 100 (deviceCoreScaling > 1).
+		if k.Coresreq == 100 && dev.Used > 0 && (isHAMiCore || effectiveTotalCore == 100) {
 			reason[common.ExclusiveDeviceAllocateConflict]++
 			klog.V(5).InfoS(common.ExclusiveDeviceAllocateConflict, "pod", klog.KObj(pod), "device", dev.ID, "device index", i, "used", dev.Used)
+			continue
+		}
+		if isHAMiCore && dev.Used == 1 && dev.Usedcores >= 100 {
+			reason[common.ExclusiveDeviceAllocateConflict]++
+			klog.V(5).InfoS(common.ExclusiveDeviceAllocateConflict, "pod", klog.KObj(pod), "device", dev.ID, "device index", i, "used", dev.Used, "usedcores", dev.Usedcores)
 			continue
 		}
 		// You can't allocate core=0 job to an already full GPU
--- a/pkg/device/ascend/device.go
+++ b/pkg/device/ascend/device.go
@@ -460,6 +460,39 @@ func hamiCoreFitBudget(advertisedTotalcore int32) int32 {
 	return hamiCorePercentBase
 }
 
+// hamiCoreExclusiveOccupant reports whether a single Pod already holds the
+// whole percentage base on dev.
+//
+// dev.Used cannot answer that. CollapseInitContainerUsage takes the peak core
+// usage and the peak slot count independently, so a Pod whose init container
+// reserves the whole card and whose app containers then run alongside is
+// recorded as Usedcores at the base with two slots. Reading dev.Used == 1 would
+// miss the reservation and let an oversold budget admit a second tenant onto a
+// card that is still exclusively held.
+//
+// Per-Pod entries are collapsed, so a Pod spreading the base across several of
+// its own containers counts as an occupant too. Such a card is full at the base
+// budget without oversell anyway, so this only declines to oversell it. Two
+// Pods holding half the card each stay below the base individually and keep
+// sharing an oversold card.
+func hamiCoreExclusiveOccupant(dev *device.DeviceUsage) bool {
+	for _, podInfo := range dev.PodInfos {
+		if podInfo == nil {
+			continue
+		}
+		for _, podSingle := range podInfo.Devices {
+			for _, ctrDevs := range podSingle {
+				for _, ctrDev := range ctrDevs {
+					if ctrDev.UUID == dev.ID && ctrDev.Usedcores >= hamiCorePercentBase {
+						return true
+					}
+				}
+			}
+		}
+	}
+	return false
+}
+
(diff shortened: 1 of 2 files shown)


curry30000:
@DSFans2014 could you take a look when you have time?

DSFans2014 (commented):
I am not sure whether `hami-vnpu-core` currently supports core(`NPU_PRIORITY`) oversubscription. @archlitchi

DSFans2014:
please resolve the conflicts

pkg/device/ascend/device.go
```diff
@@ -553,6 +580,17 @@ func (npu *Devices) Fit(devices []*device.DeviceUsage, request device.ContainerD
 			klog.V(5).InfoS(common.CardComputeUnitsExhausted, "pod", klog.KObj(pod), "device", dev.ID, "device index", i)
 			continue
 		}
+		// A card held by a single full-core occupant rejects every later request.
+		// Oversell lifts the budget above that occupant's share, so the check above
+		// no longer recognizes the card as full. Running after it keeps the
+		// non-oversold rejection reason as CardComputeUnitsExhausted. The requesting
+		// pod's own mode is irrelevant: only hami-core pods on non-hami-core nodes
+		// are filtered out, so a legacy vNPU pod still reaches an oversold card.
+		if dev.Used == 1 && dev.Usedcores >= hamiCorePercentBase {
```
xrwang8: `Used == 1` misses the init-container case. A Pod can have an init container requesting 100 and two app containers requesting 30 each on the same card. `CollapseInitContainerUsage` takes the peak of each resource separately, so it records `Usedcores=100, Slots=2`. That makes `dev.Used` equal to 2 even before the init container finishes.

With a budget of 150, another Pod requesting 50 passes both the capacity check and this guard. I checked this with a test using `CollapseInitContainerUsage` and `Fit`: `be3d1848` rejects it with `CardInsufficientCore`, while this branch accepts it.

Please cover this case and preserve the exclusive reservation across init/app phases. Dropping `Used == 1` alone would block normal shared workloads once their combined usage reaches 100, so the aggregate counters aren't enough to tell these cases apart.

xrwang8 (changes requested):
The init-container case below needs to be fixed before merging.

pkg/device/ascend/device.go
curry30000: You're right, and the trace checks out: CollapseInitContainerUsage takes the peak cores and the peak slots independently, so the 100 / 30 / 30 pod lands as Usedcores=100, Slots=2 and the scheduler rebuilds it as Used=2. The guard never fires.

Instead of adding a field to carry exclusivity, I kept it local to the ascend path and read the per-pod collapsed entries already hanging off the device: if a single pod holds the full base on that card, oversell is off for it. Two pods at 50 each stay below the base individually, so they keep sharing — the case you warned about removing Used == 1 would break.

curry30000:
> The init-container case below needs to be fixed before merging.

@xrwang8 I've updated the PR based on your comments. Please take another look when you have a chance. Thanks.

pkg/device/ascend/device.go
```diff
@@ -545,19 +599,28 @@ func (npu *Devices) Fit(devices []*device.DeviceUsage, request device.ContainerD
...
 			continue
 		}
-		// Set dev.Totalcore to 100 if vnpuMode is hami-core
+		// hami-core Coresreq is a percentage. Do not treat physical AICore
+		// (dev.Totalcore) as that budget; honor plugin oversell only when the
+		// advertised Devcore is above the percentage base.
 		effectiveTotalCore := dev.Totalcore
 		if isHAMiCore {
-			effectiveTotalCore = 100
+			effectiveTotalCore = hamiCoreFitBudget(dev.Totalcore)
```
DSFans2014: It's better to check/modify `devcore` in `GetNodeDevices`

curry30000: > It's better to check/modify `devcore` in `GetNodeDevices`

@DSFans2014 Done — Devcore is now normalized in GetNodeDevices, and Fit uses the stored Totalcore. Please take another look when you have a chance. Thanks.

pkg/device/ascend/device.go
```diff
@@ -439,6 +452,70 @@ func (dev *Devices) GetResourceNames() device.ResourceNames {
...
 
+// hamiCorePercentBase is the fixed 0-100 percentage scale hami-core Coresreq
+// is expressed in.
+const hamiCorePercentBase = 100
+
+func (dev *Devices) nodeSupportsHamiCore(n *corev1.Node) bool {
+	supported := dev.hamiVnpuCore
+	if n != nil && n.Annotations != nil {
+		if val, ok := n.Annotations[VNPUNodeSelectorAnnotation]; ok {
+			supported = val == "true"
```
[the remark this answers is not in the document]
curry30000: `hami-vnpu-core` is the existing node annotation used by both this scheduler and ascend-device-plugin; this PR only reads it. Renaming to `hami.io/...` would be a breaking cross-repo change and is out of scope here.

DSFans2014 (commented):
/lgtm
@archlitchi @Shouren please trigger the workflow

pkg/device/ascend/device.go
```diff
@@ -545,29 +620,45 @@ func (npu *Devices) Fit(devices []*device.DeviceUsage, request device.ContainerD
...
+		// A card whose full-core occupancy belongs to one pod rejects every later
+		// request. Oversell lifts the budget above that occupant's share, so the
+		// check above no longer recognizes the card as full. Running after it keeps
+		// the non-oversold rejection reason as CardComputeUnitsExhausted. The
+		// requesting pod's own mode is irrelevant: only hami-core pods on
+		// non-hami-core nodes are filtered out, so a legacy vNPU pod still reaches
+		// an oversold card. dev.Used == 1 covers usage rebuilt without pod detail;
+		// hamiCoreExclusiveOccupant covers the multi-slot collapsed entry an
+		// init-container reservation produces.
+		if dev.Usedcores >= hamiCorePercentBase && (dev.Used == 1 || hamiCoreExclusiveOccupant(dev)) {
```
Shouren, line 717: @curry30000 The exclusivity check makes scheduling depend on Pod arrival order. With a 150-core budget, Pod A has two containers requesting 50 each, and Pod B requests 50:

- A first, then B: B is rejected because A’s collapsed usage is 100.
- B first, then A: A is admitted because each container requests only 50, and A is not yet in `dev.PodInfos`.

Could we apply the same exclusivity rule to incoming and existing Pods so both orders produce a consistent result?

curry30000: > @curry30000 The exclusivity check makes scheduling depend on Pod arrival order. With a 150-core budget, Pod A has two containers requesting 50 each, and Pod B requests 50:
> 
> * A first, then B: B is rejected because A’s collapsed usage is 100.
> * B first, then A: A is admitted because each container requests only 50, and A is not yet in `dev.PodInfos`.
> 
> Could we apply the same exclusivity rule to incoming and existing Pods so both orders produce a consistent result?

@Shouren you're right, the extra occupant check is asymmetric.

It treats a collapsed existing pod (two 50s → 100, or init 100 + app slots) as exclusive, but Fit still sees an incoming pod per container. That's why A-then-B and B-then-A disagree.

There are two consistent options:

1. Match NVIDIA. Exclusive is only `Totalcore == 100 && Coresreq == 100 && Used > 0`. Once the plugin advertises 150, both orders admit A+B (100+50 <= 150). The init-container case xrwang8 raised (Used=2, Usedcores=100, later 50) is admitted too.

2. Keep that occupant exclusive, and apply it to the incoming pod as well (pod-total 100 on the card cannot share). Both orders then reject A+B, which is stricter than GPU.

Which do you prefer?

Shouren: @curry30000 I prefer to the second solution.

curry30000: > @curry30000 I prefer to the second solution.

@Shouren thanks，I went with option 2.

A pod that reaches 100 on one card no longer shares an oversold card, so both arrival orders reject A+B. Two 50-core tenants can still share. I also made sure in-flight usage from the same pod (sidecar / earlier containers) is not treated as another tenant.

Pushed in 69ae2af and 7ff376b. Tests for the ascend device path and scheduler are green. Please take another look when you have a chance.

pkg/device/ascend/device.go
```diff
@@ -490,6 +503,130 @@ func (dev *Devices) GetResourceNames() device.ResourceNames {
...
+func incomingHamiCoreOnDevice(devID string, allocated *device.PodDevices, tmpDevs map[string]device.ContainerDevices) int32 {
+	var sum int32
+	add := func(devs []device.ContainerDevice) {
+		for _, d := range devs {
+			if d.UUID == devID {
+				sum += d.Usedcores
+			}
+		}
+	}
+	if allocated != nil {
```
Shouren, line 601: @curry30000 Could we calculate concurrent per-Pod usage here instead of summing all allocation records? During init allocation, `allocated` includes earlier ordinary init containers that run sequentially; during app allocation, it excludes sidecars that are still running. This can falsely reject sequential init containers or allow a sidecar + app totaling 100 to share with another Pod. Please account for these lifecycle differences and add scheduler-level regression tests for both cases.

pkg/device/ascend/device.go
```diff
@@ -596,29 +731,54 @@ func (npu *Devices) Fit(devices []*device.DeviceUsage, request device.ContainerD
...
+		if dev.Usedcores >= hamiCorePercentBase && (dev.Used == 1 || hamiCoreExclusiveOccupant(dev)) {
+			reason[common.ExclusiveDeviceAllocateConflict]++
+			klog.V(5).InfoS(common.ExclusiveDeviceAllocateConflict, "pod", klog.KObj(pod), "device", dev.ID, "device index", i, "used", dev.Used, "usedcores", dev.Usedcores)
+			continue
+		}
+		// Incoming pods use the same occupant rule. Fit is per container, so a
+		// pod that reaches the base across several of its own requests must
+		// not share an oversold card once another tenant is already there.
+		if (isHAMiCore || nodeSupportHamiCore || dev.Totalcore >= hamiCorePercentBase) &&
+			hamiCoreIncomingExclusive(dev, pod, incomingHamiCoreOnDevice(dev.ID, allocated, tmpDevs), k.Coresreq) {
```
Shouren: @curry30000 The app-phase `allocated` map does not include running sidecars: their allocations remain in `initAllocs` until all app containers have been fitted. With a 150-core budget and another Pod using 50, an incoming Pod with a 50-core sidecar and a 50-core app is incorrectly admitted, despite reaching the 100-core exclusivity threshold. Could we include running sidecar usage in this check and add a regression test?

curry30000: @Shouren yeah that was wrong — I was summing everything in allocated, so sequential inits looked concurrent, and during the app pass the sidecar wasn't even in the map (still sitting in initAllocs).

Switched it to live occupancy only (ordinary init already gone, sidecar stays), and the scheduler now puts those init/sidecar rows into allocated before fitting app containers. Added scheduler tests for both: sequential inits next to a 50-core tenant still go through, sidecar 50 + app 50 next to a tenant gets rejected.

Could you take another look when you have a minute?

Shouren (approved):
/lgtm

PR #2953, opened 2026-09-02
fix(ascend): do not overwrite ASCEND_VISIBLE_DEVICES for Ascend containers

### LINKED ISSUE #2954: Extra ASCEND_VISIBLE_DEVICES env injection supress mounting npu device
**What happened**:

When a container requests one Ascend resource (e.g. `huawei.com/Ascend910B4: 1`), the requested instance works fine, but the other 6 instances each see "no Ascend resource requested" and each injects an empty
`ASCEND_VISIBLE_DEVICES` into the container env. Inside the container `ASCEND_VNPU_SPECS` is set correctly (the slice was allocated) but `ASCEND_VISIBLE_DEVICES` ends up empty — ascend-docker-runtime reads it
last-wins, so the empty entries win and no `/dev/davinci*` gets mounted. The pod still schedules and runs, it just has no NPU.

npu-smi info:
Image

pod resources:
Image

after webhook mutatation
Image

container has empty `ASCEND_VISIBLE_DEVICES` env and davinci* device not mounted
Image

inspect container env list:

Image

A container that requests an Ascend resource keeps the device plugin's real `ASCEND_VISIBLE_DEVICES` value and gets its devices mounted.

1. Create a pod whose container requests a vNPU slice, e.g.   `huawei.com/Ascend910B4: 1` + `huawei.com/Ascend910B4-memory: 8192`   (matches the `vir05_1c_8g` template).
2. The pod schedules and runs, but:
   - the pod spec contains 6 duplicate empty `ASCEND_VISIBLE_DEVICES` entries  (one per non-requested chip instance);
   - inside the container `ASCEND_VNPU_SPECS=vir05_1c_8g` (the plugin did allocate a slice) while `ASCEND_VISIBLE_DEVICES` is empty;
   - no `/dev/davinci*` is mounted.

The `overwriteEnv` injection should only apply to containers that request no Ascend resource at all; containers that request a chip must not be touched. I have a fix validated on real 910B hardware and will send a PR.

**Environment**:

- HAMi version: current master
- Kubernetes version: v1.33.4 (k3s)
- Accelerator vendor and model: Huawei Ascend 910B4
- Container runtime and version: containerd 2.0.5
- Operating system and kernel version: Linux arm64

peachest:
**What type of PR is this?**

/kind bug

Fixes a bug in the ascend `overwriteEnv` injection (introduced in #1738):
when multiple Ascend chip types are registered, a container that requests one of them still gets `ASCEND_VISIBLE_DEVICES` clobbered by empty values injected by the *other* chip instances, so ascend-docker-runtime mounts no device at all.

Fixes #2954

Each registered Ascend chip type has its own `Device` instance, and the scheduler webhook calls `MutateAdmission` once per instance. #1738 modeled the injection on the nvidia implementation, which has a single instance, so the per-instance aspect was missed: for a container that requests one chip, the other N-1 instances each take the `!ok` branch and each append an empty `ASCEND_VISIBLE_DEVICES` to `container.Env`.

Why that breaks mounting, based on #1738 and re-validation on a real Ascend
cluster — when Kubernetes starts a container, env entries are appended to
`Config.Env` in this order:

- variables from the original image;
- variables injected by the device plugin via kubelet;
- variables defined in `pod.spec.containers[].env`.

ascend-docker-runtime then interprets them as follows:

- `ASCEND_VISIBLE_DEVICES`: parsed from the end backwards — a later value  overrides an earlier one;
- all other variables (including `ASCEND_VNPU_SPECS`): parsed from the beginning — values defined in the image win.

So the extra empty `ASCEND_VISIBLE_DEVICES` entries from the other Ascend instances override the device plugin's real value: the pod still schedules, but ascend-docker-runtime does not actually mount the Ascend devices.

The fix (two guards in the `!ok` branch):

1. skip the injection when the container requests *any* registered Ascend resource (limits or requests) — the clearing value is only meant for containers that request nothing;
2. make the injection idempotent: skip it when the last `ASCEND_VISIBLE_DEVICES` entry already equals the value we would inject, so even a multi-instance loop appends at most one entry.

Containers that request no Ascend resource still get exactly one empty `ASCEND_VISIBLE_DEVICES` — that is the intended `overwriteEnv` behavior, unchanged by this PR.

Hardware validation:

- Huawei Ascend 910B4 (single physical card), ascend-docker-runtime, k3s  v1.33.4, containerd 2.0.5, 7 chip types registered with `overwriteEnv: true`.
- Before: a pod requesting `huawei.com/Ascend910B4: 1` + `Ascend910B4-memory: 8192` (vNPU slice `vir05_1c_8g`) schedules and runs, pod spec carries 6 empty `ASCEND_VISIBLE_DEVICES` entries, container env  shows `ASCEND_VNPU_SPECS=vir05_1c_8g` but empty  `ASCEND_VISIBLE_DEVICES`, and no `/dev/davinci*` is mounted.
- After: same pod gets `ASCEND_VISIBLE_DEVICES=<phyid>` (single value,  no duplicates), `ASCEND_VNPU_SPECS=vir05_1c_8g`, and the assigned  `/dev/davinci*` is mounted.

image

image

Yes — with multiple Ascend chip types registered and `overwriteEnv: true`, containers that request an Ascend resource now correctly keep the device plugin's `ASCEND_VISIBLE_DEVICES` value and get their devices mounted.

---

AI assistance disclosure: this fix was developed with AI assistance for code editing and this description's English wording. The root-cause analysis, reproduction, and hardware validation on a real 910B cluster are my own work.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Ascend device environment handling during admission.
  * Prevented empty `ASCEND_VISIBLE_DEVICES` values from being injected when a container requests Ascend resources.
  * Ensured non-Ascend containers receive the expected empty value when overwrite is enabled.
  * Corrected handling of repeated environment entries and values supplied through references.

### DIFF
--- a/pkg/device/ascend/device.go
+++ b/pkg/device/ascend/device.go
@@ -49,12 +49,13 @@ const (
 )
 
 type Devices struct {
-	config           VNPUConfig
-	nodeRegisterAnno string
-	useUUIDAnno      string
-	noUseUUIDAnno    string
-	handshakeAnno    string
-	hamiVnpuCore     bool
+	config                 VNPUConfig
+	nodeRegisterAnno       string
+	useUUIDAnno            string
+	noUseUUIDAnno          string
+	handshakeAnno          string
+	hamiVnpuCore           bool
+	allAscendResourceNames []corev1.ResourceName
 }
 
 type RuntimeInfo struct {
@@ -83,15 +84,20 @@ func InitDevices(vnpus VNPUs) []*Devices {
 	if !enableAscend {
 		return devs
 	}
+	allAscendResourceNames := make([]corev1.ResourceName, 0, len(vnpus.Configs))
+	for _, vnpu := range vnpus.Configs {
+		allAscendResourceNames = append(allAscendResourceNames, corev1.ResourceName(vnpu.ResourceName))
+	}
 	for _, vnpu := range vnpus.Configs {
 		commonWord := vnpu.CommonWord
 		dev := &Devices{
-			config:           vnpu,
-			nodeRegisterAnno: fmt.Sprintf("hami.io/node-register-%s", commonWord),
-			useUUIDAnno:      fmt.Sprintf("hami.io/use-%s-uuid", commonWord),
-			noUseUUIDAnno:    fmt.Sprintf("hami.io/no-use-%s-uuid", commonWord),
-			handshakeAnno:    fmt.Sprintf("hami.io/node-handshake-%s", commonWord),
-			hamiVnpuCore:     vnpus.HamiVnpuCore,
+			config:                 vnpu,
+			nodeRegisterAnno:       fmt.Sprintf("hami.io/node-register-%s", commonWord),
+			useUUIDAnno:            fmt.Sprintf("hami.io/use-%s-uuid", commonWord),
+			noUseUUIDAnno:          fmt.Sprintf("hami.io/no-use-%s-uuid", commonWord),
+			handshakeAnno:          fmt.Sprintf("hami.io/node-handshake-%s", commonWord),
+			hamiVnpuCore:           vnpus.HamiVnpuCore,
+			allAscendResourceNames: allAscendResourceNames,
 		}
 		sort.Slice(dev.config.Templates, func(i, j int) bool {
 			return dev.config.Templates[i].Memory < dev.config.Templates[j].Memory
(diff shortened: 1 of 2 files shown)


DSFans2014 (commented):
please fix the lint error

DSFans2014:
hi @peachest , I noticed that `runtimeClassName` and `overwriteEnv` were added under `config in` #1738.   From my perspective, these configurations apply uniformly to all ascend devices. If that's the case, it might be better to move them under `vnpu`.  Please let me know if there are other considerations.
https://github.com/Project-HAMi/HAMi/blob/f47cd28ac603e9bb1a6bbfccfa310cc4747f1af4/charts/hami/templates/scheduler/device-configmap.yaml#L114-L127

peachest:
> hi @peachest , I noticed that `runtimeClassName` and `overwriteEnv` were added under `config in` #1738. From my perspective, these configurations apply uniformly to all ascend devices. If that's the case, it might be better to move them under `vnpu`. Please let me know if there are other considerations.
> 
> https://github.com/Project-HAMi/HAMi/blob/f47cd28ac603e9bb1a6bbfccfa310cc4747f1af4/charts/hami/templates/scheduler/device-configmap.yaml#L114-L127

I think this is a good idea. At  the time of #1738, there is no global config for all ascend chips so I put these two fields under each ascend chip. There is another pr #2966  to enable more granular control over the behavior of overwriting env

I am wondering is there a better method to do this only once instead of multiple times in each chips for each container  in the `MutateAdmission` path. 

https://github.com/Project-HAMi/HAMi/blob/9e75c87f46c3b58a142d43fe3fc4b508c1b11824/pkg/scheduler/webhook.go#L97-L104

DSFans2014 (commented):
/lgtm

Shouren (approved):
/lgtm

PR #2959, opened 2026-09-02
fix(scheduler): fail closed on undecodable live Pod allocations

### LINKED ISSUE #2958: Fail safely when a live Pod's allocation annotation cannot be decoded

/kind bug

**What happened**:

When the scheduler starts, or receives an update for a Pod that is missing from its cache, `onAddPod` decodes the Pod's allocated-device annotations. If decoding fails, the scheduler logs the error and returns without recording the Pod's device usage.

The Pod may already be running and using a device, but that usage is then missing from scheduler accounting. The same capacity may be offered to another Pod.

A malformed allocation annotation on a live, bound Pod should fail safely. The scheduler should temporarily reject new HAMi allocations on that Pod's node until it can account for the Pod again.

A later update with valid allocation data should restore the node automatically. Deleting the Pod should also clear its failure record. If several Pods on the same node have invalid annotations, the node should remain blocked until every failure is cleared.

This was found through source review and the existing unit test; it has not yet been reproduced on a live cluster.

1. Register a node with available HAMi device capacity.
2. Send `onAddPod` a running, bound Pod with `hami.io/vgpu-node` set to that node and a malformed supported allocated-device annotation.
3. Confirm that `DecodePodDevices` returns an error and the Pod is absent from `podManager`.
4. Run `Filter` for another Pod requesting the same device capacity.

Current result: the node remains eligible because the first Pod's usage was not recorded.

Expected result: the node is rejected for new HAMi allocations until the first Pod receives a valid update or is deleted.

The relevant path is `pkg/scheduler/scheduler.go` in `onAddPod`. The existing `Test_onAddPod_BadDeviceAnnotation` confirms that a running Pod with an invalid annotation is omitted from `podManager`, but it does not protect the node or test recovery.

A safe implementation should track failures per Pod UID rather than with a single node flag. It should only block the node for a live Pod that is actually bound there, so an arbitrary annotation cannot block a node. Node removal should also clear any related failure records.

This is different from #2756, which validates annotation contents, and #2923, which reports stale usage after Pod deletion. Neither handles a live Pod whose allocation cannot be decoded during accounting recovery.

AI assistance disclosure: I used OpenAI Codex to verify the current code path, search for duplicates, and format this report. I reviewed the final description.

The runtime information below is unavailable because this is currently a source-level finding:

- Pod manifest and events: Not available
- Vendor diagnostics: Not applicable
- Container runtime configuration: Not applicable
- Device-plugin, scheduler, kubelet, and kernel logs: Not available

**Environment**:
- HAMi version: `master` at `95530c6ad09c4f3cf8651cfe53a89eda69238a85`
- Kubernetes version: Not tested on a live cluster
- Accelerator vendor and model: Backend-independent
- HAMi device backend and requested resource names: Any backend with an allocated-device annotation
- Accelerator driver and runtime version: Not applicable
- Workload image and tag: Not applicable
- Operating system and kernel version: Not applicable
- Others: Source analysis and focused unit-test verification only

asadjan4611:
**What type of PR is this?**

/kind bug

When HAMi cannot decode the allocation annotation of a live, bound Pod that is missing from scheduler accounting, it now temporarily excludes that Pod's node from new HAMi allocations.

Failures are tracked by Pod UID. A valid Pod update or deletion removes the failure automatically. If several Pods on the same node have invalid annotations, the node remains unavailable until every failure is cleared.

Malformed updates for Pods with an existing cached allocation continue using the last valid allocation instead of blocking the node.

```text
Live bound Pod
      |
Decode allocation
   /        \
Success     Failure
  |            |
Record usage   Track Pod UID and node
  |            |
Clear failure  Exclude node from Filter
                   |
             Valid update or deletion
                   |
               Restore node
```

Only Pods whose `spec.nodeName` matches HAMi's assigned-node annotation can exclude a node.

Fixes #2958

Tests cover:

- Excluding a node after an uncached bound Pod cannot be decoded.
- Restoring and accounting the node after a valid update.
- Keeping the node excluded while another invalid Pod remains.
- Clearing failures on Pod and node deletion.
- Ignoring an assignment annotation on an unbound Pod.
- Preserving a previously cached valid allocation after a malformed resync.
- Concurrent access under the Go race detector.

No device backend, resource calculation, dependency, configuration, or API behavior is changed.

AI assistance disclosure:

I used an AI Assistance for reviewing the PR Description and  also  for the verification of feature behaviour.

```release-note
HAMi now temporarily excludes a node when it cannot reconstruct a live bound Pod's device allocation, and restores the node after a valid update or deletion.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of invalid device-allocation data for scheduled HAMi workloads, including running pods.
  * Nodes with unresolved allocation failures remain excluded from scheduling to prevent incorrect placement.
  * Allocation failures are cleared when pods become unassigned, terminate, are deleted, recover successfully, or their node is removed.
  * Invalid data from unbound or non-HAMi pods no longer unnecessarily blocks nodes.
  * Valid cached allocations are preserved when resynchronization encounters invalid data.

### DIFF
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -192,13 +192,32 @@ func (s *Scheduler) doNodeNotify() {
 }
 
 func (s *Scheduler) recordAllocationDecodeFailure(pod *corev1.Pod, nodeID string) bool {
-	// Do not let the assigned-node annotation alone quarantine a node. Require
-	// it to agree with the pod's binding recorded in spec.nodeName.
 	if pod.Spec.NodeName == "" || pod.Spec.NodeName != nodeID {
 		return false
 	}
-	s.allocationDecodeFailures.record(pod.UID, nodeID)
-	return pod.UID != ""
+
+	// A Pod author can set spec.nodeName and HAMi annotations directly. Only
+	// quarantine the node when Kubernetes confirms that the Pod was scheduled
+	// and the Pod actually requests a device managed by this scheduler.
+	scheduled := false
+	for _, condition := range pod.Status.Conditions {
+		if condition.Type == corev1.PodScheduled &&
+			condition.Status == corev1.ConditionTrue {
+			scheduled = true
+			break
+		}
+	}
+	if !scheduled {
+		return false
+	}
+	for _, dev := range device.GetDevices() {
+		if device.PodRequiresDevice(dev, pod) {
+			s.allocationDecodeFailures.record(pod.UID, nodeID)
+			return pod.UID != ""
+		}
+	}
+
+	return false
 }
 
 func (s *Scheduler) onAddPod(obj any) {
@@ -270,6 +289,7 @@ func (s *Scheduler) onUpdatePod(oldObj, newObj any) {
 	}
 
 	if _, ok := newPod.Annotations[util.AssignedNodeAnnotations]; !ok {
+		s.allocationDecodeFailures.clearPod(newPod.UID)
 		return
 	}
 
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -32,6 +32,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
+	k8stypes "k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
@@ -53,9 +54,59 @@ import (
 )
 
 const (
-	defaultResync    = 1 * time.Hour
-	syncedPollPeriod = 100 * time.Millisecond
+	defaultResync                  = 1 * time.Hour
+	syncedPollPeriod               = 100 * time.Millisecond
+	unaccountedPodAllocationReason = "node has an unaccounted pod device allocation"
 )
+
+// podAllocationDecodeFailures tracks bound pods whose device allocations
+// cannot be reconstructed. Its zero value is ready for use so Scheduler
+// values constructed directly by tests remain valid.
+type podAllocationDecodeFailures struct {
+	mutex sync.RWMutex
+	pods  map[k8stypes.UID]string
+}
+
+func (f *podAllocationDecodeFailures) record(uid k8stypes.UID, nodeID string) {
+	if uid == "" || nodeID == "" {
+		return
+	}
+	f.mutex.Lock()
+	defer f.mutex.Unlock()
+	if f.pods == nil {
+		f.pods = make(map[k8stypes.UID]string)
+	}
+	f.pods[uid] = nodeID
+}
+
+func (f *podAllocationDecodeFailures) clearPod(uid k8stypes.UID) {
+	f.mutex.Lock()
+	defer f.mutex.Unlock()
+	delete(f.pods, uid)
+}
+
+func (f *podAllocationDecodeFailures) clearNode(nodeID string) {
+	f.mutex.Lock()
+	defer f.mutex.Unlock()
+	for uid, failedNodeID := range f.pods {
+		if failedNodeID == nodeID {
+			delete(f.pods, uid)
+		}
+	}
+}
+
+func (f *podAllocationDecodeFailures) nodes() map[string]struct{} {
+	f.mutex.RLock()
+	defer f.mutex.RUnlock()
+	if len(f.pods) == 0 {
+		return nil
+	}
+	nodes := make(map[string]struct{}, len(f.pods))
+	for _, nodeID := range f.pods {
+		nodes[nodeID] = struct{}{}
+	}
+	return nodes
+}
 
 type Scheduler struct {
 	*nodeManager
@@ -86,6 +137,8 @@ type Scheduler struct {
 	// cycle, so in the common path this adds no contention; it exists so these
 	// paths cannot observe or produce half-applied accounting.
 	allocLock sync.Mutex
+
+	allocationDecodeFailures podAllocationDecodeFailures
 }
 
 func NewScheduler() *Scheduler {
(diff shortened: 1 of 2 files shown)


pkg/scheduler/scheduler.go
```diff
@@ -870,7 +967,12 @@ func (s *Scheduler) getNodesUsage(nodes *[]string, task *corev1.Pod) (*map[strin
 	if nodes == nil {
 		return &cachenodeMap, &overallnodeMap, failedNodes, nil
 	}
+	blockedNodes := s.allocationDecodeFailures.nodes()
```
mesutoezdil: pr #3024 removes this exact loop, did you know?

asadjan4611:
@mesutoezdil now fixed. Each UID is now recorded and cleared in the same goroutine, so the final state is deterministic and i also verified it with repeated race tests.

pkg/scheduler/scheduler.go
asadjan4611: i moved the failed node exclusion outside loop, so it does not depend on the loop removed by #3024.

archlitchi:
The coderabbit comment is reasonable" 'Running HAMi Pods with a missing assignment annotation can unnecessarily block their node from new allocations until cleanup. Align the tracking condition with the assignment requirement before merging'. please resolve this

asadjan4611:
> The coderabbit comment is reasonable" 'Running HAMi Pods with a missing assignment annotation can unnecessarily block their node from new allocations until cleanup. Align the tracking condition with the assignment requirement before merging'. please resolve this

fixed! please review it now

Shouren:
> When HAMi cannot decode the allocation annotation of a live, bound Pod that is missing from scheduler accounting, it now temporarily excludes that Pod's node from new HAMi allocations.

@asadjan4611 My concern is that whether blocking all new HAMi allocations on the node is an overly defensive response, especially since this issue has not yet been reproduced on a live cluster. Depending on Pod creation/update permissions and admission policies, this could potentially be abused as a denial-of-service mechanism.

Could we require stronger evidence of a HAMi-owned allocation before blocking the node, or consider a narrower fallback? Checking spec.schedulerName alone would not establish this provenance either, since it can be specified by the Pod author.

asadjan4611:
@Shouren  I agree that the current checks do not prove that the allocation was created by HAMi. A matching scheduler name, node name, or annotation can be insufficient in some cluster permission models, so node-wide quarantine could affect availability.
I will not treat the current conditions as sufficient origin.
 Would you prefer that I narrow this PR to reporting the undecodable allocation without quarantining the node, or should I first design a persistent scheduler-owned allocation record that can be verified before applying node quarantine?

Shouren:
> @Shouren I agree that the current checks do not prove that the allocation was created by HAMi. A matching scheduler name, node name, or annotation can be insufficient in some cluster permission models, so node-wide quarantine could affect availability. I will not treat the current conditions as sufficient origin. Would you prefer that I narrow this PR to reporting the undecodable allocation without quarantining the node, or should I first design a persistent scheduler-owned allocation record that can be verified before applying node quarantine?

@asadjan4611 I prefer to narrow this PR to reporting the undecodable allocation. But if you any ideas for verifying allocation owned by hami-scheduler, please share it to me. 😊

asadjan4611:
> > @Shouren I agree that the current checks do not prove that the allocation was created by HAMi. A matching scheduler name, node name, or annotation can be insufficient in some cluster permission models, so node-wide quarantine could affect availability. I will not treat the current conditions as sufficient origin. Would you prefer that I narrow this PR to reporting the undecodable allocation without quarantining the node, or should I first design a persistent scheduler-owned allocation record that can be verified before applying node quarantine?
> 
> @asadjan4611 I prefer to narrow this PR to reporting the undecodable allocation. But if you any ideas for verifying allocation owned by hami-scheduler, please share it to me. 😊

@Shouren  I understand and i'll narrow this PR to reporting undecodable live Pod allocations and will remove the node-quarantine behavior.

asadjan4611:
@Shouren in future ownership verification, one possible aproach is a schedulr created allocation receipt tied to the Pod UID, assigned node, and allocation data hash. HAMi could verify that receipt before taking a nod wide safety action. This would need separate design discussion because it requires secure storage or signing and lifecycle handling.
I will keep that idea out of this PR and make the current change focused on reporting only.

pkg/scheduler/scheduler.go
```diff
@@ -151,16 +145,16 @@ func (s *Scheduler) onAddPod(obj any) {
...
-	if !ok {
-		return
-	}
 	if util.IsPodInTerminatedState(pod) {
 		if pi, ok := s.podManager.TakeAndDeletePod(pod); ok {
 			s.quotaManager.RmUsage(pod, pi.Devices)
 		}
 		return
 	}
+	nodeID, ok := pod.Annotations[util.AssignedNodeAnnotations]
```
Shouren, line 154: @asadjan4611 Please replace the `nodeID` with `nodeName`

Shouren:
> @Shouren in future ownership verification, one possible aproach is a schedulr created allocation receipt tied to the Pod UID, assigned node, and allocation data hash. HAMi could verify that receipt before taking a nod wide safety action. This would need separate design discussion because it requires secure storage or signing and lifecycle handling. I will keep that idea out of this PR and make the current change focused on reporting only.

@asadjan4611 Agree with "separate design discussion", so can you please submit a issue about this?

pkg/scheduler/scheduler.go
```diff
@@ -188,17 +191,17 @@ func (s *Scheduler) onUpdatePod(oldObj, newObj any) {
...
-	}
-
 	if util.IsPodInTerminatedState(newPod) {
 		if pi, ok := s.podManager.TakeAndDeletePod(newPod); ok {
 			s.quotaManager.RmUsage(newPod, pi.Devices)
 		}
 		return
 	}
 
+	if _, ok := newPod.Annotations[util.AssignedNodeAnnotations]; !ok {
```
mesutoezdil, line 201: this guard moved down, so terminated pods without it now reach rmusage, intended?

pkg/scheduler/scheduler.go
```diff
@@ -169,7 +169,10 @@ func (s *Scheduler) onAddPod(obj any) {
 
 	rawDevices, err := device.DecodePodDevices(device.SupportDevices, pod.Annotations)
 	if err != nil {
-		klog.ErrorS(err, "failed to decode pod devices", "pod", klog.KObj(pod))
+		if pod.Spec.NodeName == nodeID {
```
mesutoezdil, line 172: why stay silent when the annotation names another node?

pkg/scheduler/event.go
```diff
@@ -83,6 +87,16 @@ func (s *Scheduler) recordNumaRefitResultEvent(pod *corev1.Pod, successMsg strin
 	}
 }
 
+// recordAllocationDecodeFailureEvent reports an allocation that could not be
+// reconstructed without changing candidate-node selection.
+func (s *Scheduler) recordAllocationDecodeFailureEvent(pod *corev1.Pod, nodeID string, decodeErr error) {
```
mesutoezdil, line 92: the shrink decode in onupdatepod fails the same way, no event?

PR #2961, opened 2026-09-02
fix(iluvatar): read the scaled core limit as a per card value

mesutoezdil:


pkg/device/iluvatar/device.go
```diff
@@ -220,9 +220,27 @@ func (dev *IluvatarDevices) GenerateResourceRequests(ctr *corev1.Container) devi
...
+					klog.ErrorS(nil, "iluvatar core request is not a non-negative integer", "container", ctr.Name, "request", core.String())
+					return device.ContainerDeviceRequest{}
+				}
+				// Coresreq is a per card percentage. MutateAdmission rewrites
+				// this limit to count*100 when more than one device is
+				// requested, so a value above 100 is a total across the cards
+				// and has to be divided back. A value at or below 100 is
+				// already per card, which is also what an operator writes when
+				// the admission webhook is disabled or bypassed, so it is left
+				// alone.
```
DSFans2014, line 234: Since we're dividing corenum by n here, what was the purpose of multiplying it by n earlier? Could we just remove that modification in `MutateAdmission`?

https://github.com/Project-HAMi/HAMi/blob/d96e7e8ccaf85b15cb47709b319509b0a9a53eda/pkg/device/iluvatar/device.go#L95-L97

mesutoezdil: that value ends up as the pod's real resource limit, which iluvatar's own device plugin probably reads as the total across cards, so removing it would change what's exposed there

archlitchi (approved):
/lgtm

PR #2970, opened 2026-09-04
Add template to vNPU-mode

archlitchi:
/kind bug

Mode selection 'huawei.com/vnpu-mode' can only specify 'hami-core' but not legacy 'template' mode.

As a result, you can't assign a pod to use 'legacy template' mode, it can still be scheduled to a hami-core mode and uses soft-partition.

## Summary by CodeRabbit

* **New Features**
  * Added support for identifying Ascend vNPU workloads that use template mode.
  * Template-mode workloads can be scheduled on compatible nodes.

* **Bug Fixes**
  * Prevented template-mode workloads from being scheduled on nodes configured for hami-core.
  * Added handling for global hami-core settings, rejecting template workloads unless explicitly supported by the node.


archlitchi:
CC @FouoF @Shouren

FouoF (approved):

PR #2978, opened 2026-09-07
perf(scheduler): stop deep-copying whole Pods into the usage snapshot

togettoyou:
**What type of PR is this?**

/kind enhancement

The HAMi scheduler deep-copies the entire Pod cache on every `Filter` call, even though it only ever reads annotations and identity fields off those Pods. This PR replaces the whole-Pod deep copy with pointer sharing, while still copying the fields the manager owns.

`getNodesUsage` calls `ListPodsInfo()` once per `Filter`. `ListPodsInfo` calls `PodInfo.DeepCopy()` for every cached GPU Pod, and that includes `corev1.Pod.DeepCopy()` — a full copy of the PodSpec, containers, env, volumes, status, and the `managedFields`/`FieldsV1` the API server generates.

`getNodesUsage` then attaches each `*PodInfo` to the matching device's `DeviceUsage.PodInfos`. When scoring runs per candidate node, `scoreNode` deep-copies the `NodeUsage` again — once for the app containers and once more per non-sidecar init container — and the old `DeviceUsage.DeepCopy` deep-copied every attached `PodInfo`, whole Pod included, **a second time**.

So one scheduling attempt performs roughly:

```
cached GPU pods × (1 + candidate nodes × (1 + non-sidecar init containers))
```

whole-Pod copies. The product grows with both cluster size and GPU Pod count. Nothing on the scheduling path mutates those Pods.

### The change

1. `PodInfo.DeepCopy` → `PodInfo.Snapshot`: still deep-copies `Devices` (`PodDevices`) and the scalar fields the manager owns and rewrites in place, but shares the `Pod` pointer.
2. `DeviceUsage.DeepCopy`: `slices.Clone(d.PodInfos)` replaces the per-element `PodInfo.DeepCopy()`. The copy gets its own slice header and shares the read-only `*PodInfo` entries.

Net production change is those two spots, about 10 lines.

## Performance validation

**Environment**: Apple M1 / 16 GiB / macOS 25.5.0 / Go 1.27.1 darwin/arm64

KWOK + [fake-gpu-operator](https://github.com/run-ai/fake-gpu-operator) 

- **128 nodes × 8 fake H100 = 1,024 GPUs**, 10 shared slots each
- **4,096 resident Pods** (4 per GPU), loaded into the HAMi cache through the real API server and informers
- **500 new Pods** per round, 8 concurrent workers, each requesting 1 shared slot / 1,024 MiB / 10% cores
- `GOMAXPROCS=4` for both HAMi and kube-scheduler; HAMi with `kube-qps=200`, `kube-burst=400`, spread policy; kube-scheduler with `percentageOfNodesToScore: 100`
- New Pods traverse the full path: `API server → kube-scheduler → HAMi Filter/Score → HAMi Bind → API server → KWOK Running`
- Six interleaved rounds: `before-1 → after-1 → before-2 → after-2 → before-3 → after-3`. Each round deletes its own Pods and restarts HAMi; resident Pods and nodes stay put.

**Steady-state results** (3 rounds each, medians):

| Metric | before `aa6f39d` | after (this PR) | Change |
|---|---:|---:|---:|
| Wall time for 500 Pods | 34.89 s | 20.26 s | **−41.9%** |
| Successful throughput | 14.33 pod/s | 24.69 pod/s | **+72.2%** |
| End-to-end P50 | 545.5 ms | 343.4 ms | −37.1% |
| End-to-end P95 | 676.8 ms | 379.1 ms | −44.0% |
| End-to-end P99 | 862.4 ms | 396.8 ms | −54.0% |
| Process allocation / new Pod | 124.29 MiB | 9.49 MiB | **−92.4%** |
| Objects allocated (500-Pod window) | 284,416,706 | 50,336,234 | −82.3% |
| GC cycles | 356 | 65 | −81.7% |
| kube-scheduler scheduling-algorithm mean | 69.60 ms | 40.18 ms | −42.3% |

**pprof**:

|  | before `aa6f39d` | after (this PR) |
|---|---|---|
| **Allocation** `alloc_space`<br>60.84 GiB → 4.59 GiB | image | image |
| **CPU** 60 s window<br>68.69 → 10.03 CPU·s | image | image |

**benchmark**:

| `BenchmarkListPodsInfo` (pkg/device) | sec/op | B/op | allocs/op |
|---|---|---|---|
| pods=100 | 195.6 µs → 31.4 µs (**−83.96%**) | 537.6 KiB → 67.3 KiB (−87.48%) | 1,902 → 802 (−57.83%) |
| pods=1000 | 2.019 ms → 310.8 µs (**−84.61%**) | 5,375 KiB → 672 KiB (−87.50%) | 19,003 → 8,003 (−57.89%) |
| pods=5000 | 13.040 ms → 1.901 ms (**−85.42%**) | 26.25 MiB → 3.28 MiB (−87.50%) | 95,003 → 40,003 (−57.89%) |

| `BenchmarkScoreNode*` (pkg/scheduler) | sec/op | B/op | allocs/op |
|---|---|---|---|
| `ScoreNodeOccupied` podsPerGPU=0 *(control)* | 10.36 µs → 10.49 µs (~, p=0.721) | 11.00 KiB → 11.00 KiB (~) | 154 → 154 (~) |
| `ScoreNodeOccupied` podsPerGPU=2 | 35.74 µs → 10.61 µs (**−70.31%**) | 87.47 KiB → 11.31 KiB (−87.07%) | 368 → 175 (−52.45%) |
| `ScoreNodeOccupied` podsPerGPU=4 | 61.08 µs → 11.28 µs (**−81.53%**) | 163.85 KiB → 11.56 KiB (−92.94%) | 562 → 177 (−68.51%) |
| `ScoreNode` initContainers=0/1/4 *(control, idle devices)* | ~ (p=0.743 / 0.314 / 0.328) | ~ | ~ (identical per sample) |

---

**AI assistance disclosure**

This PR was written with assistance from Claude Code, covering the code change, the benchmarks, and the setup and execution of the performance validation experiments. I have reviewed and understood every change line by line, and independently verified the correctness argument (the ownership boundary around the shared Pod and the read-only nature of each consumer).

## Behavior Changes

* Pod information retrieval now returns lightweight snapshots that share read-only pod data while keeping manager-owned fields independent.
* Device usage copies duplicate the pod information list while sharing its read-only entries.

## Performance

* Reduced copying overhead for pod and device usage information.

## Tests

* Added benchmarks for pod snapshot generation and scoring nodes with existing workloads.
* Updated tests to verify snapshot isolation and shared read-only data semantics.

### DIFF
--- a/pkg/device/devices.go
+++ b/pkg/device/devices.go
@@ -256,6 +256,8 @@ func init() {
 	SupportDevices = make(map[string]string)
 }
 
+// DeepCopy returns a copy of DeviceUsage. It clones the PodInfos slice but
+// shares its read-only PodInfo entries, which callers must not mutate.
 func (d *DeviceUsage) DeepCopy() *DeviceUsage {
 	if d == nil {
 		return nil
@@ -291,12 +293,8 @@ func (d *DeviceUsage) DeepCopy() *DeviceUsage {
 	}
 	dup.MigUsage = d.MigUsage.DeepCopy()
 
-	if d.PodInfos != nil {
-		dup.PodInfos = make([]*PodInfo, len(d.PodInfos))
-		for i, pi := range d.PodInfos {
-			dup.PodInfos[i] = pi.DeepCopy()
-		}
-	}
+	// Copy the slice while sharing its read-only PodInfo snapshots.
+	dup.PodInfos = slices.Clone(d.PodInfos)
 
 	if d.CustomInfo != nil {
 		dup.CustomInfo = make(map[string]any, len(d.CustomInfo))
--- a/pkg/device/devices_test.go
+++ b/pkg/device/devices_test.go
@@ -1989,9 +1989,10 @@ func TestDeviceUsageDeepCopy(t *testing.T) {
 			}
 
 			if len(copy.PodInfos) > 0 {
-				originalNodeID := tt.original.PodInfos[0].NodeID
-				copy.PodInfos[0].NodeID = "mutated-node"
-				assert.Equal(t, tt.original.PodInfos[0].NodeID, originalNodeID)
+				assert.Assert(t, tt.original.PodInfos[0] == copy.PodInfos[0], "PodInfos entries should be shared with the copy")
+				originalEntry := tt.original.PodInfos[0]
+				copy.PodInfos[0] = &PodInfo{NodeID: "replacement-node"}
+				assert.Assert(t, tt.original.PodInfos[0] == originalEntry, "replacing a copied slice entry must not affect the original")
 			}
 
 			if copy.CustomInfo != nil {
--- a/pkg/device/pod_test.go
+++ b/pkg/device/pod_test.go
@@ -486,7 +486,7 @@ func TestUpdatePod(t *testing.T) {
 	}
 }
 
-func TestPodInfoDeepCopy(t *testing.T) {
+func TestPodInfoSnapshot(t *testing.T) {
 	tests := []struct {
 		name     string
 		original *PodInfo
@@ -523,7 +523,7 @@ func TestPodInfoDeepCopy(t *testing.T) {
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			copy := tt.original.DeepCopy()
+			copy := tt.original.Snapshot()
 
 			if tt.original == nil {
 				if copy != nil {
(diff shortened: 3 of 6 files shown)


pkg/device/devices.go
```diff
@@ -291,12 +291,8 @@ func (d *DeviceUsage) DeepCopy() *DeviceUsage {
...
 	dup.MigUsage = d.MigUsage.DeepCopy()
 
-	if d.PodInfos != nil {
-		dup.PodInfos = make([]*PodInfo, len(d.PodInfos))
-		for i, pi := range d.PodInfos {
-			dup.PodInfos[i] = pi.DeepCopy()
-		}
-	}
+	// Copy the slice while sharing its read-only PodInfo snapshots.
+	dup.PodInfos = slices.Clone(d.PodInfos)
```
Shouren, line 297: @togettoyou I'm concerned that using `slice.Clone` here might break the semantics of `DeviceUsage.DeepCopy`, shall we keep it or rename it?

Shouren (commented):
@togettoyou This PR looks like a good starting point for further work on ownership in the scheduler, which is also part of my plan. Would you be interested in exploring how to share read-only informer objects while keeping HAMi’s mutable accounting state isolated in the following PRs? That could help reduce unnecessary copies and clarify the copy/snapshot semantics.

togettoyou:
> @togettoyou This PR looks like a good starting point for further work on ownership in the scheduler, which is also part of my plan. Would you be interested in exploring how to share read-only informer objects while keeping HAMi’s mutable accounting state isolated in the following PRs? That could help reduce unnecessary copies and clarify the copy/snapshot semantics.

Certainly, I would be happy to get involved.

pkg/device/devices.go
togettoyou: I tend to keep the `slice.Clone` , as this is also part of a high-frequency "hot path."

Would renaming the method to `DeviceUsage.Snapshot` make the semantics clearer?

Shouren: @togettoyou  I think we can keep both `slice.Clone` and `DeviceUsage.DeepCopy` and a comment to `DeviceUsage.DeepCopy` about the PodInfo snapshots.

togettoyou: that works too.

Shouren: So please add another comment above the `func (d *DeviceUsage) DeepCopy() *DeviceUsage`

Shouren (approved):
/lgtm

PR #2979, opened 2026-09-07
fix(chart): target device plugin monitor port by name

magic-peach:
**What type of PR is this?**
/kind bug

The device plugin monitor Service hardcodes `targetPort: 9394` instead of targeting the container port by name, unlike the scheduler's equivalent monitor Service (fixed in #2895) which targets by name to avoid drift if the container port number ever changes. This applies the same fix here: adds `devicePlugin.service.monitorTargetPort` (default `metrics`) to `values.yaml`, uses it in `monitorservice.yaml`, and documents it in the chart README, mirroring #2895's pattern exactly.

Fixes #

Confirmed via grep that this was the only remaining hardcoded `targetPort` of this kind in the chart. Verified rendered output is functionally identical at default values (`helm template` diff shows only the port field changing from the numeric literal to the equivalent name), and `helm lint` passes.

No behavior change at default values; adds a new `devicePlugin.service.monitorTargetPort` value for consistency with the scheduler service.

---
This PR was written primarily by Claude Code, an AI assistant, under my direction and review. I verified the change by comparing rendered chart output before and after, and reviewed the diff before submitting.

## Summary by CodeRabbit

* **New Features**
  * Added configuration for the device plugin monitor service’s target port.
  * The target port defaults to the named `metrics` port and can be customized through Helm values.

* **Documentation**
  * Updated the Helm chart configuration reference with the new monitor target port setting.

### DIFF
--- a/charts/hami/README.md
+++ b/charts/hami/README.md
@@ -197,6 +197,7 @@ become ready and restart the HAMi device plugin DaemonSet.
 |-----------|-------------|---------------|
 | `devicePlugin.service.type` | Service type | `NodePort` |
 | `devicePlugin.service.httpPort` | HTTP port | `31992` |
+| `devicePlugin.service.monitorTargetPort` | Monitor target port | `metrics` |
 
 ### Device Plugin Deployment Configuration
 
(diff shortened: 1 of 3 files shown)


mesutoezdil:
Please do the minor changes in a single pr.

PR #2980, opened 2026-09-07
test(util): cover MarkAnnotationsToDelete

magic-peach:
**What type of PR is this?**
/kind cleanup

`util.MarkAnnotationsToDelete` is the only exported function in `pkg/util/util.go` without a direct unit test. It's a thin wrapper over `GetNode` and `RemoveNodeAnnotation`, both already tested individually, but its own composition and error-propagation path (when `GetNode` fails) were untested. Adds `TestMarkAnnotationsToDelete` and `TestMarkAnnotationsToDelete_NodeNotFound` following the existing fake-clientset pattern used by the neighboring `GetNode`/`RemoveNodeAnnotation` tests.

Fixes #

Test-only change, no production code touched. Ran `go test ./pkg/util/... -run TestMarkAnnotationsToDelete -v --race -count=1`, both new tests pass.

NONE

---
This PR was written primarily by Claude Code, an AI assistant, under my direction and review. I verified the tests by running them locally and reviewed the diff before submitting.

## Summary by CodeRabbit

- **Tests**
  - Added coverage for successfully removing node annotations.
  - Added coverage for handling cases where the specified annotation is not found.

### DIFF
--- a/pkg/util/util_test.go
+++ b/pkg/util/util_test.go
@@ -1131,3 +1131,34 @@ func TestGetNode(t *testing.T) {
 		})
 	}
 }
+
+func TestMarkAnnotationsToDelete(t *testing.T) {
+	node := &corev1.Node{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "test-node",
+			Annotations: map[string]string{
+				"hami.io/nvidia": "handshake",
+			},
+		},
+	}
+	oldClient := client.KubeClient
+	t.Cleanup(func() { client.KubeClient = oldClient })
+	client.KubeClient = fake.NewClientset(node)
+
+	err := MarkAnnotationsToDelete("hami.io/nvidia", "test-node")
+	assert.NilError(t, err)
+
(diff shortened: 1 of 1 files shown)


pkg/util/util_test.go
```diff
@@ -1131,3 +1131,34 @@ func TestGetNode(t *testing.T) {
 		})
 	}
 }
+
+func TestMarkAnnotationsToDelete(t *testing.T) {
```
Eshiv-Pandey: this duplicates TestRemoveAnnotation. could you rename or like strengthen that one instead? It's mis-named and its success case never asserts the annotation is actually removed.

Eshiv-Pandey:
Thanks! But MarkAnnotationsToDelete isn't actually untested... TestRemoveAnnotation (from #2052) already covers both the success and GetNode-not-found paths.

magic-peach:
You're right, thanks for catching that. TestRemoveAnnotation from #2052 already covers both branches through MarkAnnotationsToDelete directly, I missed it because the test function isn't named after the function it exercises. Closing this, no new coverage here.

PR #2985, opened 2026-09-08
fix(chart): add standard labels to webhook, monitor role, and runtime class

magic-peach:
**What type of PR is this?**
/kind cleanup

`MutatingWebhookConfiguration` (charts/hami/templates/scheduler/webhook.yaml), the device plugin monitor `ClusterRole` (charts/hami/templates/device-plugin/monitorrole.yaml), and the `RuntimeClass` (charts/hami/templates/device-plugin/runtime-class.yaml) had no labels at all, unlike nearly every other resource in the chart, including other cluster scoped ones like the scheduler ClusterRole and the device plugin ClusterRoleBinding. This adds the same standard labels those already have. The webhook config in particular is the most operationally significant resource in the chart, so missing labels there makes it harder to find with label selector based tooling.

Fixes #

Verified with `helm lint` and `helm template --show-only` on all three files: labels render correctly, no other output changed.

No behavior change; adds standard `app.kubernetes.io/*` and `helm.sh/chart` labels to three resources that previously had none.

---
This PR was written primarily by Claude Code, an AI assistant, under my direction and review. I verified the change with helm lint and helm template and reviewed the diff before submitting.

## Summary by CodeRabbit

* **Enhancements**
  * Added standard Kubernetes metadata labels to device plugin monitoring, runtime class, and scheduler webhook resources.
  * Improved resource identification and consistency across deployed components.

### DIFF
--- a/charts/hami/templates/device-plugin/monitorrole.yaml
+++ b/charts/hami/templates/device-plugin/monitorrole.yaml
@@ -3,6 +3,9 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
   name:  {{ include "hami-vgpu.device-plugin" . }}-monitor
+  labels:
+    app.kubernetes.io/component: "hami-device-plugin"
+    {{- include "hami-vgpu.labels" . | nindent 4 }}
 rules:
   - apiGroups:
       - ""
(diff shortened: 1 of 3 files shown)


mesutoezdil:
Please do the minor changes in a single pr.

PR #2986, opened 2026-09-08
fix(device-plugin): skip ld.so.preload for sandboxed runtime classes

### LINKED ISSUE #2581: Feature Request: google gVisor (runsc) compatibility for NVIDIA vGPU enforcement
## Summary

HAMi's vGPU enforcement mechanism is currently incompatible with google gVisor (runsc) sandboxed containers. We would like to request native gVisor support so that memory and core limits can be enforced for GPU workloads running inside gVisor sandboxes.

## Current Behavior

HAMi enforces vGPU limits (`nvidia.com/gpumem`, `nvidia.com/gpucores`) by injecting `libvgpu.so` via `/etc/ld.so.preload` at `Allocate()` time (`pkg/device-plugin/nvidiadevice/nvinternal/plugin/server.go`):

```go
response.Mounts = append(response.Mounts,
    &kubeletdevicepluginv1beta1.Mount{
        ContainerPath: "/etc/ld.so.preload",
        HostPath:      hostHookPath + "/vgpu/ld.so.preload",
        ReadOnly:      true,
    },
)
```

This works correctly for standard `runc`-based containers (`runtimeClassName: nvidia`). However, when a pod uses `runtimeClassName: gvisor`, the container crashes immediately on startup:

```
bash: error while loading shared libraries: libcuda.so.1: cannot open shared object file: No such file or directory
```

## Root Cause

gVisor's `sentry` (its userspace kernel) intentionally does not honour `/etc/ld.so.preload` as a security boundary — allowing host-injected library preloading would undermine the sandbox isolation model. As a result, `libvgpu.so` is never loaded and vGPU enforcement does not apply.

## Expected Behavior

HAMi should be able to enforce vGPU limits for containers running inside gVisor sandboxes, either by:

1. **Detecting `runtimeClassName: gvisor`** at `Allocate()` time and skipping the `ld.so.preload` mount while using an alternative enforcement mechanism compatible with gVisor (e.g. NRI plugin, CDI-based injection, or gVisor extension)
2. **Or at minimum**, skipping the `ld.so.preload` mount for gVisor pods (without crashing) and documenting that vGPU enforcement is not available in this configuration

## Environment

- Kubernetes: v1.35.4
- HAMi: v2.9.0
- gVisor: enabled with `nvproxy: "true"`
- nvproxy confirmed working: `/dev/nvidia0` is correctly visible inside gVisor sandbox when HAMi is not involved

## Additional Context

We confirmed through testing that nvproxy correctly passes GPU devices into gVisor sandboxes when HAMi is not involved (plain NVIDIA device plugin). The blocker is exclusively HAMi's `ld.so.preload` injection mechanism.

IIITManjeet:
**What type of PR is this?**

/kind bug

`Allocate()` mounts `/etc/ld.so.preload` into every non-MIG container unless the container sets `CUDA_DISABLE_CONTROL=true`. The file points at `libvgpu.so`, which is linked with `-lcuda -lnvidia-ml`:

```
$ readelf -d libvgpu.so.v2.9.0 | grep NEEDED
 (NEEDED)  Shared library: [libcuda.so.1]
```

That is a load-time dependency, so every process in the container has to resolve `libcuda.so.1` before it starts. Under `runtimeClassName: gvisor` it cannot, so every binary fails, including bash:

```
bash: error while loading shared libraries: libcuda.so.1: cannot open shared object file: No such file or directory
```

The pod never starts. HAMi also cannot enforce limits in it, because `libvgpu.so` is what does the enforcement.

This PR checks the pod's runtimeClassName in `Allocate()` and skips the `ld.so.preload` mount for sandbox runtimes, with a warning in the log saying limits are not enforced. Nothing else in the allocation changes. The device, the `CUDA_DEVICE_MEMORY_LIMIT_*` and `CUDA_DEVICE_SM_LIMIT` envs, the `libvgpu.so` mount, the cache directory and
`/tmp/vgpulock` are all still set.

The default list is `gvisor` and `runsc`. It can be changed with `sandboxRuntimeClassNames` in the device config. Existing installs get the fix without editing config.

Fixes #2581

I match on the RuntimeClass name instead of reading the RuntimeClass object's handler. Reading the object would be exact, but it needs a `node.k8s.io` read added to the device plugin RBAC and an API call during allocation. I can switch it if you would rather have that.

The issue says gVisor does not honour `/etc/ld.so.preload`. I do not think that is what happened. Only two things can put `libvgpu.so` into bash's dependency list: the `LD_PRELOAD` env var, or `/etc/ld.so.preload`. HAMi never sets `LD_PRELOAD` anywhere, and `vgpu-init.sh` only rewrites the preload file, so the file must have been read. If `libcuda.so.1` had been in the sandbox it would have resolved and there would be no error. So the preload was honoured, and the driver libraries that the NVIDIA container runtime puts into a runc container are not there under `runsc`.

If those libraries were injected, `libvgpu.so` might load and enforcement might work under nvproxy. That is option 1 in the issue and I have not tried it. I did option 2 because the issue asks for it as a minimum, and because HAMi should not mount a preload whose dependency it cannot satisfy either way.

Yes. A pod with `runtimeClassName` of `gvisor` or `runsc` no longer gets the `/etc/ld.so.preload` mount, so it starts instead of failing on `libcuda.so.1`. Memory and core limits are not enforced inside it. The scheduler still counts the request. The device
plugin logs a warning on each such allocation. Other runtime classes are not affected.

## Validation

Linux x86_64 (WSL2, Ubuntu), Go 1.27.0, CGO enabled:

* `go test ./pkg/device-plugin/... ./pkg/scheduler/config/... ./pkg/device/nvidia/... -count=1` — pass
* `make verify` — exit 0, 0 lint issues
* `make -C charts validate` and `hack/verify-chart-version.sh` with Helm v3.21.4 — pass
* Reverted the new condition and re-ran the two Allocate tests to check they fail against the
  unfixed code

For the root cause I extracted `ld.so.preload` and `libvgpu.so` from `projecthami/hami:v2.9.0` and checked the `DT_NEEDED`. I then built a library with the same shape, with a `DT_NEEDED` that cannot be resolved, and preloaded it: bash, ls, cat and id all
fail with the same error. With the dependency resolvable the same preload does nothing.

**Hardware:** I do not have a GPU node or a gVisor cluster. I have not reproduced the original crash or checked that a gVisor pod starts with this change, so gate 2 is not met. The runc path is covered by the repo's `e2e_test (nvidia, tesla-p4)` job, which starts a vGPU pod and runs a CUDA sample, and that needs the preload to be in place.

@marcellvasanczki-sap if you can run this on the cluster from the issue, the check is whether a pod with `runtimeClassName: gvisor` and `nvidia.com/gpumem` starts now. This also shows the driver libraries are missing in the sandbox, which is what the diagnosis rests on:

```
kubectl exec <gvisor-pod> -- ldconfig -p | grep -E "libcuda|libnvidia-ml"
```

## AI Assistance Disclosure
This PR was written with AI assistance (Claude Code), covering the codebase investigation, implementation, and tests. I reviewed the change against the existing code and ran the validation listed above.

## Summary by CodeRabbit

* **New Features**
  * Added configuration for sandboxed runtime classes, including gVisor and runsc by default.
  * Pods using configured sandboxed runtimes can start without the NVIDIA preload injection.
  * Added Helm chart values and documentation for configuring sandbox runtime classes.

* **Behavior Changes**
  * Memory and core limits are not enforced for pods using sandboxed runtimes.

### DIFF
--- a/pkg/device-plugin/nvidiadevice/nvinternal/plugin/util.go
+++ b/pkg/device-plugin/nvidiadevice/nvinternal/plugin/util.go
@@ -453,3 +453,26 @@ func createSpecFile(outputPath string) error {
 	}
 	return nil
 }
+
+// isSandboxedRuntimeClass reports whether the pod runs under one of the configured sandbox
+// runtime classes. libvgpu.so carries a DT_NEEDED on libcuda.so.1, which those sandboxes do
+// not provide, so preloading it makes every binary in the container fail to start.
+func isSandboxedRuntimeClass(pod *corev1.Pod, configured []string) bool {
+	if pod == nil || pod.Spec.RuntimeClassName == nil {
+		return false
+	}
+	runtimeClass := strings.TrimSpace(*pod.Spec.RuntimeClassName)
+	if runtimeClass == "" {
+		return false
+	}
+	names := configured
+	if len(names) == 0 {
+		names = nvidia.DefaultSandboxRuntimeClassNames
+	}
+	for _, name := range names {
+		if strings.EqualFold(strings.TrimSpace(name), runtimeClass) {
+			return true
+		}
+	}
+	return false
+}
--- a/charts/hami/README.md
+++ b/charts/hami/README.md
@@ -240,6 +240,7 @@ become ready and restart the HAMi device plugin DaemonSet.
 |-----------|-------------|---------------|
 | `devices.nvidia.gpuCorePolicy` | GPU core policy | `default` |
 | `devices.nvidia.libCudaLogLevel` | CUDA library log level | `1` |
+| `devices.nvidia.sandboxRuntimeClassNames` | RuntimeClass names whose sandbox does not provide the `libcuda.so.1` that `libvgpu.so` links against (gVisor); pods using them skip the preload injection, and vGPU limits are not enforced. Empty means the defaults `gvisor` and `runsc` | `[]` |
 
 ### Huawei Ascend
 | Parameter | Description | Default Value |
--- a/charts/hami/templates/scheduler/device-configmap.yaml
+++ b/charts/hami/templates/scheduler/device-configmap.yaml
@@ -32,6 +32,10 @@ data:
       gpuCorePolicy: {{ .Values.devices.nvidia.gpuCorePolicy }}
       libCudaLogLevel: {{ .Values.devices.nvidia.libCudaLogLevel }}
       runtimeClassName: "{{ .Values.devicePlugin.runtimeClassName }}"
+      {{- with .Values.devices.nvidia.sandboxRuntimeClassNames }}
+      sandboxRuntimeClassNames:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
       migProfileAllowlist:
       - models: [ "A30" ]
         profiles: [ "1g.6gb", "2g.12gb", "4g.24gb" ]
--- a/charts/hami/values.yaml
+++ b/charts/hami/values.yaml
@@ -531,6 +531,11 @@ devices:
   nvidia:
     gpuCorePolicy: default
     libCudaLogLevel: 1
+    # RuntimeClass names, gVisor for example, whose sandbox does not provide the
+    # libcuda.so.1 that libvgpu.so links against. Pods using one of these skip the preload
+    # injection so the container can start; memory and core limits are not enforced in them.
+    # Leave empty for the built-in defaults: gvisor and runsc.
+    sandboxRuntimeClassNames: []
   ascend:
     enabled: false
     image: ""
--- a/pkg/device-plugin/nvidiadevice/nvinternal/plugin/alloc_sandbox_runtime_test.go
+++ b/pkg/device-plugin/nvidiadevice/nvinternal/plugin/alloc_sandbox_runtime_test.go
@@ -1,0 +1,160 @@
+/*
+ * Copyright (c) 2026, HAMi.  All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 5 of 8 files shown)


FouoF:
@IIITManjeet Great work! I also doubt whether the root cause is described as #2581, so we may require @marcellvasanczki-sap to confirm it before merging this PR.

pkg/device-plugin/nvidiadevice/nvinternal/plugin/util.go
```diff
@@ -453,3 +453,26 @@ func createSpecFile(outputPath string) error {
 	}
 	return nil
 }
+
+// isSandboxedRuntimeClass reports whether the pod runs under one of the configured sandbox
+// runtime classes. libvgpu.so carries a DT_NEEDED on libcuda.so.1, which those sandboxes do
+// not provide, so preloading it makes every binary in the container fail to start.
+func isSandboxedRuntimeClass(pod *corev1.Pod, configured []string) bool {
```
mesutoezdil, line 460: can the default runtimeClassName trigger this silently?

mesutoezdil, line 460: what if someone renames the RuntimeClass object?

IIITManjeet: Hi @mesutoezdil,
Not from a pod which doesn't have a `runtimeClass`, it returns false on the very first check. But it can happen through the HAMi webhook if `devicePlugin.runtimeClassName` is set to gvisor or runsc, since it writes that into the spec of every GPU pod. Then all of them skip the preload and the only sign is the device plugin logs. Also the defaults I added apply when sandboxRuntimeClassNames is empty, so setting [] can't turn it off.
Please do suggest if I can fail at config load when runtimeClassName is in the sandbox list?

IIITManjeet: Yes this seems to be a shortcoming, I'll suggest checking the handler would cover both, but the device plugin needs RBAC to read runtimeclasses and a cached lookup in Allocate. Should I switch to that?

mesutoezdil:
point 4

PR #1419, opened 2025-10-17
upgrade promethues api metric label,add  devicetype  label

zhegemingzimeibanquan:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**:
Adds GPU model identification to metric collection by introducing a devicetype  label. This enables proper resource classification and analysis of GPU workloads across different hardware types (e.g., NVIDIA Tesla P100-PCIE-12GB,NVIDIA A100-PCIE-40GB). Specifically implements nvml.Device.GetName() to capture the exact hardware model string

**Which issue(s) this PR fixes**:
Fixes none

**Special notes for your reviewer**:
- ✅ **NVIDIA-only implementation**: Uses `nvml.Device.GetName()` directly from `go-nvml` (no `nvidia-smi`/`lspci` dependencies)
- ✅ **Exact model string**: Returns full driver-reported name (e.g., "NVIDIA H100-SXM5-80GB", not just "H100")
- ✅ **Zero performance impact**: Caches model string during first metric collection
- ✅ **Backward compatible**: New `gputype` label added to all GPU metrics (no existing metrics changed)

**Does this PR introduce a user-facing change?**: Yes

PR #1139, opened 2025-06-20
nvidia model duplicate NVIDIA prefix

lengrongfu:
**What type of PR is this?**

**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

PR #2991, opened 2026-09-09
fix(monitor): drop duplicate `NVIDIA` prefix from `device_type` label

### LINKED ISSUE #2771: bug(vGPUmonitor): hami_host_gpu_memory_controller_utilization_ratio  missing  node  label

### LINKED ISSUE #2990: vGPUmonitor host GPU metrics: device_type is double-prefixed and memory-controller utilization is missing the node label
## What happened

Two label defects in `cmd/vGPUmonitor/metrics.go` make the `hami_host_gpu_*` series unusable alongside the scheduler's `hami_gpu_*` series.

### 1. `device_type` is double-prefixed on modern drivers

The device plugin normalizes the NVML model name conditionally — it only prepends the vendor when NVML did not already report it ([`register.go:173-177`](https://github.com/Project-HAMi/HAMi/blob/master/pkg/device-plugin/nvidiadevice/nvinternal/plugin/register.go#L173-L177)):

```go
if !strings.HasPrefix(Model, "NVIDIA") {
    Model = fmt.Sprintf("NVIDIA-%s", Model)
}
```

vGPUmonitor prepends it unconditionally ([`metrics.go:330`](https://github.com/Project-HAMi/HAMi/blob/master/cmd/vGPUmonitor/metrics.go#L330)):

```go
deviceName: "NVIDIA-" + deviceName,
```

Since driver R470, `nvmlDeviceGetName` already returns the vendor in the name (`NVIDIA A100-SXM4-40GB`, `NVIDIA GeForce RTX 4090`). So on any current GPU the same physical card is labelled two different ways:

| source | metric | `device_type` |
|---|---|---|
| scheduler (from the node register annotation) | `hami_gpu_memory_limit_bytes` | `NVIDIA A100-SXM4-40GB` |
| vGPUmonitor | `hami_host_gpu_memory_used_bytes` | `NVIDIA-NVIDIA A100-SXM4-40GB` |

Any PromQL that joins or groups the two sets on `device_type` silently returns nothing. Older cards whose NVML name has no vendor (`Tesla V100-SXM2-16GB`) are unaffected, which is presumably why this went unnoticed.

The intent was clearly to keep the two in sync — #1419, which introduced the label, has the commit message *"与hami注解中的gpu显卡型号进行写法同步"* ("synchronize with the GPU model notation in the HAMi annotation"). The implementation just duplicated the rule instead of sharing it, and got the condition wrong.

### 2. `hami_host_gpu_memory_controller_utilization_ratio` is missing the `node` label

Every other host metric declares `{"node", "device_index", "device_uuid", "device_type"}`. This one declares only `{"device_index", "device_uuid", "device_type"}` ([`metrics.go:69-73`](https://github.com/Project-HAMi/HAMi/blob/master/cmd/vGPUmonitor/metrics.go#L69-L73)), and the emit site passes three label values accordingly, even though `identity.nodeName` is right there. It cannot be grouped by node or joined against its siblings without falling back to the scrape target's relabeled `instance`.

- `device_type` identifies a GPU identically in vGPUmonitor and scheduler metrics.
- All `hami_host_gpu_*` metrics carry the `node` label.

## How to reproduce it

On a node with an A100 (or any GPU on driver ≥ R470):

```
curl -s <vGPUmonitor>:9394/metrics | grep hami_host_gpu_memory_used_bytes
# device_type="NVIDIA-NVIDIA A100-SXM4-40GB"

kubectl get node <node> -o jsonpath='{.metadata.annotations.hami\.io/node-nvidia-register}'
# ...,NVIDIA A100-SXM4-40GB,...
```

and:

```
curl -s <vGPUmonitor>:9394/metrics | grep hami_host_gpu_memory_controller_utilization_ratio
# no node= label
```

Fix: move the normalization into one shared helper in `pkg/device/nvidia` used by both the plugin and the monitor, and add the missing `node` label. PR follows.

Note that fixing (1) changes the emitted `device_type` value on modern GPUs (and on the legacy `HostGPUMemoryUsage` / `HostCoreUtilization` names too), so existing dashboards that hard-code `NVIDIA-NVIDIA ...` will need updating — but that value was never meant to be there.

## Environment

- HAMi version: master (`d96e7e8`)
- Affects: NVIDIA GPUs on driver R470 and later

rapsealk:
Fixes #2990

## What this PR does

`device_type` on the `hami_host_gpu_*` series emitted by vGPUmonitor was double-prefixed on modern drivers, so it did not match the scheduler's `hami_gpu_*` series for the same physical GPU.

The device plugin only prepends `NVIDIA-` when `nvmlDeviceGetName` did not already report the vendor; vGPUmonitor prepended it unconditionally. Since driver R470 the NVML name already contains it, so the same card was labelled `NVIDIA A100-SXM4-40GB` in the node register annotation (and thus in `hami_gpu_*`) but `NVIDIA-NVIDIA A100-SXM4-40GB` in `hami_host_gpu_*`.

This completes #1139. That PR (Nov 2025, "nvidia model duplicate NVIDIA prefix") replaced the unconditional `fmt.Sprintf("%v-%v", "NVIDIA", Model)` in `register.go` with the conditional check — three weeks after #1419 had added the unconditional copy to vGPUmonitor — and the monitor copy was missed. The rule now lives in one place, `nvidia.NormalizeDeviceModel`, and both call sites use it. The plugin's behaviour is byte-for-byte unchanged; only the monitor's label changes.

**Scope note:** #2990 also reports the missing `node` label on `hami_host_gpu_memory_controller_utilization_ratio`. That is already fixed by #2776 (opened earlier, fixes #2771), so this PR no longer touches it; the two PRs share no hunks.

## Behaviour change

On GPUs with a driver ≥ R470, the `device_type` label on the `hami_host_gpu_*` metrics — and on the legacy `HostGPUMemoryUsage` / `HostCoreUtilization` names — changes from `NVIDIA-NVIDIA <model>` to `NVIDIA <model>`, matching the scheduler metrics and the node annotation.

To be precise about impact: `device_uuid` and `node` were always consistent between the two metric families, so joins on those keys were never broken. What was broken is grouping or labelling by `device_type` across the two families, and any dashboard that hard-codes the doubled value. Those need updating. Older cards whose NVML name has no vendor prefix (e.g. `Tesla V100-SXM2-16GB`) are unaffected.

## Tests

- `TestResolveGPUDeviceIdentityDoesNotDoublePrefixModel` — fails on master, where the identity returns `NVIDIA-NVIDIA A100-SXM4-40GB`.
- `TestNormalizeDeviceModel` — unit test for the new helper (already-prefixed, legacy, and empty names).

Existing `cmd/vGPUmonitor` and `pkg/device-plugin/.../plugin` tests pass unchanged, including the pedantic-registry Describe/Collect check.

## AI assistance disclosure

Per CONTRIBUTING.md: this PR was investigated and written with Claude Code, including an independent review pass that surfaced the #1139 provenance. I reviewed the diff, confirmed the double-prefix against the NVML naming behaviour and the two call sites, and verified the regression test fails on master before the fix. Review responses will be my own words, not pasted model output.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01CbWVFcwM1e4NschReqYvHr

## Summary by CodeRabbit

- **Bug Fixes**
  - Standardized NVIDIA GPU model names across device-plugin metadata and vGPU monitoring metrics.
  - Prevented duplicate vendor prefixes for GPU models already reported with “NVIDIA.”
  - Added the appropriate NVIDIA prefix for model names reported without one.
  - Improved consistency of GPU identification across different driver generations.

- **Tests**
  - Added coverage for prefixed, unprefixed, and empty GPU model names.


mesutoezdil:
remove the llm from history pls

mesutoezdil:
the node-label half of this (hostGPUMemoryUtilizationdesc) is the same fix as the still-open #2776, so whichever merges first will need the other rebased

rapsealk:
@mesutoezdil Thanks for the comment! I have dropped the duplicate changes.

archlitchi (approved):
/lgtm

PR #2993, opened 2026-09-09
test(scheduler): add tests for LoadConfig

magic-peach:
LoadConfig reads and parses the scheduler's YAML config file but had no test
coverage: not for a missing file, not for invalid YAML, not for the success
path every scheduler component depends on at startup. Adds TestLoadConfig
with four subtests: file not found, invalid YAML, an empty file (valid YAML,
zero-value config), and a valid config asserting the parsed nvidia and hygon
fields.

Fixes #

Test-only change, no production code touched. Ran go build, go vet, and the
full pkg/scheduler/config test suite locally, all pass.

No

This PR was written primarily by Claude Code, an AI coding assistant, under
my direction and review. I have read, tested, and take responsibility for
this change.

## Summary by CodeRabbit

* **Tests**
  * Expanded configuration loading coverage for missing files, invalid YAML, empty configurations, and valid Nvidia and Hygon settings.
  * Added verification that configuration parsing returns appropriate errors and correctly preserves supported hardware settings.

### DIFF
--- a/pkg/scheduler/config/config_test.go
+++ b/pkg/scheduler/config/config_test.go
@@ -18,6 +18,7 @@ package config
 
 import (
 	"fmt"
+	"os"
 	"slices"
 	"testing"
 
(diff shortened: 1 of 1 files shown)


Eshiv-Pandey (commented):
imo, this adds coverage for a thin os.ReadFile + yaml.Unmarshal wrapper whose parsing is already covered by Test_LoadConfig in the same file, with no linked issue. idts this pr will survive. don't close it tho.. wait for the maintainers to review this

pkg/scheduler/config/config_test.go
```diff
@@ -786,3 +787,48 @@ func Test_Resourcereqs(t *testing.T) {
 		})
 	}
 }
+
+func TestLoadConfig(t *testing.T) {
```
maishivamhoo123, line 791: There is already a existing Test_LoadConfig func i think it would be very confusing to add a new TestLoadConfig func  , 
It would be better if you change the func name and in new function call only error case.
What do you think?

pkg/scheduler/config/config_test.go
```diff
@@ -786,3 +787,48 @@ func Test_Resourcereqs(t *testing.T) {
...
+	t.Run("file does not exist", func(t *testing.T) {
+		_, err := LoadConfig("/nonexistent/path/config.yaml")
+		assert.ErrorContains(t, err, "no such file or directory")
```
maishivamhoo123, line 794: I think it would be better if we use the existing   errors.Is(err, fs.ErrNotExist) instead "no such file or directory" will be better am i correct?.

maishivamhoo123 (changes requested):

pkg/scheduler/config/config_test.go
mesutoezdil, line 791: did you search for existing LoadConfig tests first?

pkg/scheduler/config/config_test.go
mesutoezdil, line 794: why does this break on windows machines?

Shouren:
@magic-peach Could we consolidate these into one `TestLoadConfig`? The existing `Test_LoadConfig` calls `yaml.Unmarshal` directly. We could write its existing fixture to a temporary file, call `LoadConfig`, and retain all its field assertions, then add subtests for missing files, invalid YAML, and empty files. This would cover the actual loading path while avoiding overlapping tests and confusing names.

PR #2740, opened 2026-08-19
fix: reject invalid device count (0/negative) across biren, cambricon…

iron-prog:
**What type of PR is this?**

/kind bug

**What this PR does / why we need it**:

Fixes invalid device-count validation in 8 device backends: `biren`, `cambricon`, `kunlun`, `vastai`, `nvidia`, `iluvatar`, `awsneuron`, and `metax`.

Previously, requesting `0` or a negative device count could return `Nums: 0`, causing `fitResourceQuota` to skip GPU quota validation. For `biren` and `vastai`, node locking was skipped as well.

The fix validates device counts using the existing `hygon` pattern:

    if n, ok := v.AsInt64(); ok {
        if n <= 0 || n > math.MaxInt32 {
            ...
            return device.ContainerDeviceRequest{}
        }
    }

**Which issue(s) this PR fixes**:

Fixes #2739

**Special notes for reviewers**:

- Added regression tests for all 8 affected backends.
- Tests cover `0`, negative values, `math.MaxInt32`, and `math.MaxInt32 + 1`.
- Verified all affected backend tests pass.
- `gofmt` and `git diff --check` are clean.
- `nvidia`, `iluvatar`, `awsneuron`, and `metax` were added after review identified the same issue in those backends.
- `ascend` is already covered by #2601 and is not included in this PR.
- `hygon` and `metax-sgpu` already have the required validation.

**Does this PR introduce a user-facing change?**

    Invalid device counts (zero or negative) are now rejected instead of silently bypassing GPU quota validation.

*Prepared with AI assistance; the implementation and tests were reviewed and verified by me.*

## Summary by CodeRabbit

- **Bug Fixes**
  - Added validation for accelerator device counts, rejecting zero, negative, and values beyond supported limits.
  - Improved memory request validation for supported accelerator types, rejecting malformed, fractional, negative, or overflowing values.
  - Improved Kunlun device allocation with stricter UUID matching and clearer failures when requested devices are unavailable.

- **Tests**
  - Expanded coverage for boundary values, invalid requests, memory parsing, and Kunlun UUID allocation scenarios.

PR #2994, opened 2026-09-09
fix(scheduler): fail closed on invalid HAMi device requests instead of binding GPU-less pods

### LINKED ISSUE #2987: mthreads: a request for 7 or more cards is scheduled with no device
**What happened**:

`MutateAdmission` in `pkg/device/mthreads/device.go` rewrites the core limit to `count * coresPerMthreadsGPU` (16) for a multi-card request, so the annotation holds a pod total.

`GenerateResourceRequests` then range-checks that same value as a per-card percentage and rejects anything over 100. At 7 cards the value is 112, the check fails, and the pod is scheduled with no device.

Either the request is satisfied with 7 cards, or it is rejected. Not scheduled with no device.

Request `mthreads.com/vgpu: 7` in a container. `zz_repro_cores_test.go` covers it: fails at count 7 and above, passes at 6.

Regression from #2784. Found by reading the code, not on a cluster, so no vendor diagnostics or node logs. mthreads hardware not available here.

**Environment**:
- HAMi version: `main`
- Accelerator vendor and model: mthreads
- HAMi device backend and requested resource names: `mthreads.com/vgpu`

Eshiv-Pandey:
What type of PR is this?

/kind bug

Before this change an invalid HAMi device request (for example an mthreads core limit outside 0-100) was silently dropped, the pod looked device-less to the scheduler, was bound to a node, and ran with no GPU at all.

- `Devices.GenerateResourceRequests` now returns `(request, error)`, so each backend can distinguish an invalid request from no request, and all 16 backends return `device.ErrInvalidDeviceRequest` on invalid input.
- `Resourcereqs` propagates the error; the scheduler Filter rejects the pod with a FilteringFailed event instead of returning all nodes, and the admission webhook returns the real error instead of a generic quota message.
- mthreads and iluvatar: the MutateAdmission `count*cores` total is divided back to a per card value, and uneven totals are rejected.

Fixes #2987

**Special notes for your reviewers:**

Mirrors the mthreads part of open PR #2961 (iluvatar) but fixes the fail-open chain for every backend.

Yes: pods with invalid HAMi device requests are now rejected at admission or filtering with a descriptive error instead of being scheduled without devices.

AI assistance was used to prepare this PR.

## Bug Fixes
- Invalid accelerator requests now return descriptive errors instead of empty requests.
- Scheduler filtering and admission checks reject invalid device, memory, and core requests with clearer messages.
- Multi-device core requests are distributed correctly when evenly divisible; uneven requests are rejected.
- Explicit zero-device counts are now accepted as valid no-device requests.
- Validation is applied consistently across supported accelerators.

## Compatibility
- Device resource request APIs now return error details for invalid requests.

### DIFF
--- a/pkg/device/cambricon/device_test.go
+++ b/pkg/device/cambricon/device_test.go
@@ -422,9 +422,11 @@ func Test_checkType(t *testing.T) {
 
 func Test_GenerateResourceRequests(t *testing.T) {
 	tests := []struct {
-		name string
-		args corev1.Container
-		want device.ContainerDeviceRequest
+		name        string
+		args        corev1.Container
+		want        device.ContainerDeviceRequest
+		wantErr     bool
+		errContains string
 	}{
 		{
 			name: "don't set to limits and request",
@@ -516,7 +518,9 @@ func Test_GenerateResourceRequests(t *testing.T) {
 					},
 				},
 			},
-			want: device.ContainerDeviceRequest{},
+			want:        device.ContainerDeviceRequest{},
+			wantErr:     true,
+			errContains: "not a plain integer",
 		},
 		{
 			name: "zero count must not silently bypass quota",
@@ -527,7 +531,9 @@ func Test_GenerateResourceRequests(t *testing.T) {
 					},
 				},
 			},
-			want: device.ContainerDeviceRequest{},
+			want:        device.ContainerDeviceRequest{},
+			wantErr:     true,
+			errContains: "out of range",
 		},
 		{
 			name: "decimal-form memory request is rejected, not treated as zero",
@@ -539,7 +545,9 @@ func Test_GenerateResourceRequests(t *testing.T) {
 					},
 				},
 			},
-			want: device.ContainerDeviceRequest{},
+			want:        device.ContainerDeviceRequest{},
+			wantErr:     true,
+			errContains: "not a plain integer",
 		},
 		{
 			name: "negative count must be rejected",
@@ -550,7 +558,9 @@ func Test_GenerateResourceRequests(t *testing.T) {
 					},
 				},
 			},
-			want: device.ContainerDeviceRequest{},
+			want:        device.ContainerDeviceRequest{},
+			wantErr:     true,
+			errContains: "out of range",
 		},
 		{
 			name: "max int32 count is accepted",
@@ -578,7 +588,9 @@ func Test_GenerateResourceRequests(t *testing.T) {
 					},
 				},
 			},
-			want: device.ContainerDeviceRequest{},
+			want:        device.ContainerDeviceRequest{},
+			wantErr:     true,
+			errContains: "out of range",
 		},
 		{
 			name: "memory overflowing int32 is rejected, not truncated to zero",
@@ -590,7 +602,9 @@ func Test_GenerateResourceRequests(t *testing.T) {
 					},
 				},
 			},
-			want: device.ContainerDeviceRequest{},
+			want:        device.ContainerDeviceRequest{},
+			wantErr:     true,
+			errContains: "not a plain integer",
 		},
 		{
 			name: "zero count must not silently bypass quota",
@@ -601,7 +615,9 @@ func Test_GenerateResourceRequests(t *testing.T) {
 					},
 				},
 			},
-			want: device.ContainerDeviceRequest{},
+			want:        device.ContainerDeviceRequest{},
+			wantErr:     true,
+			errContains: "out of range",
 		},
 		{
 			name: "decimal-form memory request is rejected, not treated as zero",
@@ -613,7 +629,9 @@ func Test_GenerateResourceRequests(t *testing.T) {
 					},
 				},
 			},
-			want: device.ContainerDeviceRequest{},
+			want:        device.ContainerDeviceRequest{},
+			wantErr:     true,
+			errContains: "not a plain integer",
 		},
 		{
 			name: "negative count must be rejected",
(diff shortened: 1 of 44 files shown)


maishivamhoo123:
@Eshiv-Pandey please resolve the conflict!

pkg/device/nvidia/device.go
```diff
@@ -538,7 +538,7 @@ func (dev *NvidiaGPUDevices) GenerateResourceRequests(ctr *corev1.Container) dev
 		if n, ok := v.AsInt64(); ok {
 			if n <= 0 || n > math.MaxInt32 {
```
maishivamhoo123, line 539: Here the n <= 0 branch in every backend now returns an error. Since the API already rejects negative values, this only fires for  n == 0 and nvidia.com/gpu: 0 is a common way to say no GPU. The webhook runs fitResourceQuota on every pod, so those pods now get denied at admission.
what do you think?

Eshiv-Pandey:
done

pkg/device/cambricon/device_test.go
```diff
@@ -664,7 +664,7 @@ func Test_GenerateResourceRequests(t *testing.T) {
 			}
 			InitMLUDevice(config)
 			dev := CambriconDevices{}
-			result := dev.GenerateResourceRequests(&test.args)
+			result, _ := dev.GenerateResourceRequests(&test.args)
```
maishivamhoo123: I think this drops the new error, the zero count case a few lines above still passes even though it now returns `ErrInvalidDeviceRequest`. Please capture it.

maishivamhoo123 (changes requested):

pkg/device/cambricon/device_test.go
Eshiv-Pandey: fixed in ae968d5.. the runner now captures the error

pkg/device/nvidia/device.go
Eshiv-Pandey: just checked pr 2740 introduced the n <= 0 rejection, so 0-as-invalid is an existing merged decision. i am here only making it visible.

pkg/device/mthreads/device.go
```diff
@@ -241,9 +241,26 @@ func (dev *MthreadsDevices) GenerateResourceRequests(ctr *corev1.Container) devi
...
+					klog.ErrorS(nil, "mthreads core request is not a non-negative integer", "container", ctr.Name, "request", core.String())
+					return device.ContainerDeviceRequest{}, &device.ErrInvalidDeviceRequest{Container: ctr.Name, Device: "mthreads", Reason: fmt.Sprintf("core request %s is not a non-negative integer", core.String())}
+				}
+				// Coresreq is a per card percentage, but MutateAdmission rewrites
+				// the limit to count*coresPerMthreadsGPU (16) whenever more than
+				// one device is requested, so with multiple devices the value
+				// is a total across the cards and has to be divided back. With
+				// a single device MutateAdmission never writes a total, so a
+				// value above 100 is plain invalid.
+				if n > 1 {
```
mesutoezdil, line 266: this divides by n whenever n>1 with no >100 guard unlike iluvatar right above, so a real per-card value like 60 with 2 cards silently becomes 30

mesutoezdil:
the n<=0 rejection only had maintainer buy-in for nvidia's gpu:0 case (per #2740), did anyone confirm 0 isn't a similar no-op convention for the other 15 backends before rejecting it everywhere?

mesutoezdil:
and resolve the conversations

pkg/device/mthreads/device.go
Eshiv-Pandey: mthreads writes count16, not count100 that gate would never fire. rewrote comment plus test.

Eshiv-Pandey:
> the n<=0 rejection only had maintainer buy-in for nvidia's gpu:0 case (per #2740), did anyone confirm 0 isn't a similar no-op convention for the other 15 backends before rejecting it everywhere?

yes i was wrong here. Shivam did mentioned it, i misread #2740. it returns an empty req, not a denial .. have made this match the other backends. also had to force push cause forgot to signoff one of the commits

mesutoezdil:
resolve conflicts pls

archlitchi:
please resolve this conflicts

Eshiv-Pandey:
resolved

mesutoezdil:
please resolve this conflicts

pkg/device/iluvatar/device.go
```diff
@@ -220,28 +225,27 @@ func (dev *IluvatarDevices) GenerateResourceRequests(ctr *corev1.Container) devi
...
 						klog.ErrorS(nil, "iluvatar core request does not divide evenly across the requested devices", "container", ctr.Name, "request", core.String(), "devices", n)
-						return device.ContainerDeviceRequest{}
+						return device.ContainerDeviceRequest{}, &device.ErrInvalidDeviceRequest{Container: ctr.Name, Device: "iluvatar", Reason: fmt.Sprintf("core request %d does not divide evenly across %d devices", corenums, n)}
 					}
 					corenums /= n
-				}
-				if corenums > 100 {
-					klog.ErrorS(nil, "iluvatar core request is out of range (must be 0-100 per device)", "container", ctr.Name, "request", core.String(), "perDevice", corenums)
-					return device.ContainerDeviceRequest{}
+				} else if corenums > 100 {
```
Shouren: @Eshiv-Pandey I think we should remove the `else` here to keep the check after dividing `coresnums`

Eshiv-Pandey: @Shouren yeah correct, fixed in the latest commit

Shouren (approved):
/lgtm

PR #2999, opened 2026-09-10
fix(device): align sparse init-container annotations from the end

### LINKED ISSUE #2965: When a container with an initcontainer requests to share a GPU, it will be scheduled to a GPU that is already occupied in exclusive mode
**What happened**:
A pod (with init container) first requests a GPU in exclusive mode and completes the allocation;
After 6 minutes, Pod B (with init container) applied for 22% of CPU cores and 20GB of memory in shared mode, and Hami allocated the GPU already used by Pod A to Pod B.

Pod B should not use GPUs that are already occupied by exclusive mode

A pod (with init container) first requests a GPU in exclusive mode and completes the allocation;
After 6 minutes, Pod B (with init container) applied for 22% of CPU cores and 20GB of memory in shared mode
The problem may lie in AppcontainersOnlyDeviceUsage()
Please include the following information:

- The Pod manifest, or at least its accelerator resource requests and limits
- Pod events and relevant accelerator-related node annotations
- Output from the vendor's diagnostic tool, when available (for example, `nvidia-smi -a` for NVIDIA devices)
- Your container runtime configuration (for example, `/etc/docker/daemon.json` or the relevant containerd configuration)
- Logs from the device plugin for the affected accelerator
- The hami-scheduler logs
- The kubelet logs from the affected node (for example, `sudo journalctl -r -u kubelet`)
- Relevant kernel output from `dmesg`

> **Security note:** Remove credentials, tokens, registry secrets, private image names, and other sensitive information before posting manifests, configuration files, or logs.

**Environment**:
- HAMi version:2.10.0
- Kubernetes version:1.29.5
- Accelerator vendor and model:
- HAMi device backend and requested resource names (for example, `nvidia.com/gpu`):
- Accelerator driver and runtime version:
- Container runtime and version:
- Workload image and tag (redact private image names and state when redacted):
- Operating system and kernel version (for example, the output of `uname -a`):
- Others:
The issue does not exist in version 2.9.0, and it should have been newly introduced in version 2.10
AppcontainersonlyDeviceusage(initcontainer.go:11s) assumes that the annotation entry index corresponds to the container order. However, in the actual main solution, the init container that does not occupy GPU does not have a placeholder entry (the user log testshare annotation only has one segment=app container). Therefore, when len(podsingle) <= numInit, the only real app occupancy entry (cidx=0) is mistakenly judged as init (a<nunInit=1) and skipped, returning an empty occupancy, resulting in GPU occupancy being cleared,

Samurai007AK:
@archlitchi 
Fixes #2965

Summary:
SteadyStateDeviceUsage treated annotation index as container order. A pod with a non-GPU init container and a GPU app container decodes to a single segment, which was skipped as init, returning empty occupancy and clearing exclusive GPU usage. Align sparse entries from the end after trimming the encoder trailing empty, with a steady-state retry for decoded sparse shapes.

Test plan:
- New regression TestSteadyStateDeviceUsage_SparseInitWithoutPlaceholder (direct and decoded shapes)
- go test ./pkg/device -count=1 -short passes
- go vet ./pkg/device clean; scheduler package does not build on Windows due to go-nvml (pre-existing, unrelated)

Validation:
- Scheduler-extender scoped change, verified with unit tests (no GPU hardware available).

---

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected device usage accounting for sparse annotations when init containers lack placeholder entries.
  - Preserved application container resource usage instead of incorrectly clearing it.
  - Improved handling of device usage data with trailing empty entries.

- **Tests**
  - Added regression coverage for sparse init-container annotations, including wire-decoded data.

### DIFF
--- a/pkg/device/initContainer.go
+++ b/pkg/device/initContainer.go
@@ -43,6 +43,25 @@ func isSidecarAt(pod *corev1.Pod, cidx int) bool {
 	return util.IsSidecarContainer(&pod.Spec.InitContainers[cidx])
 }
 
+// alignOffset maps a sparse annotation back onto container order. The encoder
+// is expected to emit a placeholder per container, but older annotations omit
+// leading non-GPU init entries, so index 0 would misclassify the first app
+// entry as init and clear exclusive usage. Trim the encoder's trailing empty
+// (exactly the surplus beyond total containers) and align any remainder
+// shortfall from the end.
+func alignOffset(podSingle PodSingleDevice, numInit, numApp int) (PodSingleDevice, int) {
+	total := numInit + numApp
+	end := len(podSingle)
+	for end > total && len(podSingle[end-1]) == 0 {
+		end--
+	}
+	trimmed := podSingle[:end]
+	if offset := total - len(trimmed); offset > 0 {
+		return trimmed, offset
+	}
+	return trimmed, 0
+}
+
 // CollapseInitContainerUsage returns the effective device usage for a pod.
 func CollapseInitContainerUsage(pod *corev1.Pod, raw PodDevices) PodDevices {
 	if raw == nil {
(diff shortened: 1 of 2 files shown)


mesutoezdil:

PR #3000, opened 2026-09-10
test(util): cover EmitNodeWarningEvent client-filter and nil-client paths

magic-peach:
**What type of PR is this?**

/kind failing-test

EmitNodeWarningEvent has a client-side filter that walks every existing event and keeps the one with the latest LastTimestamp among those matching the node UID and reason, plus an early return when no Kubernetes client is set. Neither path had a test: existing tests only ever had zero or one matching event in the fake client, so the loop that picks the most recent match and skips events with a different UID or reason never actually branched. This adds a case with several events, including ones that must be ignored by UID or reason, and a case for the uninitialized client early return.

This PR was written primarily by Claude Code, reviewed and tested locally before submitting.

Fixes #

No existing issue tracks this, found while reading pkg/util/util.go for test coverage gaps.

Verified by temporarily breaking the filter logic (skipping the UID/reason check) and confirming the new test fails, then restoring it. go test ./pkg/util/... -run TestEmitNodeWarningEvent passes with all 5 subtests.

No

## Summary by CodeRabbit

- **Tests**
  - Added coverage ensuring node warning event emission safely handles an unavailable Kubernetes client.
  - Added coverage verifying that only the latest matching event is updated, while unrelated or older events remain unchanged.

### DIFF
--- a/pkg/util/util_test.go
+++ b/pkg/util/util_test.go
@@ -923,6 +923,73 @@ func TestEmitNodeWarningEvent(t *testing.T) {
 		// Old event still present plus one new event.
 		assert.Equal(t, 2, len(events.Items))
 	})
+
+	t.Run("uninitialized client is a no-op", func(t *testing.T) {
+		oldClient := client.KubeClient
+		client.KubeClient = nil
+		t.Cleanup(func() { client.KubeClient = oldClient })
+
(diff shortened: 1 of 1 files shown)


mesutoezdil:

PR #3001, opened 2026-09-10
fix(device): enforce min limit across multiple ResourceQuota objects

### LINKED ISSUE #2997: Only one ResourceQuota per namespace is enforced, and deleting any one of them removes the limit
**What happened**:

`QuotaManager` keys limits by namespace and device resource name only (`pkg/device/quota.go:38-40`), with nothing recording which ResourceQuota object set them. `addQuotaLocked` writes straight into the slot and `delQuotaLocked` clears it unconditionally, so with two objects in one namespace that both set `limits.hami.io/gpumem`:

```
after adding strict(1000) and loose(5000): {Used:0 Limit:5000 LimitSet:true}
  -> a 3000 MiB pod is admitted, which "strict" forbids
after deleting loose:                      {Used:0 Limit:0 LimitSet:false}
  -> a 1000000 MiB pod is admitted, although "strict" is still in the cluster
```

That is master (`b5ec6b1`), driven through `s.onAddQuota` / `s.onDelQuota` and then `fitResourceQuota`. Which object wins is informer delivery order, so it can flip on a scheduler restart with no cluster change. `NewQuotaManager` hands back the same `localCache` as `GetLocalCache`, so the webhook check and the per-backend re-check added in #2397 all read this one map. #2386 made the del/add pair atomic, which does not help here: two objects share one slot, this is not a gap in time.

Kubernetes applies every ResourceQuota in a namespace, so a platform baseline quota plus a tenant quota should leave the more restrictive limit in force.

Before I send a patch I would like your call on the shape. Keying the limit side by object name and recomputing the minimum on add and delete is contained, but `Spec.Scopes` and `Spec.ScopeSelector` are not read anywhere in the tree, so a minimum taken over all objects would let a scoped quota restrict pods outside its scope. Happy to write it either way.

Two ResourceQuota objects in one namespace, both with a `limits.<hami memory resource>` entry and different values.

**Environment**:
- HAMi version: master (`b5ec6b1`)

lj361588364:
## Summary

Fixes #2997

`QuotaManager` kept a single limit slot per `(namespace, resource)`, and `addQuotaLocked`/`delQuotaLocked` overwrote/zeroed it unconditionally. When two `ResourceQuota` objects in one namespace both set e.g. `limits.hami.io/gpumem`, the surviving limit was decided purely by informer event order and flipped across scheduler restarts — deleting the larger-limit object left the resource effectively unlimited even though the stricter object still existed.

This PR records limits **per ResourceQuota object** (`namespace -> object name -> resource`) and derives the effective limit as the **minimum** across all objects, matching upstream kube quota semantics where each object is enforced independently and a pod must satisfy all of them.

## Design notes

- `Used` stays namespace-scoped: usage is a property of the pods in the namespace, not of any quota object, so it does not need per-object bookkeeping (and keeping it shared avoids losing usage when the last object is deleted).
- `FitQuota` is untouched — it keeps reading the derived `(Limit, LimitSet)`, so the scheduling hot path and the webhook check are unchanged.
- `UpdateQuota`'s atomic swap (the fix from #2386) is preserved: the swap now updates one entry in the per-object map, and an update that drops a resource from an object's spec releases that object's claim.
- Deleting an object's limits is driven by its spec, so a stale/never-added object cannot wipe another object's recorded limits.
- An explicit `"0"` from one object blocks the namespace even when other objects allow more (`min(0, x) = 0`), preserving the explicit-zero semantics from the existing `LimitSet` handling.

## Known limitation (unchanged, deterministic now)

`Spec.Scopes` / `ScopeSelector` are still not consulted. A scoped quota's limit is applied namespace-wide via the min — but the previous last-writer-wins behavior had the same over-restriction risk, this just makes it deterministic. Proper scope support (filter objects by scope when computing the min, check pods against matching scopes in `FitQuota`) is left as a follow-up.

## Tests

New cases in `pkg/device/quota_test.go`:

1. two objects, different limits → effective = min, in **both add orders**
2. delete the looser object → strict limit stays enforced (the bug scenario)
3. delete the stricter object → limit rises to the remaining value
4. in-place update of one object's limit → effective follows; dropping a resource from the spec releases it
5. delete all objects → `LimitSet=false`, usage slot preserved
6. order-independence: same object set added in forward/reverse/rotated orders → identical final state
7. explicit `0` + large limit from another object → still blocked

All pass with `-short --race -count=1`. Existing tests (`TestUpdateQuotaKeepsLimitEnforced`, `TestDelQuotaNonLimitsKey`, `TestUpdateQuota`, …) pass unmodified.

## AI Assistance Disclosure

Per CONTRIBUTING.md: this PR was developed with AI coding assistance (Kscc / Claude Code); the design was proposed and reviewed by @lj361588364.

## Summary by CodeRabbit

- **Bug Fixes**
  - Resource quota limits are now calculated as the most restrictive limit across all applicable quota objects.
  - Updating or removing a quota object correctly recalculates available capacity without affecting remaining quotas.
  - Quota results no longer depend on the order in which quota objects are processed.
  - Explicit zero limits are correctly enforced, preventing requests when any applicable quota blocks the resource.
  - Removing stale, unrelated, or emptied quota objects no longer leaves behind incorrect resource limits.

### DIFF
--- a/pkg/device/quota.go
+++ b/pkg/device/quota.go
@@ -39,7 +39,15 @@ type DeviceQuota map[string]*Quota
 
 type QuotaManager struct {
 	Quotas map[string]*DeviceQuota
-	mutex  sync.RWMutex
+	// objectLimits records what each ResourceQuota object contributes:
+	// namespace -> object name -> resource -> limit. Kubernetes enforces every
+	// ResourceQuota object in a namespace independently, so the effective limit
+	// for a resource is the minimum across all objects. Storing per object (not
+	// a single slot) keeps the derived state a pure function of the current set
+	// of objects, so the result no longer depends on informer event order or on
+	// which object happened to be added/removed last.
+	objectLimits map[string]map[string]map[string]int64
+	mutex        sync.RWMutex
 }
 
 var localCache QuotaManager
@@ -53,7 +61,8 @@ var once sync.Once
 func NewQuotaManager() *QuotaManager {
 	once.Do(func() {
 		localCache = QuotaManager{
-			Quotas: make(map[string]*DeviceQuota),
+			Quotas:       make(map[string]*DeviceQuota),
+			objectLimits: make(map[string]map[string]map[string]int64),
 		}
 	})
 	return &localCache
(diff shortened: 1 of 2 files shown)


asadjan4611:
@lj361588364  could we remove the complete recorded object entry and recalculate using the resource keys stored in obj, instead of iterating over quota.Spec.Hard...

asadjan4611:
could we avoid storing empty entries and remove the object/namespace map whn an update drops its final managed limit bcz **addQuotaLocked** currently stores an empty entry for every ResourceQuota that has noHAMi-managed limits. Since the scheduler watches all ResourceQuota objects, this can retain unnecessary state.

@lj361588364 can you please also take a look on it 
@archlitchi please corrrect to me if i wrong

pkg/device/quota.go
```diff
@@ -238,51 +247,101 @@ func (q *QuotaManager) UpdateQuota(oldQuota, newQuota *corev1.ResourceQuota) {
 	q.logQuotasLocked()
 }
 
-// addQuotaLocked requires q.mutex to be held.
+// addQuotaLocked requires q.mutex to be held. It records the object's limits
+// under its name and recomputes the effective limit for every resource the
+// object (previously or now) carries.
 func (q *QuotaManager) addQuotaLocked(quota *corev1.ResourceQuota) {
```
mesutoezdil, line 253: what happens with thousands of unrelated ResourceQuota objects?

pkg/device/quota.go
```diff
@@ -238,51 +247,101 @@ func (q *QuotaManager) UpdateQuota(oldQuota, newQuota *corev1.ResourceQuota) {
...
 	}
 }
 
-// delQuotaLocked requires q.mutex to be held.
+// delQuotaLocked requires q.mutex to be held. It removes the object's recorded
+// limits for the resources its spec mentions and recomputes the effective
+// limit from the remaining objects. Deletion is driven by the spec so an
+// object that was never added (or a stale event) cannot wipe another object's
+// recorded limits.
 func (q *QuotaManager) delQuotaLocked(quota *corev1.ResourceQuota) {
```
mesutoezdil, line 315: did you test a delete event with stale spec?

pkg/device/quota.go
lj361588364: Good catch — fixed in 15814c9. An object with no HAMi-managed limits (or one updated down to zero managed limits) now leaves no state at all: its entry is pruned, and maps the pruning leaves empty are cleaned up. Covered by TestUnrelatedQuotaLeavesNoEntry.

pkg/device/quota.go
lj361588364: Also good catch. Deletion is now driven by the recorded entry, not the event's spec: delQuotaLocked removes the whole recorded entry and recomputes from the recorded keys, so a delete event carrying a stale spec cannot leave entries behind. Covered by TestDelQuotaWithStaleSpec (added in 15814c9).

lj361588364:
@asadjan4611 Thanks for the review!

Re [your first point](https://github.com/Project-HAMi/HAMi/pull/3001#issuecomment-5651584534): that's exactly the shape delQuotaLocked has now (15814c9) — it removes the complete recorded entry for the object and recomputes the effective limit from the resource keys stored in that entry, rather than iterating over the event's `Spec.Hard`. The recorded state is the single source of truth, so a delete event carrying a stale spec can't leave anything behind (covered by `TestDelQuotaWithStaleSpec`).

Re [your second point](https://github.com/Project-HAMi/HAMi/pull/3001#issuecomment-5651624813): agreed and fixed. `addQuotaLocked` no longer stores an entry for objects without HAMi-managed limits, and an object updated down to zero managed limits is pruned along with any maps the pruning empties. The scheduler watches every ResourceQuota in the cluster, so unrelated objects now leave zero state. Covered by `TestUnrelatedQuotaLeavesNoEntry`.

mesutoezdil:
Point 4

PR #3003, opened 2026-09-10
test(nodelock): cover the remaining untested branches

magic-peach:
**What type of PR is this?**

/kind failing-test

pkg/util/nodelock had six branches with no test: SetNodeLock and ReleaseNodeLock rejecting a lock that is already held before entering the retry loop, both functions propagating an initial Get failure straight through, ReleaseNodeLock and GenerateNodeLockKeyByPod handling a nil pod, and ParseNodeLock rejecting a malformed two part annotation. Adds a test for each. Coverage for this package goes from 86.9 percent to 90.3 percent.

Fixes #

No existing issue tracks this, found while reviewing pkg/util/nodelock for test coverage gaps.

Verified with go test -race -short -count=1, all tests pass. Confirmed each new test actually exercises the intended branch by checking the coverage profile before and after.

No

## Summary by CodeRabbit

* **Tests**
  * Added coverage for node lock error handling, including nil pod inputs and malformed lock annotations.
  * Verified lock contention behavior avoids unnecessary updates.
  * Verified initial retrieval errors are propagated.
  * Confirmed releasing an already-removed lock succeeds without an additional update.

### DIFF
--- a/pkg/util/nodelock/nodelock_test.go
+++ b/pkg/util/nodelock/nodelock_test.go
@@ -1347,3 +1347,124 @@ func TestSetNodeLockNilPod(t *testing.T) {
 		t.Fatalf("expected error %q, got %q", expectedErrMsg, err.Error())
 	}
 }
+
+func TestReleaseNodeLockNilPod(t *testing.T) {
+	nodeLocks = newNodeLockManager()
+	nodeName := "nil-pod-release-node"
+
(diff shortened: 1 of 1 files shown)


maishivamhoo123:
@magic-peach please read the https://github.com/Project-HAMi/HAMi/blob/master/CONTRIBUTING.md  before raising the PR.

Issue #1903, opened 2026-05-25
MIG feature does not work


almazkhalikov:
Hi,

I am trying to use HAMi Dynamic MIG on NVIDIA A800 GPUs with the following setup:

* HAMi 2.8.0
* NVIDIA driver 580.119.02
* Kubernetes device plugin uses:
    * deviceListStrategy=cdi-annotations
    * device-id-strategy=uuid
* GPU model: NVIDIA A800 80GB PCIe

My node configuration:
```json
{
  "nodeconfig": [
    {
      "name": "nodename.ru",
      "operatingmode": "mig",
      "migstrategy": "mixed",
      "filterdevices": {
        "uuid": [
          "GPU-995b8cf7-2da8-4410-1d54-0176dae35e47"
        ],
        "index": [0]
      }
    }
  ]
}
```

Relevant device-plugin runtime flags:
```
--mig-strategy="mixed"
--device-list-strategy="[cdi-annotations]"
--device-id-strategy="uuid"
--nvidia-driver-root="/"
--resource-name="nvidia.com/gpu"
```

My HELM custom values
```yaml
devicePlugin:
   image:
    registry: "*****.ru"
    repository: "*****"
    tag: ""
    pullSecrets:
      - *****
  monitor:
    image:
      registry: "*****"
      repository: "*****"
      tag: ""
      pullSecrets:
        - *****
    resyncInterval: "1m"
  runtimeClassName: "nvidia"
  migStrategy: "mixed"
  passDeviceSpecsEnabled: true
  deviceListStrategy: "cdi-annotations"
  nvidiaHookPath: "/usr/bin/nvidia-ctk"
  nvidiaDriverRoot: "/"

  nodeConfiguration:
    config: |
      {
        "nodeconfig": [
          {
            "name": "nodename.ru",
            "operatingmode": "mig",
            "preconfigureddevicememory": 0,
            "migstrategy": "mixed",
            "filterdevices": {
              "uuid": ["GPU-995b8cf7-2da8-4410-1d54-0176dae35e47"],
              "index": [0]
            }
          }
        ]
      }

  extraArgs:
    - --device-list-strategy=cdi-annotations

```

Observed behavior:
1. MIG mode is enabled successfully:
```bash
nvidia-smi -q | grep -i "MIG Mode" -A5


MIG Mode
    Current : Enabled
    Pending : Disabled
```

2. During startup, device-plugin logs:
```bash
device 0 has no MIG devices configured
```

[...]

PR #3007, opened 2026-09-11
Change default deviceListStrategy from envvar to volume-mounts

shakir91:
**What type of PR is this?**
/kind bug

This PR changes the default value of `deviceListStrategy` from `envvar` to `volume-mounts`.

The `envvar` strategy passes the device list to the container via the `NVIDIA_VISIBLE_DEVICES` environment variable. In multi-tenant clusters, this environment variable can be overridden from inside an untrusted container image, allowing a workload to request access to GPUs it was not allocated. This has been raised as a concern in NVIDIA/k8s-device-plugin#1282, where `volume-mounts` is recommended specifically because it does not expose this override path and is required for secure GPU isolation in shared environments.

Since HAMi is commonly deployed in multi-tenant clusters where device isolation between workloads matters, defaulting to `volume-mounts` gives users a safer out-of-the-box configuration without requiring them to discover and manually override this setting.

Related: NVIDIA/k8s-device-plugin#1282

- Users who rely on the current `envvar` default (for example, in combination with Dynamic MIG per Project-HAMi/HAMi#1903) can still explicitly set `deviceListStrategy: envvar` in their values.yaml — this PR only changes the default, not the available options.
- Happy to also update the chart README / docs/config.md if this direction is agreed on.

## Summary by CodeRabbit

- **Configuration**
  - Changed the default NVIDIA device list strategy to use volume mounts instead of environment variables.

### DIFF
--- a/charts/hami/values.yaml
+++ b/charts/hami/values.yaml
@@ -440,7 +440,7 @@ devicePlugin:
   migStrategy: "none"
   disablecorelimit: "false"
   passDeviceSpecsEnabled: true
-  deviceListStrategy: "envvar"
+  deviceListStrategy: "volume-mounts"
   nvidiaHookPath: null
   # Root of the NVIDIA driver on the host. Set to "auto" to read the driver
   # and device roots from GPU Operator's driver-ready contract. If the


shakir91:
"Closing — per maintainer feedback, this should be a docs-only change, moving to Project-HAMi/website instead."

Issue #3006, opened 2026-09-11
test(device): add a shared conformance suite for the Devices interface


im-Toqeer-506:
**What would you like to be added**:

Add a shared conformance test suite for the `device.Devices` interface.

The proposed package would be:

```go
pkg/device/conformance
```

with a simple entry point:

```go
func Run(t *testing.T, dev device.Devices, f Fixture)
```

Each vendor under `pkg/device/<vendor>` could use this suite in its existing tests.

The suite should check common behaviours that every `device.Devices` implementation is expected to follow, such as:

* `Fit` refusal reasons can be parsed correctly.
* `spread` and `binpack` select devices in the expected order.
* Device selection does not depend on the input device order.
* `AddResourceUsage` does not exceed the device's total resources.
* `PatchAnnotations` produces annotations that can be decoded correctly.
* Empty resource requests return `Nums == 0`.
* `GetNodeDevices` returns an error for missing or malformed annotations instead of panicking.

The fixture would provide only vendor-specific data that the suite cannot create itself, such as a sample node, container resource request, and expected resource names.

**Why is this needed**:

There are many vendor implementations of the same `device.Devices` interface, but each backend currently tests its behaviour separately. This means the same type of bug can appear in multiple backends and require separate fixes.

For example, issues such as #2973, #2811, #2809, #2810, and #2815 exposed similar behavioural problems in different vendors.

A shared conformance suite would:

* Catch the same class of bug across multiple backends.
* Reduce duplicated test code.
* Give maintainers a common behavioural contract to review.
* Make it easier for contributors to add and test new device backends.

This would be test-only and would not change production behaviour.

**Anything else we need to know?**:

I suggest implementing this in stages:

1. Add `pkg/device/conformance` with the `Fixture` type and a few initial invariants.
2. Integrate it with one vendor backend and validate the approach.
3. Add the remaining invariants.
4. Integrate the suite with the remaining backends.

Before implementation, I would like maintainer feedback on two points:

1. Should `Fixture` be a struct containing sample inputs, or should the backend tests provide a small interface?
2. If a backend has a legitimate behavioural difference, should the test provide an explicit documented skip, or should the invariant be changed so it applies universally?

Issue #2973, opened 2026-09-05
Node spread policy inverts topology scoring for Ascend 910, Vastai and Kunlun


Gurkaran18:
**What happened**:

`NodeScoreList.Less` (`pkg/scheduler/policy/node_policy.go:48-53`) sorts descending under the `spread` node policy, and `Filter` takes the last element (`pkg/scheduler/scheduler.go:1154`) — so under `spread` the **lowest** total score wins.

`OverrideScore` (`node_policy.go:67`) compensates for that sign flip only for backends implementing the `policyNeutralScorer` marker (`node_policy.go:63-65`, weight `-10000` at line 86).

Three backends return a policy-independent, higher-is-better topology score, ignore the `policy` argument, and do **not** implement the marker. Their contribution is therefore added with the wrong sign under `spread`:

| Backend | Location | Score range |
|---|---|---|
| Ascend 910 | `pkg/device/ascend/device.go:379-414` | `0-10` (network locality x `Ascend910NetworkWeight`, `device.go:45`) |
| Vastai | `pkg/device/vastai/device.go:172-210` | `0-1` per container (AIC locality) |
| Kunlun | `pkg/device/kunlun/device.go:159-180` | `0-3000` (`calcscore`, `topo.go:76`) |

`MetaxSDevices` (`pkg/device/metax/sdevice.go:292`) is currently the only backend carrying the marker.

The result is that with `spread`, HAMi prefers the node with the **worst** device topology.

Kunlun is the severe case. `ComputeDefaultScore` caps the utilisation signal at `util.Weight * 3 = 30` (`pkg/util/types.go:103`), while `calcscore` contributes up to `3000`. The topology term does not merely invert the ranking — it swamps the spread signal entirely, so the policy stops spreading and degenerates into "always pick the worst topology".

**What you expected to happen**:

When two nodes are equally utilised and differ only in device topology, the same node should win under both `binpack` and `spread`. The node policy should decide *how utilisation is distributed*, not invert a backend's topology preference.

**How to reproduce it (as minimally and precisely as possible)**:

Reproduced against `master` at commit `42cdc78` with a unit test — no cluster or accelerator hardware required, since this is scheduler-extender scoring logic.

Save as `pkg/scheduler/policy/spread_topology_repro_test.go` and run
`go test ./pkg/scheduler/policy/ -run TestSpreadInvertsTopologyScore -v`:

```go
package policy

import (
	goflag "flag"
	"sort"
	"testing"

[...]

PR #3008, opened 2026-09-11
test(device): add a shared conformance suite for the Devices interface

im-Toqeer-506:
**What type of PR is this?**

/kind feature

Every vendor package implements the same `device.Devices` interface, but each is tested in isolation, so one conceptual defect has to be found and fixed once per backend. #2973 named three backends at once (Ascend 910, Vastai and Kunlun); #2811, #2810, #2809 and #2815 are the same shapes one vendor at a time.

This adds `pkg/device/conformance`, a suite each backend runs from its own tests. `Fixture` carries the vendor-specific inputs the suite cannot build for itself, and `Run` executes the checks as named subtests.

The six initial invariants are derived from merged fixes rather than invented:

| Invariant | Derived from |
| --- | --- |
| A refusal from `Fit` must parse under `common.ParseReason`, so the Pod gets an event naming the cause | #2881 |
| Applying what `Fit` returned must leave every device inside its own totals | #2810, #2809 |
| Reversing the candidate list must not change whether the request fits | #2811 |
| A backend whose `ScoreNode` returns a non-zero score must implement `PolicyNeutralScore()` | #2973 |
| A container requesting none of a backend's resources must yield `Nums == 0` | — |
| `GetNodeDevices` must report no devices for a node with no register annotation instead of panicking | — |

`conformance_test.go` runs every check against a backend that deliberately breaks it, so the suite is shown to fail for a broken backend rather than only to pass for a correct one.

iluvatar is wired as the first caller. Remaining invariants and backends follow in later PRs, per the staging in the issue.

Part of #3006 (stages 1 and 2 of 4 — this does not close the issue)

```release-note
NONE
```

---
**AI assistance disclosure:**
AI was used to assist with the analysis, implementation, and tests. I reviewed the changes and verified the results.

## Summary by CodeRabbit

- **New Features**
  - Added a shared policy-neutral scoring contract for device backends, enabling consistent scheduler scoring across policies.

- **Tests**
  - Added comprehensive conformance checks for device backends, including resource requests, feasibility, allocation limits, refusal reasons, scoring, and safe node handling.
  - Added fixture validation, panic-safe device queries, device-order independence, empty requests, and nodes without registration annotations.
  - Added conformance coverage for the Iluvatar and Metax device backends.

### DIFF
--- a/pkg/device/conformance/conformance.go
+++ b/pkg/device/conformance/conformance.go
@@ -1,0 +1,337 @@
+/*
+Copyright 2026 The HAMi Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package conformance holds the behavioural contract shared by every
+// device.Devices implementation.
+//
+// Each backend calls Run from its own tests. Because the checks live here
+// rather than in one vendor's test file, a violation is reported against every
+// backend that has it, instead of being found and fixed one vendor at a time
+// as happened with #2973, #2811, #2809, #2810 and #2815.
+package conformance
+
+import (
+	"fmt"
+	"slices"
+	"testing"
+
+	corev1 "k8s.io/api/core/v1"
+
+	"github.com/Project-HAMi/HAMi/pkg/device"
+	"github.com/Project-HAMi/HAMi/pkg/device/common"
+	"github.com/Project-HAMi/HAMi/pkg/util"
+)
+
+// reporter is the subset of *testing.T the checks use. The checks take this
+// interface rather than *testing.T so this package's own tests can assert that
+// a check reports a violation, which a failing subtest cannot express.
+type reporter interface {
+	Helper()
+	Errorf(format string, args ...any)
+}
+
+// policyNeutralScorer mirrors the unexported marker interface in
+// pkg/scheduler/policy. A backend whose ScoreNode returns a "higher is better"
+// score must implement it so OverrideScore can invert that score under the
+// spread policy. Go interfaces are structural, so redeclaring it here lets the
+// conformance suite check for it without importing the scheduler.
+type policyNeutralScorer interface {
+	PolicyNeutralScore()
+}
+
+// Fixture carries the vendor-specific inputs the suite cannot build for
+// itself. Node and NodeInfo are optional; the rest are required.
+type Fixture struct {
+	// Devices is a list of healthy devices the backend can allocate from. It
+	// needs at least two entries so that order-dependent behaviour is
+	// observable.
+	Devices []*device.DeviceUsage
+
+	// Request is a device request that Devices can satisfy.
+	Request device.ContainerDeviceRequest
+
+	// Pod is passed through to Fit and ScoreNode. Its annotations must not
+	// select a scheduling policy that changes whether Request fits, because
+	// several checks rely on the fixture being allocatable.
+	Pod *corev1.Pod
+
+	// Node is passed to ScoreNode. An empty Node is used when nil.
+	Node *corev1.Node
+
+	// NodeInfo is passed to Fit. One derived from Node is used when nil.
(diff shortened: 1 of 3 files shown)


pkg/device/conformance/conformance.go
```diff
@@ -0,0 +1,337 @@
...
+		util.NodeSchedulerPolicySpread.String(),
+	} {
+		if dev.ScoreNode(f.node(), podDevices, previous, policy) != 0 {
+			scores = true
+		}
+	}
+	if !scores {
+		return
+	}
+	if _, ok := dev.(policyNeutralScorer); !ok {
```
Eshiv-Pandey: this check is wrong metax inverts correctly, marker would double-invert

pkg/device/conformance/conformance.go
```diff
@@ -0,0 +1,337 @@
...
+		r.Errorf("Fit refused the fixture's own request (%q); the fixture must describe an allocatable request", reason)
+		return
+	}
+
+	byID := make(map[string]*device.DeviceUsage, len(devices))
+	for _, d := range devices {
+		byID[d.ID] = d
+	}
+
+	for _, containerDevice := range containerDevices {
```
Eshiv-Pandey: Overcounts! `score.go` ..119 applies usage only for request's type.

pkg/device/conformance/conformance.go
```diff
@@ -0,0 +1,337 @@
...
+
+type nodeDevicesResult struct {
+	infos      []*device.DeviceInfo
+	err        error
+	panicValue any
+}
+
+// safeGetNodeDevices calls GetNodeDevices and captures a panic, so that a
+// backend which panics fails its own check instead of aborting the test binary.
+func safeGetNodeDevices(dev device.Devices, node corev1.Node) (result nodeDevicesResult) {
```
Eshiv-Pandey: only GetNodeDevices guarded AWSneuron nil-map panic aborts binary

pkg/device/conformance/conformance.go
```diff
@@ -0,0 +1,337 @@
...
+			forwardFit, forwardReason, reversedFit, reversedReason)
+	}
+}
+
+// checkScoreNodeDeclaresPolicyNeutrality asserts that a backend returning a
+// non-zero node score also implements PolicyNeutralScore. NodeScoreList sorts
+// descending and Filter takes the last entry, so under the spread policy the
+// lowest score wins. An undeclared "higher is better" score is therefore
+// inverted and the worst node is picked (#2973).
+func checkScoreNodeDeclaresPolicyNeutrality(r reporter, dev device.Devices, f Fixture) {
```
mesutoezdil, line 255: did you test this against metax's existing implementation?

pkg/device/conformance/conformance.go
```diff
@@ -0,0 +1,337 @@
...
+	{"EmptyContainerRequestsNoDevices", checkEmptyContainerRequestsNoDevices},
+	{"GetNodeDevicesRejectsMissingAnnotation", checkGetNodeDevicesRejectsMissingAnnotation},
+}
+
+// Run executes every conformance check against dev.
+//
+// Do not call it from a parallel test. Backend constructors register into the
+// package-level device.InRequestDevices and device.SupportDevices maps, which
+// are not safe for concurrent use.
+func Run(t *testing.T, dev device.Devices, f Fixture) {
```
mesutoezdil, line 134: what happens if two backends run in parallel?

pkg/device/conformance/conformance.go
im-Toqeer-506: You are right thank you. The check now compares the binpack and spread scores. A marker with a per-policy score fails, and the same non-zero score without the marker also fails (#2973). A policy-aware scorer such as `MetaxDevices` is now accepted. Fixed in `7e75213`.

pkg/device/conformance/conformance.go
im-Toqeer-506: Not before. When I did, the old check wrongly rejected `MetaxDevices`. Fixed in `7e75213`. I added `TestConformance` for Metax, with topology annotations on the node so the score is non-zero, along with stub cases for both incorrect combinations.

pkg/device/conformance/conformance.go
im-Toqeer-506: In the same package, it is a data race on `device.InRequestDevices` and `SupportDevices`. I reproduced `concurrent map writes` with `-race`, which is why `Run` says not to call it in parallel. Each backend runs it from its own package, and `go test` gives each package its own process, so different backends never share those maps.

mesutoezdil:
This is being closed because it does not comply with the contribution guidelines. Point 4.

PR #3010, opened 2026-09-11
feat(scheduler): validate device resource names and defaults at startup

### LINKED ISSUE #3009: A device config with an empty or malformed resource name is accepted at startup and the backend then never matches a request
**What would you like to be added?**

Improve `validateConfig` in `pkg/scheduler/config/config.go:262` so it validates the actual values inside each backend configuration, instead of only checking whether at least one backend configuration exists.

For every backend with a non-empty configuration:

* `ResourceCountName` must not be empty.
* Every configured resource name must be a valid Kubernetes qualified name.
* `DefaultGPUNum` must be greater than `0`.
* `DefaultMemory` must be `>= 0`.
* `DefaultCores` must be between `0` and `100`.
* `MemoryFactor` must be `>= 0`.
* Two backends must not use the same `ResourceCountName`, because this would make `device.DevicesMap` routing ambiguous.

If there are multiple configuration errors, report them together and include the backend name and the invalid field, similar to the existing `initErrors` aggregation in `InitDevicesWithConfig`.

**Why is this needed?**

Currently, `validateConfig` only checks whether at least one backend configuration is present:

```go
func validateConfig(config *Config) error {
	if !reflect.DeepEqual(config.NvidiaConfig, nvidia.NvidiaConfig{}) ||
		...
		len(config.VNPUs.Configs) > 0 {
		return nil
	}
	return fmt.Errorf("all configurations are empty")
}
```

This means invalid values can pass configuration validation.

For example, an empty `ResourceCountName` or a typo in `ResourceMemoryName` can be accepted during startup. The backend is then registered, but it may never match container requests correctly or may use a resource name that does not exist on the node.

The result can be Pods that never schedule, without a clear configuration error being reported during scheduler startup.

This has already caused the need for backend-specific checks:

* **#2889** added a guard in Ascend against an empty `ResourceMemoryName`.
* `pkg/device/nvidia/device.go:443` has a similar check for an empty `ResourceMemoryName`.

These checks protect individual code paths, but the underlying configuration problem can still occur in other backends.

Validating the configuration once during startup is simpler and more reliable than adding similar checks separately to every backend.

There are three points where maintainer feedback would help before implementation:

1. **Should invalid configuration fail startup or only produce warnings?**

   My preference is to reject clearly invalid configuration during startup, including:

   * empty `ResourceCountName`
   * duplicate `ResourceCountName`
   * invalid numeric ranges

   For resource-name syntax, I'm open to either rejecting it immediately or initially warning and enforcing it in a later release.

2. **How should this relate to #2941?**

   #2941 handles unknown configuration keys, while this change validates the values of known keys.

   They solve different problems and could work together:

   * #2941 → validates **which fields exist**
   * this change → validates **whether those field values are valid**

im-Toqeer-506:
**What type of PR is this?**

/kind feature

**What this PR does / why we need it**

Previously, `validateConfig` only checked whether at least one backend configuration existed. It did not validate the actual values inside those configurations.

This PR adds validation for configuration values that can cause the scheduler to behave incorrectly.

The scheduler now **rejects the configuration at startup** when:

* Two device backends use the same `ResourceCountName`. This can cause multiple backends to handle the same container resource.
* `MemoryFactor` is negative.
* `nvidia.defaultCores` is outside the `0-100` range.
* `nvidia.defaultMemory` is negative.

The scheduler also **logs an error but continues starting** when a configured resource name is not a valid Kubernetes qualified name. This is reported because the name cannot match a Kubernetes resource, but the operator should still be able to choose the resource name.

Fixes #3009

**Special notes for the reviewer**

While implementing the validation, I checked each proposed rule against the existing code. Two rules from the original issue turned out to be incorrect, so they are not included in this PR:

1. **`ResourceCountName` cannot always be required.**

   Some backends are registered even when they are not configured. In that case, an empty `ResourceCountName` means the backend is simply inactive.

   For example, the default configuration leaves Biren, Enflame GCU, Metax-SGPU, and Vastai unconfigured. Requiring a non-empty value would break the default installation.

2. **`DefaultGPUNum` does not need to be greater than `0`.**

   For NVIDIA, `0` is intentionally used to mean that no default GPU count should be added.

   The code already checks:

   ```go
   if dev.config.DefaultGPUNum > 0
   ```

   So changing this to require a value greater than `0` would change the existing behavior.

I also checked #2889. It was not caused by an empty resource name simply failing to match a request. The issue was that Ascend could create a `Limits[""]` entry. Its fix explicitly skips the empty resource name, which is another indication that an empty resource name can be valid for an inactive or specific device configuration.

**How the validation works**

The backend configurations are different from each other, so there is no single common configuration structure that can validate everything.

Instead, resource names are checked **after the devices are initialized** using the existing `GetResourceNames()` method from the `device.Devices` interface.

This has two benefits:

* It validates the final values after backend-specific defaults have been applied.
* It works consistently across all backends, including backends that can register multiple devices from one configuration.

The NVIDIA-specific numeric checks remain in `validateConfig`, because those values are only available in the NVIDIA configuration.

`anyDeviceConfigured` is simply the existing configuration check extracted into a variable so that validation can continue after determining that at least one backend is configured. Its behavior is unchanged.

```release-note
The scheduler now rejects invalid device configuration when resource count names are duplicated, or when nvidia defaultCores, nvidia defaultMemory, or a backend memoryFactor is out of range. Invalid Kubernetes resource names are logged as errors but do not prevent startup.
```

**AI assistance disclosure**

I used an AI assistant to verify the implementation and test cases.

## Summary by CodeRabbit

- **New Features**
  - Added support for configuring and initializing Remote GPU devices.
  - Added configuration options for device extender binding and device plugin settings.

- **Bug Fixes**
  - Configuration validation now rejects empty configurations, invalid NVIDIA defaults, negative memory settings, and out-of-range core values.
  - Reports all detected configuration violations together for easier troubleshooting.
  - Detects conflicting device resource names, invalid memory factors, malformed resource names, and missing device backends.

### DIFF
--- a/pkg/scheduler/config/config.go
+++ b/pkg/scheduler/config/config.go
@@ -19,11 +19,16 @@ package config
 import (
 	"flag"
 	"fmt"
+	"maps"
 	"os"
 	"reflect"
+	"slices"
+	"strings"
 	"time"
 
 	"gopkg.in/yaml.v2"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/apimachinery/pkg/util/validation"
 	"k8s.io/klog/v2"
 
 	"github.com/Project-HAMi/HAMi/pkg/device"
@@ -254,13 +259,18 @@ func InitDevicesWithConfig(config *Config) error {
 		return fmt.Errorf("errors occurred during initialization: %v", initErrors)
 	}
 
+	if err := validateRegisteredDevices(device.DevicesMap); err != nil {
+		return fmt.Errorf("invalid device configuration: %w", err)
+	}
+
 	klog.Info("All devices initialized successfully")
 	return nil
 }
 
-// validateConfig validates the configuration object to ensure it is complete.
-func validateConfig(config *Config) error {
-	if !reflect.DeepEqual(config.NvidiaConfig, nvidia.NvidiaConfig{}) ||
+// anyDeviceConfigured reports whether at least one backend has been given a
+// configuration block.
+func anyDeviceConfigured(config *Config) bool {
+	return !reflect.DeepEqual(config.NvidiaConfig, nvidia.NvidiaConfig{}) ||
 		!reflect.DeepEqual(config.CambriconConfig, cambricon.CambriconConfig{}) ||
 		!reflect.DeepEqual(config.HygonConfig, hygon.HygonConfig{}) ||
 		len(config.IluvatarConfig) > 0 ||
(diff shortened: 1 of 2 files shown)


pkg/scheduler/config/config.go
```diff
@@ -272,10 +282,85 @@ func validateConfig(config *Config) error {
...
-		return nil
+		len(config.VNPUs.Configs) > 0
+}
+
+// validateConfig validates the configuration object to ensure it is complete.
+//
+// It checks only the values that no backend can reinterpret later. Resource
+// names are deliberately left to validateRegisteredDevices, which reads them
+// back after each backend has applied its own defaults.
+func validateConfig(config *Config) error {
```
mesutoezdil, line 293: are you aware of the #2996 hold on this?

pkg/scheduler/config/config.go
```diff
@@ -272,10 +282,85 @@ func validateConfig(config *Config) error {
...
+//
+// Reading the names back through GetResourceNames keeps this uniform across
+// backends whose configuration structs differ, covers the ones that register
+// more than one device from a single configuration block, and sees the value a
+// backend substituted for an empty field rather than the empty field itself.
+//
+// An empty name is not an error: it means the backend does not offer that
+// dimension, or has not been configured at all, and the default configuration
+// registers several backends in exactly that state.
+func validateRegisteredDevices(devices map[string]device.Devices) error {
```
mesutoezdil, line 321: did you check every default config for collisions?

pkg/scheduler/config/config.go
im-Toqeer-506: Yes i have viewed PR #2996 is about scheduler policy flags and my pr only check the device config values

pkg/scheduler/config/config.go
im-Toqeer-506: yes each default resourceCount name is in the chart is unique . Backends that are not configured are simply skiped and the deafult setup would not be rejected .

archlitchi (dismissed):
/lgtm

archlitchi:
please resolve this conflict

im-Toqeer-506:
@archlitchi 
would you prefer me to merge master into the PR branch to the history as it is or to rebase it onto the master for the cleaner history

archlitchi (approved):
/lgtm

PR #3018, opened 2026-09-12
fix(device-plugin): propagate GPU health events to every replica

0z5a:
## What changed

A critical NVML event from a physical GPU previously reached only the last device stored for that parent UUID. Track every logical device for each parent and notify all affected MIG slices and time-sliced replicas. A MIG-scoped event still filters by GPU and compute instance; unrelated parents, unknown devices, and disabled XIDs remain excluded.

Use a pointer receiver for `Device.IsMigDevice` so the health reader does not copy the concurrently updated `Health` field. Add parent-wide, MIG-scoped, unrelated-device, disabled-event, replica, and concurrent-health-update regressions.

## Validation

- `go test -race ./pkg/device-plugin/nvidiadevice/nvinternal/rm` passes on the submitted sources, including the new receiver regression.
- Earlier Linux integration exercised the actual device-plugin gRPC `ListAndWatch` path with injected NVML events: expected scenarios improved from **16/39 to 39/39**. All seven MIG slices under one parent and all four sharing replicas received the expected state. The fixed receiver passed the integration race check.
- The events were injected through the NVML interface; no physical XID fault was induced. This is a health-propagation correctness result, with no model-throughput or scheduler-performance claim.

`make verify` passed with golangci-lint 2.13.1, including the full-tree staticcheck, license-header, import-alias, and RBAC checks. The initial dependency download timeout was resolved before the successful rerun.

### DIFF
--- a/pkg/device-plugin/nvidiadevice/nvinternal/rm/devices.go
+++ b/pkg/device-plugin/nvidiadevice/nvinternal/rm/devices.go
@@ -268,7 +268,7 @@ func (d Device) AlignedAllocationSupported() bool {
 }
 
 // IsMigDevice returns checks whether d is a MIG device or not.
-func (d Device) IsMigDevice() bool {
+func (d *Device) IsMigDevice() bool {
 	return strings.Contains(d.Index, ":")
 }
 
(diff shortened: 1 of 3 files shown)


mesutoezdil:

PR #2761, opened 2026-08-21
fix(skill): align vGPU metrics summary with actual codebase definitions

haitwang-cloud:
**What type of PR is this?**
/kind documentation

**What this PR does / why we need it**:
Aligns the `hami-vgpu-metrics-summary` skill file with the actual Prometheus metrics defined in the codebase. The skill was outdated — it only listed 4 out of 12 vGPU monitor metrics and was missing the `hami_resource_quota_limit` scheduler metric, which would cause incomplete analysis when using the skill against live clusters.
**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

Doc-only change to` skill/hami-vgpu-metrics-summary/SKILL.md`. Metrics were cross-checked against `cmd/scheduler/metrics.go` and `cmd/vGPUmonitor/metrics.go`.

**Does this PR introduce a user-facing change?**:

No

## Summary by CodeRabbit

- **Documentation**
  - Added the `hami_resource_quota_limit` scheduler metric to the main and quick-reference lists.
  - Expanded vGPU monitoring documentation to cover current and legacy runtime metrics.
  - Clarified host, container, memory, utilization, kernel, context, module, buffer, and MIG metrics.
  - Documented how runtime metrics complement scheduler allocation metrics.

Issue #2685, opened 2026-08-16
docs: 7 vGPU monitor metrics undocumented in skill/hami-vgpu-metrics-summary/SKILL.md


KrishnaParihar1:
## Summary
`cmd/vGPUmonitor/metrics.go` exports several `hami_*` Prometheus metrics
that were missing from `skill/hami-vgpu-metrics-summary/SKILL.md`'s
"Runtime-only vGPU monitor metrics" list and "Quick Reference: Important
HAMi Metrics" table — the maintainer-written reference doc for
interpreting HAMi's exposed metrics.

## Undocumented metrics

| Metric | Defined at | Labels |
|---|---|---|
| `hami_container_device_memory_bytes` | `cmd/vGPUmonitor/metrics.go` | namespace, pod, container, vdevice_index, device_uuid |
| `hami_container_device_utilization_ratio` | `cmd/vGPUmonitor/metrics.go` | namespace, pod, container, vdevice_index, device_uuid |
| `hami_container_last_kernel_elapsed_seconds` | `cmd/vGPUmonitor/metrics.go` | namespace, pod, container, vdevice_index, device_uuid |
| `hami_mig_device_info` | `cmd/vGPUmonitor/metrics.go` | namespace, pod, container, vdevice_index, device_uuid, mig_uuid, profile, gpu_instance_id, compute_instance_id |
| `hami_vgpu_memory_context_bytes` | `cmd/vGPUmonitor/metrics.go` | namespace, pod, container, vdevice_index, device_uuid |
| `hami_vgpu_memory_module_bytes` | `cmd/vGPUmonitor/metrics.go` | namespace, pod, container, vdevice_index, device_uuid |
| `hami_vgpu_memory_buffer_bytes` | `cmd/vGPUmonitor/metrics.go` | namespace, pod, container, vdevice_index, device_uuid |

Verified by grepping every `.md` file in the repo for each metric name —
none appeared anywhere outside the Go source and its test file.
`hami_container_last_kernel_elapsed_seconds` in particular is useful for
detecting idle GPU allocations, and its absence from the reference doc
means it's currently invisible to anyone using the skill for GPU
usage analysis.

## A related question for maintainers
While tracing the emission code (`collectContainerMetrics` in
`cmd/vGPUmonitor/metrics.go`), I noticed `hami_container_device_memory_bytes`
is set to the same value as `hami_vgpu_memory_used_bytes` (`memoryTotal`),
[...]

PR #3023, opened 2026-09-12
test(metrics): fail when skill docs reference an unregistered metric

### LINKED ISSUE #3022: Nothing catches it when SKILL.md metric names drift from the collectors

**Labels:** `kind/feature`

Add a Go test that verifies every `hami_*` metric mentioned in `skill/*/SKILL.md` is actually registered by one of the binaries.

The test should:

- Extract all `hami_*` metric names from each `SKILL.md`.
- Compare them with the metrics registered by the collectors.
- Fail if a documented metric does not exist.
- Report registered metrics that are not documented, without failing.

Follow the existing test pattern in `dashboards/dashboard_test.go`.

**Why is this needed**:

`skill/hami-vgpu-metrics-summary/SKILL.md` currently documents 29 `hami_*` metrics, but there is no test to ensure they still exist.

Metric names are defined as string literals in the collectors, so renaming or removing one can make the skill documentation stale without any test catching it.

This has already happened in #2685 and #2761. A test would prevent this kind of documentation drift from happening again.

The test needs a canonical list of metric names. I suggest moving the metric names into `pkg/metrics` as constants and using those constants in both collectors and the test.

The scope is metric names only. No validation of query semantics or documentation prose is needed.

The test should run as part of the existing `make test`.

im-Toqeer-506:
**What type of PR is this?**

/kind feature

Adds a test that checks whether the metrics documented in `skill/*/SKILL.md` are actually registered by the collectors.

The test:

* Fails if a documented metric is not registered.
* Reports registered metrics that are missing from the documentation without failing.
* Reads metric names directly from the collector source code.

This helps prevent documentation from becoming outdated when a metric is renamed or removed. This has already happened in #2685 and #2761.

Fixes #3022

**Special notes for the reviewer:**

The test is placed in `pkg/metrics/` so it is included in the existing `go test` command without changing `hack/unit-test.sh`.

I chose to scan the collector source code instead of moving all metric names into shared constants. This keeps the change limited to testing and avoids changing 29 existing metric declarations.

The test only matches metric names inside backticks. This avoids false positives such as:

* `hami_vgpu_metrics_summarizer`
* `hami_host_gpu_`

The test currently finds 27 documented metrics, and all of them are registered.

It also reports three registered metrics that are not documented:

* `hami_host_gpu_ecc_errors_total`
* `hami_host_gpu_power_usage_watts`
* `hami_host_gpu_temperature_celsius`

These are reported but do not fail the test.

I also verified that renaming `hami_gpu_shared_count` in the collector makes the test fail and correctly points to the outdated documentation.

Validation completed successfully with `go build ./...`, `go vet`, `gofmt`, `goimports`, `addlicense`, import-alias checks, and `golangci-lint`. The full package test also passes except for the existing unrelated `TestPrepareHostPIDLockParent` failure, which also fails on clean master.

```release-note
NONE
```

**AI assistance disclosure**:

I used an AI assistant to verify the implementation and test cases.

## Summary by CodeRabbit

* **Tests**
  * Improved automated validation that metrics referenced in skill documentation correspond to registered metrics.
  * Expanded coverage to recognize supported metric definitions while ignoring comments, unrelated text, and unsupported dynamically constructed names.
  * Added scanning across command and package sources.
  * Limited undocumented-metric reporting to applicable `hami_` metrics without failing the test suite.

### DIFF
--- a/pkg/metrics/skill_metrics_test.go
+++ b/pkg/metrics/skill_metrics_test.go
@@ -1,0 +1,247 @@
+/*
+Copyright 2026 The HAMi Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package metrics
+
+import (
+	"maps"
+	"os"
+	"path/filepath"
+	"regexp"
+	"slices"
+	"strings"
+	"testing"
+)
+
+// The skill documents under skill/ name the metrics an operator is told to
+// query, but metric names are plain string literals in the collectors, so
+// renaming one leaves the documents stale with nothing to catch it. That has
+// happened twice already, in #2685 and #2761.
+//
+// These tests compare the two sides. They live in this package rather than
+// beside the documents because hack/unit-test.sh runs
+// `go test $(go list ./pkg/... ./cmd/...)`, which does not reach skill/ or
+// dashboards/.
+
+// repoRoot is this package's path back to the repository root.
+const repoRoot = "../.."
+
+var (
+	// documentedMetricPattern matches a metric name written as inline code.
+	// Metric names in the skill documents are always in backticks, which is
+	// what separates them from the skill's own identifier in the YAML front
+	// matter and from family wildcards such as `hami_host_gpu_*`, whose
+	// trailing character stops the name short.
+	documentedMetricPattern = regexp.MustCompile("`(hami_[a-z0-9_]+)`")
+
+	// registeredMetricPattern matches a metric name written as a Go string
+	// literal, which is how every collector passes it to prometheus.NewDesc.
+	registeredMetricPattern = regexp.MustCompile(`"(hami_[a-z0-9_]+)"`)
+)
+
+// documentedMetrics returns every metric name referenced by a skill document,
+// mapped to the documents that reference it.
+func documentedMetrics(t *testing.T) map[string][]string {
+	t.Helper()
+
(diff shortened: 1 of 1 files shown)


pkg/metrics/skill_metrics_test.go
```diff
@@ -0,0 +1,247 @@
...
+// registeredMetrics returns every metric name that appears as a string literal
+// in non-test Go sources under cmd/ and this package, which is where the
+// collectors declare them.
+func registeredMetrics(t *testing.T) map[string]string {
+	t.Helper()
+
+	registered := map[string]string{}
+	for _, dir := range []string{filepath.Join(repoRoot, "cmd"), filepath.Join(repoRoot, "pkg", "metrics")} {
+		err := filepath.WalkDir(dir, func(path string, entry os.DirEntry, err error) error {
+			if err != nil {
```
mesutoezdil, line 170: does this miss metrics registered outside cmd and pkg/metrics?

pkg/metrics/skill_metrics_test.go
```diff
@@ -0,0 +1,247 @@
...
+
+	// registeredMetricPattern matches a metric name written as a Go string
+	// literal, which is how every collector passes it to prometheus.NewDesc.
+	registeredMetricPattern = regexp.MustCompile(`"(hami_[a-z0-9_]+)"`)
+)
+
+// documentedMetrics returns every metric name referenced by a skill document,
+// mapped to the documents that reference it.
+func documentedMetrics(t *testing.T) map[string][]string {
+	t.Helper()
```
mesutoezdil, line 62: what about metric names built with fmt.Sprintf?

pkg/metrics/skill_metrics_test.go
im-Toqeer-506: Today every hami_ metric is declared in cmd/scheduler, cmd/vGPUmonitor or pkg/metrics, but a new collector elsewhere would have been missed.
Since 9bd174b the scan covers all of cmd/ and pkg/.

pkg/metrics/skill_metrics_test.go
im-Toqeer-506: None do today. Such a name cannot be read from the source. So the scan now records where it is along with names from constants or Namespace and Subsystem. If a doc references a metric the scan cannot find. The failure lists those places so you know where to look (9bd174b).

mesutoezdil:
Point 4.

PR #2936, opened 2026-09-01
fix(nvidia): make managed admission env mutation idempotent

asadjan4611:
**What type of PR is this?**

/kind bug

**What this PR does / why we need it**:

Makes `CUDA_TASK_PRIORITY` and `GPU_CORE_UTILIZATION_POLICY` mutation idempotent during webhook reinvocation.

The last same-name environment entry is checked because kubelet uses the last value. This avoids duplicates while still replacing a stale effective value. NVIDIA and Ascend visible-device handling from #2966 is preserved unchanged.

**Which issue(s) this PR fixes**:

Fixes #2929

#2966 covers the NVIDIA and Ascend visible-device variables. This PR completes the remaining NVIDIA-managed variables.

**Does this PR introduce a user-facing change?**:

No. Repeated admission mutation now leaves these HAMi-managed variables unchanged.

**AI assistance disclosure**:

AI assistance was used for code analysis, test development, and wording. The changes were reviewed and validated by the contributor.

Issue #2929, opened 2026-09-01
[Bug] Make NVIDIA and Ascend admission mutation safe for webhook reinvocation


asadjan4611:
## What happened?

HAMi supports `scheduler.admissionWebhook.reinvocationPolicy: IfNeeded`, but the NVIDIA and Ascend admission handlers append HAMi-managed environment variables without checking whether they were already added.

If Kubernetes reinvokes the webhook, duplicate variables may be added, including:

- `CUDA_TASK_PRIORITY`
- `GPU_CORE_UTILIZATION_POLICY`
- `NVIDIA_VISIBLE_DEVICES`
- `ASCEND_VISIBLE_DEVICES`

The default policy is `Never`, so this affects installations that explicitly use `IfNeeded`.

## What did you expect to happen?

Admission mutation should be idempotent. Mutating the same Pod twice should produce the same result as mutating it once.

## How can it be reproduced?

1. Configure the webhook with `reinvocationPolicy: IfNeeded`.
2. Create a Pod that triggers NVIDIA or Ascend environment mutation.
3. Run the admission mutation twice on the same Pod.
4. Observe that the same HAMi-managed environment variables can be appended again.

```text
First invocation  -> Add the required environment variable
Second invocation -> Find the existing value
                  -> Leave the Pod unchanged
```

## Proposed solution

Before appending a HAMi-managed environment variable, check whether the same name and value already exist.

The implementation should preserve the result and precedence of the first mutation. Only the confirmed NVIDIA and Ascend paths should be changed. Backends that are already idempotent should remain unchanged.

## Affected components

- `pkg/device/nvidia/device.go`
- `pkg/device/nvidia/device_test.go`
- `pkg/device/ascend/device.go`
- `pkg/device/ascend/device_test.go`
- `pkg/scheduler/webhook_test.go`, if webhook-level coverage is needed

## Additional information

- **Implementation complexity:** Low to Medium
- **Breaking change:** No
- **Documentation change:** No
- **Maintainer acceptance probability:** Medium–High

Issue #2558, opened 2026-08-10
[BUG] Mutating admission webhook ignores InitContainers, causing quota bypass and scheduling failures


Rickydama3:
# [BUG] Mutating admission webhook completely ignores InitContainers, causing quota bypass and scheduling failures

**What happened**:
The HAMi mutating admission webhook (`pkg/scheduler/webhook.go`) completely ignores `InitContainers` during its validation and mutation phases. It currently only iterates through `pod.Spec.Containers`. 

Because of this, if a user deploys a Pod that requests HAMi GPU resources *exclusively* in an `InitContainer` (e.g. `nvidia.com/gpu: 1`), the webhook fails to identify the GPU request. The webhook's `hasResource` flag remains `false`. Consequently, the webhook does not mutate `pod.Spec.SchedulerName` to use the HAMi scheduler, and it does not parse the requests in its `fitResourceQuota` checks.

**What you expected to happen**:
The mutating admission webhook should process `InitContainers` exactly as it processes regular containers, ensuring that:
- `MutateAdmission` is called for every `InitContainer`.
- `pod.Spec.SchedulerName` is mutated if *any* container (init or regular) requests HAMi resources.
- `fitResourceQuota` correctly evaluates the pod's total GPU resource request. Since `InitContainers` run sequentially before regular containers, the correct Kubernetes formula for the effective pod request should be applied: `MAX( MAX(InitContainer_Reqs), SUM(Container_Reqs) )`.

**How to reproduce it (as minimally and precisely as possible)**:
1. Create a pod specification that requests a HAMi GPU (e.g., `nvidia.com/gpu: 1`) inside an `InitContainer`, but does NOT request a GPU in the regular `Containers` list.
2. Apply the pod to the cluster.
3. Observe that the pod's `SchedulerName` is not patched to `hami-scheduler`, meaning the pod bypasses HAMi entirely.
4. Observe that the GPU request is not checked against the user's quota limits.

**Anything else we need to know?**:
This directly contradicts the behavior of the HAMi scheduler backend (`pkg/device/devices.go`), which *does* correctly iterate over and account for `InitContainers`. This is a critical quota/security bypass and bug that needs fixing.

**Environment**:
- HAMi version: Latest main branch
- nvidia driver or other AI device driver version: N/A
- Docker version from `docker version`: N/A
- Docker command, image and tag used: N/A
- Kernel version from `uname -a`: N/A
- Others: N/A

PR #3026, opened 2026-09-13
test(e2e): cover the mutating webhook's admission denials

### LINKED ISSUE #3020: The mutating webhook's admission denials have no end-to-end coverage

**/kind feature**

Add an E2E test suite under `test/e2e/admission/` to cover the main decisions made by the mutating webhook. Follow the structure used by `test/e2e/policy/` and reuse helpers from `test/utils/`.

The suite should cover:

* Privileged pod with a device request is denied; privileged pod without a device request is admitted.
* Invalid `hami.io/numa-alignment` value is denied during admission.
* Pod exceeding a namespace `ResourceQuota` for a HAMi resource is denied.
* Pod requesting HAMi resources has `spec.schedulerName` rewritten; a pod without HAMi resources is unchanged.
* Submitting the same pod spec twice produces the same mutation, covering webhook reinvocation.

These tests only verify admission behaviour, so they do not require physical accelerators.

**Why is this needed:**

The admission webhook currently has no E2E coverage for these paths. Several previous bugs have affected webhook behaviour, including quota bypass, non-idempotent mutations, invalid resource injection, and incorrect resource validation.

Adding E2E coverage will help catch these regressions before they reach users.

**Anything else we need to know:**

`hack/e2e-test.sh` already runs Ginkgo recursively over `./test/e2e/`, so the new suite should be picked up automatically.

The `ResourceQuota` test should create and clean up its own namespace to avoid affecting other E2E tests.

The scope is limited to admission outcomes; device allocation and scheduling are already covered by existing E2E suites.

im-Toqeer-506:
**What type of PR is this?**

/kind feature

Adds `test/e2e/admission/`, an end-to-end test suite for the decisions made by the mutating webhook when a Pod is created. It follows the structure of `test/e2e/policy/` and reuses the existing helpers in `test/utils/`.

The existing `test/e2e/` suites cover `node`, `pod`, and `policy`, but the webhook admission paths in `pkg/scheduler/webhook.go` had no E2E coverage. This is important because the webhook has had several admission-related regressions:

- #2558 — InitContainers were ignored, allowing a quota bypass.
- #2929 / #2936 — admission mutation was not idempotent during webhook reinvocation.
- #2889 — an empty `ResourceMemoryName` was injected as a map key.
- #2975 — the Ascend webhook rejected a Pod with a valid `vnpu-core` resource.

The new suite covers both negative and positive cases so that each restriction is shown to be selective:

| Case | Expected result |
| --- | --- |
| Privileged container with a device request | Denied |
| Privileged container without a device request | Admitted |
| Invalid `hami.io/numa-alignment` value | Denied |
| Valid `best-effort` NUMA alignment value | Admitted |
| Pod exceeding the namespace device-memory quota | Denied |
| Pod within the device-memory quota | Admitted |
| HAMi Pod | `schedulerName` is rewritten |
| Non-HAMi Pod | `schedulerName` is unchanged |
| Same Pod spec submitted twice | Mutations are identical |

Fixes #3020

**No accelerator is required.**

Pods expected to be admitted use a node selector that matches no nodes, so they remain Pending. The tests only verify admission and mutation, which occur before scheduling. This also prevents the suite from competing for real GPUs with other E2E tests.

`go list ./pkg/... ./cmd/...` does not include the E2E packages, so `make test` will not attempt to run these tests without a cluster.

The existing unit suite is unaffected. Its only failure, `TestPrepareHostPIDLockParent`, reproduces identically on clean master.

```release-note
NONE
```
**AI assistance disclosure**:
 I used an AI assistant to verify the implementation and check whether the test cases were passing.

## Summary by CodeRabbit

* **Tests**
  * Added end-to-end coverage for admission webhook behavior, including privileged Pods, device requests, NUMA alignment, scheduler assignment, mutation consistency, and device-memory quotas.
  * Expanded mutation consistency checks to include complete environment entries and runtime class configuration.
  * Added validation coverage for both admitted and rejected Pods.
  * Added a dedicated test suite for running admission webhook end-to-end tests.

### DIFF
--- a/test/e2e/admission/test_admission.go
+++ b/test/e2e/admission/test_admission.go
@@ -1,0 +1,276 @@
+/*
+Copyright 2026 The HAMi Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package admission covers the decisions the HAMi mutating webhook makes when
+// a Pod is created. The webhook is the first component every HAMi workload
+// meets and several of its guards fail silently when they regress: a quota
+// bypass (#2558), non-idempotent mutation under reinvocation (#2929, #2936),
+// an empty resource name injected as a map key (#2889) and a valid resource
+// rejected outright (#2975).
+//
+// The specs assert admission outcomes only, so they need no physical
+// accelerator. Pods that are expected to be admitted carry a node selector
+// that matches nothing, which keeps them Pending and stops the suite from
+// competing for real GPUs.
+package admission
+
+import (
+	"context"
+	"time"
+
+	"github.com/onsi/ginkgo/v2"
+	"github.com/onsi/gomega"
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+
+	"github.com/Project-HAMi/HAMi/test/utils"
+)
+
+const (
+	admissionNamespacePrefix = "hami-admission-e2e-"
+	admissionTestImage       = "registry.k8s.io/pause:3.9"
+	admissionScheduler       = "hami-scheduler"
+	admissionTimeout         = 2 * time.Minute
+	admissionInterval        = 2 * time.Second
+
+	gpuResourceName    = corev1.ResourceName("nvidia.com/gpu")
+	gpuMemResourceName = corev1.ResourceName("nvidia.com/gpumem")
+
+	// unschedulableNodeSelectorKey keeps an admitted Pod Pending so the suite
+	// never consumes a real device. Admission has already happened by the time
+	// scheduling is attempted, which is all these specs assert on.
+	unschedulableNodeSelectorKey = "hami.io/e2e-admission-never-matches"
+)
+
+// newPod returns a Pod that is admitted but never scheduled.
+func newPod(name string) *corev1.Pod {
+	return &corev1.Pod{
+		ObjectMeta: metav1.ObjectMeta{Name: name},
+		Spec: corev1.PodSpec{
+			RestartPolicy: corev1.RestartPolicyNever,
+			NodeSelector:  map[string]string{unschedulableNodeSelectorKey: "true"},
+			Containers: []corev1.Container{{
+				Name:  "app",
+				Image: admissionTestImage,
+			}},
+		},
+	}
+}
+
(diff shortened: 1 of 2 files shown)


test/e2e/admission/test_admission.go
```diff
@@ -0,0 +1,266 @@
...
+		deleteNamespace(client, namespace)
+	})
+
+	// A privileged container can escape the in-container limits HAMi relies on,
+	// so requesting a device from one is refused. Without a device request the
+	// Pod is none of HAMi's business and must still be admitted.
+	ginkgo.It("denies a privileged Pod that requests a device", func() {
+		pod := withPrivileged(withDeviceRequest(newPod("privileged-with-device"), 1))
+		_, err := client.CoreV1().Pods(namespace).Create(context.TODO(), pod, metav1.CreateOptions{})
+		gomega.Expect(err).To(gomega.HaveOccurred(), "a privileged Pod requesting a device must be denied")
```
mesutoezdil, line 163: your ci shows this failing, why wasn't it denied?

test/e2e/admission/test_admission.go
```diff
@@ -0,0 +1,266 @@
...
+	ginkgo.It("leaves schedulerName alone for a Pod requesting no HAMi resources", func() {
+		created := createAndCleanup(client, namespace, newPod("scheduler-untouched"))
+		gomega.Expect(created.Spec.SchedulerName).NotTo(gomega.Equal(admissionScheduler))
+	})
+
+	// Regression cover for #2929 and #2936: mutation must be a function of the
+	// submitted spec, not of how many times it has been applied. Submitting the
+	// same spec twice is a proxy for webhook reinvocation, which cannot be
+	// forced from outside the API server.
+	ginkgo.It("produces the same mutation for an identical spec", func() {
```
mesutoezdil: does this test reinvocation, or just two admissions?

test/e2e/admission/test_admission.go
```diff
@@ -0,0 +1,266 @@
...
+				if err != nil {
+					return err.Error()
+				}
+				_ = client.CoreV1().Pods(quotaNamespace).Delete(context.TODO(), created.Name, metav1.DeleteOptions{})
+				return ""
+			}, admissionTimeout, admissionInterval).Should(gomega.ContainSubstring("exceeding resource quota"),
+				"a Pod requesting more device memory than the namespace quota allows must be denied")
+		})
+
+		ginkgo.It("admits a Pod that fits within the device memory quota", func() {
```
mesutoezdil, line 262: is this ordering actually guaranteed without Ordered here?

test/e2e/admission/test_admission.go
im-Toqeer-506: Okay I am resolving this issue

test/e2e/admission/test_admission.go
im-Toqeer-506: It only tests two admissions. #2936, the fix it claims to cover, isn't merged either. I changed this locally but haven't pushed yet: the spec is renamed to mutates two admissions of an identical spec the same way and its comment no longer claims reinvocation coverage. go vet passes.

test/e2e/admission/test_admission.go
im-Toqeer-506: Yes. The Context sits inside the Ordered, Serial Describe, so it inherits the ordering; Ginkgo won't even allow BeforeAll outside an ordered container. The two quota specs also don't depend on each other. Quota usage is only counted when a Pod gets devices, and these Pods never schedul

mesutoezdil:
Point 4

PR #3029, opened 2026-09-14
fixing the awsneuron: derive core geometry per node

### LINKED ISSUE #2762: awsneuron: derive core geometry per node instead of caching the first node
## What happened

The AWS Neuron implementation stores `coresPerAWSNeuron` and `coremask` in a shared `AWSNeuronDevices` object.

When `GetNodeDevices` processes the first AWS Neuron node, it calculates the number of cores per device:

```go
if dev.coresPerAWSNeuron == 0 {
    dev.coresPerAWSNeuron = uint(coresTotal) / uint(counts)
}
```

This value is calculated only once. All nodes processed afterward reuse the value calculated from the first node.

This is a problem in clusters that contain different AWS Neuron instance types because the number of cores per device can be different:

- Inferentia1 devices have 4 NeuronCores per device.
- Inferentia2 and Trainium1 devices have 2 NeuronCores per device.

For example:

1. Node A has 1 Neuron device and 4 NeuronCores.
2. HAMi processes Node A first and stores `coresPerAWSNeuron = 4`.
3. Node B has 1 Neuron device and 2 NeuronCores.
4. HAMi processes Node B using the same `AWSNeuronDevices` object.
5. Node B incorrectly reuses the value `4` from Node A.

The `coremask` field is also stored in the shared object. It is reset and rebuilt each time `GetNodeDevices` is called:

```go
dev.coremask = 0
for i < int(dev.coresPerAWSNeuron) {
    dev.coremask *= 2
    dev.coremask++
    i++
}
```

These shared fields are not protected from concurrent access. If two code paths call `GetNodeDevices` at the same time, they may read or modify the fields concurrently.

The cached `coresPerAWSNeuron` value is also used by `PatchAnnotations` when calculating core indexes for `AWS_NEURON_VISIBLE_CORES`. Therefore, a workload can receive core indexes calculated using another node's hardware geometry.

Relevant code:

- [`AWSNeuronDevices` shared fields](https://github.com/Project-HAMi/HAMi/blob/4707fb02c91c545bc7343ce26dba4c32919f9a3e/pkg/device/awsneuron/device.go#L36-L40)
- [`GetNodeDevices`](https://github.com/Project-HAMi/HAMi/blob/4707fb02c91c545bc7343ce26dba4c32919f9a3e/pkg/device/awsneuron/device.go#L156-L197)
- [`PatchAnnotations`](https://github.com/Project-HAMi/HAMi/blob/4707fb02c91c545bc7343ce26dba4c32919f9a3e/pkg/device/awsneuron/device.go#L200-L225)

HAMi should calculate AWS Neuron hardware geometry separately for every node.

The following values must be based on the node where the workload is scheduled:

- Number of cores per device
- Core mask
- Core indexes used in `AWS_NEURON_VISIBLE_CORES`

Processing one node must not change the geometry used for another node.

For a cluster containing different AWS Neuron instance types:

- An Inferentia1 node should use its own 4-core device geometry.
- An Inferentia2 or Trainium1 node should use its own 2-core device geometry.
- The order in which nodes are processed should not change their results.
- `AWS_NEURON_VISIBLE_CORES` should contain indexes calculated from the selected node's geometry.
- Concurrent calls should not cause data races or mix information from different nodes.

## How to reproduce it

This problem can be reproduced with a focused unit test. AWS Neuron hardware should not be required for this test.

### First-node caching case

1. Create one `AWSNeuronDevices` instance.
2. Create Node A with the following resources:

```yaml
aws.amazon.com/neuron: 1
aws.amazon.com/neuroncore: 4
```

3. Call `GetNodeDevices` for Node A.
4. Create Node B with the following resources:

```yaml
aws.amazon.com/neuron: 1
aws.amazon.com/neuroncore: 2
```

5. Call `GetNodeDevices` for Node B using the same `AWSNeuronDevices` instance.
6. Inspect the device information returned for Node B.

Expected result for Node B:

```text
Count = 2
Devcore = 3
```

The core mask value `3` represents two available cores.

Current source-level result after processing Node A first:

```text
Count = 4
Devcore = 15
```

Node B incorrectly uses the four-core geometry cached from Node A.

The test should also process the nodes in the opposite order. A four-core node must not inherit the two-core geometry from a previously processed node.

### Core-index case

1. Process a node with four cores per device.
2. Process a node with two cores per device using the same `AWSNeuronDevices` instance.
3. Select a device from the two-core node.
4. Call `PatchAnnotations`.
5. Inspect `AWS_NEURON_VISIBLE_CORES`.

For device index `1` on a two-core node, the core indexes must be calculated using the two-core geometry of that node. They must not use the four-core offset cached from the previously processed node.

### Concurrency case

Add a focused test that calls `GetNodeDevices` concurrently for nodes with different core counts while using the same `AWSNeuronDevices` instance.

Run the test with:

```bash
go test -race ./pkg/device/awsneuron/...
```

The test should:

- Complete without a race-detector report.
- Always return the correct core count for each node.
- Always return the correct core mask for each node.
- Never mix geometry from different nodes.

I found this issue through source-code analysis. I have not yet tested it on real AWS Neuron hardware. The focused race-detector test has also not yet been completed.

### Root cause

The root cause is that node-specific hardware information is stored in fields belonging to one shared `AWSNeuronDevices` object.

`coresPerAWSNeuron` is initialized from the first node and is not recalculated for later nodes. `coremask` is also modified on the shared object for every call.

Hardware geometry belongs to a specific node or device. It should not be stored as one global value for the whole AWS Neuron backend.

### Suggested solution

The implementation should keep hardware geometry separate for each node.

A possible solution is:

1. Calculate cores per device separately whenever `GetNodeDevices` processes a node.
2. Calculate the core mask from that node's geometry.
3. Store or propagate the calculated geometry with the node or allocated device information.
4. Make `PatchAnnotations` use the geometry belonging to the selected node or device.
5. Make resource-request generation independent of whichever node was processed first.
6. Remove the shared mutable geometry fields, or replace them with synchronized information keyed by node or device identity.
7. Add regression tests for both possible node-processing orders.
8. Add a focused concurrent test and run it with Go's race detector.

Adding only a mutex around the current global value would not completely fix the issue. A mutex could prevent a data race, but nodes with different hardware geometry would still share the wrong value.

### Scope

This issue covers:

- Preventing the first processed node from defining the geometry of every other node.
- Calculating core counts separately for each node.
- Calculating core masks separately for each node.
- Generating core indexes from the selected node's geometry.
- Removing or safely replacing shared mutable geometry state.
- Adding regression tests for heterogeneous nodes and concurrent access.

This issue does not request a change to HAMi's existing policy for how many NeuronCores a workload may request from one device.

Supporting allocation of all four cores on an Inferentia1 device can be handled as separate work. This issue should only make the existing behavior correct for each node and independent of node-processing order.

### User impact

This problem can cause:

- Incorrect NeuronCore capacity to be recorded for a node.
- Incorrect core masks in scheduler device records.
- Incorrect core indexes in `AWS_NEURON_VISIBLE_CORES`.
- Results that change depending on which node is processed first.
- Possible data races during concurrent access.
- Workloads receiving AWS Neuron configuration based on another node's hardware.

The issue mainly affects clusters containing different AWS Neuron instance types. The concurrency concern can also affect code paths that inspect Neuron nodes at the same time.
AWS documentation confirms that the number of NeuronCores per device differs between AWS Neuron hardware generations:

https://awsdocs-neuron.readthedocs-hosted.com/en/latest/deploy/faq.html

## Acceptance criteria

- [ ] Cores per device are calculated separately for every AWS Neuron node.
- [ ] A node does not reuse geometry calculated from a previously processed node.
- [ ] Processing nodes in a different order produces the same per-node results.
- [ ] Core masks are correct for each node's hardware geometry.
- [ ] `PatchAnnotations` calculates core indexes using the selected node or device geometry.
- [ ] Resource-request behavior does not depend on which node was processed first.
- [ ] Shared AWS Neuron geometry state is removed or made node-specific and race-free.
- [ ] A unit test covers a four-core node followed by a two-core node.
- [ ] A unit test covers a two-core node followed by a four-core node.
- [ ] A regression test verifies the generated `AWS_NEURON_VISIBLE_CORES` indexes.
- [ ] A focused concurrent test passes with Go's race detector.
- [ ] Existing AWS Neuron tests continue to pass.
- [ ] HAMi's existing core-request limits are not changed unintentionally.

## Environment

- HAMi revision reviewed: `4707fb02c91c545bc7343ce26dba4c32919f9a3e`
- HAMi version: Current source revision listed above
- Kubernetes version: Not tested; source-level finding
- Accelerator vendor: AWS Neuron
- Accelerator models involved: Inferentia1, Inferentia2, and Trainium1
- Backend: AWS Neuron
- Resource names:
  - `aws.amazon.com/neuron`
  - `aws.amazon.com/neuroncore`
- AWS Neuron driver/runtime version: Not tested
- Container runtime and version: Not applicable to this source-level finding
- Container image and tag: Not applicable
- Operating system and kernel: Not tested
- Real hardware validation: Not yet performed
- Race-detector validation: Not yet completed
- Validation completed: Source-code analysis

## Logs and diagnostic information

This is a scheduler source-code issue and has not yet been reproduced in a live AWS Neuron cluster.

- Pod manifest and resource limits: Not applicable to the unit-level reproduction
- Pod events and accelerator node annotations: Not available
- Vendor diagnostic output: Not available
- Container runtime configuration: Not applicable
- Device-plugin logs: Not available
- HAMi scheduler logs: Not available
- Kubelet logs: Not applicable
- Kernel `dmesg` output: Not applicable

asadjan4611:

Fixes #2762

/kind bug

The AWS Neuron backend stored the core geometry from the first discovered node in shared fields.

In a mixed cluster, Inferentia1 nodes may expose four cores per device while Inferentia2 nodes expose two. Therefore, the first registered node could incorrectly control the geometry and core-index calculation for every later node.

This PR:

- Calculates Neuron core geometry separately for every node.
- Removes the shared mutable geometry cache and its possible data race.
- Preserves each node’s physical core stride during allocation.
- Uses the selected node’s geometry when producing `AWS_NEURON_IDS`.
- Rejects invalid capacities where total cores cannot be divided evenly by device count.
- Adds mixed-node, allocation-workflow, registration-order, invalid-capacity, and concurrency tests.

The existing maximum-two-addressable-cores allocation policy remains unchanged. Full four-core allocation support is a separate v2.11 enhancement.

**Behavior before this PR**:

If an Inferentia1 node was registered first, a later Inferentia2 node could use the wrong four-core geometry; registering nodes in the opposite order could similarly give Inferentia1 the wrong geometry.

**Behavior after this PR**:

Every node uses geometry derived from its own advertised device and core capacities, regardless of registration order or concurrent discovery.

The component workflow test covers node discovery, resource-request generation, device fitting, and final annotation creation for both Inf1-to-Inf2 and Inf2-to-Inf1 registration orders.

AI assistance disclosure:  I  used an AI Assistance(codex)  for the PR description and also for code writing solution. 

```release-note
Fixed AWS Neuron scheduling in mixed-node clusters by deriving core geometry independently for each node.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved AWS Neuron device discovery and allocation across nodes with different core layouts.
  * Added validation for invalid, inconsistent, or unsupported device capacity configurations.
  * Corrected core offset calculations when placing workloads on mixed-geometry Neuron devices, including large core counts.
  * Preserved device metadata during core usage updates.

* **Tests**
  * Added coverage for mixed-device layouts, allocation behavior, concurrent discovery, resource requests, and invalid capacity handling.


asadjan4611:
Special Note for  the reviewer:

for now i am just solving this issue #2762 then after this PR i'll implment my proposed feature  mentioned in HAMi V2.11 (Support Inferentia(4cores) for AWS Neuron devices)

archlitchi (approved):
/lgtm

PR #3037, opened 2026-09-15
fix(util): add nil guard to EmitNodeWarningEvent to prevent panic

Norway-02:
**What type of PR is this?**

/kind bug

Adds a nil guard for `node` at the entry of `EmitNodeWarningEvent` in `pkg/util/util.go`. If `node` is nil, accessing `node.Name` previously caused a runtime panic when logging or emitting warning events.

Fixes # NONE

* Modified `pkg/util/util.go` to safely return when `node == nil`.
* Added unit test `TestEmitNodeWarningEvent/nil_node_does_not_panic` in `pkg/util/util_test.go`.
* Verified via `golangci-lint run ./pkg/util/... --concurrency=2`, `go vet ./pkg/util/...`, and `go test -v ./pkg/util/...`.
* @Shouren @FouoF @archlitchi @mesutoezdil PTAL!

```release-note
NONE
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented node warning events from causing a panic when no node is provided.
  * Added a warning log and safely exits when the node information is unavailable.

* **Tests**
  * Added coverage to verify that handling a missing node completes without errors.


Norway-02:
@Shouren @FouoF @archlitchi @mesutoezdil PTAL! 

This PR adds a nil guard for `node` at the entry of `EmitNodeWarningEvent` in `pkg/util/util.go` to prevent runtime panics when accessing `node.Name` on nil node pointers. All tests and linters pass cleanly.

mesutoezdil:

PR #3040, opened 2026-09-15
fix(scheduler): verify pod identity before the extender acts on a request

mesutoezdil:
## Summary by CodeRabbit

* **Bug Fixes**
  * Scheduler requests now require a valid pod UID and verify pod identity, node assignment, and scheduling state before filtering or binding.
  * Invalid or forged requests no longer release another pod’s device reservation.
  * Pod reservations remain intact when identity or assignment checks fail.
  * Pod annotation updates are now tied to the intended pod identity, preventing updates to a different object with the same name.
  * Mismatched pod identities are rejected without deleting cached reservations.


archlitchi (approved):
/lgtm

Issue #3041, opened 2026-09-15
A pod using another scheduler can write its own GPU allocation


mesutoezdil:
If a pod uses a `schedulerName` that is not HAMi's, the webhook lets it through early, at webhook.go#L65-L70. So the quota check, the check that blocks privileged GPU pods, and the `nodeName` check never run for it. The pod also keeps any `hami.io/` annotations it was created with, nothing removes them.

Nothing later checks these annotations either. `GetPendingPod` (util.go#L81-L123) just looks at `hami.io/bind-time`, `hami.io/bind-phase` and `hami.io/vgpu-node`. Then the device plugin reads the device list right from `hami.io/vgpu-devices-to-allocate` (server.go#L964). The only check left is that the device count matches what kubelet asked for, and that number comes from the pod's own `nvidia.com/gpu` field.

So anyone who can create a pod can just write their own GPU UUID and their own memory/core numbers into these annotations. HAMi never counts this usage and never checks quota. The scheduler never even sees the pod, so it can also put another pod on the same GPU.

To reproduce, create a pod with `schedulerName: default-scheduler`, `nvidia.com/gpu: 1`, and write these annotations by hand:

```yaml
hami.io/bind-time: "1757923200"
hami.io/bind-phase: allocating
hami.io/vgpu-node: gpu-node-1
hami.io/vgpu-devices-to-allocate: "GPU-<uuid>,NVIDIA,20000,100:;"
```

I found this by reading the code on master @ 092a827 (v2.10.0 has the same code), I have not tested it on a real cluster yet. We could fix it by rejecting such pods at admission, or by removing their `hami.io/` annotations. But the real root cause is that the device plugin trusts these annotations without checking if the scheduler actually wrote them. I would like your opinion before I send a PR.

PR #3042 went with rejecting the pod at admission (on create). But that only makes the problem smaller, it does not fully close it. Three things are still open, and each one needs its own design discussion before I send a PR:

- The same annotations can still be added after the pod is created, with a simple PATCH. This works because the admission webhook only watches CREATE (see webhook.yaml), not UPDATE.
- A pod with the label `hami.io/webhook: ignore` skips admission completely. The API server does this before any HAMi code runs.
- The device plugin still trusts these annotations without checking that the scheduler wrote them. This is the root cause behind all three points above.

mesutoezdil:
[...]

PR #3042, opened 2026-09-15
fix(scheduler): deny pods created with scheduler-owned allocation annotations

mesutoezdil:
## Summary by CodeRabbit

* **Bug Fixes**
  * Pods with forged scheduler-owned device allocation annotations are now rejected during admission.
  * Validation occurs even when a pod targets a different scheduler.
  * Pods using unrelated annotations continue to be admitted normally.
  * Rejection messages identify the annotation that caused the admission failure.
  * This prevents manually supplied allocation metadata from being accepted as valid scheduling state.


asadjan4611:
I see this PR covers forged annotations during Pod creation, while the webhook-ignore, later PATCH, and device-plugin trust paths remain. Should these be handled in a follow-up PR?

mesutoezdil:
> I see this PR covers forged annotations during Pod creation, while the webhook-ignore, later PATCH, and device-plugin trust paths remain. Should these be handled in a follow-up PR?

kept those out on purpose since all three (identity-checked update validation, restricting the ignore label, device-plugin trust) need their own design discussion
therefore there is an issue for it #3041

archlitchi (approved):
/lgtm

PR #3044, opened 2026-09-15
feat(chart): expose scheduler hostNetwork and dnsPolicy

mehmetafsar510:
**What type of PR is this?**

/kind feature

`devicePlugin.hostNetwork` exists and pairs itself with `ClusterFirstWithHostNet`, but the scheduler Deployment has no equivalent.

On clusters where the pod network cannot reach kube-apiserver, the scheduler is the one that suffers: the extender never finishes registering. That is the normal state of affairs on managed offerings running Calico with restrictive network policies — Gardener shoots are the case I hit, and some kOps setups behave the same way.

The only way out today is to re-apply this after **every** `helm upgrade`:

```bash
kubectl patch deployment hami-scheduler -n kube-system --type=strategic \
  -p '{"spec":{"template":{"spec":{"hostNetwork":true,"dnsPolicy":"ClusterFirstWithHostNet"}}}}'
```

This adds `scheduler.hostNetwork` and `scheduler.dnsPolicy`, mirroring the device plugin:

```yaml
scheduler:
  hostNetwork: false
  dnsPolicy: ""
```

An empty `dnsPolicy` selects `ClusterFirstWithHostNet` when `hostNetwork` is on, and otherwise leaves the Kubernetes default in place. It can still be set explicitly — `Default`, for instance, to use the node's own `resolv.conf`.

Defaults are unchanged. `hostNetwork` stays `false` and no `dnsPolicy` is rendered, so existing releases render exactly as they do today.

One caveat worth a look, and it is documented in `values.yaml` rather than left for users to discover: with `hostNetwork: true` the extender/webhook port (`scheduler.service.httpTargetPort`, 9443 by default) and the metrics port (`scheduler.metricsBindAddress`, `:9395`) bind on the node itself, so both must be free on every node the scheduler can land on.

Testing — `helm lint` passes, and the chart was rendered three ways:

| values | result |
|---|---|
| defaults | `hostNetwork: false`, no `dnsPolicy` — identical to before |
| `scheduler.hostNetwork=true` | `hostNetwork: true` + `dnsPolicy: ClusterFirstWithHostNet` |
| `scheduler.hostNetwork=true,scheduler.dnsPolicy=Default` | override honoured |

Chart README updated.

```release-note
Added `scheduler.hostNetwork` and `scheduler.dnsPolicy` chart values, so clusters whose pod network cannot reach kube-apiserver no longer need a manual `kubectl patch` after every `helm upgrade`.
```

## Summary by CodeRabbit

* **New Features**
  * Added Helm chart options to run the scheduler on the host network when required for Kubernetes API server connectivity.
  * Added configurable DNS policy support, including an automatic host-network-compatible default.
  * Documented the new scheduler networking settings and their defaults.


mesutoezdil:

PR #3045, opened 2026-09-15
feat(chart): expose scheduler.strategy

mehmetafsar510:
**What type of PR is this?**

/kind feature

The scheduler Deployment leaves `strategy` unset, so Kubernetes applies its default `RollingUpdate` with 25% maxSurge. That is the right default for a multi-node cluster, but there are two setups where the surge pod can never become ready while the old one is still running:

- **a single schedulable node** — a test cluster, or `scheduler.nodeName` / a node label selector pinning the scheduler to one node
- **hostNetwork** — the surge pod cannot bind the extender/webhook port while the old pod still holds it

In both cases the rollout stalls until the old ReplicaSet is deleted by hand, and there is no value to set `type: Recreate` instead.

This exposes `scheduler.strategy`, rendered only when set:

```yaml
{{- with .Values.scheduler.strategy }}
strategy:
  {{- toYaml . | nindent 4 }}
{{- end }}
```

The default is `{}`, which renders nothing, so the Deployment keeps getting the Kubernetes default and no existing release changes. I chose that over shipping an explicit `RollingUpdate` block so that the chart does not pin values Kubernetes may evolve.

Testing — `helm lint` passes, and the chart was rendered with defaults (no `strategy` key emitted, verified by grep), with `type: Recreate`, and with an explicit `RollingUpdate` carrying `maxUnavailable: 0` / `maxSurge: 1`. Chart README updated.

```release-note
Added the `scheduler.strategy` chart value, allowing `type: Recreate` where the old and new scheduler pods cannot run side by side.
```

## Summary by CodeRabbit

* **New Features**
  * Added configurable deployment strategies for the scheduler.
  * The default remains Kubernetes’ `RollingUpdate` behavior.
  * Set the strategy to `Recreate` when old and new scheduler pods cannot run simultaneously.
  * Documented the new Helm setting and configuration examples.


mesutoezdil:

PR #3046, opened 2026-09-15
feat(chart): bound the gpuOperatorToolkitReady wait

mehmetafsar510:
**What type of PR is this?**

/kind feature

The `toolkit-validation` init container loops on `toolkit-ready` with no bound:

```sh
until [ -f /run/nvidia/validations/toolkit-ready ]; do
  sleep 5
done
```

That is correct while NVIDIA GPU Operator is the thing installing nvidia-container-toolkit, because the file is guaranteed to appear.

It is a trap on nodes whose image already ships the toolkit. Several managed offerings pre-bake it, and so do hand-built GPU node images. Nothing ever writes the file, the init container waits forever, and the device plugin never reaches Running. The pod sits in `Init:0/1` with no event and no failure to point at — which is the part that costs time, because nothing in the cluster says what it is waiting for.

This adds a bound and a choice of what to do when it is hit:

```yaml
devicePlugin:
  gpuOperatorToolkitReady:
    enabled: false
    hostPath: "/run/nvidia/validations"
    timeoutSeconds: 0      # NEW - 0 waits forever
    skipIfMissing: false   # NEW - on timeout: fail (false) or continue (true)
```

Past the timeout the init container either **fails**, so the condition is visible in the pod status instead of hiding in a wait, or **logs and continues**, for nodes known to be ready without GPU Operator having said so.

Two defaults chosen deliberately, both in the conservative direction:

- `timeoutSeconds: 0` waits forever, exactly as today, so GPU Operator clusters are unaffected.
- Of the two timeout outcomes, failing is the default, because a silent skip can mask a genuinely broken toolkit install.

The loop stays POSIX sh and parses clean under `dash -n`. It was exercised as a standalone script across five cases:

| case | result |
|---|---|
| file present, `timeout=0` | exits 0 immediately |
| file missing, `timeout=0` | still waiting when killed at 12s (current behaviour preserved) |
| file missing, `timeout=10`, `skipIfMissing=false` | exits 1 with the path on stderr |
| file missing, `timeout=10`, `skipIfMissing=true` | exits 0, logs that it continued |
| file appears at 7s, `timeout=20` | exits 0 as soon as it appears |

`helm lint` passes and the rendered init container was inspected.

The chart README did not document `gpuOperatorToolkitReady` at all, so this adds rows for the whole block, not only the two new fields. Happy to split that into its own PR if you would rather keep this one to the template change.

```release-note
Added `devicePlugin.gpuOperatorToolkitReady.timeoutSeconds` and `skipIfMissing`, so the toolkit-ready wait can be bounded on nodes that ship nvidia-container-toolkit without NVIDIA GPU Operator. The default is unchanged and still waits indefinitely.
```

## Summary by CodeRabbit

* **New Features**
  * Added configurable readiness checks for NVIDIA GPU Operator toolkit installation before starting the device plugin.
  * Added timeout controls, including an option to continue when the readiness marker is unavailable.
  * Added configuration for the validation container’s security context and marker file location.

* **Documentation**
  * Documented the new readiness-check settings, defaults, and timeout behavior in the Helm chart documentation.


mesutoezdil:

PR #3047, opened 2026-09-15
fix(nodelock): release a lock held by a terminated pod

mehmetafsar510:
**What type of PR is this?**

/kind bug

`LockNode` already recovers a lock whose owner has been deleted, but the check is existence-only — it calls `Pods().Get()` and treats only `NotFound` as dangling:

```go
if _, err := client.GetClient().CoreV1().Pods(ns).Get(ctx, previousPodName, metav1.GetOptions{}); err != nil {
    if !apierrors.IsNotFound(err) { ... }
    // dangling -> take over
    skipOwnerCheck = true
}
```

A pod that is still an object in the API yet has reached a terminal phase does not match, so the lock survives.

That is the common case rather than a corner one. A pod rejected at admission (`UnexpectedAdmissionError` surfaces as `Failed`), evicted, preempted, or simply finished keeps its Pod object until something removes it — a Job history limit, a user, or nothing at all for a bare pod. It will never reach `Allocate`, so it will never release the lock itself.

Meanwhile every other HAMi pod targeting that node contends until `NodeLockTimeout` expires — 5m by default (`scheduler.nodeLockExpire`). On a cluster with few GPU nodes that is a visible scheduling stall, and it repeats on every admission failure.

This treats `Succeeded` and `Failed` owners as dangling, the same as a deleted one. Both phases are terminal and neither holds devices, so taking the lock over is safe. Owners in `Pending` or `Running` still contend exactly as before, so a live pod cannot lose the lock out from under it.

On scope — this deliberately does **not** add a controller or a reconcile loop. The existing recovery point in `LockNode` already runs at exactly the right moment, when someone actually wants the lock, so widening its notion of "dangling" is all that is needed.

Both directions are covered by tests, and both were confirmed to fail before the fix:

- `TestLockNodeTerminalOwner` — a `Failed` and a `Succeeded` owner must both be taken over, with the annotation verified to name the new pod. It uses a **fresh** lock timestamp so the `NodeLockTimeout` path cannot mask the result, and it creates the owner pod so the `NotFound` path cannot either.
- `TestLockNodePreservesRunningOwner` — a `Pending` and a `Running` owner must still return `ErrNodeLockContention`.

`go test -race ./pkg/util/nodelock/...` passes, as do `./pkg/scheduler/...`, `./pkg/device/...` and `./pkg/util/...`. `gofmt` and `go vet` are clean.

```release-note
Fixed a node lock held by a pod that has already terminated (for example one rejected at admission) blocking other HAMi pods from that node until the lock timeout expired.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Node locks are now released when the owning pod has completed, failed, been evicted, or otherwise reached a terminal state.
  * Other pods can re-acquire the node without waiting for the lock timeout.
  * Locks held by pending or running pods continue to be preserved.


mesutoezdil:

PR #3043, opened 2026-09-15
fix: disable MIG when using hami-core

FouoF:
**What type of PR is this?**

/kind bug

**What this PR does / why we need it**:

Add MIG status assert in hami-core mode. Trying disable MIG card by card. Retry when still program running. If need reset GPU guide the user to deal with it.

**Which issue(s) this PR fixes**:
Fixes #2974 

**Special notes for your reviewer**:

Generated by Grok, tested in A100 instance. 

**Does this PR introduce a user-facing change?**:

## Summary by CodeRabbit

* **New Features**
  * Non-MIG operating modes can automatically disable MIG on idle GPUs.
  * MIG startup handles idle GPU cleanup and adopts existing annotated allocations.
  * NVML-backed device discovery and topology scoring work across operating modes.

* **Bug Fixes**
  * Startup now reports reset-required MIG changes with reboot instructions and documentation guidance.
  * MIG remains enabled when GPUs have active allocations, preventing unsafe changes.
  * Startup fails instead of advertising hami-core when MIG cannot be disabled.

* **Documentation**
  * Clarified dynamic MIG session ownership, startup behavior, health checks, and failure handling.

PR #3049, opened 2026-09-15
refactor(vGPUMonitor): cleanup the temporary files of HAMi-Core in Device Plugin

Shouren:
**What type of PR is this?**
/kind feature

While reviewing the Flexible MIG feature earlier, I identified an opportunity to improve its ownership model. This refactor clarifies ownership of vGPU cache management: the device plugin manages cache directories as part of the container allocation lifecycle, while vGPUmonitor focuses on reading usage data and exposing metrics. 

It also reduces repeated Kubernetes API calls and makes cleanup and MIG recovery safer during startup and restarts.

The main changes in this PR:
1. Move cache management into the device plugin
2. Share a node-scoped Pod informer
3. Harden MIG recovery
4. Improve monitor resource handling

Fixes #

An independent toggle for vGPUmonitor is not included in this PR because it still does more than collect metrics. Its feedback loop updates the recentKernel and utilizationSwitch fields in shared memory based on task priority and GPU contention. Disabling it would therefore also disable this feedback control, not just monitoring.

Moving cache ownership to the Device Plugin removes one dependency on vGPUmonitor, but does not yet make it a purely observational component. The feedback responsibilities need to be addressed before introducing a toggle that affects observability only.

This PR may come into conflict with #3043, so I will make ready after that PR got merged.

No

**AI Assistance Notice**:
I designed the overall refactoring architecture, and Codex generated most of the code implementation. I have thoroughly reviewed the code and successfully completed local testing.


pkg/device-plugin/nvidiadevice/nvinternal/vgpucache/manager.go
```diff
@@ -0,0 +1,264 @@
...
+				scanErr = errors.Join(scanErr, fmt.Errorf("recheck vGPU cache entry %s: %w", target, err))
+			}
+			continue
+		}
+		if !os.SameFile(before, current) || !before.ModTime().Equal(current.ModTime()) {
+			klog.InfoS("Skipping vGPU cache entry changed during GC", "directory", target)
+			continue
+		}
+		if err := os.RemoveAll(target); err != nil {
+			scanErr = errors.Join(scanErr, fmt.Errorf("remove vGPU cache directory %s: %w", target, err))
```
mesutoezdil, line 316: does an undeletable entry relist the api every scan forever?

cmd/device-plugin/nvidia/main.go
```diff
@@ -270,9 +279,31 @@ func loadConfig(c *cli.Context, flags []cli.Flag) (*spec.Config, error) {
...
+		return fmt.Errorf("create device-plugin node Pod informer: %w", err)
+	}
+	listLiveNodePods := func() ([]*corev1.Pod, error) { return nodePods.ListFresh(processCtx) }
+	cacheManager, err := vgpucache.New(resolveVGPUCacheConfig(), nodePods.List, listLiveNodePods)
+	if err != nil {
+		return fmt.Errorf("create vGPU cache manager: %w", err)
+	}
+	nodePods.Start(processCtx)
+	syncCtx, cancelSync := context.WithTimeout(processCtx, 30*time.Second)
+	err = nodePods.WaitForSync(syncCtx)
```
mesutoezdil: should a slow api server at boot block serving devices?

cmd/device-plugin/nvidia/main.go
```diff
@@ -53,10 +57,15 @@ type options struct {
...
-	hostNvidiaDriverRoot        = "/"
-	gpuOperatorNvidiaDriverRoot = "/run/nvidia/driver"
-	hostContainerRoot           = "/host"
+	autoNvidiaDriverRoot         = "auto"
+	hostNvidiaDriverRoot         = "/"
+	defaultVGPUCacheRoot         = "/usr/local/vgpu/containers"
+	defaultVGPUCacheScanInterval = 5 * time.Second
+	defaultVGPUCacheGracePeriod  = 5 * time.Minute
+	vgpuCacheRootEnvName         = "HAMI_VGPU_CACHE_ROOT"
+	vgpuCacheGracePeriodEnvName  = "HAMI_RESYNC_INTERVAL"
```
mesutoezdil: lowering this for fresher metrics also shortens the gc grace, intended?

cmd/device-plugin/nvidia/main.go
Shouren: @mesutoezdil It is required for MigMode but not for `hami-core,` so I add a check for operating mode. Now it only syncs for MigMode.

pkg/device-plugin/nvidiadevice/nvinternal/vgpucache/manager.go
Shouren: @mesutoezdil Yes, it will lead to a period scan which lasts forever. I added a backoff mechanism for retrying and an `VGPUCacheCleanupAbandoned` warning event after the last retry failed, so now the unrecoverable deletion error would not make the scan last forever.

cmd/device-plugin/nvidia/main.go
Shouren: @mesutoezdil No, it was inherited from previous vGPUMonitor. So i added a new env for it.

PR #3053, opened 2026-09-15
fix(scheduler): count GPU app containers correctly with init containers

### LINKED ISSUE #3052: fix(scheduler): count GPU app containers correctly with init containers
**What happened**:

HAMi can accept a Pod above its GPU-memory `ResourceQuota` when the Pod contains a non-GPU init container.

A focused scheduler test uses one healthy NVIDIA GPU with 40000 MB memory, a namespace quota of `limits.hami.io/gpumem: 24000`, and two app containers that each request `hami.io/gpu: 1` and `hami.io/gpumem-percentage: 40`.

Each app container receives 16000 MB. The total app memory is therefore 32000 MB.

Without an init container, HAMi rejects the placement because 32000 MB exceeds the 24000 MB quota.

After adding one init container with no GPU request, HAMi accepts the same placement.

Adding an init container that does not request a GPU must not change how HAMi counts the app containers.

Both cases should be rejected because the app containers need:

- App container A: 16000 MB
- App container B: 16000 MB
- Total: 32000 MB
- Quota: 24000 MB

This was reproduced with a focused scheduler unit test.

1. Register one healthy NVIDIA GPU with 40000 MB memory and enough sharing slots.
2. Create a namespace quota with `limits.hami.io/gpumem: 24000`.
3. Create a Pod with two app containers.
4. Set the following limits on both app containers:

   `hami.io/gpu: 1`

   `hami.io/gpumem-percentage: 40`

5. Add one init container without accelerator resource requests.
6. Run the normal scheduler scoring path.

Observed result:

- Without the init container, HAMi rejects the 32000 MB allocation.
- With the non-GPU init container, HAMi accepts the same 32000 MB allocation.

`allocateAppContainers` creates a temporary allocation list containing only app containers.

NVIDIA `fitQuota` passes this list to `CollapseInitContainerUsage`. That helper expects entries to have the same positions as the original Pod containers.

When a Pod has an init container, the first app allocation can be treated as an init-container allocation. HAMi then uses a peak value instead of adding both app-container allocations. This allows an allocation above the namespace quota.

Suggested fix:

- Preserve the original container positions in the temporary allocation used for quota checks.
- Keep empty entries for containers without GPU allocations.
- Add regression tests for:
  - A non-GPU init container followed by GPU app containers
  - A GPU init container followed by GPU app containers
  - Native sidecar init containers
  - Valid sequential init and app GPU reuse

Related issues:

- #2828 and #2836 fixed a similar container-position problem in the NUMA-refit path.
- This issue affects normal scheduling and does not require NUMA refitting.

This issue was reproduced with mocks and a focused unit test. Pod events, node annotations, vendor diagnostics, runtime configuration, device-plugin logs, scheduler logs, kubelet logs, and `dmesg` are unavailable because no live cluster was used.

**Environment**:

- HAMi revision: `a7e3ffd66891d0fddfc5e331572cab3375c6bf3e`
- Kubernetes version: Not tested on a live cluster
- Accelerator vendor and model: Simulated NVIDIA GPU with 40000 MB memory
- HAMi backend and requested resource names: NVIDIA; `hami.io/gpu`, `hami.io/gpumem-percentage`
- Accelerator driver and runtime version: Not applicable to the mock test
- Container runtime and version: Not applicable to the mock test
- Workload image and tag: Not applicable to the mock test
- Operating system and kernel version: Linux `6.8.0-138-generic`
- Go version: `1.27.0`
- Validation: Reproduced three times; the no-init-container control case was rejected correctly

asadjan4611:
**What type of PR is this?**

/kind bug

This PR fixes GPU-memory quota accounting for Pods with init containers.

Previously, HAMi could treat the first app container as an init container and under-count GPU memory. This could allow a Pod to exceed its GPU-memory `ResourceQuota`.

The fix preserves the original positions of init and app containers during quota checks.

```text
Pod allocations
      ↓
Preserve container positions
      ↓
Calculate correct GPU usage
      ↓
Accept within quota or reject above quota
```

Fixes #3052

Only the scheduler allocation logic and its regression tests are changed. No dependencies or configuration are changed.

```release-note
Fixed GPU-memory ResourceQuota accounting for Pods containing init containers.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved device placement for pods with init containers and sidecars.
  * Ensured regular init containers do not unnecessarily constrain application container placement.
  * Improved quota validation using complete pod allocation history.
  * Prevented incorrect node selection when resource quota limits are exceeded.
  * Standardized NVIDIA device model names for consistent reporting.

* **Tests**
  * Added coverage for init-container placement, sidecar allocations, quota validation, and application resource allocation.


DSFans2014 (commented):
please resolve the conflict

asadjan4611:
@DSFans2014 resolved the conflicts

Shouren:
@asadjan4611 After investigating, The "GPU-memory quota accounting for Pods with init containers" issue has been fixed by #2952.

PR #3056, opened 2026-09-16
feat: add Inf1 four-core Neuron support

### LINKED ISSUE #2873: Support all 4 NeuronCores on AWS Inferentia (Inf1) devices

Remove the hardcoded 2-core-per-device limitation in the `awsneuron` backend so AWS Inferentia (Inf1) instances can use all 4 NeuronCores per chip instead of only 2.

Specifically, in `pkg/device/awsneuron/device.go`, `maxCoresPerNeuronDevice` is currently fixed at `2`. `coresPerDevice()` therefore clamps the observed per-device core capacity to 2, even when the underlying Inf1 device provides 4 NeuronCores.

The same two-core assumption is also present in the allocation and annotation logic:

- `addCoreUsage` uses a fixed two-bit representation for available cores.
- `PatchAnnotations` only checks/emits the first two core indexes.

The proposed change is to:

1. Remove the fixed `maxCoresPerNeuronDevice = 2` ceiling and use the observed per-device capacity.
2. Generalize `addCoreUsage` to support an N-bit core allocation mask.
3. Generalize `PatchAnnotations` to emit all allocated core indexes rather than only the first two.
4. Add regression tests for 1-, 2-, 3-, and 4-core requests on a 4-core Inf1-shaped device.
5. Verify that existing 2-core behavior remains unchanged for Inf2/Trn1.
6. Check whether Trn2 requires the same generalization.

No device-plugin-side change is expected to be required because HAMi uses AWS's official Neuron Device Plugin rather than providing its own Neuron device plugin.

**Why is this needed**:

This follows up on #2762, which appears to have been closed by the stale-issue bot rather than resolved on the technical merits, and addresses the "Support Inferentia(4cores) for AWS Neuron devices" item in the #2764 roadmap.

The current limitation is also reflected in the existing code and tests. `device_test.go` contains a `Test_splitCoreRequest` case describing that an Inferentia chip has four cores but only two are currently addressable.

As a result, an Inf1 device can currently be represented approximately as:

```text
Inf1 physical device:  4 NeuronCores
HAMi capacity:         2 cores

core 0  ✓
core 1  ✓
core 2  ✗
core 3  ✗
```
This can leave up to half of the available NeuronCore capacity on Inf1 devices unschedulable by HAMi.

The expected behavior is:
```
Inf1:
    4 physical NeuronCores → 4 schedulable cores

Inf2/Trn1:
    2 physical NeuronCores → 2 schedulable cores
```

**Anything else we need to know?:**

I'd like to take this issue. For the implementation, my initial preference is to continue deriving the per-device core count from the observed node capacity (`coresTotal / counts`) and remove the unnecessary fixed clamp, rather than introducing an instance-type lookup table. This avoids maintaining a separate mapping of AWS instance types and keeps the scheduler based on the capacity reported by the cluster.

Before implementing, I would also like to verify whether there are any known cases where Neuron core-capacity reporting can be unreliable and whether Trn2 needs the same treatment.
The intended change is limited to the `awsneuron` scheduler/backend and its tests, with particular attention to preserving existing `Inf2/Trn1` behavior and adding regression coverage for 4-core Inf1 devices.
cc @archlitchi  @mesutoezdil

iron-prog:
**What type of PR is this?**

/kind feature

This PR adds support for AWS Inf1 Neuron devices with four addressable cores per device.

The existing AWS Neuron implementation assumes a fixed two-core-per-device model. This assumption affects device capacity calculation, core-mask handling, allocation, and annotation generation when running on Inf1 hardware.

This PR updates the Neuron device handling to derive the addressable core capacity from the observed device information instead of relying on the fixed two-core assumption. It also adds support for four-core devices while preserving the existing behavior for two-core Neuron devices.

Additional tests cover:
- four-core Inf1 device capacity
- partial core requests on four-core devices
- whole-device allocation
- core-mask generation
- core usage tracking
- existing two-core device behavior

Fixes #2873

The main point to review is the change from the fixed two-core Neuron assumption to using the observed/addressable core capacity.

The implementation needs to keep the existing two-core Neuron behavior unchanged while correctly representing four-core Inf1 devices.

The AWS Neuron package tests pass, including the new four-core cases.

Yes.

This adds AWS Inf1 four-core Neuron support, allowing HAMi to correctly represent and allocate Neuron cores on supported Inf1 hardware. Existing two-core Neuron behavior remains unchanged.

## Summary by CodeRabbit

* **Bug Fixes**
  * AWS Neuron devices now recognize and allocate their full reported core capacity instead of being limited to two cores.
  * Core masks and resource annotations now include all available and allocated core indexes.
  * Core requests support partial allocations and full four-core device allocations, with clearer validation for unsupported odd-count requests.
  * Core allocation results now accurately identify newly assigned cores while preserving overall device usage tracking.


mesutoezdil:

iron-prog:

@mesutoezdil  could you point me to which guideline this didn't meet? I want to make sure I follow the right process before resubmitting.

Issue #3019, opened 2026-09-12
Scheduler resolves device UUIDs with a linear scan on every Filter call


im-Toqeer-506:
**Labels:** `kind/feature`

**What would you like to be added:**

Add a UUID-based index for node devices when `NodeUsage` is created in `buildNodeUsage`. Use this index instead of scanning the full device list in `getNodesUsage` and `fitInDevices`.

**Why is this needed:**

Both `getNodesUsage` and `fitInDevices` currently find devices by looping through the entire `DeviceLists` slice.

This happens on every `Filter` call, and `getNodesUsage` does it while processing cached pods. Using a UUID map will make device lookups faster and remove the manual linear search and `targetIdx == -1` checks.

The improvement is more noticeable on nodes with more devices, such as 16 GPUs instead of 4.

The index must also be rebuilt when `NodeUsage` is copied. Each scoring goroutine gets its own deep copy, so sharing the map could cause allocations to update the wrong `DeviceUsage`.

No scheduling behaviour should change. `DeviceLists` will remain the slice used for ordering and sorting.

An explicit test should verify that the index is correctly rebuilt after copying.

This should be merged after **#2978**, which already has `lgtm`/`approved` and modifies the same functions. Its new benchmarks can also be used to measure the improvement.

im-Toqeer-506:
Closing this after benchmarking. The UUID index showed no meaningful performance improvement and slightly increased allocations. The real bottleneck is `buildNodeUsage` snapshot construction, so the original assumption was incorrect. I may open a separate issue for that.

PR #3064, opened 2026-09-17
 index device UUIDs once per scrape

### LINKED ISSUE #3058: perf(metrics): index device UUIDs once per scrape instead of scanning all nodes per container device

Improve device UUID lookup in the scheduler metrics collector.

Currently, `findNodeDeviceUsage` scans all nodes and their devices every time an allocated device is processed. Since this happens inside `collectContainerMetrics`, the same cluster-wide scan can run many times during a single Prometheus scrape.

Instead, build a device UUID lookup map once per scrape and use it for the remaining lookups.

The proposed change is:

* Build a `map[string]deviceMeta` from the snapshot returned by `InspectAllNodesUsage`.
* Pass this map to `collectContainerMetrics`.
* Replace `findNodeDeviceUsage` with a simple map lookup.
* Keep the existing fallback behavior when a UUID is not found.
* Remove `findNodeDeviceUsage` after the change.

This should keep the current metric behavior unchanged while reducing repeated device scans.

## Why is this needed?

The current lookup is approximately:

`O(allocated devices × nodes × devices per node)`

For larger clusters, this can result in millions of comparisons during every Prometheus scrape.

The lookup is mainly needed to get `Totalcore` and `Type` for AMD core metric normalization. However, all devices currently pay the cost of the full scan.

Building the lookup map once changes the repeated lookup to approximately `O(1)` per device.

This is similar to the optimization made in #3019, where a repeated device UUID scan was replaced with a lookup.

im-Toqeer-506:
**What type of PR is this?**

/kind feature

`collectContainerMetrics` called `findNodeDeviceUsage` for every allocated device. Each call scanned all nodes and their devices, making the lookup increasingly expensive as the cluster grew.

This PR builds a device UUID → metadata map once per Prometheus scrape and uses it for constant-time lookups.

The map contains the `Totalcore` and `Type` values needed by `normalizeAMDCoreMetrics`. This avoids scanning the cluster for every device, especially when most devices are not AMD.

This follows the same approach used in #3019 for the Filter path.

**Behaviour is unchanged:**

- The lookup remains node-agnostic, matching the previous behaviour.
- Unknown UUIDs still fall back to zero-value metadata, so raw values are preserved.
- For duplicate UUIDs, the first entry is kept, matching the old lookup behaviour.

**Benchmarks:**

`BenchmarkCollectContainerMetrics` with 2000 allocated pods:

| Nodes | Before | After | Change |
| --- | --- | --- | --- |
| 10 | 14.50 ms/op | 13.24 ms/op | −9% |
| 100 | 24.00 ms/op | 12.78 ms/op | −47% |
| 1000 | 95.86 ms/op | 14.51 ms/op | −85% |

`BenchmarkCollect` with 200 pods:

| Nodes | Before | After | Change |
| --- | --- | --- | --- |
| 10 | 2.96 ms/op | 3.09 ms/op | +4% |
| 100 | 16.93 ms/op | 16.15 ms/op | −5% |
| 1000 | 143.28 ms/op | 124.89 ms/op | −13% |

The index adds some memory usage at larger cluster sizes, but significantly reduces lookup time. At 10 nodes, the full scrape is slightly slower because the cluster scan is still small.

Fixes #3058

* Changes are limited to the scheduler extender, with `collectRemoteGPUMetrics` unchanged. Added tests and benchmarks for the new device metadata index, and all local checks passed.
* Benchmarks were run with `go test ./cmd/scheduler/ -run '^$' -bench .`.

```release-note
NONE
```

**AI assistance disclosure**:
AI assistance was used for code analysis, test development, and wording. The changes were reviewed and validated by the contributor

## Summary by CodeRabbit

- **Performance**
  - Improved scheduler metrics collection efficiency, especially in larger clusters with many nodes, devices, and pods.

- **Bug Fixes**
  - Improved consistency of device-related metrics when devices have duplicate identifiers or are not present in the advertised node data.
  - Preserved raw core values for devices whose metadata cannot be resolved, preventing those metrics from being omitted.


archlitchi:
nice optimize

archlitchi (approved):
/lgtm

PR #3069, opened 2026-09-17
fix(ascend): defer mode-agnostic memory trimming

### LINKED ISSUE #3057: [Ascend] Defer memory trimming for mode-agnostic Pods until node filtering
/kind bug

**What happened**:

For an Ascend Pod without the `huawei.com/vnpu-mode` annotation, `MutateAdmission` currently treats the empty mode like template mode and rounds the requested memory to a template before a node is selected.

Example with 8192 MB and 16384 MB templates:

```yaml
resources:
  limits:
    huawei.com/Ascend910B4: 1
    huawei.com/Ascend910B4-memory: 10000
```

The webhook changes `10000` to `16384`. If Filter later evaluates a HAMi-core node, the original `10000` value is already lost and HAMi unnecessarily reserves `16384`.

```text
Current:
No mode + 10000 MB
        ↓
Webhook trims to 16384
        ↓
Template node: 16384 ✅
HAMi-core node: 16384 ❌ (should be 10000)
```

An unspecified mode should remain mode-agnostic:

- Explicit `template`: trim memory at the webhook.
- Explicit `hami-core`: preserve the exact memory.
- No mode: preserve the exact memory at the webhook and resolve it per candidate node in Filter.

```text
Expected:
No mode + 10000 MB
        ↓
Webhook keeps 10000
        ↓
Filter checks candidate node
        ├── Template node → trim to 16384
        └── HAMi-core node → keep 10000
```

Explicit mode/node mismatches should continue to be rejected.

1. Configure Ascend templates containing `8192` and `16384`.
2. Create a Pod requesting `10000` MB without `huawei.com/vnpu-mode`.
3. Call `MutateAdmission`.
4. Observe that the request is changed to `16384` before Filter knows the candidate node mode.

This is reproducible with unit tests and does not require Ascend hardware.

**Suggested implementation and tests**:

- In `MutateAdmission`, call `trimMemory` only when the mode is explicitly `template`.
- In `Fit`, trim a mode-agnostic request only when the candidate node uses template mode; preserve it on a HAMi-core node.
- Keep the calculation local to each candidate so evaluating one node does not modify another node’s request.
- Ensure allocation accounting and downstream metadata use the effective memory selected by Filter.
- Add tests for explicit template, explicit HAMi-core, and unspecified mode against both node modes.

This is the remaining node-aware behavior after #2696: that PR stopped inappropriate HAMi-core trimming, but an empty mode is still trimmed at admission. This issue tracks the v2.11 roadmap item “Ascend only trimming memory for templates mode” in #2764.

**Environment**:

- HAMi version: master @ `684c5f74f6a6`
- Component: Ascend admission webhook and scheduler Filter
- Hardware required for reproduction: No

git-jxj:
**What type of PR is this?**

/kind bug

Ascend admission currently rounds a mode-agnostic memory request before a node is selected. That loses the original request and can make the same Pod fit incorrectly when candidate nodes use different virtualization modes.

This change marks mode-agnostic requests for deferred resolution and resolves a copy of each request during node filtering:

- template-mode nodes round to the matching vNPU template;
- hami-core nodes keep the raw requested memory;
- the resolved mode is carried into allocation annotations and accounting without mutating the shared request.

The regression test evaluates the same `10000 MiB` request against both node modes and verifies `16384 MiB` for template mode and `10000 MiB` for hami-core mode.

Fixes #3057

This is limited to scheduler admission/filtering logic, so it was validated with the scheduler/device unit suite rather than in-container Ascend isolation tests.

Validation:

- `go test ./pkg/device/... -short`
- `make verify`

AI assistance disclosure: OpenAI Codex was used to assist with implementation and tests. I reviewed the complete diff, reproduced the old behavior, and validated the final scheduler logic locally.

```release-note
Fix Ascend mode-agnostic memory requests so they are resolved against each candidate node's virtualization mode.
```

## Summary by CodeRabbit

* **New Features**
  * Memory requests without a specified vNPU mode are now resolved according to the selected node’s capabilities.
  * Supported nodes preserve the requested memory, while template-based nodes apply the appropriate memory rounding.
  * Allocation details now accurately reflect the resolved mode and memory settings.

* **Bug Fixes**
  * Improved validation for insufficient or incompatible memory requests across different node types.


asadjan4611:
you can't raise the PR without my permission or without maintainer  permisssions
@mesutoezdil  please close this PR its againts the contriubtion guidelines.

bcz i am still discussing some stuff about the issue with the Menguxan

mesutoezdil:
it was still assigned to sb. therefore closed. thanks!

PR #3089, opened 2026-09-20
fix(enflame): stop binding two resource-name flags to one variable

### LINKED ISSUE #3016: enflame: resource name flags use the same variable
I noticed that these two flags in `pkg/device/enflame/config.go` are using the same variable:

```
--enflame-drs-gcu-resource-name
--enflame-vgcu-resource-name
```

Both are bound to `EnflameResourceNameDRSGCU`.

So if both flags are passed, the last one wins. tor example, this:

```text
--enflame-drs-gcu-resource-name=custom-a --enflame-vgcu-resource-name=custom-b
```

gives `custom-b`, while reversing them gives `custom-a`.

This is because both flags are registered on the same `FlagSet` and Go's `flag` package doesn't care that they're pointing to the same variable.

This could be an issue for users migrating from the old VGCU flag since the final resource name can change just based on argument order.

I also noticed that:

```text
--enflame-vgcu-legacy-resource-name
--enflame-vgcu-percentage-resource-name
```

seem to be unused. They're assigned in `InitEnflameDevice`, but I couldn't find anywhere reading those values afterwards.

I think we should define which flag takes precedence here and add a test for it. The unused flags probably also need to be checked to see if they're still needed.

No Enflame hardware should be needed since this can be tested at the config/flag level.

Ggrryta:
**What type of PR is this?**

/kind bug

--enflame-drs-gcu-resource-name and the legacy --enflame-vgcu-resource-name
were both bound to &EnflameResourceNameDRSGCU, so when both flags were passed
the last one parsed silently won. Each flag now parses into its own variable,
and InitEnflameDevice resolves them in a fixed order: yaml (new key >
legacy key) > new flag > legacy flag alias > default. A warning is logged
when both flags are set or when the yaml config overrides provided flags.

Fixes #3016

- make verify passes locally.
- The yaml config still wins over the flags, same as before. In standard
  chart deployments the configmap always sets resourceNameDRSGCU, so the
  flags only matter for deployments managing their own device config.
- This fixes the DRS-GCU resource-name flags only; the other enflame flags
  have pre-existing issues best addressed separately.
- Tested with unit tests only, no enflame hardware available to me.

Yes: when both flags are set the new flag wins now. Setups that relied on
the legacy flag winning by argument order must drop the new flag.

**AI assistance disclosure:**

This fix was developed with an AI coding assistant (investigation, code and
tests). I reviewed all the changes.

## Summary by CodeRabbit

- **Bug Fixes**
  - Enflame device resource-name configuration now resolves consistently regardless of flag order.
  - Configuration precedence is deterministic: YAML settings take priority, followed by the current flag, legacy flag, and default resource name.
  - Existing legacy VGCU resource-name configuration remains supported.
  - Added warnings when conflicting flags are provided or command-line values are overridden by YAML configuration.
  - Containers requesting a different resource name are skipped as expected.


asadjan4611:
@Ggrryta first of all please read the contribution guidelines then rasie the PR .
this PR does not comply with the contribution guidelines

kube-gopher (commented):
I don't think it's necessary to write such complicated comments, and there's no need to write separate comments for the old code either

Ggrryta:
@asadjan4611 Thanks for pointing this out

Ggrryta:
@kube-gopher Agreed. I removed the redundant comments and kept the one in InitEnflameDevice about the fixed resolution order — that is what someone reading this code later needs first.

mesutoezdil:

PR #3092, opened 2026-09-20
fix(device-plugin): auto driverroot make mount unchange

FouoF:
**What type of PR is this?**

/kind bug

The driver detection running after container creating, so it can not change the mount. So pre mount two possible path to the container and driver detection select them to the device plugin via ENV. Just follow what gpu-operator do.

Fixes #

## Summary by CodeRabbit

- **New Features**
  - Updated automatic NVIDIA driver path detection for host-installed drivers and GPU Operator environments.
  - Automatic mode now resolves driver paths through a unified host-root mapping, reducing configuration complexity.
  - Unsupported custom driver-root configurations are now rejected with a clear error.

- **Documentation**
  - Revised configuration guidance to describe automatic driver-path handling for host and GPU Operator installations.


cmd/device-plugin/nvidia/main.go
```diff
@@ -476,15 +479,33 @@ func resolveNvidiaDriverRoot(config *spec.Config) error {
...
 }
 
+func autoContainerDriverRoot(driverRoot string) (string, error) {
+	switch filepath.Clean(driverRoot) {
+	case hostNvidiaDriverRoot:
+		return hostDriverContainerRoot, nil
+	case gpuOperatorNvidiaDriverRoot:
+		return filepath.Join(gpuOperatorRunContainerRoot, "driver"), nil
+	default:
+		return "", fmt.Errorf("unsupported auto-detected NVIDIA driver root %q", driverRoot)
```
mesutoezdil, line 507: since /host-driver-root already mounts the full host root, would filepath.Join(hostDriverContainerRoot, driverRoot) handle any gpu operator reported root, instead of failing startup on anything besides these two exact paths?

FouoF: That's true. This is designed for isolation originally but actually it doesn't work. I will remove it.

FouoF: But mount root seems dangerous. I will check if it is necessary.

FouoF:
@mesutoezdil Remove the mount of root is possible but more complex. I will discuss it in the new issue #3099.

Shouren (approved):
/lgtm

PR #3093, opened 2026-09-21
refactor(device): share the allocation check with every vendor plugin

### LINKED ISSUE #3078: vendor device plugins do not validate the allocation annotation
The checks added in #3074 and #3075 live in this repository, so they only cover the in-tree NVIDIA device plugin. The Ascend, Cambricon, Hygon, Metax, Enflame, Kunlun, AWS Neuron, Biren and VastAI plugins read the same `hami.io/vgpu-devices-to-allocate` annotation from their own repositories and act on it without comparing it against the pod's limits. A pod forging that annotation on any of those vendors is still served exactly what it names, which is the hole reported in #3041. Moving the comparison into a shared helper under `pkg/device` that every plugin calls would keep it in one place, otherwise each repository needs its own copy.

mesutoezdil:
Closes #3078

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved validation of device allocations to prevent containers from exceeding requested memory or compute resources.
  - Added support for validating percentage-based memory requests when device capacity is available.
  - Absolute memory requests now take precedence when both absolute and percentage-based limits are provided.
  - Allocation errors now provide clearer details, including the affected container and device.
  - Requests that cannot be meaningfully bounded continue without an invalid comparison.
  - Validation now consistently applies across allocations involving multiple devices.


pkg/device/allocation.go
```diff
@@ -0,0 +1,71 @@
...
+// whose slices are charged a rounded-up capacity rather than the raw request,
+// as NVIDIA MIG is, has nothing to compare against and should not call this.
+func ValidateContainerAllocation(ctrName string, req ContainerDeviceRequest, allocated ContainerDevices, cardMemory CardMemoryMB) error {
+	for _, each := range allocated {
+		limit, bounded := requestedMemoryMB(req, each.UUID, cardMemory)
+		if bounded && each.Usedmem > limit {
+			return fmt.Errorf("container %s is allocated %d MB on device %s but requests %d MB",
+				ctrName, each.Usedmem, each.UUID, limit)
+		}
+		if req.Coresreq > 0 && each.Usedcores > req.Coresreq {
```
Shouren: @mesutoezdil I think we can not simply compare `req.Coresreq` and `each.Usedcores` for AMD and AWS Neuron device. AFAIK, `Usedcores` is the number of CU for AMD and the Bitmask for AWS Neuron.

mesutoezdil: yeah, right, done

pkg/device/allocation.go
```diff
@@ -0,0 +1,90 @@
...
+	if req.Memreq > 0 {
+		return req.Memreq, true
+	}
+	if req.MemPercentagereq >= 1 && req.MemPercentagereq <= 100 {
+		if cardMemory == nil {
+			return 0, false
+		}
+		total, ok := cardMemory(uuid)
+		if !ok {
+			return 0, false
```
blackdragoon26, line 85: Is this intentional,
the skipping of percentage-memory validation, when card memory is unknown?

PR #3094, opened 2026-09-21
fix(device): enforce minimum limit across multiple ResourceQuota objects

### LINKED ISSUE #2997: Only one ResourceQuota per namespace is enforced, and deleting any one of them removes the limit

ManthanNimodiya:
**What type of PR is this?**
/kind bug

Tracks device limits per ResourceQuota object in `QuotaManager` and recalculates the effective limit as the minimum across all ResourceQuota objects in the namespace upon additions, updates, and deletions. Previously, adding a second ResourceQuota overwrote the limit flatly and deleting any object cleared enforcement entirely.

Fixes #2997

- Validation: Verified locally across `pkg/device/...` and `pkg/scheduler/...` test suites. Added unit tests in `pkg/device/quota_test.go` covering multi-quota minimum enforcement, partial deletions, stale delete specs, and unrelated ResourceQuota handling.
- AI Assistance Disclosure: AI assistance was consulted for codebase exploration and test suite structuring; the per-object tracking logic, minimum recalculation, and unit test assertions were authored and validated locally.

NONE

## Summary by CodeRabbit

* **Bug Fixes**
  * Multiple resource quotas now enforce the most restrictive limit for each resource.
  * Removing a stricter quota correctly restores the remaining applicable limit.
  * Removing all applicable quotas clears the enforced limit.
  * Quota deletions with outdated or empty specifications now correctly remove previously tracked limits.
  * Quotas containing only unrelated resource settings no longer create misleading resource entries.


pkg/device/quota.go
```diff
@@ -238,51 +240,149 @@ func (q *QuotaManager) UpdateQuota(oldQuota, newQuota *corev1.ResourceQuota) {
...
+func (q *QuotaManager) recalcLimitLocked(namespace, resourceName string) {
+	var minLimit int64
+	var found bool
+
+	if nsObjs, ok := q.objectLimits[namespace]; ok {
+		for _, objLimits := range nsObjs {
+			if limit, ok := objLimits[resourceName]; ok {
+				if !found || limit < minLimit {
+					minLimit = limit
+					found = true
```
mesutoezdil, line 255: spec.scopes is never read, does a scoped quota now clamp every pod?

pkg/device/quota.go
```diff
@@ -238,51 +240,149 @@ func (q *QuotaManager) UpdateQuota(oldQuota, newQuota *corev1.ResourceQuota) {
...
-		if ok {
-			dn, ok := managedQuotaName(idx)
-			if !ok {
-				continue
+	if quota == nil || q.objectLimits == nil {
+		return
+	}
+
+	// For anonymous quotas (Name == ""), only proceed if the quota carries managed keys.
+	if quota.Name == "" && !hasManagedKeys(quota) {
```
mesutoezdil: can a resourcequota ever reach here with an empty name?

pkg/device/quota.go
ManthanNimodiya: Yeah, taking the min would've clamped the whole namespace. I updated QuotaManager to ignore scoped quotas so they don't over-restrict unrelated pods, and added a test covering it.

pkg/device/quota.go
ManthanNimodiya: You're right, in a real cluster that never happens since k8s always requires a name. That was just a leftover band-aid because an older unit test fixture didn't set metadata.Name. 

Dropped the check and gave the test fixtures proper names.

Shouren:
@ManthanNimodiya Please solve the "Treat an empty scope selector as unscoped." comment posted by codeRabbit.

Issue #3098, opened 2026-09-21
Device plugin rejects the scheduler's own allocation when both gpumem and gpumem-percentage are set (regression from #3075)


adity1raut:
**What happened**:

If a container sets both `nvidia.com/gpumem` and `nvidia.com/gpumem-percentage`, the device plugin rejects the allocation the scheduler booked for it, and the pod fails admission.

The scheduler (`pkg/device/nvidia/device.go:1023`) sizes the slice from `gpumem` and uses the percentage only when `gpumem` is 0. The plugin's `memoryLimitMB` (`pkg/device-plugin/nvidiadevice/nvinternal/plugin/util.go:157`) checks the percentage first. For example, on an 80 GB GPU with `gpumem: 20000` and `gpumem-percentage: 10`:

```
container main is allocated 20000 MB on device GPU-xxx but requests 8192 MB
```

**What you expected to happen**:

The plugin should compute the limit in the same order as the scheduler, accept the allocation, and start the pod with a 20000 MB limit.

**How to reproduce it (as minimally and precisely as possible)**:

Deploy HAMi from master (6a15b13) and apply:

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: gpumem-both
spec:
  containers:
  - name: main
    image: nvidia/cuda:12.4.1-base-ubuntu22.04
    command: ["sleep", "infinity"]
    resources:
      limits:
        nvidia.com/gpu: 1
        nvidia.com/gpumem: 20000
        nvidia.com/gpumem-percentage: 10
```

The pod is scheduled, then fails with `UnexpectedAdmissionError`.

**Anything else we need to know?**:

Suggested fix: in `memoryLimitMB`, check `req.Memreq > 0` before the percentage, as `Fit` does. The code comment there, which says the scheduler sizes from the percentage when both are set, is also wrong.

**Environment**:
- HAMi version:a regression from #3075
- Kubernetes version: any
- Accelerator vendor and model: NVIDIA, any
- HAMi device backend and requested resource names (for example, `nvidia.com/gpu`): nvidia; `nvidia.com/gpu`, `nvidia.com/gpumem`, `nvidia.com/gpumem-percentage`
- Accelerator driver and runtime version: any
- Container runtime and version: any
- Workload image and tag (redact private image names and state when redacted): nvidia/cuda:12.4.1-base-ubuntu22.04
- Operating system and kernel version (for example, the output of `uname -a`): any
- Others: found by code review and reproduced with a unit test

[...]

PR #3106, opened 2026-09-21
fix(scheduler): gpumem score before gpumem-percentage

diegodevxd:
**What type of PR is this?**

/kind bug

`ComputeScore` resolved the request memory percentage-first while `Fit`
resolves it `Memreq`-first, so a container that sets both `gpumem` and
`gpumem-percentage` was scored from one value and booked with the other.

On an 80GB card with `gpumem: 20000` and `gpumem-percentage: 10`, scoring
counted 8192MB against a 20000MB booking — a 59% undercount — so the device
ranked as emptier than it would be once the pod landed, and could win over a
device that was the better fit.

The guard now defers to the percentage only when `Memreq` is unset, which is
the same order the backends that carry a real memory percentage use in `Fit`
(nvidia, ascend, cambricon, hygon, iluvatar, metax, mthreads).

Part of #3098 — the scheduler-side half. The device-plugin half is #3093.

Scheduler-extender only, so per the hardware-validation gate this is covered by
unit testing rather than real GPU validation.

`TestComputeScore` already had two cases setting both fields
(`gpu_policy_test.go:578` and `:640`), but in both the request `Type` is
`"type2"` against a `"type1"` device, so the vendor filter drops them before the
memory branch — no existing expectation encoded the old behaviour. The new case
fails without the fix (`ComputeScore() = 2.25, want 3.6914062`) and passes with
it. `Count: 8` and `Totalmem: 81920` are chosen so every term is a binary
fraction, since the test compares float32 with exact `!=`.

Disclosure: I used AI to navigate the code, translate the recommendations, and to help generate the code. I reviewed the change, ran the tests, and I understand how it works.

```release-note
Fixed GPU scoring so that a container setting both `gpumem` and `gpumem-percentage` is scored from `gpumem`, matching the memory the scheduler books for it.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected GPU scheduling scores to prioritize fixed memory requests when both fixed and percentage-based memory requests are provided.
  * Improved consistency between memory allocation behavior and scheduling scores.


pkg/scheduler/policy/gpu_policy.go
```diff
@@ -215,7 +215,13 @@ func (ds *DeviceListsScore) ComputeScore(requests device.ContainerDeviceRequests
 
 		request += 1
 		core += container.Coresreq
-		if container.MemPercentagereq != 0 && container.MemPercentagereq != 101 {
+		// Size the request from Memreq first and fall back to the percentage
+		// only when Memreq is unset, matching how the backends that carry a
+		// real memory percentage resolve it in Fit (nvidia, ascend, cambricon,
+		// hygon, iluvatar, metax, mthreads). Scoring from the percentage while
+		// Fit books Memreq makes the device look emptier than the allocation
+		// it is about to receive.
```
mesutoezdil, line 223: enflame passes -1 here as a mode, does that score negative?

